PR #10, opened 2019-02-14
Fix y tensor shape in the XOR trivial test

JacopoMangiavacchi:
The label tensor y was defined with shape [4]
`let y: Tensor<Float> = [0, 1, 1, 0]`

It should have instead shape [4, 1]
`let y: Tensor<Float> = [[0], [1], [1], [0]]`

Resolves [TF-209](https://bugs.swift.org/browse/TF-209).


JacopoMangiavacchi:
I signed it!

rxwei:
Thanks! Have you verified this actually fixed the problem? I'd suggest adding some XCAsserts.

JacopoMangiavacchi:
I've added some XCAsserts **BUT** I don't have an environment (gpu) right now to test with "swift test". 

I only tested on a Colab notebook.

rxwei:
Great. I've tested it locally and it works. Could you apply this change to simplify the test please?

```diff
-        XCTAssertLessThanOrEqual(predicted[0][0], 0.3)
-        XCTAssertGreaterThanOrEqual(predicted[1][0], 0.8)
-        XCTAssertGreaterThanOrEqual(predicted[2][0], 0.8)
-        XCTAssertLessThanOrEqual(predicted[3][0], 0.3)
+        let ŷ = classifier.applied(to: [[0, 0], [0, 1], [1, 0], [1, 1]])
+        XCTAssertEqual(ceil(ŷ), y)
```

JacopoMangiavacchi:
I've updated the change but I think it should call round() and not ceil() on the prediction.

Anyway trying with round on Colab with the last snapshot round() gave me this error:
error: argument type 'Tensor<Float>' does not conform to expected type 'FloatingPoint'

ceil() works but as expected it is returning [[1], [1], [1], [1]] witch of course is not the expected result.

rxwei:
Ah right, `round` is not defined in the standard library yet. I'll add them.

rxwei:
Filed [TF-248](https://bugs.swift.org/browse/TF-248).

rxwei:
I won't be able to get to it today, but if you are free, feel free to self-assign that bug and add it in the DeepLearning module! You can add a file called Operators.swift

rxwei:
Awesome. Thanks Jacopo!

PR #11, opened 2019-02-14
Use valueWithPullback directly rather than the special curry version.

pschuh:


Sources/DeepLearning/Layer.swift
```diff
@@ -36,12 +36,16 @@ public protocol Layer: Differentiable & KeyPathIterable
     func applied(to input: Input) -> Output
 }
 
-public extension Layer {
+public extension Layer
+    where TangentVector : AdditiveArithmetic, CotangentVector : AdditiveArithmetic,
```
rxwei: To make this more error-prone, let's add these constraints to all differential operators defined in [AutoDiff.swift](https://github.com/apple/swift/blob/tensorflow/stdlib/public/core/AutoDiff.swift) and add FIXME's explaining that they are workarounds for [SR-9595](https://bugs.swift.org/browse/SR-9595).

Sources/DeepLearning/Layer.swift
```diff
@@ -36,12 +36,16 @@ public protocol Layer: Differentiable & KeyPathIterable
...
+          Input.TangentVector : AdditiveArithmetic, Input.CotangentVector : AdditiveArithmetic {
     func valueWithPullback(at input: Input)
         -> (output: Output,
             pullback: (Output.CotangentVector)
                 -> (layerGradient: CotangentVector, inputGradient: Input.CotangentVector)) {
-        let (out, pullback) = _valueWithPullback(at: self, input, in: Self.applied(to:))
+        let (out, pullback) = valueWithPullback(at: input) { layer, input in
```
rxwei: Just return it directly.

Sources/DeepLearning/Layer.swift
dan-zheng: Adding constraints to differential operators is a very good idea.

Sources/DeepLearning/Layer.swift
pschuh: Please note that the returned pullback is a labeled tuple: (layerGradient: CotangentVector, inputGradient: Input.CotangentVector)

rxwei: I see. I'd expect tuple coercion to work implicitly, but apparently my memory is from years ago.

PR #12, opened 2019-02-17
TF-245: Add the sequential combinator.

saeta:
It's common to want to combine network layers sequentially. This commit adds
the >> operator that makes it trivial to feed the output of one layer to the
input of the next layer.

Note: this commit also adds in the skeleton for a parallel combinator (e.g.
one that could be easily used to form residual networks) but this triggers a
compiler crash. (TF-244)

Related bugs: TF-235

### DIFF
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -82,15 +82,46 @@ public struct Sequential<LHS: Layer, RHS: Layer>: Layer
     }
 }
 
+/// Sequencing two layers to form a chain.
+///
+/// The sequential combinator composes two layers sequentially, feeding the
+/// output of the first layer as input to the second layer.
 public func >> <LHS: Layer, RHS: Layer>(_ lhs: LHS, _ rhs: RHS) -> Sequential<LHS, RHS> {
     return Sequential(lhs, rhs)
 }
 
+/// A sequencing of a Layer and an activation function.
+public struct SequentialFunction<LHS: Layer, OutputType: Differentiable>: Layer
+    where LHS.TangentVector: AdditiveArithmetic,
+          LHS.CotangentVector: AdditiveArithmetic,
+          LHS.Input.CotangentVector: AdditiveArithmetic,
+          LHS.Output.CotangentVector: AdditiveArithmetic {
+    let lhs: LHS
+
+    @noDerivative
+    let rhs: @differentiable (LHS.Output) -> OutputType
+
+     init(_ lhs: LHS, _ rhs: @escaping @differentiable (LHS.Output) -> OutputType) {
+        self.lhs = lhs
+        self.rhs = rhs
+    }
+
+    @differentiable(wrt: (self, input))
+    public func applied(to input: LHS.Input) -> OutputType {
+        return rhs(lhs.applied(to: input))
+    }
+}
+
+/// Sequence a Layer with a differentiable activation function.
+public func >> <LHS: Layer, OutputType>(
+        _ lhs: LHS,
+        _ rhs: @escaping @differentiable (LHS.Output) -> OutputType) -> SequentialFunction<LHS, OutputType> {
+    return SequentialFunction(lhs, rhs)
+}
+
 // TODO(TF-244): Uncomment once TF-244 is resolved.
-//public struct Parallel<LHS: Layer, RHS: Layer, LHSOutput, RHSOutput, AggregateOutput: Differentiable>: Layer
-//    where LHS.Output == LHSOutput,
-//          RHS.Output == RHSOutput,
-//          LHS.Input == RHS.Input,
+//public struct Parallel<LHS: Layer, RHS: Layer, AggregateOutput: Differentiable>: Layer
+//    where LHS.Input == RHS.Input,
 //          LHS.TangentVector: AdditiveArithmetic,
 //          RHS.TangentVector: AdditiveArithmetic,
 //          LHS.CotangentVector: AdditiveArithmetic,
@@ -99,12 +130,13 @@ public func >> <LHS: Layer, RHS: Layer>(_ lhs: LHS, _ rhs: RHS) -> Sequential<LH
 //          LHS.Output.CotangentVector: AdditiveArithmetic,
 //          RHS.Output.CotangentVector: AdditiveArithmetic,
 //          RHS.Output.TangentVector: AdditiveArithmetic {
+//    typealias CombinerFunction = @differentiable (LHS.Output, RHS.Output) -> AggregateOutput
 //    let lhs: LHS
 //    let rhs: RHS
 //    @noDerivative
-//    let combiner: @differentiable (LHSOutput, RHSOutput) -> AggregateOutput
+//    let combiner: CombinerFunction
 //
-//     init(_ lhs: LHS, _ rhs: RHS, combinedWith combiner: @escaping @differentiable (LHSOutput, RHSOutput) -> AggregateOutput) {
+//     init(_ lhs: LHS, _ rhs: RHS, combinedWith combiner: @escaping CombinerFunction) {
 //        self.lhs = lhs
 //        self.rhs = rhs
 //        self.combiner = combiner
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -51,6 +51,71 @@ public extension Layer
     }
 }
 
+/// A sequencing of two Layers to form a chain.
+///
+/// The sequential combinator composes two layers sequentially, feeding the
+/// output of the first layer as input to the second layer.
+///
+/// It is normally constructed via the >> operator.
+public struct Sequential<LHS: Layer, RHS: Layer>: Layer
+    where LHS.Output == RHS.Input,
+          LHS.TangentVector: AdditiveArithmetic,
+          RHS.TangentVector: AdditiveArithmetic,
+          LHS.CotangentVector: AdditiveArithmetic,
+          RHS.CotangentVector: AdditiveArithmetic,
+          LHS.Input.CotangentVector: AdditiveArithmetic,
+          LHS.Output.CotangentVector: AdditiveArithmetic,
+          RHS.Output.CotangentVector: AdditiveArithmetic,
+          RHS.Output.TangentVector: AdditiveArithmetic {
+    let lhs: LHS
+    let rhs: RHS
+
+     init(_ lhs: LHS, _ rhs: RHS) {
+        self.lhs = lhs
+        self.rhs = rhs
+    }
+
+    @differentiable(wrt: (self, input))
+    public func applied(to input: LHS.Input) -> RHS.Output {
+        let intermediateValue = lhs.applied(to: input)
+        return rhs.applied(to: intermediateValue)
+    }
+}
+
+public func >> <LHS: Layer, RHS: Layer>(_ lhs: LHS, _ rhs: RHS) -> Sequential<LHS, RHS> {
+    return Sequential(lhs, rhs)
+}
+
+// TODO(TF-244): Uncomment once TF-244 is resolved.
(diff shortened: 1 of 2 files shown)


saeta:
Note: I currently don't know of the right way to refer to the type of the combined model... do folks have any thoughts?

Sources/DeepLearning/Layer.swift
```diff
@@ -51,6 +51,71 @@ public extension Layer
...
+        return rhs.applied(to: intermediateValue)
+    }
+}
+
+public func >> <LHS: Layer, RHS: Layer>(_ lhs: LHS, _ rhs: RHS) -> Sequential<LHS, RHS> {
+    return Sequential(lhs, rhs)
+}
+
+// TODO(TF-244): Uncomment once TF-244 is resolved.
+//public struct Parallel<LHS: Layer, RHS: Layer, LHSOutput, RHSOutput, AggregateOutput: Differentiable>: Layer
```
dan-zheng: It should not be necessary to add `LHSOutput` and `RHSOutput` generic type parameters because you can refer to them as `LHS.Output` and `RHS.Output`.

Have you tried that? The type of `combiner` would then be:
```swift
let combiner: @differentiable (LHS.Output, RHS.Output) -> AggregateOutput
```

dan-zheng:
> Note: I currently don't know of the right way to refer to the type of the combined model... do folks have any thoughts?

Are you referring to the name of the combinator types (currently `Sequential` and `Parallel`)?
If so:
* I think `Sequential` is a fine term-of-art name. An alternative is "literal" names like `ComposedLayers`.
* I'm not familiar with the `Parallel` combinator - does it have precedent in other frameworks, or do you have a specific use case in mind?

Sources/DeepLearning/Layer.swift
saeta: Good point. Thanks!

I've updated the code. Haven't checked yet to see if it compiles because I'm currently running into [TF-244](https://bugs.swift.org/browse/TF-244).

saeta:
@dan-zheng Thanks for the speedy feedback! I was thinking more parallel would be useful for things like the [`ResidualConvBlock`](https://github.com/tensorflow/swift-models/blob/stable/ResNet/ResNet50.swift#L67). Does that make sense?

Sources/DeepLearning/Layer.swift
rxwei: This is not the right direction. Layer identity and existence as a property are two critical part of this API. The former is for clarify, while the latter is for being *even able* to differentiate w.r.t. parameters inside chained layers.

rxwei:
Thanks for realizing the combinator idea! There are two kinds of things we should consider adding, and I’d prefer landing the application operator first:
* Composition, like the one you wrote here: lhs >> rhs gives you a new layer that you can put in a struct as a stored property. 
* Application: x >> layer1 >> layer2 applies layer 1 and layer 2 sequentially. 

I’d prefer to land the application operator as a first step because it has the least impact to the programming model and type complexity. The composition operator would introduce nested long generic parameters that are beyond what can be considered as being idiomatic in Swift, let alone the type checking complexity when the user defines a chain of layers as a store peoperty initializer expression. So we’ll need to carefully evaluate that. 

Lastly, both these operators will be infected by the ‘isTraining’ parameter change. I’ll leave a response on Ed’s proposal PR and land the code patch in the API repo soon.

dan-zheng:
> Thanks for realizing the combinator idea! There are two kinds of things we should consider adding, and I’d prefer landing the application operator first:
> 
> * Composition, like the one you wrote here: lhs >> rhs gives you a new layer that you can put in a struct as a stored property.
> * Application: x >> layer1 >> layer2 applies layer 1 and layer 2 sequentially.

I completely agree with Richard's assessment. We should have composition and application combinators.

The `|>` "pipe-forward" operator is the term-of-art for the "application combinator". It's originally from [`F#`](https://theburningmonk.com/2011/09/fsharp-pipe-forward-and-pipe-backward):
```fsharp
// Pipe forward operator.
let (|>) x f = f x

// Forward composition operator (added in this patch).
// TensorFlow Scala has this operator too.
let (>>) f g x = g (f x)

// There are also backward versions of these operators.
// I don't think we need them though.
```

dan-zheng:
> @dan-zheng Thanks for the speedy feedback! I was thinking more parallel would be useful for things like the [`ResidualConvBlock`](https://github.com/tensorflow/swift-models/blob/stable/ResNet/ResNet50.swift#L67). Does that make sense?

I see. Pasting here for reference:
```swift
    @differentiable(wrt: (self, input))
    func applied(to input: Tensor<Float>) -> Tensor<Float> {
        var tmp = relu(layer1.applied(to: input))
        tmp = relu(layer2.applied(to: tmp))
        tmp = layer3.applied(to: tmp)
        // This next line is the reason for `Parallel`.
        // We want behavior that deviates from `shortcut.applied(to:)`.
        return relu(tmp + shortcut.applied(to: input))
    }
```

Again, I agree with Richard's assessment: this is a slightly tricky problem and requires more design, so we should land the core "composition" (`>>`) and "application"  (`|>`) combinators first.

"Merge layers" a la Keras may help mitigate related problems. https://keras.io/layers/merge
These are utility layers that perform common n-ary operations so that users can express simple arbitrary computation while continuing to work with `Layer`.

rxwei:
> The `|>` "pipe-forward" operator is the term-of-art for the "application combinator".

As usual, I'd like to push back on defining operators that come from ML-inspired languages in Swift, and this "term of art" is rather tied to those languages. I'm enjoy using those languages, but in Swift we have to think more about clarity and consistency with the standard conventions.

> let prediction = input |> dense1 >> dropout >> dense2 >> dropout >> dense3
> Again, I agree with Richard's assessment: this is a slightly tricky problem and requires more design, so we should land the core "composition" (`>>`) and "application" (`|>`) combinators first.

This is _not_ what I mean, and is not the right direction for the first patch. I suggested adding only one operator: the application operator. The application combinator is left associative and should never require a composition operator to work.  If you introduce a composition operator you'll end up with a lot of problems I mentioned above, here plus the necessity of adding/synthesizing a `@differentiable` attribute on the types.

dan-zheng:
> This is _not_ what I mean, and is not the right direction for the first patch. I suggested adding only one operator: the application operator. The application combinator is left associative and should never require a composition operator to work. If you introduce a composition operator you'll end up with a lot of problems I mentioned above, here plus the necessity of adding/synthesizing a `@differentiable` attribute on the types.

Oops, sorry for not reading your original response more carefully, and for suggesting potentially misleading ideas. My ideas are not polished, please take them with a grain of salt.

I fully agree with adding minimal functionality/complexity in the first patch. I'm dropping out of this conversation to focus on other work! @saeta: I would recommend coordinating closely with @rxwei before further development to prevent throwaway effort.

Sources/DeepLearning/Layer.swift
rxwei: BTW, I do acknowledge the validity of this idea in the fullness of time. Differentiating w.r.t. inner layers' parameters will fall out of the _currying and differentiable closure contexts_ semantics that @pschuh is working on. But even then, this will not be a preferable approach because
- `AnyDerivative` existential is highly inefficient than concrete layer types. This is already worth a thousand words I suppose...
- Users can't refer to sublayers, parameters, per-sublayer, or per-parameter gradients in that model.

Sources/DeepLearning/Layer.swift
```diff
@@ -51,6 +51,103 @@ public extension Layer
     }
 }
 
+/// A sequencing of two Layers to form a chain.
+///
+/// The sequential combinator composes two layers sequentially, feeding the
+/// output of the first layer as input to the second layer.
+///
+/// It is normally constructed via the >> operator.
```
rxwei:
```suggestion
/// It is normally constructed via the `>>` operator.
```

Sources/DeepLearning/Layer.swift
```diff
@@ -51,6 +51,103 @@ public extension Layer
...
+          LHS.CotangentVector: AdditiveArithmetic,
+          RHS.CotangentVector: AdditiveArithmetic,
+          LHS.Input.CotangentVector: AdditiveArithmetic,
+          LHS.Output.CotangentVector: AdditiveArithmetic,
+          RHS.Output.CotangentVector: AdditiveArithmetic,
+          RHS.Output.TangentVector: AdditiveArithmetic {
+    let lhs: LHS
+    let rhs: RHS
+
+     init(_ lhs: LHS, _ rhs: RHS) {
```
rxwei:
```suggestion
    init(_ lhs: LHS, _ rhs: RHS) {
```

Sources/DeepLearning/Layer.swift
```diff
@@ -51,6 +51,103 @@ public extension Layer
     }
 }
 
+/// A sequencing of two Layers to form a chain.
+///
+/// The sequential combinator composes two layers sequentially, feeding the
+/// output of the first layer as input to the second layer.
+///
+/// It is normally constructed via the >> operator.
+public struct Sequential<LHS: Layer, RHS: Layer>: Layer
```
rxwei: I'd suggest renaming `LHS` and `RHS` (along with their corresponding properties, `lhs` and `rhs`) to unabbreviated words, e.g. `First` and `Second`.

rxwei:
Ok, I'm gonna help you land the first sequential combinator, and improve over time!

rxwei:
Updates: 
- I changed `>>` to `>>>`. `>>` conflicts with the bitwise shift operator and is defined as not having any precedence. Operator declarations cannot shadow one another. I defined operator precedence group `FunctionCompositionPrecedence` for `>>>`.
- I moved `>>>` to an extension to `Layer` since that's the most idiomatic way to add an operator.
- I dropped combinators other than `Sequential` because even this alone is non-trivial to unblock.
- `Sequential` is blocked by [TF-266](https://bugs.swift.org/browse/TF-266).

saeta:
Closing this pull request as this is not the direction we want to go.

eaplatanios:
@saeta @rxwei I currently have a sequential combinator implementation that seems to compile and work fine. Should I open a PR? What was the main reason that this PR was closed?

rxwei:
Combinator style APIs or combinator operators are not super common or idiomatic in Swift. While we are open to discussing this direction, I'd suggest putting a hold on this for now and prioritizing the evolution of the foundation (e.g. protocols and optimizers).

eaplatanios:
The reason I brought this up is that currently there is no simple way to compose layers. In order to create a layer that is defined as the composition of two other layers one needs to create an entirely new layer struct. For example, a multi-layer perceptron could be simply defined as something like this if we allowed for a composition operator:
```swift
let mlp = Dense<Float>(inputSize: 784, outputSize: 128) >>> ReLU<Float>() >>>
  Dense<Float>(inputSize: 128, outputSize: 64) >>> ReLU<Float>() >>>
  Dense<Float>(inputSize: 64, outputSize: 32) >>> ReLU<Float>() >>>
  Dense<Float>(inputSize: 32, outputSize: 10) >>> Softmax<Float>()
```
This is useful when defining layer "wrappers", which I currently do for the contextual parameter generation methods that I propose in my research. Such wrappers could also be more generally useful. Currently one needs to define a new layer struct per such composition (I cannot use the `sequenced` API directly because I'm not applying the composition to an input).

Note that the above example compiles and works fine in my current implementation so I could open a PR to continue the discussion there if you want.

I agree that math-related protocols are super important and should be prioritized. I just mentioned this because I already have a working version. Regarding the protocols, I want to think more about the use cases and will post some more thoughts later on.

rxwei:
I agree that defining new structs isn't ideal. Sure, opening a new PR sounds good to me, though we may not proceed with merging it in the short term. You could also add it to your own library to start using it :)

eaplatanios:
Sounds good. And yes that's what I'm already doing and do not depend on merging this. I just thought it might useful for others too and was curious to know why this PR was closed in the past. I was under the impression that the main reason was a limitation of auto-diff that's not relevant anymore. :)

PR #13, opened 2019-02-19
Replace `LearningPhaseIndicator` with a `context` argument to `Layer.applied(to:in:)`.

rxwei:
This PR address proposal https://github.com/tensorflow/swift/pull/115.

~It's currently blocked by bug [TF-256](https://bugs.swift.org/browse/TF-256).~


Sources/DeepLearning/Layer.swift
```diff
@@ -16,51 +16,51 @@
 @_exported import TensorFlow
 #endif
 
+public enum LearningPhase {
```
rxwei: Comments to be added when crashers get fixed.

rxwei:
Merging to unblock progress. Happy to address any comments later.

PR #14, opened 2019-02-20
[PRNG] Initialize tensors using Philox PRNG

austinzheng:
Changes:
- Added global instances of Philox and Threefry PRNGs (seeded using arc4random)
- Tensor initializers now use Philox if no PRNG is specified

### DIFF
--- a/Sources/DeepLearning/Random.swift
+++ b/Sources/DeepLearning/Random.swift
@@ -125,6 +125,10 @@ private typealias UInt32x4 = (UInt32, UInt32, UInt32, UInt32)
 /// share state. The random data generated is of high-quality, but is not
 /// suitable for cryptographic applications.
 public struct ThreefryRandomNumberGenerator: SeedableRandomNumberGenerator {
+    public static var global = ThreefryRandomNumberGenerator(
+        uint64Seed: ARC4RandomNumberGenerator.global.next()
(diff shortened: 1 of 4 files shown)


Sources/DeepLearning/Random.swift
```diff
@@ -125,6 +125,10 @@ private typealias UInt32x4 = (UInt32, UInt32, UInt32, UInt32)
 /// share state. The random data generated is of high-quality, but is not
 /// suitable for cryptographic applications.
 public struct ThreefryRandomNumberGenerator: SeedableRandomNumberGenerator {
+    public static var global = ThreefryRandomNumberGenerator(
+        uint64Seed: ARC4RandomNumberGenerator.global.next()
```
dan-zheng: Sorry for speculation since I'm not familiar with the code.
Is depending on `ARC4RandomNumberGenerator.global.next()` as the seed intentional here?

austinzheng: Yeah. Now that I think about it, though, I don't know if it's a good idea. The original intent was to seed the PRNG with a 64-bit value (as opposed to the 32-bit value returned by `time(nil)`). However, doing this doesn't actually buy us anything, since arc4random is also deterministic. I'll change it.

PR #16, opened 2019-02-21
Aggregate improvements.

rxwei:
1. Introduce a `RandomDistribution` protocol that unifies all random distribution classes in the library.
2. Clean up tensor initializers to reduce code duplication.
3. Add `identity(_:)` function.
4. Make two `Context` initializers be `required` since they can be subclassed.
5. Enhance `Layer` methods and make backward-compatible. 
  * Introduce `inferring(from:)`, which is equivalent to `applied(to:in:)` with a default inference context.
  * Add back `applied(to:)` that simply calls `inferring(from:)`. Add a deprecation warning that tells people to switch to either `applied(to:in:)` or `inferring(from:)`.
  * Rename `Layer.valueWithPullback(at:in:)` to `appliedForBackpropagation(to:in:)` to ease the learning curve.
6. Improve documentation on everything.

### DIFF
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -16,19 +16,28 @@
 @_exported import TensorFlow
 #endif
 
+/// A value that indicates either a training phase or an inference phase for a layer.
 public enum LearningPhase {
     case training
     case inference
 }
 
+/// A context that stores contextual information used for the application of layers.
 open class Context {
+    /// The current learning phase.
     public var learningPhase: LearningPhase
 
-    public init(learningPhase: LearningPhase) {
+    /// Creates a context.
+    ///
+    /// - Parameter learningPhase: The current learning phase.
+    public required init(learningPhase: LearningPhase) {
         self.learningPhase = learningPhase
     }
 
-    public init(_ other: Context) {
+    /// Creates a context by copying all information from an existing context.
+    ///
+    /// - Parameter context: The existing context to copy from.
+    public required init(_ other: Context) {
         self.learningPhase = other.learningPhase
     }
 }
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -101,14 +101,19 @@ public extension Layer {
     }
 }
 
-/// A mutable, shareable reference to a tensor
-public final class Parameter<T: TensorFlowScalar> {
-    public var value: Tensor<T>
-    public init(_ value: Tensor<T>) {
+/// A mutable, shareable, owning reference to a tensor.
+public final class Parameter<Scalar: TensorFlowScalar> {
+    public var value: Tensor<Scalar>
+    public init(_ value: Tensor<Scalar>) {
         self.value = value
     }
 }
 
+/// A densely-connected neural network layer.
+///
+/// `Dense` implements the operation `activation(matmul(input, weight) + bias)` where `activation`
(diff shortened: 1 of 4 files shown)


rxwei:
I'm gonna merge now to unblock progress. Comments welcome!

I'm going to merge each commit instead of squashing here, because each commit in this PR is completely functionally independent and has a more descriptive commit message.

Sources/DeepLearning/Layer.swift
```diff
@@ -293,12 +317,10 @@ public struct AvgPool2D<Scalar: TensorFlowFloatingPoint>: Layer {
 public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
     /// The offset value, also known as beta.
     public var offset: Tensor<Scalar>
-
     /// The scale value, also known as gamma.
     public var scale: Tensor<Scalar>
-
+    /// The axis.
```
dan-zheng: How about slightly more description? e.g. `The axis of normalization.`

Sources/DeepLearning/Layer.swift
```diff
@@ -47,16 +56,43 @@ public protocol Layer: Differentiable & KeyPathIterable
...
+               message: "Switch to 'applied(to:in:)' for training, or 'inferring(from:)' for inference")
+    func applied(to input: Input) -> Output {
+        return inferring(from: input)
+    }
+
+    /// Returns the inference output obtained from applying the layer to the given input.
+    ///
+    /// - Parameter input: The input to the layer.
+    /// - Returns: The inference output.
+    func inferring(from input: Input) -> Output {
```
dan-zheng: How about `inferred(from:)` in the past tense to be consistent with `applied(to:)`?

Sources/DeepLearning/Operators.swift
```diff
@@ -16,8 +16,12 @@
...
-// This is a manual definition.
-func round<T : BinaryFloatingPoint>(_ t: Tensor<T>) -> Tensor<T> {
-    return Raw.round(t)
+// Rounds the values of a tensor to the nearest integer, element-wise.
+func round<Scalar: BinaryFloatingPoint>(_ x: Tensor<Scalar>) -> Tensor<Scalar> {
+    return Raw.round(x)
+}
+
+// Return a tensor with the same shape and contents as input.
+func identity<Scalar>(_ x: Tensor<Scalar>) -> Tensor<Scalar> {
```
dan-zheng: Do you intend to use `identity` as a default activation function (e.g. default argument to `Dense` initializer)?

FYI: shape errors due to differentiating an identity function activation could be improved: https://bugs.swift.org/projects/TF/issues/TF-243

Sources/DeepLearning/Layer.swift
rxwei: That would not be a semantically correct name -- past participle verbs should only be used when `self` is an object as in "apply a layer to something" (hence `applied(to:)`). If you use `inferred(from:)` then it will read like "infer a layer from something", which is clearly incorrect.

Sources/DeepLearning/Operators.swift
rxwei: I do not intend to do so because 1) default activation as identity is not common and 2) it's better to show the activation argument in code completion as the first candidate.

Sources/DeepLearning/Layer.swift
rxwei: Will def'ly fix when we have a systematic documentation pass over it.

PR #17, opened 2019-02-21
Execution context

ewconnell:
With the available time, I went with more code less talk. I believe this solution meets the desired goals and is pleasing to use.

- The execution context is fully extensible
- The context is passed as a parameter to the `applied(in:to:` function replacing the use of `LearningPhaseIndicator`
- The context is not held by the model, so models will now copy correctly without side effects
- The use is clean and nicely readable

Example use in a composite layer
```swift
public func applied(in context: ExecutionContext,
                    to input: Tensor<Float>) -> Tensor<Float> {
  let h0 = conv1.applied(in: context, to: input)
  let h1 = maxPool1.applied(in: context, to: h0)
  let bn = batchNorm.applied(in: context, to: h1)
  let h2 = conv2.applied(in: context, to: bn)
  let h3 = maxPool2.applied(in: context, to: h2)
  let dense1InputShape = Tensor<Int32>([h3.shape[0], 800])
  let h4 = dense1.applied(in: context, to: h3.reshaped(toShape: dense1InputShape))
  return dense2.applied(in: context, to: h4)
}
```
Selection within an operator such as `BatchNorm` is clear 
```swift
public func applied(in context: ExecutionContext,
                    to input: Tensor<Scalar>) -> Tensor<Scalar> {
    if context.executionMode == .training {
        return applyTraining(in: context, to: input)
    } else {
        return applyInference(in: context, to: input)
    }
}
```
Note: The apple/swift repo for the TensorFlowRuntime tests will need to be updated as well, once these changes are agreed on.

### DIFF
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -16,13 +16,43 @@
 @_exported import TensorFlow
 #endif
 
+/// Specifies whether applied(in:to: should perform inference or training
+public enum ExecutionMode { case inference, training }
+
+/// The ExecutionContext is a base class used to configure
+/// model resources for execution in different environments such as
+/// CPU, GPU, Google Cloud, AWS, Azure, etc..
+open class ExecutionContext {
+    public let executionMode: ExecutionMode
+    
+    public init(executionMode: ExecutionMode) {
+        self.executionMode = executionMode
+    }
+}
+
+/// TODO: move this to wherever TensorShape is defined
+public extension TensorShape {
+    func flattened(axis: Int = 0) -> TensorShape {
+        precondition(axis < self.count)
+        var dims: [Int32]
+        switch axis {
+        case 0: dims = [self.indices.reduce(1, *)]
+        case 1: dims = [self[0], self.indices.suffix(axis).reduce(1, *)]
+        default: dims =
+            [Int32](self.indices.prefix(upTo: Int32(axis))) +
+            [self.indices.suffix(axis).reduce(1, *)]
+        }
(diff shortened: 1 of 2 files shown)


ewconnell:
trying to rerun googlebot to resolve CLA error

rxwei:
Thanks Ed!  I have actually added the proposed API in this repo (therefore there's merge conflict) as I was migrating existing models and tests. It's here: https://github.com/tensorflow/swift-apis/blob/8b922f5da515ff7ab451df1254b67dca865d1b47/Sources/DeepLearning/Layer.swift#L19. Could you take a look and see if you are happy with it?

There's some difference from your implementation
- `ExecutionMode` is called `LearningPhase` instead. I did this because it's more aligned with Keras. 
- I put the `in:` argument at last, because it reads more like a trailing argument that's not going to vary across different calls to `applied`. Ideally we really want to have Scala's implicit arguments in Swift.
Let me know if these makes sense to you!

Plus, I added `inferring(from:)` as a short hand for `applied(to:in:)` with an inference context.

ewconnell:
Looks like you have everything taken care of already. Great job :)

rxwei:
Thanks for pushing this through! I was experimenting with your proposed change earlier to see whether it would trigger bugs, and it did ([TF-256](https://bugs.swift.org/browse/TF-256)) so we fixed that. Then I committed the API in the repo to work on migrating the models it broke. Sorry for not letting you know earlier.

Let me know if you have any concerns about the current implementation or naming, and feel free to create PRs to improve them.

PR #20, opened 2019-02-22
Add a helper function for sequential models.

saeta:
Many deep learning models are composed of sequential layers stacked one on
top of each other. It can be relatively tedious to write out the explicit
`applied(to:)` function because it's fairly repetitive and the underlying
intent is relatively obscured. (It can be especially bothersome because
it's the 2nd (or 3rd) time you're writing out all the layers. (The first time
is to declare all the instance variables, and the second time (if necessary)
is in the initializer.)

Fortunately, with a single helper functions, we can make everything both type
safe as well as convenient and easily expressible & readable!

This commit adds a family of `Sequential` functions that take in a context, an
input, and a variable number of layers. It chains through the output of one
layer into the input of the next.

This API approach has a number of advantages:

 1. It avoids introducing new symbolic operators, which can be very confusing
    to new users.

 2. It works with today's AutoDiff implementation. (Yay!)

 3. It is very readable and clean.

 4. It avoids users "getting stuck". Concretely, if someone implemented a model
    using my previously proposed `>>>` operator, if they wanted to add a
    residual (or skip) connection, they would have to basically re-write their
    whole model using a struct, etc. With this API structure, only "local"
    changes are required. (e.g. If only one skip-connection is required, they
    can split the sequential chain into two pieces.)

Downsides of this approach:

 1. It doesn't DRY-out the types required to define a model. (I have some
    thoughts here, but there isn't enough room in this
    margin^H^H^H^H^H^Hcommit message.)

 2. We should think hard about how things should look when we have loops.

 3. I'm sure there's a better way to code-gen out all the different Sequential
    airities. (I got bored hand-writing them out after 4...) Suggestions
    welcome!

### DIFF
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -101,8 +101,8 @@ public extension Layer {
     }
 }
 
+/// Adds helpers for standard feed-forward, sequential models.
 public extension Differentiable {
-
     @differentiable(wrt: (self, l1, l2))
     func sequenced<L1: Layer, L2: Layer>(
         in context: Context, through l1: L1, _ l2: L2)
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -101,6 +101,37 @@ public extension Layer {
     }
 }
 
+@differentiable(wrt: (input, l1, l2))
+public func Sequential<L1: Layer, L2: Layer>(in context: Context, from input: L1.Input, _ l1: L1, _ l2: L2) -> L2.Output where L1.Output == L2.Input {
+  let o1 = l1.applied(to: input, in: context)
+  return l2.applied(to: o1, in: context)
+}
+
+@differentiable(wrt: (input, l1, l2, l3))
+public func Sequential<L1: Layer, L2: Layer, L3: Layer>(in context: Context, from input: L1.Input, _ l1: L1, _ l2: L2, _ l3: L3) -> L3.Output where L1.Output == L2.Input, L2.Output == L3.Input {
+  let o1 = l1.applied(to: input, in: context)
+  let o2 = l2.applied(to: o1, in: context)
+  return l3.applied(to: o2, in: context)
+}
+
+@differentiable(wrt: (input, l1, l2, l3, l4))
+public func Sequential<L1: Layer, L2: Layer, L3: Layer, L4: Layer>(in context: Context, from input: L1.Input, _ l1: L1, _ l2: L2, _ l3: L3, _ l4: L4) -> L4.Output where L1.Output == L2.Input, L2.Output == L3.Input, L3.Output == L4.Input {
+  let o1 = l1.applied(to: input, in: context)
+  let o2 = l2.applied(to: o1, in: context)
+  let o3 = l3.applied(to: o2, in: context)
+  return l4.applied(to: o3, in: context)
+}
+
+@differentiable(wrt: (input, l1, l2, l3, l4, l5))
+public func Sequential<L1: Layer, L2: Layer, L3: Layer, L4: Layer, L5: Layer>(in context: Context, from input: L1.Input, _ l1: L1, _ l2: L2, _ l3: L3, _ l4: L4, _ l5: L5) -> L5.Output where L1.Output == L2.Input, L2.Output == L3.Input, L3.Output == L4.Input, L4.Output == L5.Input {
+  let o1 = l1.applied(to: input, in: context)
+  let o2 = l2.applied(to: o1, in: context)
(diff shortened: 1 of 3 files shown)


Sources/DeepLearning/Layer.swift
```diff
@@ -101,6 +101,37 @@ public extension Layer {
     }
 }
 
+@differentiable(wrt: (input, l1, l2))
+public func Sequential<L1: Layer, L2: Layer>(in context: Context, from input: L1.Input, _ l1: L1, _ l2: L2) -> L2.Output where L1.Output == L2.Input {
```
rxwei: Function names should always use camel case.

Sources/DeepLearning/Layer.swift
```diff
@@ -101,6 +101,80 @@ public extension Layer {
     }
 }
 
+extension Differentiable {
```
rxwei: Use `public extension` so you can drop `public`s from individual methods.

Sources/DeepLearning/Layer.swift
```diff
@@ -101,6 +101,80 @@ public extension Layer {
...
+
+    @differentiable(wrt: (self, l1, l2))
+    public func sequenced<L1: Layer, L2: Layer>(
+        in context: Context, through l1: L1, _ l2: L2)
+        -> L2.Output
```
rxwei: I was about to leave comments about indentation, but I thought "it's gonna be gyb'd anyway" :)

Sources/DeepLearning/Layer.swift
saeta: Thanks; fixed!

Sources/DeepLearning/Layer.swift
saeta: Yeah, I don't have my editor set up to do the indentation automatically for me. :-( Sorry!

Tests/DeepLearningTests/SequentialTests.swift
```diff
@@ -0,0 +1,48 @@
...
+            let 𝛁model = model.gradient { model -> Tensor<Float> in
+                let ŷ = model.applied(to: x, in: context)
+                return meanSquaredError(predicted: ŷ, expected: y)
+            }
+            optimizer.update(&model.allDifferentiableVariables, along: 𝛁model)
+        }
+        print(model.inferring(from: [[0, 0], [0, 1], [1, 0], [1, 1]]))
+    }
+
+     static var allTests = [
```
rxwei:
```suggestion
    static var allTests = [
```

Tests/DeepLearningTests/SequentialTests.swift
```diff
@@ -0,0 +1,48 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import XCTest
+@testable import DeepLearning
+
+final class SequentialTests: XCTestCase {
+    func testSequential() {
+
```
rxwei: Remove extra line? We don't really have a "leave a leading line" convention.

Tests/DeepLearningTests/SequentialTests.swift
```diff
@@ -0,0 +1,48 @@
...
+            }
+            optimizer.update(&model.allDifferentiableVariables, along: 𝛁model)
+        }
+        print(model.inferring(from: [[0, 0], [0, 1], [1, 0], [1, 1]]))
+    }
+
+     static var allTests = [
+        ("testSequential", testSequential)
+    ]
+}
```
rxwei: Add a trailing line.

Sources/DeepLearning/Layer.swift
```diff
@@ -101,6 +101,80 @@ public extension Layer {
     }
 }
 
+public extension Differentiable {
+
```
rxwei: Same about the leading line here.

Tests/DeepLearningTests/SequentialTests.swift
saeta: Fixed!

Tests/DeepLearningTests/SequentialTests.swift
saeta: Fixed.

Tests/DeepLearningTests/SequentialTests.swift
saeta: D'oh. Yeah, my editor is not properly configured. Sorry for all the formatting gaffes!

Sources/DeepLearning/Layer.swift
saeta: Removed. Thanks!

PR #21, opened 2019-02-22
Add default padding values.

saeta:
In the interest of progressive disclosure of complexity, and good default
values, set the padding on the layer types to be `.valid`. This follows
the Keras convention as well (https://keras.io/layers/convolutional/).


rxwei:
I think you missed AvgPool.

saeta:
Good catch. Thanks; fixed!

PR #23, opened 2019-02-22
Add a trainer abstraction to provide a default training loop.

saeta:
Because most supervised models have similar training loops, to help users get
started more easily, this commit adds a sketch for what a pre-built training
loop might look like. This will be used to help flesh out the API surface
based on real usage. The API should not be assumed to be frozen!

> Note: the implementation of the training loop should be completely re-written
> to be callback driven. It also should have reasonable defaults for outputing
> information to the console, and should have direct support for high
> performance input pipelines. These are considered out of scope for the
> initial check-in, in order to facilitate parallel development.


Sources/DeepLearning/Trainer.swift
```diff
@@ -0,0 +1,47 @@
...
+          Loss: FloatingPoint,
+          Loss == Loss.CotangentVector {
+    // TODO: Convert to protocol to take advantage of argument labels to disambiguate.
+    typealias LossFn = @differentiable (Model.Output, Model.Output) -> Loss
+
+    var model: Model
+    var optimizer: Opt
+    var lossFn: LossFn
+
+    init(model: Model, with optimizer: Opt, loss: @escaping LossFn) {
```
rxwei: The `Optimizer` protocol is not yet a class protocol, so this semantically copies the optimizer. Ideally we want a move here, but before we have the ownership model we should probably change the `Optimizer` protocol to just be a class protocol (i.e. refining `AnyObject`).

Sources/DeepLearning/Trainer.swift
```diff
@@ -0,0 +1,47 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// SupervisedTrainer implements a standard training loop for a model.
+public struct SupervisedTrainer<Model, Opt: Optimizer, Loss: Differentiable>
+    where Opt.Model == Model,
+          Loss: FloatingPoint,
+          Loss == Loss.CotangentVector {
+    // TODO: Convert to protocol to take advantage of argument labels to disambiguate.
```
rxwei: I don't think that's nice though. It kind of complicates the first-class concept just for argument labels, then we would need to add a bunch of conforming types that really just represent a function.

You can add argument disambiguation for function types here, FWIW:
```swift
@differentiable (_ predicted: Model.Output, _ expected: Model.Output) -> Loss
```

Sources/DeepLearning/Trainer.swift
```diff
@@ -0,0 +1,47 @@
...
+    typealias LossFn = @differentiable (Model.Output, Model.Output) -> Loss
+
+    var model: Model
+    var optimizer: Opt
```
rxwei: If `Optimizer` is guaranteed to be a class then this can be `let`.

Sources/DeepLearning/Trainer.swift
```diff
@@ -0,0 +1,47 @@
...
+    var lossFn: LossFn
+
+    init(model: Model, with optimizer: Opt, loss: @escaping LossFn) {
+        self.model = model
+        self.optimizer = optimizer
+        self.lossFn = loss
+    }
+
+    mutating func fit(x: Model.Input, y: Model.Output, forSteps numSteps: Int) {
```
rxwei: It would make the current "copying model at the training initializer" semantics (which is confusing) go away if you make model an inout argument taken in the `fit` function.

Sources/DeepLearning/Trainer.swift
saeta: I agree with you; I've removed the TODO (it'll show up the next time I push) and will resolve this conversation.

Sources/DeepLearning/Trainer.swift
saeta: Right; I didn't particularly like having the model be copied into the trainer (because the local variable in the user's program is untrained, but would be very easy to use by accident.

Does it instead make sense for this to be a freestanding function (or better yet a type method to avoid polluting the global namespace)?

rxwei: I think defining this as a freestanding function is actually a good idea.

Sources/DeepLearning/Trainer.swift
```diff
@@ -13,26 +13,18 @@
...
-        self.lossFn = loss
-    }
-
-    mutating func fit(x: Model.Input, y: Model.Output, forSteps numSteps: Int) {
+public struct SupervisedTrainer {
+    static func fit<Model, Opt: Optimizer, Loss: Differentiable>(
+        model: inout Model,
+        with optimizer: Opt,
+        loss lossFn: @escaping @differentiable (Model.Output, Model.Output) -> Loss,
+        x: Model.Input,
```
rxwei: Rename argument label to "input".

Sources/DeepLearning/Trainer.swift
```diff
@@ -13,26 +13,18 @@
...
+        y: Model.Output,
```
rxwei: Rename argument label to "output".

Sources/DeepLearning/Trainer.swift
```diff
@@ -13,26 +13,18 @@
...
+        stepCount numSteps: Int)
```
rxwei: "num..." is not idiomatic even if it's an argument name or local variable name. Argument names are being exported in API interface and shown in code completion. You can just use `stepCount`.

Sources/DeepLearning/Trainer.swift
```diff
@@ -13,26 +13,18 @@
...
-        self.model = model
-        self.optimizer = optimizer
-        self.lossFn = loss
-    }
-
-    mutating func fit(x: Model.Input, y: Model.Output, forSteps numSteps: Int) {
+public struct SupervisedTrainer {
+    static func fit<Model, Opt: Optimizer, Loss: Differentiable>(
+        model: inout Model,
+        with optimizer: Opt,
```
rxwei: The "with:" label is not super idiomatic here and is generally not used unless it's absolutely necessary. I suggest renaming it to "using:".

Sources/DeepLearning/Trainer.swift
```diff
@@ -13,26 +13,18 @@
...
-    var lossFn: LossFn
-
-    init(model: Model, with optimizer: Opt, loss: @escaping LossFn) {
-        self.model = model
-        self.optimizer = optimizer
-        self.lossFn = loss
-    }
-
-    mutating func fit(x: Model.Input, y: Model.Output, forSteps numSteps: Int) {
+public struct SupervisedTrainer {
```
rxwei: Define a `private init() {}` so that the default initializer won't be used anywhere. Or make this an `open class` so that this can be subclassed and extended. The `fit` function should be `open` in that case too.

Sources/DeepLearning/Trainer.swift
```diff
@@ -13,26 +13,18 @@
...
+    static func fit<Model, Opt: Optimizer, Loss: Differentiable>(
+        model: inout Model,
```
rxwei: Omit needless words. Argument labels should not merely repeat type information. When this argument is a direct object forming a phrase with the function base name and when it has clear type information, omit the argument label.

Sources/DeepLearning/Trainer.swift
```diff
@@ -13,26 +13,18 @@
...
+        with optimizer: Opt,
+        loss lossFn: @escaping @differentiable (Model.Output, Model.Output) -> Loss,
```
rxwei: "lossFn" -> "lossFunction"

Sources/DeepLearning/Trainer.swift
```diff
@@ -13,26 +13,18 @@
...
+        x: Model.Input,
+        y: Model.Output,
+        stepCount numSteps: Int)
+        where Opt.Model == Model, Loss: FloatingPoint, Loss == Loss.CotangentVector {
         // TODO: Rewrite training loop with callbacks.
         let context = Context(learningPhase: .training)
-
+
```
rxwei: Remove trailing spaces.

Sources/DeepLearning/Trainer.swift
```diff
@@ -13,26 +13,18 @@
...
-    mutating func fit(x: Model.Input, y: Model.Output, forSteps numSteps: Int) {
+public struct SupervisedTrainer {
+    static func fit<Model, Opt: Optimizer, Loss: Differentiable>(
+        model: inout Model,
+        with optimizer: Opt,
+        loss lossFn: @escaping @differentiable (Model.Output, Model.Output) -> Loss,
+        x: Model.Input,
+        y: Model.Output,
+        stepCount numSteps: Int)
+        where Opt.Model == Model, Loss: FloatingPoint, Loss == Loss.CotangentVector {
```
rxwei: If you constrain `Loss` to `FloatingPoint`, this will not work with loss functions that we have in swift-apis. The result should be a tensor.
Remove `Loss: FloatingPoint, Loss == Loss.CotangentVector`.

Sources/DeepLearning/Trainer.swift
```diff
@@ -13,26 +13,18 @@
...
-
-    init(model: Model, with optimizer: Opt, loss: @escaping LossFn) {
-        self.model = model
-        self.optimizer = optimizer
-        self.lossFn = loss
-    }
-
-    mutating func fit(x: Model.Input, y: Model.Output, forSteps numSteps: Int) {
+public struct SupervisedTrainer {
+    static func fit<Model, Opt: Optimizer, Loss: Differentiable>(
```
rxwei: Replace `Loss: Differentiable` with `LossScalar: TensorFlowFloatingPoint`. This is the differentiable type that does not require any extra constraints.

Sources/DeepLearning/Trainer.swift
rxwei:
```suggestion
        loss lossFn: @escaping @differentiable (Model.Output, Model.Output) -> Tensor<LossScalar>,
```

Sources/DeepLearning/Optimizer.swift
```diff
@@ -16,11 +16,11 @@
...
 #endif
 
-public protocol Optimizer {
+public protocol Optimizer: AnyObject {
     associatedtype Model: Layer
     associatedtype Scalar: FloatingPoint
     var learningRate: Scalar { get }
-    mutating func update(_ variables: inout Model.AllDifferentiableVariables,
+    func update(_ variables: inout Model.AllDifferentiableVariables,
                          along direction: Model.CotangentVector)
```
rxwei: Fix indentation.

Sources/DeepLearning/Trainer.swift
saeta: Fixed, thanks! (Marking all related comments as resolved.)

Sources/DeepLearning/Optimizer.swift
saeta: Thanks; fixed!

Sources/DeepLearning/Trainer.swift
saeta: Fixed.

Sources/DeepLearning/Trainer.swift
saeta: Marking as resolved, as the code has been completely restructured.

Sources/DeepLearning/Trainer.swift
saeta: Refactored; resolving now.

Sources/DeepLearning/Trainer.swift
saeta: Fixed.

Sources/DeepLearning/Trainer.swift
saeta: Fixed.

Sources/DeepLearning/Trainer.swift
saeta: Fixed.

Sources/DeepLearning/Trainer.swift
saeta: Fixed.

Sources/DeepLearning/Trainer.swift
```diff
@@ -14,20 +14,20 @@
...
         stepCount numSteps: Int)
-        where Opt.Model == Model {
+        where Opt.Model.AllDifferentiableVariables == Opt.Model.CotangentVector {
         // TODO: Rewrite training loop with callbacks.
         let context = Context(learningPhase: .training)
 
         // TODO: Implement shuffling, randomization, etc!
         for _ in 0..<numSteps {
-            let (_, 𝛁model) = model.valueWithGradient { model -> LossScalar in
+            let (𝛁model, _) = model.gradient(at: y) { model, y -> LossScalar in
```
rxwei: Why do you need to differentiate w.r.t. `y`?

Sources/DeepLearning/Trainer.swift
saeta: Since I don't think it makes sense to override the fit function, I've left it as a public struct, and added a private initializer.

Sources/DeepLearning/Trainer.swift
saeta: This is to work around a bug in autodiff. I'll try and put together a small example & file a bug. CC @dan-zheng who came up with this workaround.

saeta:
Note: about to add a test.

Shashi456:
@saeta I'd like to continue work on this, Could I?

rxwei:
@Shashi456 Thanks a lot for offering to help! Please hold off on this because the library abstractions we rely on are not ready yet. We’ll file a GitHub issue with a more concrete and future-proof design later when the underlying library is more complete later this year.

Shashi456:
@rxwei Alright, I'll continue with #54 and #127 then.

PR #24, opened 2019-02-23
Add Sigmoid Cross Entropy Loss

tanmayb123:
a.k.a. "binary_crossentropy" from Keras

### DIFF
--- a/Sources/DeepLearning/Loss.swift
+++ b/Sources/DeepLearning/Loss.swift
@@ -19,8 +19,8 @@ import TensorFlow
 /// Computes the Mean Squared Error loss between logits and labels
 ///
 /// - Parameters
-///     - logits: one-hot encoded outputs from a neural network.
-///     - labels: one-hot encoded values that correspond to the correct output.
+///   - logits: one-hot encoded outputs from a neural network.
+///   - labels: one-hot encoded values that correspond to the correct output.
 @differentiable
 public func meanSquaredError<Scalar: TensorFlowFloatingPoint>(
     predicted: Tensor<Scalar>, expected: Tensor<Scalar>) -> Tensor<Scalar> {
@@ -30,8 +30,8 @@ public func meanSquaredError<Scalar: TensorFlowFloatingPoint>(
 /// Computes the Softmax Cross Entropy (Categorical Cross Entropy) loss between logits and labels
 ///
 /// - Parameters
-///     - logits: one-hot encoded outputs from a neural network.
-///     - labels: one-hot encoded values that correspond to the correct output.
+///   - logits: one-hot encoded outputs from a neural network.
+///   - labels: one-hot encoded values that correspond to the correct output.
 @differentiable
 public func softmaxCrossEntropy<Scalar: TensorFlowFloatingPoint>(
     logits: Tensor<Scalar>, labels: Tensor<Scalar>) -> Tensor<Scalar> {
@@ -41,8 +41,8 @@ public func softmaxCrossEntropy<Scalar: TensorFlowFloatingPoint>(
 /// Computes the Sigmoid Cross Entropy (Binary Cross Entropy) loss between logits and labels
 ///
 /// - Parameters
-///     - logits: single continuous values from 0 to 1.
-///     - labels: integer values that correspond to the correct output.
+///   - logits: single continuous values from 0 to 1.
+///   - labels: integer values that correspond to the correct output.
 @differentiable
 public func sigmoidCrossEntropy<Scalar: TensorFlowFloatingPoint>(
     logits: Tensor<Scalar>, labels: Tensor<Scalar>) -> Tensor<Scalar> {
--- a/Sources/DeepLearning/Loss.swift
+++ b/Sources/DeepLearning/Loss.swift
@@ -46,7 +46,7 @@ public func softmaxCrossEntropy<Scalar: TensorFlowFloatingPoint>(
 @differentiable
 public func sigmoidCrossEntropy<Scalar: TensorFlowFloatingPoint>(
     logits: Tensor<Scalar>, labels: Tensor<Scalar>
-    ) -> Tensor<Scalar> {
+) -> Tensor<Scalar> {
     let loss = labels * log(logits) +
         (Tensor<Scalar>(ones: labels.shape) - labels) *
         log(Tensor<Scalar>(ones: logits.shape) - logits)
--- a/Sources/DeepLearning/Loss.swift
+++ b/Sources/DeepLearning/Loss.swift
@@ -16,18 +16,33 @@
 import TensorFlow
 #endif
 
+/// Computes the Mean Squared Error loss between logits and labels
+///
+/// - Parameters
+///     - logits: one-hot encoded outputs from a neural network.
+///     - labels: one-hot encoded values that correspond to the correct output.
 @differentiable
 public func meanSquaredError<Scalar: TensorFlowFloatingPoint>(
     predicted: Tensor<Scalar>, expected: Tensor<Scalar>) -> Tensor<Scalar> {
     return (expected - predicted).squared().mean()
 }
 
+/// Computes the Softmax Cross Entropy (Categorical Cross Entropy) loss between logits and labels
+///
+/// - Parameters
+///     - logits: one-hot encoded outputs from a neural network.
+///     - labels: one-hot encoded values that correspond to the correct output.
 @differentiable
 public func softmaxCrossEntropy<Scalar: TensorFlowFloatingPoint>(
     logits: Tensor<Scalar>, labels: Tensor<Scalar>) -> Tensor<Scalar> {
     return -(labels * logSoftmax(logits)).mean(alongAxes: 0).sum()
 }
 
+/// Computes the Sigmoid Cross Entropy (Binary Cross Entropy) loss between logits and labels
+///
+/// - Parameters
+///     - logits: single continuous values from 0 to 1.
+///     - labels: integer values that correspond to the correct output.
 @differentiable
 public func sigmoidCrossEntropy<Scalar: TensorFlowFloatingPoint>(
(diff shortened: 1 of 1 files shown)


tanmayb123:
I signed it!

Sources/DeepLearning/Loss.swift
```diff
@@ -27,3 +27,12 @@ public func softmaxCrossEntropy<Scalar: TensorFlowFloatingPoint>(
     logits: Tensor<Scalar>, labels: Tensor<Scalar>) -> Tensor<Scalar> {
     return -(labels * logSoftmax(logits)).mean(alongAxes: 0).sum()
 }
+
```
saeta: We haven't been doing this well yet, but could you add a Swift documentation comment (`///`-prefixed) describing this function? Maybe something along the lines of:

```
/// Computes the sigmoid cross entropy between logits and labels.
///
/// Logits are one-hot encoded outputs from a neural network. Labels are integer values that correspond to the correct output.
```

or something along those lines?

tanmayb123: Alright, will do.

tanmayb123:
Thanks @saeta :)

I've added documentation comments for each loss function.

Sources/DeepLearning/Loss.swift
```diff
@@ -16,14 +16,38 @@
 import TensorFlow
 #endif
 
+/// Computes the Mean Squared Error loss between logits and labels
+///
+/// - Parameters
+///     - logits: one-hot encoded outputs from a neural network.
```
dan-zheng: Please indent individual parameter doc comments by 2, to match the existing style.
```diff
 /// - Parameters
-///     - logits: one-hot encoded outputs from a neural network.
+///   - logits: one-hot encoded outputs from a neural network.
```
Thank you!

tanmayb123: Done

Sources/DeepLearning/Loss.swift
```diff
@@ -16,14 +16,38 @@
 import TensorFlow
 #endif
 
+/// Computes the Mean Squared Error loss between logits and labels
```
dan-zheng: Please end each doc comment with a period.
```suggestion
/// Computes the mean squared error between logits and labels.
```

Also, I think you can drop the word "loss".
Example from [TensorFlow Python docs](https://www.tensorflow.org/api_docs/python/tf/nn/softmax_cross_entropy_with_logits_v2):
>  Computes softmax cross entropy between logits and labels.

Thank you!

tanmayb123: Done

Sources/DeepLearning/Loss.swift
```diff
@@ -16,14 +16,38 @@
 import TensorFlow
 #endif
 
+/// Computes the mean squared error between logits and labels.
+///
+/// - Parameters
```
```suggestion
/// - Parameters:
```

Sources/DeepLearning/Loss.swift
```diff
@@ -16,14 +16,38 @@
...
+///   - labels: one-hot encoded values that correspond to the correct output.
 @differentiable
 public func meanSquaredError<Scalar: TensorFlowFloatingPoint>(
     predicted: Tensor<Scalar>, expected: Tensor<Scalar>) -> Tensor<Scalar> {
     return (expected - predicted).squared().mean()
 }
 
+/// Computes the softmax cross entropy (categorical cross entropy) between logits and labels.
+///
+/// - Parameters
```
```suggestion
/// - Parameters:
```

Sources/DeepLearning/Loss.swift
```diff
@@ -16,14 +16,38 @@
...
+///   - labels: one-hot encoded values that correspond to the correct output.
 @differentiable
 public func softmaxCrossEntropy<Scalar: TensorFlowFloatingPoint>(
     logits: Tensor<Scalar>, labels: Tensor<Scalar>) -> Tensor<Scalar> {
     return -(labels * logSoftmax(logits)).mean(alongAxes: 0).sum()
 }
+
+/// Computes the sigmoid cross entropy (binary cross entropy) between logits and labels.
+///
+/// - Parameters
```
```suggestion
/// - Parameters:
```

Sources/DeepLearning/Loss.swift
```diff
@@ -16,14 +16,38 @@
...
+///   - logits: single continuous values from 0 to 1.
+///   - labels: integer values that correspond to the correct output.
+@differentiable
+public func sigmoidCrossEntropy<Scalar: TensorFlowFloatingPoint>(
+    logits: Tensor<Scalar>, labels: Tensor<Scalar>) -> Tensor<Scalar> {
+    let loss = labels * log(logits) +
+               (Tensor<Scalar>(ones: labels.shape) - labels) *
```
```suggestion
         (Tensor<Scalar>(ones: labels.shape) - labels) *
```

Sources/DeepLearning/Loss.swift
```diff
@@ -16,14 +16,38 @@
...
+               log(Tensor<Scalar>(ones: logits.shape) - logits)
```
```suggestion
        log(Tensor<Scalar>(ones: logits.shape) - logits)
```

Sources/DeepLearning/Loss.swift
```diff
@@ -16,14 +16,38 @@
...
 @differentiable
 public func softmaxCrossEntropy<Scalar: TensorFlowFloatingPoint>(
     logits: Tensor<Scalar>, labels: Tensor<Scalar>) -> Tensor<Scalar> {
     return -(labels * logSoftmax(logits)).mean(alongAxes: 0).sum()
 }
+
+/// Computes the sigmoid cross entropy (binary cross entropy) between logits and labels.
+///
+/// - Parameters
+///   - logits: single continuous values from 0 to 1.
```
rxwei: Start every parameter description with an uppercase letter.

Sources/DeepLearning/Loss.swift
```diff
@@ -16,14 +16,38 @@
...
+///   - labels: integer values that correspond to the correct output.
+@differentiable
+public func sigmoidCrossEntropy<Scalar: TensorFlowFloatingPoint>(
+    logits: Tensor<Scalar>, labels: Tensor<Scalar>) -> Tensor<Scalar> {
```
rxwei: Move `) -> Tensor<Scalar> {` to the next line.

See line wrapping guidelines here: https://google.github.io/swift/#line-wrapping.

Sources/DeepLearning/Loss.swift
rxwei: And wrap backticks around numeric values (`0`, `1`).

Sources/DeepLearning/Loss.swift
rxwei: The call to `init(ones:)` is unnecessary and inefficient. Since `-` is broadcasting, yon can just use `1 -`. The whole statement can be simplified to:
```swift
let loss = labels * log(logits) + (1 - labels) * log(1 - logits)
```

tanmayb123: I did try that at first, but got this error:

```
error: repl.swift:23:2: error: function is not differentiable
@differentiable
~^~~~~~~~~~~~~~

repl.swift:24:13: note: when differentiating this function definition
            ^

repl.swift:26:42: note: function call is not differentiable because generic requirements are not met
    let loss = labels * log(logits) + (1 - labels) * log(1 - logits)
                                         ^


```

rxwei: Oh, in that case it's definitely a high-priority bug.  Could you file a bug on bugs.swift.org for this as well?

rxwei: Could you also try:
```swift
let loss = labels * log(logits) + (Scalar(1) - labels) * log(Scalar(1) - logits)
```
?

Sources/DeepLearning/Loss.swift
tanmayb123: Done - thanks!

Sources/DeepLearning/Loss.swift
tanmayb123: Same error (had to split it up into chunks because `the compiler is unable to type-check this expression in reasonable time`):

```
error: repl.swift:23:2: error: function is not differentiable
@differentiable
~^~~~~~~~~~~~~~

repl.swift:24:13: note: when differentiating this function definition
            ^

repl.swift:28:29: note: function call is not differentiable because generic requirements are not met
        let p2 = (Scalar(1) - labels)
                            ^


```

tanmayb123: Filed the bug report:

Sources/DeepLearning/Loss.swift
```diff
@@ -16,14 +16,39 @@
...
+
+/// Computes the sigmoid cross entropy (binary cross entropy) between logits and labels.
+///
+/// - Parameters:
+///   - logits: Single continuous values from `0` to `1`.
+///   - labels: Integer values that correspond to the correct output.
+@differentiable
+public func sigmoidCrossEntropy<Scalar: TensorFlowFloatingPoint>(
+    logits: Tensor<Scalar>, labels: Tensor<Scalar>
+    ) -> Tensor<Scalar> {
```
```suggestion
) -> Tensor<Scalar> {
```

Sources/DeepLearning/Loss.swift
tanmayb123: This does work:

```
@differentiable
    logits: Tensor<Scalar>, labels: Tensor<Scalar>
) -> Tensor<Scalar> {
    let loss = labels * log(logits) +
        (Tensor<Scalar>(1) - labels) *
        log(Tensor<Scalar>(1) - logits)
    return -loss.mean(alongAxes: 0).sum()
}
```

rxwei: This looks good, please use that for now. You can move `log(Tensor<Scalar>(1) - logits)` to the end of the second line of the statement. We are enforcing a horizontal limit of 100 columns.

tanmayb123: Done. Thanks!

rxwei:
Thanks Tanmay!

PR #26, opened 2019-02-23
Add default (identity)  activations to the Dense layer initializers.

saeta:
This change follows along with the defaulted values for Conv and other layers.


rxwei:
> (Side note: [related issue](https://bugs.swift.org/browse/TF-243) about identity activation and debuggability of `gradient` tensor seed shape mismatch.)

You can check these pretty easily by inserting a `precondition` or `assert` in each primitive VJP's pullback to make sure the seed matches the shape of the output. However, that would cause the pullback to capture the original result, which is bad for memory usage. The proper way is to make the differentiation transform insert these checks based on the optimization level.

PR #29, opened 2019-02-24
Update README sample code to reflect #28.

rxwei:
Should go in after #28 gets merged.


rxwei:
Temporarily closed because #28 is still problematic.

PR #31, opened 2019-02-25
Add Triplet loss function

tanmayb123:


tanmayb123:
One question: what should I call this? Is `triplet` just fine? Maybe `tripletLoss`?

eaplatanios:
I feel it's nice if we can add citations to the papers that introduced the various loss functions, layers, etc., whenever they're not one of the super standard/common ones.

eaplatanios:
@tanmayb123 Could you add a citation to [this paper](https://arxiv.org/abs/1503.03832) and also add the functional form of the loss in the documentation string?

tanmayb123:
Sure thing, but @rxwei had decided some time ago _not_ to include loss functions that aren't available in Keras, but this PR stayed open. What should I do?

rxwei:
Let's add it for now!

Sources/DeepLearning/Loss.swift
```diff
@@ -53,3 +53,21 @@ public func sigmoidCrossEntropy<Scalar: TensorFlowFloatingPoint>(
         (Tensor<Scalar>(1) - labels) * log(Tensor<Scalar>(1) - logits)
     return -loss.mean(alongAxes: 0).sum()
 }
+
+/// Computes the triplet loss.
```
```suggestion
/// Returns the triplet loss.
```

rxwei: Also add the citation as Anthony suggested.

saeta:
Hi @tanmayb123 ! It looks like this PR is coming from an unknown repository. As a result, we'll close this out for now. But if you'd like to try and address the comments and add a test, I think we'd love to get this merged in. 

All the best,
-Brennan

Kshitij09:
@tanmayb123 should I borrow your code and submit a new PR with requested changes? I was planning to submit a PR on your fork but seems like the branch having this code missing.

PR #33, opened 2019-02-26
Remove explicit differentiation parameters. They are no longer required!

rxwei:
We've recently changed the type checker to improve the ergonomics of the `@differentiable` attribute.
* https://github.com/apple/swift/pull/22915: Explicit differentiation parameters are no longer required in a `@differentiable` attribute when the function has some arguments that do not conform to `Differentiable`. Those non-differentiable parameters will be skipped and the rest will be differentiated with respect to.
* https://github.com/apple/swift/pull/22877: On an instance method, when a `wrt:` is not specified, `self` is being implicitly included as a differentiation parameter.
* https://github.com/apple/swift/pull/22877: When a `@differentiable` requirement is not met, the `@differentiable` attribute fix-it will appear exactly as written in the original declaration instead of the most complex, canonical form. For instance, `@differentiable` instead of `@differentiable(wrt: (x))`.

This greatly simplifies libraries and applications that use automatic differentiation. The protocol requirement `Layer.applied(to:in:)` becomes as simple as this:
```swift
@differentiable
func applied(to input: Input, in context: Context) -> Output
```

This PR updates deep learning APIs to use the simplest form of `@differentiable` possible. Hooray!

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ struct Model: Layer {
     var layer2 = Dense<Float>(inputSize: hiddenSize, outputSize: hiddenSize, activation: relu)
     var layer3 = Dense<Float>(inputSize: hiddenSize, outputSize: 3, activation: identity)
     
-    @differentiable(wrt: (self, input))
+    @differentiable
     func applied(to input: Tensor<Float>, in context: Context) -> Tensor<Float> {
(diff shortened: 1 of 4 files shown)


lattner:
Yay, thank you!

PR #34, opened 2019-02-26
Add public memberwise initializers to layers.

### LINKED ISSUE #30: Add public memberwise initializers for layers.

rxwei:
Add a memberwise initializer to every layer. Resolves #30.

Notes:
1. In some layers, some stored properties are not public, we should consider making them public.
2. All initializers that deal with integers take `Int`, which is good. However, some layers expose a public `Int32` stored property. This is inconsistent with the initializers and is a leaky abstraction that should be fixed. One potential solution is to rename the `Int32` property, make it private, and define a computed property that has a getter (and setter, if necessary) that mutates the private property with a conversion. The other potential solution is to use `Int32` everywhere. Neither is pretty.


rxwei:
> How about marking the initializers with `@differentiable`?

[TF-305](https://bugs.swift.org/browse/TF-305), thanks to your suggestions...

PR #36, opened 2019-02-27
add old optimizer initializers back

marcrasi:
Per @rxwei's suggestion in https://github.com/tensorflow/swift-apis/pull/28, adds the old initializers back, to maintain backwards-compatibility with a bunch of existing code and notebooks.

### DIFF
--- a/Sources/DeepLearning/Optimizer.swift
+++ b/Sources/DeepLearning/Optimizer.swift
@@ -39,9 +39,7 @@ public class Adam<Model: Layer, Scalar: TensorFlowFloatingPoint>: Optimizer
         beta1: Scalar = 0.9,
         beta2: Scalar = 0.999,
         epsilon: Scalar = 1e-8,
-        decay: Scalar = 0,
-        modelType: Model.Type = Model.self,
-        scalarType: Scalar.Type = Scalar.self
+        decay: Scalar = 0
     ) {
         precondition(learningRate >= 0, "Learning rate must be non-negative")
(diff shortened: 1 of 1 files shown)


Sources/DeepLearning/Optimizer.swift
```diff
@@ -35,13 +35,13 @@ public class Adam<Model: Layer, Scalar: TensorFlowFloatingPoint>: Optimizer
...
 
     public init(
-        for _: __shared Model,
         learningRate: Scalar = 1e-3,
         beta1: Scalar = 0.9,
         beta2: Scalar = 0.999,
         epsilon: Scalar = 1e-8,
         decay: Scalar = 0,
-        scalarType: Scalar.Type
+        modelType: Model.Type = Model.self,
```
rxwei: You can remove these two metatype args from the old optimizer initializers.

rxwei:
> As a result, in general, I am biased towards doing whatever is going to make clean APIs for the future, and also make things easier for us to maintain / evolve going forward.

If you meant to propose making the `init(for:...)` the only API available, I don't think I would agree with that. Taking a dummy instance for type inference is not idiomatic in Swift (no precedents) and taking a metatype where it can take a generic type parameter is even more unswifty. Quite the opposite, I think `SGD<Model, Float>()` is the cleanest API out there and it's sufficiently Swifty. We added `init(for:...)` just for ML programmers to ramp up and discover the API quickly.

PR #40, opened 2019-03-01
add index-based softmax cross entropy loss

marcrasi:
As per discussion in https://groups.google.com/a/tensorflow.org/d/msg/swift/ASQJJkjf-xU/7bl6HOvdBQAJ


marcrasi:
Oops, the implementation that I copied was not generic and I did not notice that. Will update it to be generic.

marcrasi:
Updated to use generic Scalar.

PR #45, opened 2019-03-05
Change implementation of softmax cross entropy loss 

xiejw:
This PR changed two things. 

1. Used a numerical stable implementation for softmax cross entropy.
2. Changed the gradient wr.t. Now the loss only generates the gradient w.r.t. logits. This is consistent with its variant (Int label version).

In addition, added numerical tests to verify the correctness.

### DIFF
--- a/Sources/DeepLearning/Loss.swift
+++ b/Sources/DeepLearning/Loss.swift
@@ -53,11 +53,20 @@ func _vjpSoftmaxCrossEntropy<Scalar: TensorFlowFloatingPoint>(
 /// - Parameters:
 ///   - logits: One-hot encoded outputs from a neural network.
 ///   - oneHotLabels: One-hot encoded values that correspond to the correct output.
-@differentiable
+@differentiable(wrt: logits, vjp: _vjpSoftmaxCrossEntropy)
 public func softmaxCrossEntropy<Scalar: TensorFlowFloatingPoint>(
     logits: Tensor<Scalar>, oneHotLabels: Tensor<Scalar>
 ) -> Tensor<Scalar> {
-    return -(oneHotLabels * logSoftmax(logits)).mean(alongAxes: 0).sum()
+    return Raw.softmaxCrossEntropyWithLogits(features: logits, labels: oneHotLabels).loss.mean()
+}
+
+@usableFromInline
+func _vjpSoftmaxCrossEntropy<Scalar: TensorFlowFloatingPoint>(
+    logits: Tensor<Scalar>, oneHotLabels: Tensor<Scalar>
+) -> (Tensor<Scalar>, (Tensor<Scalar>) -> Tensor<Scalar>) {
+    let (loss, grad) = Raw.softmaxCrossEntropyWithLogits(features: logits, labels: oneHotLabels)
+    let batchSize = Tensor<Scalar>(logits.shapeTensor[0])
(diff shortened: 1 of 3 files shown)


Sources/DeepLearning/Loss.swift
```diff
@@ -53,11 +53,20 @@ func _vjpSoftmaxCrossEntropy<Scalar: TensorFlowFloatingPoint>(
...
+    return Raw.softmaxCrossEntropyWithLogits(features: logits, labels: oneHotLabels).loss.mean()
+}
+
+@usableFromInline
+func _vjpSoftmaxCrossEntropy<Scalar: TensorFlowFloatingPoint>(
+    logits: Tensor<Scalar>, oneHotLabels: Tensor<Scalar>
+) -> (Tensor<Scalar>, (Tensor<Scalar>) -> Tensor<Scalar>) {
+    let (loss, grad) = Raw.softmaxCrossEntropyWithLogits(features: logits, labels: oneHotLabels)
+    let batchSize = Tensor<Scalar>(logits.shapeTensor[0])
+    return (loss.mean(), { v in (v / batchSize) * grad })
```
rxwei: Unnecessary parens?
```suggestion
    return (loss.mean(), { v in v / batchSize * grad })
```

Tests/DeepLearningTests/LossTests.swift
```diff
@@ -0,0 +1,72 @@
...
+            at: logits,
+            in: { softmaxCrossEntropy(logits: $0, oneHotLabels: labels) })
+        assertAllClose(expected: expectedGradients, got: gradients)
+    }
+
+    func assertAllClose(expected a: Tensor<Float>, got b: Tensor<Float>, tol: Float = 1e-6) {
+        XCTAssertEqual(a.shape, b.shape)
+        for (index, elementInA) in a.scalars.enumerated() {
+            let elementInB = b.scalars[index]
+            XCTAssertTrue(
```
rxwei: Have you considered using `XCTAssertLessThan`?

Tests/DeepLearningTests/LossTests.swift
```diff
@@ -0,0 +1,72 @@
...
+
+        // As the loss is mean loss, we should scale the golden loss numbers.
+        let expectedGradients = expectedGradientsBeforeMean / Float(logits.shape[0])
+        let gradients = gradient(
+            at: logits,
+            in: { softmaxCrossEntropy(logits: $0, oneHotLabels: labels) })
+        assertAllClose(expected: expectedGradients, got: gradients)
+    }
+
+    func assertAllClose(expected a: Tensor<Float>, got b: Tensor<Float>, tol: Float = 1e-6) {
```
rxwei: Sorry for the nitpick on the test API.
* Consider spelling out argument labels for clarity.
* How about calling it `assertElementsEqual`? The `tolerance:` parameter already indicates it's an approximation.
* It's probably longer than 100 columns after my suggested edit.

```suggestion
    func assertAllClose(expected: Tensor<Float>, actual: Tensor<Float>, tolerance: Float = 1e-6) {
```

Tests/DeepLearningTests/LossTests.swift
xiejw: Done. Sounds better.

PR #46, opened 2019-03-05
[Doc] Document optimizers.

dan-zheng:
Add API docs for optimizers. Comments are mostly adapted from Keras.

Todo: Add details about optimizer computation/initialization.

### DIFF
--- a/Sources/DeepLearning/Operators.swift
+++ b/Sources/DeepLearning/Operators.swift
@@ -16,7 +16,7 @@
 import TensorFlow
 #endif
 
-/// Returns the values of the specified tensor to the nearest integer, element-wise.
+/// Returns the values of the specified tensor rounded to the nearest integer, element-wise.
 public func round<Scalar: BinaryFloatingPoint>(_ x: Tensor<Scalar>) -> Tensor<Scalar> {
     return Raw.round(x)
 }
--- a/Sources/DeepLearning/Operators.swift
+++ b/Sources/DeepLearning/Operators.swift
@@ -16,12 +16,12 @@
 import TensorFlow
 #endif
 
-/// Rounds the values of a tensor to the nearest integer, element-wise.
+/// Returns the values of the specified tensor to the nearest integer, element-wise.
 public func round<Scalar: BinaryFloatingPoint>(_ x: Tensor<Scalar>) -> Tensor<Scalar> {
     return Raw.round(x)
 }
 
-/// Returns a tensor with the same shape and contents as the input.
+/// Returns a tensor with the same shape and scalars as specified tensor.
 @differentiable
 public func identity<Scalar>(_ x: Tensor<Scalar>) -> Tensor<Scalar> {
     return x
--- a/Sources/DeepLearning/Operators.swift
+++ b/Sources/DeepLearning/Operators.swift
@@ -16,12 +16,12 @@
 import TensorFlow
 #endif
 
-// Rounds the values of a tensor to the nearest integer, element-wise.
+/// Rounds the values of a tensor to the nearest integer, element-wise.
 public func round<Scalar: BinaryFloatingPoint>(_ x: Tensor<Scalar>) -> Tensor<Scalar> {
     return Raw.round(x)
 }
 
-// Return a tensor with the same shape and contents as input.
+/// Returns a tensor with the same shape and contents as the input.
 @differentiable
 public func identity<Scalar>(_ x: Tensor<Scalar>) -> Tensor<Scalar> {
     return x
--- a/Sources/DeepLearning/Optimizer.swift
+++ b/Sources/DeepLearning/Optimizer.swift
@@ -37,8 +37,7 @@ public protocol Optimizer {
 
 /// Adam optimizer.
 ///
-/// References:
-/// - [Adam - A Method for Stochastic Optimization](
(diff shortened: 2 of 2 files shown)


Sources/DeepLearning/Operators.swift
```diff
@@ -16,12 +16,12 @@
 import TensorFlow
 #endif
 
-// Rounds the values of a tensor to the nearest integer, element-wise.
+/// Rounds the values of a tensor to the nearest integer, element-wise.
```
rxwei: If you start with "rounds", it sounds like a mutating function, like [`FloatingPoint.round()`](https://developer.apple.com/documentation/swift/floatingpoint/2297801-round).

Instead, the comment should be like [`FloatingPoint.rounded()`](https://developer.apple.com/documentation/swift/floatingpoint/2295900-rounded).

Sources/DeepLearning/Optimizer.swift
```diff
@@ -91,11 +112,24 @@ public class Adam<Model: Layer, Scalar: TensorFlowFloatingPoint>: Optimizer
     }
 }
 
+/// RMSProp optimizer.
```
rxwei:
```suggestion
/// The RMSProp optimizer.
```

Sources/DeepLearning/Optimizer.swift
```diff
@@ -91,11 +112,24 @@ public class Adam<Model: Layer, Scalar: TensorFlowFloatingPoint>: Optimizer
...
+///
+/// It is recommended to leave the parameters of this optimizer at their default values (except the
+/// learning rate, which can be freely tuned). This optimizer is usually a good choice for recurrent
+/// neural networks.
+///
+/// References:
+/// - [rmsprop: Divide the gradient by a running average of its recent magnitude](
```
rxwei: Is there a "references" syntax?

Sources/DeepLearning/Optimizer.swift
dan-zheng: I deliberately chose not to add an article, actually.
I think `The xxx optimizer` is not quite appropriate because there can exist multiple distinct instances of `RMSProp`.

I would prefer `RMSProp optimizer` over `An optimizer that implements the RMSProp algorithm` over `An RMSProp optimizer` over `The RMSProp optimizer`.

Sources/DeepLearning/Optimizer.swift
dan-zheng: There isn't. It ends up looking like this:
image

There isn't an appropriate comment type, here's the full list: https://github.com/apple/swift/blob/master/include/swift/Markup/SimpleFields.def

Let me know if you'd like any changes, otherwise I'll keep things as is.

Sources/DeepLearning/Operators.swift
dan-zheng: Thanks! Changed documentation to `Returns the values of the specified tensor to the nearest integer, element-wise.` in https://github.com/tensorflow/swift-apis/pull/46/commits/668c9be214a67683bc79ea8ac2fd577c5114bc32.

Sources/DeepLearning/Optimizer.swift
rxwei: The issue I worry about is that the bullet is not being indented under "References:" in your rendered text.

Sources/DeepLearning/Optimizer.swift
rxwei: Okay.

Sources/DeepLearning/Operators.swift
```diff
@@ -16,12 +16,12 @@
 import TensorFlow
 #endif
 
-// Rounds the values of a tensor to the nearest integer, element-wise.
+/// Returns the values of the specified tensor to the nearest integer, element-wise.
```
rxwei:
```suggestion
/// Returns the values of the specified tensor rounded to the nearest integer, element-wise.
```

Sources/DeepLearning/Optimizer.swift
dan-zheng: Yeah, that's unfortunate. Indenting the hyphen further right doesn't help.

Sources/DeepLearning/Operators.swift
dan-zheng: Oops, thanks! Fixed in https://github.com/tensorflow/swift-apis/pull/46/commits/e4985e91fbade933588d1f685e4fa4321f3e303b.

Sources/DeepLearning/Optimizer.swift
rxwei: Can you drop the hyphen and put the reference right after the colon? Also make "references" be singular.

dan-zheng: Done in https://github.com/tensorflow/swift-apis/pull/46/commits/14de85f78cf2977697e1b470b9becc4b084eabc4.

- https://tensorflow.devsite.corp.google.com/swift/api_docs/Classes#adam
image

- https://tensorflow.devsite.corp.google.com/swift/api_docs/Classes/Adam
image

rxwei: I'm not adding this bullet. Any convention to follow here? Seems it's best to keep it flat.

PR #47, opened 2019-03-05
Document all layers.

rxwei:
Add doc comments to all layers and layer-related utilities.

### DIFF
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -108,7 +108,7 @@ public extension Differentiable {
     ///
     /// - Parameter:
     ///   - context: The context that stores contextual information used for the application of
-    /// layers.
+    ///     layers.
     ///   - l1: The first layer.
     ///   - l2: The second layer.
     /// - Returns: The final layer's output after sequential application.
@@ -127,7 +127,7 @@ public extension Differentiable {
     ///
     /// - Parameter:
     ///   - context: The context that stores contextual information used for the application of
-    /// layers.
+    ///     layers.
     ///   - l1: The first layer.
     ///   - l2: The second layer.
     ///   - l3: The third layer.
@@ -149,7 +149,7 @@ public extension Differentiable {
     ///
     /// - Parameter:
     ///   - context: The context that stores contextual information used for the application of
-    /// layers.
+    ///     layers.
     ///   - l1: The first layer.
     ///   - l2: The second layer.
     ///   - l3: The third layer.
@@ -174,7 +174,7 @@ public extension Differentiable {
     ///
     /// - Parameter:
     ///   - context: The context that stores contextual information used for the application of
-    /// layers.
+    ///     layers.
     ///   - l1: The first layer.
     ///   - l2: The second layer.
     ///   - l3: The third layer.
@@ -202,7 +202,7 @@ public extension Differentiable {
     ///
     /// - Parameter:
     ///   - context: The context that stores contextual information used for the application of
-    /// layers.
+    ///     layers.
     ///   - l1: The first layer.
     ///   - l2: The second layer.
     ///   - l3: The third layer.
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -58,7 +58,7 @@ public protocol Layer: Differentiable & KeyPathIterable
 
     /// Returns the output obtained from applying the layer to the given input.
     ///
-    /// - Parameters
+    /// - Parameters:
     ///   - input: The input to the layer.
     ///   - context: The contextual informance for the layer application, e.g. the current learning
     ///     phase.
@@ -265,7 +265,7 @@ public struct Dense<Scalar: TensorFlowFloatingPoint>: Layer {
 
     /// Returns the output obtained from applying the layer to the given input.
     ///
-    /// - Parameters
+    /// - Parameters:
     ///   - input: The input to the layer.
     ///   - context: The contextual informance for the layer application, e.g. the current learning
     ///     phase.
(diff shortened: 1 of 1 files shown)


Sources/DeepLearning/Layer.swift
```diff
@@ -102,8 +102,16 @@ public extension Layer {
     }
 }
 
-/// Adds helpers for standard feed-forward, sequential models.
 public extension Differentiable {
+    /// Returns the output computed by applying a sequence of layers to the previous layer's output,
+    /// except that the first layer's input is `self`.
+    ///
+    /// - Parameter:
```
dan-zheng:
```suggestion
    /// - Parameters:
```
Same elsewhere.

Sources/DeepLearning/Layer.swift
```diff
@@ -554,6 +721,13 @@ public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
         )
     }
 
+    /// Returns the output obtained from applying the layer to the given input.
+    ///
+    /// - Parameters
```
dan-zheng:
```suggestion
    /// - Parameters:
```

Sources/DeepLearning/Layer.swift
```diff
@@ -507,19 +652,34 @@ public struct AvgPool2D<Scalar: TensorFlowFloatingPoint>: Layer {
...
+    ///   - padding: The padding.
     public init(poolSize: (Int, Int), strides: (Int, Int), padding: Padding = .valid) {
         self.poolSize = (1, Int32(poolSize.0), Int32(poolSize.1), 1)
         self.strides = (1, Int32(strides.0), Int32(strides.1), 1)
         self.padding = padding
     }
 
+    /// Returns the output obtained from applying the layer to the given input.
+    ///
+    /// - Parameters
```
dan-zheng:
```suggestion
    /// - Parameters:
```

Sources/DeepLearning/Layer.swift
```diff
@@ -618,14 +807,25 @@ public struct Dropout<Scalar: TensorFlowFloatingPoint>: Layer {
...
+    /// Creates an upsampling layer.
+    ///
+    /// - Parameter size: The upsampling factor for rows and columns.
     public init(size: Int32) {
        self.size = size
     }
 
+    /// Returns the output obtained from applying the layer to the given input.
+    ///
+    /// - Parameters
```
dan-zheng:
```suggestion
    /// - Parameters:
```

Sources/DeepLearning/Layer.swift
```diff
@@ -636,10 +836,21 @@ public struct UpSampling2D<Scalar: TensorFlowFloatingPoint>: Layer {
...
+///
+/// A flatten layer flattens the input when applied without affecting the batch size.
 @_fixed_layout
 public struct Flatten<Scalar: TensorFlowFloatingPoint>: Layer {
+    /// Creates a flatten layer.
     public init() {}
 
+    /// Returns the output obtained from applying the layer to the given input.
+    ///
+    /// - Parameters
```
dan-zheng:
```suggestion
    /// - Parameters:
```

Sources/DeepLearning/Layer.swift
```diff
@@ -648,21 +859,36 @@ public struct Flatten<Scalar: TensorFlowFloatingPoint>: Layer {
...
+    /// Creates a reshape layer.
+    ///
+    /// - Parameter shape: The target shape.
     public init(_ shape: TensorShape) {
         self.init(shape: Tensor(shape.dimensions))
     }
 
+    /// Returns the output obtained from applying the layer to the given input.
+    ///
+    /// - Parameters
```
dan-zheng:
```suggestion
    /// - Parameters:
```

Sources/DeepLearning/Layer.swift
```diff
@@ -209,6 +263,13 @@ public struct Dense<Scalar: TensorFlowFloatingPoint>: Layer {
         self.activation = activation
     }
 
+    /// Returns the output obtained from applying the layer to the given input.
+    ///
+    /// - Parameters
```
dan-zheng:
```suggestion
    /// - Parameters:
```

Sources/DeepLearning/Layer.swift
```diff
@@ -257,7 +332,10 @@ public extension Dense {
     }
 }
 
-/// A convolutional neural network layer.
+/// A 2D convolution layer (e.g. spatial convolution over images).
```
dan-zheng: Do you prefer `2-D` over `2D`?
I prefer `2-D`.

Sources/DeepLearning/Layer.swift
```diff
@@ -257,7 +332,10 @@ public extension Dense {
...
+///
+/// This layer creates a convolution kernel that is convolved with the layer input to produce a
+/// tensor of outputs. If activation is not `nil`, it is applied to the outputs as well.
```
dan-zheng: How about:
```suggestion
/// tensor of outputs. If activation is not `nil`, it is applied to the outputs.
```

rxwei: This is copied from Keras. The whole sentence should be deleted because we do not have `Optional` activation functions yet.

Sources/DeepLearning/Layer.swift
rxwei: Of course `2-D`.

Sources/DeepLearning/Layer.swift
rxwei: Thanks for the catch. These are copied from the `applied(to:)` protocol requirement comments, which are not from this patch.

Sources/DeepLearning/Layer.swift
```diff
@@ -373,7 +482,16 @@ public struct BatchNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
 
-    /// The batch dimension.
+    /// Creates a batch normalization layer.
+    ///
+    /// - Parameters:
+    ///   - axis: The axis that should be normalized (typically the features axis).
+    ///   - momentum: The momentum for the moving average.
+    ///   - offset: The offset to be added to the normalized tensor.
+    ///   - scale: The scale to multiply the normalized tensor by.
+    ///   - epsilon: The small scalar added to variance to avoid dividing by zero.
```
dan-zheng: In https://github.com/tensorflow/swift-apis/pull/46, I used the following doc comment for `epsilon`:
`A small scalar added to the denominator to improve numerical stability.`

What do you think?

rxwei: Okay. Changed.

PR #50, opened 2019-03-05
Improved mean squared error. 

xiejw:
This PR has the following changes.
1. Fixed the docstring for the meanSquaredError. The original version is likely due to copy-paste.
2. Restricted the wrt of gradient to make it consistent with other APIs.
3. Added numerical tests for loss and gradients.

### DIFF
--- a/Sources/DeepLearning/Loss.swift
+++ b/Sources/DeepLearning/Loss.swift
@@ -52,20 +52,21 @@ func _vjpSoftmaxCrossEntropy<Scalar: TensorFlowFloatingPoint>(
 /// Computes the softmax cross entropy (categorical cross entropy) between logits and labels.
 ///
 /// - Parameters:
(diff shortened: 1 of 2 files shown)


Sources/DeepLearning/Loss.swift
```diff
@@ -16,12 +16,12 @@
...
 
-/// Computes the mean squared error between logits and labels.
+/// Computes the mean squared error between predictions and labels.
 ///
 /// - Parameters:
-///   - logits: One-hot encoded outputs from a neural network.
-///   - labels: One-hot encoded values that correspond to the correct output.
-@differentiable
+///   - predicted: Predicted outputs from a neural network.
+///   - labels: Expected values, i.e., targets, that correspond to the correct output.
```
```suggestion
///   - labels: Expected values, i.e. targets, that correspond to the correct output.
```

PR #51, opened 2019-03-08
Fix the sigmoid cross entropy with logits.

xiejw:
Fix the sigmoid cross entropy with logits. 

Particularly,
1. logits should be wrapped into a sigmoid. The original impl actually assumes the logits are outputs of sigmoid.
2. The mean loss should be reduced by number of elements not batch size. 
3. Use a numerical stable implementation to avoid Nan.

### DIFF
--- a/Sources/DeepLearning/Loss.swift
+++ b/Sources/DeepLearning/Loss.swift
@@ -88,6 +88,6 @@ public func sigmoidCrossEntropy<Scalar: TensorFlowFloatingPoint>(
     let maxLogitsWithZero = max(logits, Tensor(0))
     // TODO(TF-349): Use abs directly once `abs` is differentiable.
     let absOfLogitsbsl = max(logits, -logits)
-    let loss =  maxLogitsWithZero - logits * labels + log(1 + exp(-absOfLogitsbsl))
+    let loss = maxLogitsWithZero - logits * labels + log(1 + exp(-absOfLogitsbsl))
     return loss.mean()
 }
--- a/Sources/DeepLearning/Loss.swift
+++ b/Sources/DeepLearning/Loss.swift
@@ -83,7 +83,7 @@ public func sigmoidCrossEntropy<Scalar: TensorFlowFloatingPoint>(
     // This numerical stable implementation is based on tf.nn.sigmoid_cross_entropy_with_logits.
 
     let maxLogitsWithZero = max(logits, Tensor(0))
-    // TODO(SR-10061): Use abs directly once abc is differentiable.
+    // TODO(TF-349): Use abs directly once abc is differentiable.
     let absOfLogitsbsl = max(logits, -logits)
     let loss =  maxLogitsWithZero - logits * labels + log(1 + exp(-absOfLogitsbsl))
     return loss.mean()
--- a/Sources/DeepLearning/Loss.swift
+++ b/Sources/DeepLearning/Loss.swift
@@ -73,6 +73,9 @@ func _vjpSoftmaxCrossEntropy<Scalar: TensorFlowFloatingPoint>(
 
 /// Computes the sigmoid cross entropy (binary cross entropy) between logits and labels.
 ///
+/// The reduction is reduced over all elements. If reduced over batch size is intended, please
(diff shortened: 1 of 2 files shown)


Sources/DeepLearning/Loss.swift
```diff
@@ -74,13 +74,17 @@ func _vjpSoftmaxCrossEntropy<Scalar: TensorFlowFloatingPoint>(
 /// Computes the sigmoid cross entropy (binary cross entropy) between logits and labels.
 ///
 /// - Parameters:
-///   - logits: Single continuous values from `0` to `1`.
+///   - logits: Unscaled value from a neural network.
 ///   - labels: Integer values that correspond to the correct output.
-@differentiable
+@differentiable(wrt: logits)
 public func sigmoidCrossEntropy<Scalar: TensorFlowFloatingPoint>(
     logits: Tensor<Scalar>, labels: Tensor<Scalar>
```
saeta: Should there be 2 generic types, so that `labels` can be a Tensor of a non-`TensorFlowFloatingPoint` type?

Sources/DeepLearning/Loss.swift
```diff
@@ -74,13 +74,17 @@ func _vjpSoftmaxCrossEntropy<Scalar: TensorFlowFloatingPoint>(
...
 ) -> Tensor<Scalar> {
-    let loss = labels * log(logits) +
-        (Tensor<Scalar>(1) - labels) * log(Tensor<Scalar>(1) - logits)
-    return -loss.mean(alongAxes: 0).sum()
+    // This numerical stable implementation is based on tf.nn.sigmoid_cross_entropy_with_logits.
+
+    let maxLogitsWithZero = max(logits, Tensor(0))
+    // TODO(SR-10061): Use abs directly once abc is differentiable.
```
saeta: `s/SR-10061/TF-349/` (I moved the bug you filed to the right project: https://bugs.swift.org/browse/TF-349)

```suggestion
    // TODO(SR-10061): Use abs directly once `abs` is differentiable.
```

Sources/DeepLearning/Loss.swift
```diff
@@ -74,13 +74,17 @@ func _vjpSoftmaxCrossEntropy<Scalar: TensorFlowFloatingPoint>(
 /// Computes the sigmoid cross entropy (binary cross entropy) between logits and labels.
 ///
 /// - Parameters:
-///   - logits: Single continuous values from `0` to `1`.
+///   - logits: Unscaled value from a neural network.
```
```suggestion
///   - logits: The unscaled output of a neural network.
```

Sources/DeepLearning/Loss.swift
rxwei: I don't think so. Labels need to be multiplied with logits (see line 88).

rxwei: I don't think so. Labels need to be multiplied with logits (see line 88).

lattner:
Very nice Jianwei!

Sources/DeepLearning/Loss.swift
xiejw: Done.

Sources/DeepLearning/Loss.swift
```diff
@@ -74,13 +74,17 @@ func _vjpSoftmaxCrossEntropy<Scalar: TensorFlowFloatingPoint>(
...
 ) -> Tensor<Scalar> {
-    let loss = labels * log(logits) +
-        (Tensor<Scalar>(1) - labels) * log(Tensor<Scalar>(1) - logits)
-    return -loss.mean(alongAxes: 0).sum()
+    // This numerical stable implementation is based on tf.nn.sigmoid_cross_entropy_with_logits.
+
+    let maxLogitsWithZero = max(logits, Tensor(0))
+    // TODO(SR-10061): Use abs directly once `abs` is differentiable.
+    let absOfLogitsbsl = max(logits, -logits)
+    let loss =  maxLogitsWithZero - logits * labels + log(1 + exp(-absOfLogitsbsl))
```
```suggestion
    let loss = maxLogitsWithZero - logits * labels + log(1 + exp(-absOfLogitsbsl))
```

PR #55, opened 2019-03-12
Added the MNIST dataset for Swift for TensorFlow Datasets Library

pruthiviraj71:
Suggest any changes if needed.


Sources/DeepLearning/MNISTdataset.swift
```diff
@@ -0,0 +1,91 @@
...
+let trainImages = "train-images-idx3-ubyte"
+let trainLabels = "train-labels-idx1-ubyte"
+let testImages = "t10k-images-idx3-ubyte"
+let testLabels = "t10k-label-idx1-ubyte"
+let imageSize = 28
+let trainCount = 60000
+let testCount = 10000
+
+// Function to download and extract from online
+func download_and_extract(){
```
rxwei: To ensure consistency and readability, please use camel case for all names in Swift, including function names and variable names.

All our library code is following the [Google Swift Style](https://google.github.io/swift) with the exception that indentation is 4 spaces long. It’d be great if you take a look!

Sources/DeepLearning/MNISTdataset.swift
```diff
@@ -0,0 +1,91 @@
...
+    return (
+        trainImages: Tensor(shape: [trainrowCount, columnCount], scalars: trainImages) / 255,
+        trainLabels: Tensor(trainLabels),
+        testImages: Tensor(shape: [testrowCount, columnCount], scalars: testImages) / 255,
+        testLabels: Tensor(testLabels)
+    )
+}
+
+download_and_extract()
+let (trainx, trainy, testx, testy) = readMNIST(trainImages: trainImages, trainLabels: trainLabels, testImages: testImages, testLabels: testLabels)
```
saeta: Newline at the end of the file.

saeta: I think you'll likely want to remove these last couple lines from the file, in order to avoid downloading and extracting the MNIST dataset whenever anyone uses Swift for TensorFlow. Does that make sense?

Sources/DeepLearning/MNISTdataset.swift
```diff
@@ -0,0 +1,91 @@
...
+    print("Reading data.")
+    let trainImages = readFile(trainImages).dropFirst(16).map { Float($0) }
+    let trainLabels = readFile(trainLabels).dropFirst(8).map { Int32($0) }
+    let testImages = readFile(testImages).dropFirst(16).map { Float($0) }
+    let testLabels = readFile(testLabels).dropFirst(8).map { Int32($0) }
+        
+    let trainrowCount = Int32(trainLabels.count)
+    let testrowCount = Int32(testLabels.count)
+    let columnCount = Int32(testImages.count) / testrowCount
+    print(trainImages.count)
```
saeta: Could you remove debugging prints?

Sources/DeepLearning/MNISTdataset.swift
```diff
@@ -0,0 +1,91 @@
+import Python
+import TensorFlow
+
+let np = Python.import("numpy")
```
saeta: Do you think you could do a native-swift implementation, instead of falling back to Python?

Sources/DeepLearning/MNISTdataset.swift
```diff
@@ -0,0 +1,91 @@
...
+    }
+    
+    for extractedFile in extractedFiles {
+        
+        if !Bool(path.isfile(extractedFile))! {
+            print("Extracting \(extractedFile)")
+            f_in = gzip.open(helperFiles[counter], "rb")
+            f_out = Python.open(extractedFiles[counter], "wb")
+            sh.copyfileobj(f_in, f_out)
+            print("EXtraction Completed: \(extractedFile)")
```
saeta: Looks like you have a typo here.

Sources/DeepLearning/MNISTdataset.swift
```diff
@@ -0,0 +1,91 @@
...
+            print("File \(extractedFile): already exists")
+        }
+        counter = counter + 1
+    }
+    
+}
+
+// Reads a file into an array of bytes.
+func readFile(_ filename: String) -> [UInt8] {
+    let d = Python.open(filename, "rb").read()
```
saeta: Please use more descriptive variable names. :-)

Sources/DeepLearning/MNISTdataset.swift
```diff
@@ -0,0 +1,91 @@
+import Python
+import TensorFlow
+
+let np = Python.import("numpy")
+let path = Python.import("os.path")
+let urllib = Python.import("urllib.request")
+let gzip = Python.import("gzip")
+let sh = Python.import("shutil")
+
+let URL = "http://yann.lecun.com/exdb/mnist/"
```
saeta: To avoid polluting the namespace, could you define these constants within an object?

Sources/DeepLearning/MNISTdataset.swift
pruthiviraj71: I am not sure if there is any native function in Foundation library to decompress gz files which is the file hosted in MNIST official URL. I can do a pure swift implementation for opening the files and converting the file into swift array. Just that I may have to use a third party framework for extraction process.

Sources/DeepLearning/MNISTdataset.swift
pruthiviraj71: Hey @saeta , I have made all changes as you requested.

pruthiviraj71:
1. Script is now placed under Scripts Directory in root as suggested by @rxwei .
2. Using camel case for all variable and function names as per the [Google Swift Style guide](https://google.github.io/swift/).
3. Removed debugging printing statements and fixed few typos.
4. Added a dictionary MNIST constants to store constants to prevent namespace polluting as suggested by @saeta .

Scripts/MNISTdataset.swift
```diff
@@ -0,0 +1,97 @@
+/*
```
marcrasi: I suggest putting the usage comments on the function declarations instead of at the top level of the file (https://google.github.io/swift/#file-comments).

Also, I have a few more comment formatting requests from https://google.github.io/swift/#documentation-comments:
* Use "///" for documentation comments instead of "/* ... */"
* Use Apple's markup format (https://developer.apple.com/library/content/documentation/Xcode/Reference/xcode_markup_formatting_ref/). In particular, surround code examples with three backticks.

(same comment applies to a few other comments that you have before declarations in this file)

Scripts/MNISTdataset.swift
```diff
@@ -0,0 +1,97 @@
...
+import TensorFlow
+
+let np = Python.import("numpy")
+let path = Python.import("os.path")
+let urllib = Python.import("urllib.request")
+let gzip = Python.import("gzip")
+let sh = Python.import("shutil")
+
+// MNIST constants
+let constants: [String: String] = [
```
marcrasi: > Declaring an enum without cases is the canonical way to define a “namespace” to group a set of related declarations, such as constants or helper functions.

(swift style guide)

An advantage over this `[String: String]` approach is that the compiler will catch references to missing constants at compile time!

The enum should be `fileprivate`.

Scripts/MNISTdataset.swift
```diff
@@ -0,0 +1,97 @@
...
+let (trainX, trainY, testX, testY) = readMNIST(trainImages: constants["trainImages"]!, 
+                                                   trainLabels: constants["trainLabels"]!, 
+                                                   testImages: constants["testImages"]!, 
+                                                   testLabels: constants["testLabels"]!)
+*/
+
+import Python
+import TensorFlow
+
+let np = Python.import("numpy")
```
marcrasi: I think that `fileprivate` would be a good access level for all of these. If this code ever gets put in a module, these are probably not things that we want to expose to other files in the module.

Though I echo @saeta's point that it would be nice to do all of these using swift packages. You could probably add some dependencies to Package.swift.

Scripts/MNISTdataset.swift
```diff
@@ -0,0 +1,97 @@
...
+    "testImagesFile": "t10k-images-idx3-ubyte.gz",
+    "testLabelsFile": "t10k-labels-idx1-ubyte.gz",
+    "trainImages": "train-images-idx3-ubyte",
+    "trainLabels": "train-labels-idx1-ubyte",
+    "testImages": "t10k-images-idx3-ubyte",
+    "testLabels": "t10k-label-idx1-ubyte",
+]
+
+// Function to download and extract files from online
+func downloadAndExtract(){
```
marcrasi: If you make `readMNIST` call this automatically, then this can be `fileprivate`.

Put a space between the `)` and `{`.

Scripts/MNISTdataset.swift
```diff
@@ -0,0 +1,97 @@
...
+
```
marcrasi: We don't typically put blank lines at the beginning/end of code blocks. Only use blank lines to group statements into logical sections (https://google.github.io/swift/#vertical-whitespace).

Scripts/MNISTdataset.swift
```diff
@@ -0,0 +1,97 @@
...
+        if !Bool(path.isfile(extractedFile))! {
+            print("Extracting \(extractedFile)")
+            fin = gzip.open(helperFiles[counter], "rb")
+            fout = Python.open(extractedFiles[counter], "wb")
+            sh.copyfileobj(fin, fout)
+            print("Extraction Completed: \(extractedFile)")
+        } else {
+            print("File \(extractedFile): already exists")
+        }
+         counter = counter + 1
```
marcrasi: this has an extra space in front of it

Scripts/MNISTdataset.swift
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// Reads a file into an array of bytes.
+func readFile(_ filename: String) -> [UInt8] {
+    let buffer = Python.open(filename, "rb").read()
+    return Array(numpy: np.frombuffer(buffer, dtype: np.uint8))!
+}
+
+// Reads MNIST images and labels from specified file paths.
+func readMNIST(trainImages: String, trainLabels: String,
```
marcrasi: A good line wrapping (https://google.github.io/swift/#line-wrapping) would be:
```
func readMNIST(
    trainImages: String,
    trainLabels: String,
    testImages: String,
    testLabels: String
) -> (
    trainImages: Tensor<Float>,
    trainLabels: Tensor<Int32>,
￼    testImages: Tensor<Float>,
    testLabels: Tensor<Int32>
) {
```

Scripts/MNISTdataset.swift
```diff
@@ -0,0 +1,97 @@
...
+        } else {
+            print("File \(extractedFile): already exists")
+        }
+         counter = counter + 1
+    }
+
+}
+
+// Reads a file into an array of bytes.
+func readFile(_ filename: String) -> [UInt8] {
```
marcrasi: This can also be `fileprivate`

Scripts/MNISTdataset.swift
```diff
@@ -0,0 +1,97 @@
...
+    "trainImages": "train-images-idx3-ubyte",
+    "trainLabels": "train-labels-idx1-ubyte",
+    "testImages": "t10k-images-idx3-ubyte",
+    "testLabels": "t10k-label-idx1-ubyte",
+]
+
+// Function to download and extract files from online
+func downloadAndExtract(){
+    
+    let helperFiles = [constants["trainImagesFile"]!, constants["trainLabelsFile"]!,
```
marcrasi: I think these should be one per line, like in function call line wrapping: https://google.github.io/swift/#function-calls

marcrasi:
Also, I talked to @rxwei and we agreed that actually the best location for this would be a new directory `Sources/DeepLearningContrib`, because it's a new kind of feature that we haven't developed standard protocols for yet. Once standard protocols for data loaders get designed, data loaders in `Sources/DeepLearningContrib` can be updated to conform and then moved into `Sources/DeepLearning`.

Sorry for asking you to move around the file so many times!

pruthiviraj71:
Hey @marcrasi , I made some changes as suggested by you and @saeta to implement the script in native swift so I removed all python implementation and replaced it by swift alternative code. All that is left is downloading and extracting of gz files which I believe will be much easier with third party swift libraries or if you have any solution, I'll try implementing that.
And thanks for the code review!

Sources/DeepLearningContrib/MNISTdataset.swift
```diff
@@ -0,0 +1,86 @@
...
+                       constants.testLabelsFile]
+    let extractedFiles = [constants.trainImages,
+                          constants.trainLabels,
+                          constants.testImages,
+                          constants.testLabels]
+    var counter = 0
+    let path = FileManager.default.currentDirectoryPath
+    for helperFile in helperFiles {
+        let url = NSURL(fileURLWithPath: path)
+        if let pathComponent = url.appendingPathComponent(helperFile) {
```
marcrasi: With `URL`, this is non-optional, so it should not require an `if let`!

Also I think `pathComponent` should be named something like `helperFileLocalUrl`, to describe that it's the local system url where the file should be.

Sources/DeepLearningContrib/MNISTdataset.swift
```diff
@@ -0,0 +1,86 @@
...
+            if fileManager.fileExists(atPath: filePath) {
+                print("Not downloading \(helperFile): already exists")
+            } else {
+                print("Downloading \(helperFile)")
+                /// TODO: Add Swift Implementation for downloading files.
+                print("Downloaded \(helperFile)")
+            }
+        }    
+    }
+    for extractedFile in extractedFiles {
```
marcrasi: It might simplify the code to have a single loop that does downloading and extracting. With a single loop, you won't have to reconstruct intermediate variables like `helperFileLocalUrl` that you need in both steps.

Sources/DeepLearningContrib/MNISTdataset.swift
```diff
@@ -0,0 +1,86 @@
...
+            let filePath = pathComponent.path
+            let fileManager = FileManager.default
+            if fileManager.fileExists(atPath: filePath) {
+                print("File \(extractedFile): already exists")
+            } else {
+                print("Extracting \(extractedFile)")
+                /// TODO: Add Swift Implementation for extracting downloaded files.
+                print("Extraction Completed: \(extractedFile)")
+            }
+            counter = counter + 1
```
marcrasi: Doing `for (index, bla) in array.enumerated()` would be more canonical than manually incrementing a `counter`. Though if you're not using counter anywhere, you should just delete it.

Sources/DeepLearningContrib/MNISTdataset.swift
```diff
@@ -0,0 +1,86 @@
...
+    testImages: Tensor<Float>,
+    testLabels: Tensor<Int32>
+) {
+    downloadAndExtract()
+    print("Reading data.")
+    let trainImages = try! Data(contentsOf: URL(fileURLWithPath: constants.trainImages)).dropFirst(16).map { Float($0) }
+    let trainLabels = try! Data(contentsOf: URL(fileURLWithPath: constants.trainLabels)).dropFirst(8).map { Int32($0) }
+    let testImages = try! Data(contentsOf: URL(fileURLWithPath: constants.testImages)).dropFirst(16).map { Float($0) }
+    let testLabels = try! Data(contentsOf: URL(fileURLWithPath: constants.testLabels)).dropFirst(8).map { Int32($0) }
+    let trainrowCount = Int32(trainLabels.count)
```
marcrasi: `row` should be capitalized for camelCase (also on the next line)

pruthiviraj71:
Hey @marcrasi , I made few modifications as you requested and added gzip extraction process using the framework you referred to. Please provide feedback!

Sources/DeepLearningContrib/MNISTdataset.swift
```diff
@@ -0,0 +1,86 @@
+import Foundation
+import TensorFlow
+
+/// MNIST constants
+fileprivate enum constants {
+    static let URL = "http://yann.lecun.com/exdb/mnist/"
+    static let trainImagesFile = "train-images-idx3-ubyte.gz"
```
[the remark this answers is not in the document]
pruthiviraj71: Also @marcrasi , I am not sure if struct is better here because enum without cases does the work here. If you want, I can update it to use struct.

saeta:
Hi @pruthiviraj71 ! Thank you so much for this PR. I think for now, we'll close out this PR, as we have an alternate MNIST dataset implementation available as part of [swift-models](https://github.com/tensorflow/swift-models). If you'd like to discuss more of the design, please do let me know and we can schedule some time at an upcoming Swift design meetings! -Brennan

PR #60, opened 2019-03-21
Mark `Reshape._nonTrivial` with `@usableFromInline`.

dan-zheng:
Unblocks merging master.


rxwei:
It has to be internal

PR #66, opened 2019-03-30
Adding GlobalAveragePooling1D

Shashi456:
So I'm relatively new to the swift ecosystem, let me know if there are any errors. Afaik my understanding of GlobalAveragePooling in a single dimension is returning the mean * the no of batches if it's a multi-batch input.

### DIFF
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -1025,7 +1025,7 @@ public struct GlobalAveragePooling1D<Scalar: TensorFlowFloatingPoint>: Layer {
     /// - Returns: The output 
     @differentiable
     public func applied(to input: Tensor<Scalar>, in _: Context) -> Tensor<Scalar> {
-        return input.mean(alongAxes: axis)
+        return input.mean(alongAxes: 1).reshaped(to: [input.shape[0], input.shape[2]])
     }
 }
 
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -1013,7 +1013,7 @@ public struct AvgPool2D<Scalar: TensorFlowFloatingPoint>: Layer {
 /// A global average pooling layer for temporal data.
 @_fixed_layout
 public struct GlobalAveragePooling1D<Scalar: TensorFlowFloatingPoint>: Layer {
-    /// Creates a global average pooling layer in 1 dimension
+    /// Creates a global average pooling layer.
     public init() {}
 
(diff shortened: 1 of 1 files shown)


Sources/DeepLearning/Layer.swift
```diff
@@ -1009,12 +1009,32 @@ public struct AvgPool2D<Scalar: TensorFlowFloatingPoint>: Layer {
...
+    /// Returns the output obtained from applying the layer to the given input.
+    ///
+    /// - Parameters:
+    ///   - input: The input to the layer.
+    ///   - context: The contextual information for the layer application, e.g. the current learning
+    ///     phase.
+    /// - Returns: The output 
+    @differentiable
+    public func applied(to input: Tensor<Scalar>, in _: Context) -> Tensor<Scalar> {
+        return input.mean(alongAxes: axis)
```
tanmayb123:
```suggestion
        return input.mean(alongAxes: 1).reshaped(to: [input.shape[0], input.shape[2]])
```

tanmayb123:
```suggestion
        return input.mean(alongAxes: 1).reshaped(to: [input.shape[0], input.shape[2]])
```

tanmayb123: @Shashi456 the code should compile after you make this change. If the input has this shape:

[batchSize, timeSteps, channels]

The `mean(alongAxes:)` part will make it this shape:

[batchSize, 1, channels]

And then the reshape will make it this:

[batchSize, channels]

Shashi456: so you passed the index of time steps for the mean, which means that is where the 1D data is stored, so it computes the mean after which you have batchsize number of datapoints and hence reshaping it that way? 
Am i getting it right?

rxwei:
Did you sign the CLA?

Shashi456:
@rxwei yeah i have

Sources/DeepLearning/Layer.swift
```diff
@@ -1009,12 +1009,32 @@ public struct AvgPool2D<Scalar: TensorFlowFloatingPoint>: Layer {
     }
 }
 
+
+/// A global average pooling layer for temporal data.
+@_fixed_layout
+public struct GlobalAveragePooling1D<Scalar: TensorFlowFloatingPoint>: Layer {
+    /// Creates a global average pooling layer in 1 dimension
```
rxwei:
```suggestion
    /// Creates a global average pooling layer in 1 dimension.
```

Please end sentences with a period.

Issue #52, opened 2019-03-11
Implement Recurrent Layers


adam-r-kowalski:
It would be really nice to have a set of recurrent layers including LSTMs and GRUs as part of the core layers API. Ideally we can parameterize the activation functions for the various gates rather than hard coding in the tanh and sigmoid functions.

rxwei:
Definitely, they are on our roadmap. Thanks for opening an issue about this.

rxwei:
FWIW, here's an example RNN implemented using the Swift of TensorFlow Deep Learning Library: https://gist.github.com/rxwei/ce6644efad8f229651050e096c05ccbb.

Most people are often stuck at these two things on their first try:
* Defining a layer with multiple inputs and outputs.
    * Answer: Define a `struct Input : Differentiable { ... }` and `struct Output : Differentiable { ... }` in your layer with however many arity you need. As to variable-sized inputs/outputs, [`Array` will conform to `Differentiable`](https://github.com/apple/swift/pull/23183) soon.
* Backpropagation through time without differentiating control flow.
    * Answer: Control flow differentiation is not required for BPTT. Each `Layer` has a `Backpropagator` member type. Use `appliedForBackpropagation` to evaluate the forward pass and get a back propagator, and accumulate backpropagators in an array.

You can find how these are done in my gist. If anyone's interested in taking a stab and implement RNN cells, let me know!

rxwei:
Update: #80 added the `RNNCell` protocol and #71 added `SimpleRNN` and `LSTM`. We still need to add GRU and hopefully NTM.

dhasl002:
I'll take a shot at GRU and NTM

dhasl002:
Update: #417 has created to implement GRU cells. I intend to move on to NTM next.

PR #71, opened 2019-04-02
Add Recurrent Layers

tanmayb123:
#52

### DIFF
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -1329,7 +1329,7 @@ public struct SimpleRNNCell<Scalar: TensorFlowFloatingPoint>: RNNCell {
     /// - Returns: The hidden state.
     @differentiable
     public func applied(to input: Input) -> Output {
-        let concatenatedInput = input.stepInput.concatenated(with: input.state)
+        let concatenatedInput = input.input.concatenated(with: input.state)
         let newState = matmul(concatenatedInput, weight) + bias
         return Output(output: newState, state: newState)
     }
@@ -1391,7 +1391,7 @@ public struct LSTMCell<Scalar: TensorFlowFloatingPoint>: RNNCell {
     /// - Returns: The hidden state.
     @differentiable
     public func applied(to input: Input) -> Output {
-        let gateInput = input.stepInput.concatenated(with: input.state.hidden, alongAxis: 1)
+        let gateInput = input.input.concatenated(with: input.state.hidden, alongAxis: 1)
 
         let inputGate = sigmoid(matmul(gateInput, inputWeight))
         let updateGate = tanh(matmul(gateInput, updateWeight))
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -1297,12 +1297,12 @@ public struct SimpleRNNCell<Scalar: TensorFlowFloatingPoint>: RNNCell {
 
     @noDerivative var stateShape: TensorShape
 
-	@differentiable
-	public var zeroState: Tensor<Scalar> {
-		get {
-			return Tensor(zeros: stateShape)
-		}
-	}
+    @differentiable
+    public var zeroState: Tensor<Scalar> {
+        get {
+            return Tensor(zeros: stateShape)
+        }
+    }
 
     /// Creates a `SimpleRNNCell` with the specified input size and hidden state size.
     ///
@@ -1326,7 +1326,7 @@ public struct SimpleRNNCell<Scalar: TensorFlowFloatingPoint>: RNNCell {
     @differentiable
     public func applied(to input: RNNCellInput, in _: Context) -> RNNCellOutput {
         let concatenatedInput = input.stepInput.concatenated(with: input.state)
-		let newState = matmul(concatenatedInput, weight) + bias
+        let newState = matmul(concatenatedInput, weight) + bias
         return RNNCellOutput(output: newState, state: newState)
     }
 }
@@ -1341,12 +1341,12 @@ public struct LSTMCell<Scalar: TensorFlowFloatingPoint>: RNNCell {
 
     @noDerivative var stateShape: TensorShape
 
-	@differentiable
-	public var zeroState: Tensor<Scalar> {
-		get {
-			return State(cell: Tensor(zeros: stateShape), hidden: Tensor(zeros: stateShape))
-		}
-	}
+    @differentiable
+    public var zeroState: Tensor<Scalar> {
+        get {
+            return State(cell: Tensor(zeros: stateShape), hidden: Tensor(zeros: stateShape))
+        }
+    }
 
     /// Creates a `LSTMCell` with the specified input size and hidden state size.
     ///
@@ -1394,7 +1394,7 @@ public struct LSTMCell<Scalar: TensorFlowFloatingPoint>: RNNCell {
         let newCellState = (input.state.cell * forgetGate + inputGate * updateGate)
         let newHiddenState = tanh(newCellState) * outputGate
 
-	let newState = State(cell: newCellState, hidden: newHiddenState)
+        let newState = State(cell: newCellState, hidden: newHiddenState)
 
         return RNNCellOutput(output: newState, state: newState)
     }
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -1298,19 +1298,19 @@ public struct Reshape<Scalar: TensorFlowFloatingPoint>: Layer {
 
 struct LSTMCell<Scalar: TensorFlowFloatingPoint>: Layer {
 
-    var inputW: Tensor<Scalar>
-    var updateW: Tensor<Scalar>
-    var forgetW: Tensor<Scalar>
-    var forgetB: Tensor<Scalar>
-    var outputW: Tensor<Scalar>
+    var inputWeight: Tensor<Scalar>
+    var updateWeight: Tensor<Scalar>
+    var forgetWeight: Tensor<Scalar>
+    var forgetBias: Tensor<Scalar>
+    var outputWeight: Tensor<Scalar>
 
     init(inputSize: Int32, hiddenSize: Int32) {
         let concatenatedInputSize = inputSize + hiddenSize
-        self.inputW = Tensor<Scalar>(glorotUniform: [concatenatedInputSize, hiddenSize])
-        self.updateW = Tensor<Scalar>(glorotUniform: [concatenatedInputSize, hiddenSize])
-        self.forgetW = Tensor<Scalar>(glorotUniform: [concatenatedInputSize, hiddenSize])
-        self.forgetB = Tensor<Scalar>(zeros: [hiddenSize])
-        self.outputW = Tensor<Scalar>(glorotUniform: [concatenatedInputSize, hiddenSize])
+        self.inputWeight = Tensor<Scalar>(glorotUniform: [concatenatedInputSize, hiddenSize])
+        self.updateWeight = Tensor<Scalar>(glorotUniform: [concatenatedInputSize, hiddenSize])
+        self.forgetWeight = Tensor<Scalar>(glorotUniform: [concatenatedInputSize, hiddenSize])
+        self.forgetBias = Tensor<Scalar>(zeros: [hiddenSize])
+        self.outputWeight = Tensor<Scalar>(glorotUniform: [concatenatedInputSize, hiddenSize])
     }
 
     struct HiddenState: Differentiable {
@@ -1339,10 +1339,10 @@ struct LSTMCell<Scalar: TensorFlowFloatingPoint>: Layer {
     func applied(to input: Input, in _: Context) -> HiddenState {
         let gateInput = input.inputs.concatenated(with: input.hidden.hiddenState, alongAxis: 1)
 
-        let inputGate = sigmoid(matmul(gateInput, inputW))
-        let updateGate = tanh(matmul(gateInput, updateW))
-        let forgetGate = sigmoid(matmul(gateInput, forgetW) + forgetB)
-        let outputGate = sigmoid(matmul(gateInput, outputW))
+        let inputGate = sigmoid(matmul(gateInput, inputWeight))
+        let updateGate = tanh(matmul(gateInput, updateWeight))
+        let forgetGate = sigmoid(matmul(gateInput, forgetWeight) + forgetBias)
+        let outputGate = sigmoid(matmul(gateInput, outputWeight))
 
         let newCellState = (input.hidden.cellState * forgetGate + inputGate * updateGate)
         let newHiddenState = tanh(newCellState) * outputGate
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -1296,6 +1296,53 @@ public struct Reshape<Scalar: TensorFlowFloatingPoint>: Layer {
     }
 }
 
+/// A Simple RNN Cell.
+public struct SimpleRNNCell<Scalar: TensorFlowFloatingPoint>: Layer {
+    public var weight: Tensor<Scalar>
+    public var bias: Tensor<Scalar>
+    
+    @noDerivative var stateShape: TensorShape
+
+    /// Creates a `SimpleRNNCell` with the specified input size and hidden state size.
+    ///
+    /// - Parameters:
+    ///   - inputSize: The number of features in 2-D input tensors.
+    ///   - hiddenSize: The number of features in 2-D hidden states.
+    public init(inputSize: Int, hiddenSize: Int) {
+        let concatenatedInputSize = Int32(inputSize + hiddenSize)
+        self.weight = Tensor(glorotUniform: [concatenatedInputSize, Int32(hiddenSize)])
+        self.bias = Tensor(zeros: [Int32(hiddenSize)])
+        self.stateShape = TensorShape([1, Int32(hiddenSize)])
+    }
+
+    public struct Input: Differentiable {
+        public var stepInput: Tensor<Scalar>
+        public var state: Tensor<Scalar>
+
+        init(stepInput: Tensor<Scalar>, state: Tensor<Scalar>) {
+            self.stepInput = stepInput
+            self.state = state
+        }
+    }
+    
+    /// Returns the output obtained from applying the layer to the given input.
+    ///
+    /// - Parameters:
+    ///   - input: The input to the layer.
+    ///   - context: The contextual information for the layer application, e.g. the current learning
+    ///     phase.
+    /// - Returns: The hidden state.
+    @differentiable
+    public func applied(to input: Input, in _: Context) -> Tensor<Scalar> {
+        let concatenatedInput =  input.stepInput.concatenated(with: input.state)
+        return matmul(concatenatedInput, weight) + bias
+    }
+    
+    public func zeroState() -> Tensor<Scalar> {
+        return Tensor(zeros: stateShape)
+    }
+}
+
 /// An LSTM Cell.
 public struct LSTMCell<Scalar: TensorFlowFloatingPoint>: Layer {
     public var inputWeight: Tensor<Scalar>
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -1303,7 +1303,7 @@ public struct SimpleRNNCell<Scalar: TensorFlowFloatingPoint>: RNNCell {
     }
 
     public typealias State = Tensor<Scalar>
-    public typealias TimeStepInput = Tensor<Float>
+    public typealias TimeStepInput = Tensor<Scalar>
     public typealias TimeStepOutput = State
     public typealias Input = RNNCellInput<TimeStepInput, State>
     public typealias Output = RNNCellOutput<TimeStepOutput, State>
@@ -1314,10 +1314,10 @@ public struct SimpleRNNCell<Scalar: TensorFlowFloatingPoint>: RNNCell {
     ///   - inputSize: The number of features in 2-D input tensors.
     ///   - hiddenSize: The number of features in 2-D hidden states.
     public init(inputSize: Int, hiddenSize: Int) {
-        let concatenatedInputSize = Int32(inputSize + hiddenSize)
-        self.weight = Tensor(glorotUniform: [concatenatedInputSize, Int32(hiddenSize)])
-        self.bias = Tensor(zeros: [Int32(hiddenSize)])
-        self.stateShape = TensorShape([1, Int32(hiddenSize)])
+        let concatenatedInputSize = inputSize + hiddenSize
+        self.weight = Tensor(glorotUniform: [concatenatedInputSize, hiddenSize])
+        self.bias = Tensor(zeros: [hiddenSize])
+        self.stateShape = TensorShape([1, hiddenSize])
     }
 
     /// Returns the output obtained from applying the layer to the given input.
@@ -1361,12 +1361,12 @@ public struct LSTMCell<Scalar: TensorFlowFloatingPoint>: RNNCell {
     ///   - inputSize: The number of features in 2-D input tensors.
     ///   - hiddenSize: The number of features in 2-D hidden states.
     public init(inputSize: Int, hiddenSize: Int) {
-        let concatenatedInputSize = Int32(inputSize + hiddenSize)
-        let gateWeightShape = TensorShape([concatenatedInputSize, Int32(hiddenSize)])
+        let concatenatedInputSize = inputSize + hiddenSize
+        let gateWeightShape = TensorShape([concatenatedInputSize, hiddenSize])
         self.inputWeight = Tensor(glorotUniform: gateWeightShape)
         self.updateWeight = Tensor(glorotUniform: gateWeightShape)
         self.forgetWeight = Tensor(glorotUniform: gateWeightShape)
-        self.forgetBias = Tensor(zeros: [Int32(hiddenSize)])
+        self.forgetBias = Tensor(zeros: [hiddenSize])
         self.outputWeight = Tensor(glorotUniform: gateWeightShape)
         self.stateShape = TensorShape([1, concatenatedInputSize])
     }
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -1303,8 +1303,10 @@ public struct SimpleRNNCell<Scalar: TensorFlowFloatingPoint>: RNNCell {
     }
 
     public typealias State = Tensor<Scalar>
-    public typealias Input = RNNCellInput<Tensor<Scalar>, State>
-    public typealias Output = RNNCellOutput<State, State>
+    public typealias TimeStepInput = Tensor<Float>
+    public typealias TimeStepOutput = State
+    public typealias Input = RNNCellInput<TimeStepInput, State>
+    public typealias Output = RNNCellOutput<TimeStepOutput, State>
 
     /// Creates a `SimpleRNNCell` with the specified input size and hidden state size.
     ///
@@ -1348,8 +1350,10 @@ public struct LSTMCell<Scalar: TensorFlowFloatingPoint>: RNNCell {
         return State(cell: Tensor(zeros: stateShape), hidden: Tensor(zeros: stateShape))
     }
 
-    public typealias Input = RNNCellInput<Tensor<Scalar>, State>
-    public typealias Output = RNNCellOutput<State, State>
+    public typealias TimeStepInput = Tensor<Scalar>
+    public typealias TimeStepOutput = State
+    public typealias Input = RNNCellInput<TimeStepInput, State>
+    public typealias Output = RNNCellOutput<TimeStepOutput, State>
 
     /// Creates a `LSTMCell` with the specified input size and hidden state size.
     ///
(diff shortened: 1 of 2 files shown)


tanmayb123:
Would we prefer to have `forgetWeight` and `forgetBias` or `forgetW` and `forgetB`?

dan-zheng:
> Would we prefer to have `forgetWeight` and `forgetBias` or `forgetW` and `forgetB`?

The full names `Weight` and `Bias` are preferable.

tanmayb123:
Done.

Sources/DeepLearning/Layer.swift
```diff
@@ -1298,19 +1298,19 @@ public struct Reshape<Scalar: TensorFlowFloatingPoint>: Layer {
 
 struct LSTMCell<Scalar: TensorFlowFloatingPoint>: Layer {
```
rxwei: Remove extra line.

Sources/DeepLearning/Layer.swift
```diff
@@ -1295,3 +1295,59 @@ public struct Reshape<Scalar: TensorFlowFloatingPoint>: Layer {
         return input.reshaped(toShape: shape)
     }
 }
+
+struct LSTMCell<Scalar: TensorFlowFloatingPoint>: Layer {
```
rxwei: Make this and all properties public, otherwise people can’t use it :)

Sources/DeepLearning/Layer.swift
```diff
@@ -1295,3 +1295,59 @@ public struct Reshape<Scalar: TensorFlowFloatingPoint>: Layer {
...
+
+        @differentiable
+        init(cellState: Tensor<Scalar>, hiddenState: Tensor<Scalar>) {
+            self.cellState = cellState
+            self.hiddenState = hiddenState
+        }
+    }
+
+    struct Input: Differentiable {
+        var inputs: Tensor<Scalar>
```
rxwei: We should find a more meaningful name than “inputs” which confusingly repeats the name of the parent type “Input”.

Sources/DeepLearning/Layer.swift
```diff
@@ -1295,3 +1295,59 @@ public struct Reshape<Scalar: TensorFlowFloatingPoint>: Layer {
...
+    init(inputSize: Int32, hiddenSize: Int32) {
+        let concatenatedInputSize = inputSize + hiddenSize
+        self.inputWeight = Tensor<Scalar>(glorotUniform: [concatenatedInputSize, hiddenSize])
+        self.updateWeight = Tensor<Scalar>(glorotUniform: [concatenatedInputSize, hiddenSize])
+        self.forgetWeight = Tensor<Scalar>(glorotUniform: [concatenatedInputSize, hiddenSize])
+        self.forgetBias = Tensor<Scalar>(zeros: [hiddenSize])
+        self.outputWeight = Tensor<Scalar>(glorotUniform: [concatenatedInputSize, hiddenSize])
+    }
+
+    struct HiddenState: Differentiable {
```
rxwei: Why call it “HiddenState” instead of the existing associated type name “Output”? This way, protocol requirement “Output” is going to be inferred as a typealias of “HiddenState”, which is confusing.

Sources/DeepLearning/Layer.swift
```diff
@@ -1295,3 +1295,59 @@ public struct Reshape<Scalar: TensorFlowFloatingPoint>: Layer {
...
+
+struct LSTMCell<Scalar: TensorFlowFloatingPoint>: Layer {
+
+    var inputWeight: Tensor<Scalar>
+    var updateWeight: Tensor<Scalar>
+    var forgetWeight: Tensor<Scalar>
+    var forgetBias: Tensor<Scalar>
+    var outputWeight: Tensor<Scalar>
+
+    init(inputSize: Int32, hiddenSize: Int32) {
```
rxwei: Use Int instead of Int32 for layers.

dan-zheng: Please add doc comment:
```swift
/// Creates a `LSTMCell` with the specified input size and hidden state size.
///
/// - Parameters:
///   - inputSize: The number of features in 2-D input tensors.
///   - hiddenSize: The number of features in 2-D hidden states.
```

Sources/DeepLearning/Layer.swift
tanmayb123: I'm not too sure, or I'm not creative enough to come up with a name for it. The input to the LSTM layer consists of:

(a) the input for _that_ timestep.
(b) the previous timestep's hidden state.

What do you think we could name it?

Sources/DeepLearning/Layer.swift
tanmayb123: Because the output _is_ is the hidden state - like for example, in the Input struct, we've got `var hidden: HiddenState` which represents the previous timestep's output (which is the hidden state). Wdyt?

Sources/DeepLearning/Layer.swift
tanmayb123: Done

Sources/DeepLearning/Layer.swift
tanmayb123: Done.

Sources/DeepLearning/Layer.swift
tanmayb123: Done.

Sources/DeepLearning/Layer.swift
rxwei: I don't have a good answer here, but anything that differentiates the time step input from `Input` would be less confusing. How about the following?
```swift
public struct Input: Differentiable {
    public stepInput: Tensor<Scalar>
    public state: State
}
```

I don't feel strongly at all about `stepInput`. Feel free to suggest a better name.

Sources/DeepLearning/Layer.swift
```diff
@@ -1295,3 +1295,70 @@ public struct Reshape<Scalar: TensorFlowFloatingPoint>: Layer {
...
+    public var outputWeight: Tensor<Scalar>
+
+    /// Creates a `LSTMCell` with the specified input size and hidden state size.
+    ///
+    /// - Parameters:
+    ///   - inputSize: The number of features in 2-D input tensors.
+    ///   - hiddenSize: The number of features in 2-D hidden states.
+    public init(inputSize: Int32, hiddenSize: Int32) {
+        let concatenatedInputSize = inputSize + hiddenSize
+        self.inputWeight = Tensor<Scalar>(glorotUniform: [concatenatedInputSize, hiddenSize])
```
rxwei: Drop `<Scalar>` from all these lines. They can be type-inferred.

rxwei: Define a `let weightShape = [concatenatedInputSize, hiddenSize]` instead of creating them multiple times.

Sources/DeepLearning/Layer.swift
```diff
@@ -1295,3 +1295,70 @@ public struct Reshape<Scalar: TensorFlowFloatingPoint>: Layer {
...
+        self.updateWeight = Tensor<Scalar>(glorotUniform: [concatenatedInputSize, hiddenSize])
+        self.forgetWeight = Tensor<Scalar>(glorotUniform: [concatenatedInputSize, hiddenSize])
+        self.forgetBias = Tensor<Scalar>(zeros: [hiddenSize])
+        self.outputWeight = Tensor<Scalar>(glorotUniform: [concatenatedInputSize, hiddenSize])
+    }
+
+    public struct HiddenState: Differentiable {
```
rxwei: Regarding the naming of this, there are two issues: 
* It aliases `Output`, which can be confusing, though it's reasonable because it's technically correct.
* Type `HiddenState` does not meaningfully distinguish between a cell state and a hidden state (while it contains both as members!). Have you considered calling it just `State`?

rxwei:
```swift
public struct State: Differentiable {
    var cell: Tensor<Scalar>
    var hidden: Tensor<Scalar>
}
```

Sources/DeepLearning/Layer.swift
Shashi456: how about ```currentstate``` or ```currentInput``` or we could go change convention and go to something like ```c_0``` and ```h_0```, denoting current input and hidden state input

eaplatanios:
It would also be nice if we can have an RNN cell protocol, so that all valid cells have to conform to it and can be used in code that expects RNN cells (e.g., simple RNN, bi-directional RNN, etc).

tanmayb123:
@rxwei What do you think such a protocol would look like? i.e. What kind of functionality would it define?

Sources/DeepLearning/Layer.swift
```diff
@@ -1295,3 +1295,70 @@ public struct Reshape<Scalar: TensorFlowFloatingPoint>: Layer {
...
+            self.hiddenState = hiddenState
+        }
+    }
+
+    public struct Input: Differentiable {
+        var inputs: Tensor<Scalar>
+        var hidden: HiddenState
+
+        @differentiable
+        init(inputs: Tensor<Scalar>, hidden: HiddenState) {
```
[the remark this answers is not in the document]
tanmayb123: Sorry, missed that.

Shashi456:
@tanmayb123 I think he means a simple rnn cell, used in char RNNs and other such recurrent networks. While LSTMs mostly were designed to handle long term dependencies practically.

The four update equations for an rnn would be : 
image


While in the LSTM you have different kinds of gates, which act as a memory unit. 

image

For futrther resources, maybe you can look [here ](http://karpathy.github.io/2015/05/21/rnn-effectiveness/)and [here](http://colah.github.io/posts/2015-08-Understanding-LSTMs/). 

Char RNNs are very much still in fashion in my opinion.

@eaplatanios but it probably makes more sense to add it in a seperate PR.

tanmayb123:
I agree that Simple RNNs should be implemented (I could include that in this PR), but @eaplatanios was probably referring to an RNNCell protocol that all RNN Cells (i.e. LSTM/GRU/SimpleRNN) would conform to.

Sources/DeepLearning/Layer.swift
```diff
@@ -1295,3 +1295,70 @@ public struct Reshape<Scalar: TensorFlowFloatingPoint>: Layer {
...
+        let concatenatedInputSize = inputSize + hiddenSize
+        self.inputWeight = Tensor<Scalar>(glorotUniform: [concatenatedInputSize, hiddenSize])
+        self.updateWeight = Tensor<Scalar>(glorotUniform: [concatenatedInputSize, hiddenSize])
+        self.forgetWeight = Tensor<Scalar>(glorotUniform: [concatenatedInputSize, hiddenSize])
+        self.forgetBias = Tensor<Scalar>(zeros: [hiddenSize])
+        self.outputWeight = Tensor<Scalar>(glorotUniform: [concatenatedInputSize, hiddenSize])
+    }
+
+    public struct HiddenState: Differentiable {
+        var cellState: Tensor<Scalar>
```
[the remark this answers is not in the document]
tanmayb123: Done.

Sources/DeepLearning/Layer.swift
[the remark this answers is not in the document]
tanmayb123: Done.

Sources/DeepLearning/Layer.swift
tanmayb123: Done.

tanmayb123: Done.

Sources/DeepLearning/Layer.swift
tanmayb123: Done.

Sources/DeepLearning/Layer.swift
tanmayb123: Done.

rxwei:
I probably won't have time to review this or make suggestions until late evening.

rxwei:
Here's a sketch:
```swift
public struct RNNInput<StepInput: Differentiable, State: RNNState>: Differentiable {
    public var stepInput: StepInput
    public var previousState: State
    public init(stepInput: StepInput, previousState: State) {
        self.stepInput = stepInput
        self.previousState = previousState
    }
}

public struct RNNState<CellState: Differentiable, HiddenState: Differentiable>: Differentiable {
    public var cell: CellState
    public var hidden: HiddenState
    public init(cell: CellState, hidden: HiddenState) {
        self.cell = cell
        self.hidden = hidden
    }
}

public protocol RNNCell: Layer
    where Input == RNNInput<StepInput, State>, Output == RNNState<CellState, HiddenState> {
    associatedtype StepInput: Differentiable
    associatedtype CellState: Differentiable
    associatedtype HiddenState: Differentiable
    typealias State = Output

    init(inputSize: Int, hiddenSize: Int)
    var zeroState: State { get }
}

public extension RNNCell {
    @differentiable
    func applied(to stepInput: StepInput, _ previousState: State) -> State {
        return applied(to: Input(stepInput: stepInput, previousState: previousState))
    }
}
```

tanmayb123:
Love the sketch, Richard - thanks :)
Quick question: only LSTMs have a "cell state". Others, like GRU and SimpleRNN only have a "hidden state". Therefore, their "state" wouldn't be a struct, rather just a single Tensor. That's why I was a bit confused on how to create that protocol.

eaplatanios:
Sorry I wasn't able to respond earlier because I've been kept busy. I was thinking of something a bit more abstract where `RNNState` does not need to consist of two parts but is rather a generic type itself (called just `State` should be fine), as it can vary between different RNN cells.

rxwei:
Yeah, the cell state is definitely weird. I think that part may not need a fixed structure, like @eaplatanios said it can just be a generic type.

```swift
public struct RNNInput<StepInput: Differentiable, State: Differentiable>: Differentiable {
    public var stepInput: StepInput
    public var previousState: State
    public init(stepInput: StepInput, previousState: State) {
        self.stepInput = stepInput
        self.previousState = previousState
    }
}

public protocol RNNCell: Layer where Input == RNNInput<StepInput, State> {
    associatedtype StepInput: Differentiable
    typealias State = Output

    init(inputSize: Int, hiddenSize: Int)
    var zeroState: State { get }
}

public extension RNNCell {
    @differentiable
    func applied(to stepInput: StepInput, _ previousState: State) -> State {
        return applied(to: Input(stepInput: stepInput, previousState: previousState))
    }
}
```

rxwei:
I'm going to check the protocol in. If you can make these new layers work with the new protocol, that'd be really great!

tanmayb123:
I'll do that shortly - thanks Richard!

Sources/DeepLearning/Layer.swift
```diff
@@ -1295,3 +1295,125 @@ public struct Reshape<Scalar: TensorFlowFloatingPoint>: Layer {
...
+        let forgetGate = sigmoid(matmul(gateInput, forgetWeight) + forgetBias)
+        let outputGate = sigmoid(matmul(gateInput, outputWeight))
+
+        let newCellState = (input.state.cell * forgetGate + inputGate * updateGate)
+        let newHiddenState = tanh(newCellState) * outputGate
+
+        return State(cell: newCellState, hidden: newHiddenState)
+    }
+    
+    public func zeroState() -> State {
```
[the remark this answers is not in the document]
superbobry: I think you might want to have `zeroState` (or maybe `initialState`) accept `batchSize` as the inputs are likely to be batched.

Sources/DeepLearning/Layer.swift
```diff
@@ -1295,3 +1295,125 @@ public struct Reshape<Scalar: TensorFlowFloatingPoint>: Layer {
...
+    /// - Parameters:
+    ///   - inputSize: The number of features in 2-D input tensors.
+    ///   - hiddenSize: The number of features in 2-D hidden states.
+    public init(inputSize: Int, hiddenSize: Int) {
+        let concatenatedInputSize = Int32(inputSize + hiddenSize)
+        let gateWeightShape = TensorShape([concatenatedInputSize, Int32(hiddenSize)])
+        self.inputWeight = Tensor(glorotUniform: gateWeightShape)
+        self.updateWeight = Tensor(glorotUniform: gateWeightShape)
+        self.forgetWeight = Tensor(glorotUniform: gateWeightShape)
+        self.forgetBias = Tensor(zeros: [Int32(hiddenSize)])
```
superbobry: A rule of a thumb is to initialize the forget bias with ~1, see http://proceedings.mlr.press/v37/jozefowicz15.pdf

Also, is there a reason for not having a bias for other gates?

Sources/DeepLearning/Layer.swift
```diff
@@ -1295,3 +1295,125 @@ public struct Reshape<Scalar: TensorFlowFloatingPoint>: Layer {
...
+    /// - Parameters:
+    ///   - input: The input to the layer.
+    ///   - context: The contextual information for the layer application, e.g. the current learning
+    ///     phase.
+    /// - Returns: The hidden state.
+    @differentiable
+    public func applied(to input: Input, in _: Context) -> State {
+        let gateInput = input.stepInput.concatenated(with: input.state.hidden, alongAxis: 1)
+
+        let inputGate = sigmoid(matmul(gateInput, inputWeight))
```
superbobry: As a further optimization you could fuse all gates' weights into a single Tensor and do a single matmul:

```swift
let (input, update, forget, output) = split(matmul(gateInput, weight)) + bias, 4)  # Wishful thinking.
...
```

tanmayb123: Agreed, I just thought it was more understandable/readable this way. What do you think @rxwei?

rxwei: Definitely so, but we do not have a Swifty implementation of the `Split` op yet.

rxwei:
Hi @tanmayb123, we are preparing a Swift for TensorFlow v0.3 release this Wednesday. Would you like to update this PR and check it in so that it can be part of the release?

tanmayb123:
Sure thing Richard, working on that now. Quick concern that I didn't notice before: `RNNCellOutput` has both an `Output` and `State`, but that's a bit problematic for the cells since they don't have any outputs per time-step - their output _is_ the state. What should I do in that case? Should I pass the new state to `Output` and `State`?

rxwei:
> Should I pass the new state to `Output` and `State`?

Yes, that's the idea!

tanmayb123:
Great, working on that now.

Sources/DeepLearning/Layer.swift
```diff
@@ -1297,12 +1297,12 @@ public struct SimpleRNNCell<Scalar: TensorFlowFloatingPoint>: RNNCell {
...
 
-		@differentiable
-		public var zeroState: Tensor<Scalar> {
-				get {
-						return Tensor(zeros: stateShape)
-				}
+	@differentiable
+	public var zeroState: Tensor<Scalar> {
+		get {
+			return Tensor(zeros: stateShape)
```
rxwei: I think you are using tabs instead of spaces.

tanmayb123: Yep. Fixed.

rxwei:
BTW, the `applied(to:)` method no longer takes a `Context`.

brettkoonce:
ps hi tanmay! 🍀

rxwei:
It'd be great if you can add some tests.

Sources/DeepLearning/Layer.swift
```diff
@@ -1289,3 +1289,120 @@ public extension RNNCell {
...
+    public var updateWeight: Tensor<Scalar>
+    public var forgetWeight: Tensor<Scalar>
+    public var forgetBias: Tensor<Scalar>
+    public var outputWeight: Tensor<Scalar>
+
+    @noDerivative var stateShape: TensorShape
+
+    @differentiable
+    public var zeroState: State {
+        get {
```
rxwei: Omit `get` since it doesn't have a `set`.

Sources/DeepLearning/Layer.swift
```diff
@@ -1289,3 +1289,120 @@ public extension RNNCell {
...
+/// A Simple RNN Cell.
+public struct SimpleRNNCell<Scalar: TensorFlowFloatingPoint>: RNNCell {
+    public var weight: Tensor<Scalar>
+    public var bias: Tensor<Scalar>
+
+    @noDerivative var stateShape: TensorShape
+
+    @differentiable
+    public var zeroState: Tensor<Scalar> {
+        get {
```
rxwei: Omit `get` here as well.

Tests/DeepLearningTests/LayerTests.swift
```diff
@@ -82,6 +82,20 @@ final class LayerTests: XCTestCase {
...
+    func testSimpleRNNCell() {
+        let weight = Tensor<Float>(ones: [7, 5]) *
+                        Tensor<Float>([0.33333333, 1, 0.33333333, 1, 0.33333333])
+        let bias = Tensor<Float>(ones: [5])
+        var cell = SimpleRNNCell<Float>(inputSize: 2, hiddenSize: 5)
+        cell.weight = weight
+        cell.bias = bias
+        let simpleRNNState = Tensor<Float>(ones: [1, 5]) * Tensor<Float>([1, 0.2, 0.5, 2, 0.6])
+        let simpleRNNInput = Tensor<Float>(ones: [1, 2]) * Tensor<Float>([0.3, 0.7])
+        let output = cell.inferring(from: .init(stepInput: simpleRNNInput, state: simpleRNNState)).state
```
rxwei: Fit all lines within 100 columns.

Tests/DeepLearningTests/LayerTests.swift
```diff
@@ -82,6 +82,20 @@ final class LayerTests: XCTestCase {
         XCTAssertEqual(output.shape, expected)
     }
 
+    func testSimpleRNNCell() {
+        let weight = Tensor<Float>(ones: [7, 5]) *
+                        Tensor<Float>([0.33333333, 1, 0.33333333, 1, 0.33333333])
```
rxwei:
```suggestion
        Tensor<Float>([0.33333333, 1, 0.33333333, 1, 0.33333333])
```

Sources/DeepLearning/Layer.swift
```diff
@@ -1303,8 +1303,10 @@ public struct SimpleRNNCell<Scalar: TensorFlowFloatingPoint>: RNNCell {
     }
 
     public typealias State = Tensor<Scalar>
-    public typealias Input = RNNCellInput<Tensor<Scalar>, State>
-    public typealias Output = RNNCellOutput<State, State>
+    public typealias TimeStepInput = Tensor<Float>
```
rxwei: `Float` -> `Scalar`

Sources/DeepLearning/Layer.swift
```diff
@@ -1289,3 +1289,120 @@ public extension RNNCell {
...
+    public typealias Input = RNNCellInput<TimeStepInput, State>
+    public typealias Output = RNNCellOutput<TimeStepOutput, State>
+
+    /// Creates a `SimpleRNNCell` with the specified input size and hidden state size.
+    ///
+    /// - Parameters:
+    ///   - inputSize: The number of features in 2-D input tensors.
+    ///   - hiddenSize: The number of features in 2-D hidden states.
+    public init(inputSize: Int, hiddenSize: Int) {
+        let concatenatedInputSize = Int32(inputSize + hiddenSize)
```
rxwei: I think you can drop all the `Int32(...)` conversions now.

Sources/DeepLearning/Layer.swift
tanmayb123: Ah sorry, completely missed that. Fixing now.

Sources/DeepLearning/Layer.swift
tanmayb123: Perfect, will do.

Sources/DeepLearning/Layer.swift
jekbradbury: I'd rather leave it exploded out and rely on XLA/other optimizations to make it more efficient later.

Tests/DeepLearningTests/LayerTests.swift
```diff
@@ -82,6 +82,19 @@ final class LayerTests: XCTestCase {
...
 
+    func testSimpleRNNCell() {
+        let weight = Tensor<Float>(ones: [7, 5]) * Tensor<Float>([0.3333, 1, 0.3333, 1, 0.3333])
+        let bias = Tensor<Float>(ones: [5])
+        var cell = SimpleRNNCell<Float>(inputSize: 2, hiddenSize: 5)
+        cell.weight = weight
+        cell.bias = bias
+        let state = Tensor<Float>(ones: [1, 5]) * Tensor<Float>([1, 0.2, 0.5, 2, 0.6])
+        let input = Tensor<Float>(ones: [1, 2]) * Tensor<Float>([0.3, 0.7])
+        let output = cell.inferring(from: .init(input: input, state: state)).state
```
rxwei:
```suggestion
        let output = cell.applied(to: input, state: state).state
```

Sources/DeepLearning/Layer.swift
```diff
@@ -1289,3 +1289,120 @@ public extension RNNCell {
...
+    /// Creates a `LSTMCell` with the specified input size and hidden state size.
+    ///
+    /// - Parameters:
+    ///   - inputSize: The number of features in 2-D input tensors.
+    ///   - hiddenSize: The number of features in 2-D hidden states.
+    public init(inputSize: Int, hiddenSize: Int) {
+        let concatenatedInputSize = inputSize + hiddenSize
+        let gateWeightShape = TensorShape([concatenatedInputSize, hiddenSize])
+        self.inputWeight = Tensor(glorotUniform: gateWeightShape)
+        self.updateWeight = Tensor(glorotUniform: gateWeightShape)
```
jekbradbury: You should have biases on all four gate weights. The forget bias is "special" in that it's the only one that is frequently initialized to something other than zero, but all biases are important.

tanmayb123: Ah, got it. Working on that now.

rxwei:
I didn't see James's comment. Definitely address his concerns first :)

Sources/DeepLearning/Layer.swift
tanmayb123: Done.

Sources/DeepLearning/Layer.swift
```diff
@@ -1338,10 +1338,13 @@ public struct SimpleRNNCell<Scalar: TensorFlowFloatingPoint>: RNNCell {
 /// An LSTM Cell.
 public struct LSTMCell<Scalar: TensorFlowFloatingPoint>: RNNCell {
     public var inputWeight: Tensor<Scalar>
```
rxwei: Would be a little cleaner if you define all weights one the same line, and all biases on the same line.
```swift
var inputWeight, updateWeight, ...: Tensor<Scalar>
var inputBias, updateBias, ...: Tensor<Scalar>
```

tanmayb123: Done.

tanmayb123:
Also, related to this PR, I've opened #91 to see how to handle sequential inputs and keeping track of hidden states over time automatically.

tanmayb123:
The build is failing, but not on any of the Recurrent Cell code.

Sources/DeepLearning/Layer.swift
```diff
@@ -1289,3 +1289,121 @@ public extension RNNCell {
         return applied(to: RNNCellInput(input: input, state: state))
     }
 }
+
+/// A Simple RNN Cell.
+public struct SimpleRNNCell<Scalar: TensorFlowFloatingPoint>: RNNCell {
+    public var weight: Tensor<Scalar>
+    public var bias: Tensor<Scalar>
+
+    @noDerivative var stateShape: TensorShape
```
rxwei: I think this can be public. Same for LSTM.

tanmayb123: Sure thing.

Sources/DeepLearning/Layer.swift
```diff
@@ -1295,7 +1295,7 @@ public struct SimpleRNNCell<Scalar: TensorFlowFloatingPoint>: RNNCell {
     public var weight: Tensor<Scalar>
     public var bias: Tensor<Scalar>
 
-    @noDerivative var stateShape: TensorShape
+    @noDerivative public var stateShape: TensorShape
```
rxwei: Oh, this should be `public private(set)`, because setting it outside the cell will give you unexpected shape errors.

tanmayb123: So I should just make this a constant? i.e. `@noDerivative public let stateShape: TensorShape`

rxwei: That sounds good too. People may yell at me for making this immutable at some point. We'll see :)

tanmayb123: Or do you think it would be better to calculate the shape inside of `zeroState` based off of the shape of the weight(s)?

rxwei: Yeah, that could be better. You can make `stateShape` a computed property then.

tanmayb123: good thing I made this change, fixed a bug too.

rxwei: Great :)

rxwei:
I'm going to merge it once tests pass.

rxwei:
Tests are failing because your branch is still old. Could you pull and merge?

tanmayb123:
Sure thing.

rxwei:
Merged. Thanks for iterating on this!

tanmayb123:
Of course :)

PR #73, opened 2019-04-02
Updated the convolution ops to support explicit paddings.

eaplatanios:


Sources/DeepLearning/Operators.swift
```diff
@@ -98,12 +98,14 @@ public extension Tensor where Scalar: BinaryFloatingPoint {
...
 //===------------------------------------------------------------------------------------------===//
-// Convolution and pooling
+// Convolution and Pooling
 //===------------------------------------------------------------------------------------------===//
 
 /// A padding scheme. Used by padding, convolution, and pooling ops.
 // @_frozen // SR-9739
 public enum Padding {
+    /// The "explicit" padding scheme.
+    case explicit(_ paddings: [Int32])
```
rxwei: I left a comment about this in the previous PR.

Sources/DeepLearning/Operators.swift
```diff
@@ -98,12 +98,14 @@ public extension Tensor where Scalar: BinaryFloatingPoint {
...
 
 //===------------------------------------------------------------------------------------------===//
-// Convolution and pooling
+// Convolution and Pooling
 //===------------------------------------------------------------------------------------------===//
 
 /// A padding scheme. Used by padding, convolution, and pooling ops.
 // @_frozen // SR-9739
 public enum Padding {
+    /// The "explicit" padding scheme.
```
rxwei: We should document what "explicit" means. Adapting from [TF 2.0 docs](https://www.tensorflow.org/versions/r2.0/api_docs/python/tf/nn/conv2d):

"An array indicating the explicit padding sizes at the start and end of each dimension."

rxwei:
The build's going to fail, which is expected. I'll merge this and update API checkout in https://github.com/apple/swift/pull/23700.

Issue #54, opened 2019-03-11
Implement more layers that are available in Keras


rxwei:

tanmayb123:
@rxwei Quick question on this: Do we also want to add layers like `Add` `Subtract` `Multiply` `Concatenate` etc.? Two things to note:

1. If we do, we could just sequence these operations instead of having to run the operations on the Tensors.
2. How would a layer handle an arbitrary length of inputs? I'm aware of how to send _multiple_ pre-defined inputs using a struct, but could you just pass an array of Tensors (i.e. `[Tensor<Scalar>]`) for many inputs?

rxwei:
Hi @tanmayb123, let’s not add those operator-like layers for now and defer it to further discussions. Ideally, we would want functions to be able to conform to the layer protocol, but it’s currently impossible in Swift, so we’ll probablu need a single type wrapper that can turn any differentiable function to a layer.

As for arbitrary-length input/output, there’s ongoing work on making Array conform to Differentiable. It should be resolved next week or so.

rxwei:
Random thought: We can define layer wrappers for common arities (unary and binary), and define a `layer(from:)` factory function for turning any `Differentiable`-to-`Differentiable` function into a layer.

Rough sketch:
```swift
public struct ClosureLayer<T: Differentiable, U: Differentiable>: Layer {
    @noDerivative
    public var function: @differentiable (T) -> U
    func applied(to input: T) -> U {
        return function(input)
    }
}

public struct ClosureLayer2<T: Differentiable, U: Differentiable, V: Differentiable>: Layer {
    ...
}

func layer<T: Differentiable, U: Differentiable>(_ function: @differentiable (T) -> U) -> ClosureLayer<T, U> {
    return ClosureLayer(function: function)
}

...
```

Then, you'll be able to use functions in `sequenced(in:through:)`.

```
input.sequenced(in: context, through: conv, maxPool, layer(sin), ...)
```

What's better: You can now use a trailing closure to create anonymous layers!

```swift
let myLayer = layer { (input: Tensor<Float>) in
    sin(cos(x))
}
```

Shashi456:
@rxwei could we go onto make a list of what is available and whats to be done? so that gives a clearer picture of what layers need to be added? #53, #52 also are layer adding issues. So making a issue and referencing all of them might make the task easier.

Shashi456:
The implementations include :


- [x] Dense Layer


Convolution Layer :
- [x] Conv 1D Layer
[...]

PR #74, opened 2019-04-03
Made some more compatibility fixes to support the latest raw ops.

eaplatanios:
Hopefully fixes all remaining issues [here](https://github.com/apple/swift/pull/23700).

### DIFF
--- a/Sources/DeepLearning/Operators.swift
+++ b/Sources/DeepLearning/Operators.swift
@@ -251,7 +251,7 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
     internal func _vjpMaxPooled(
         kernelSize: (Int32, Int32, Int32, Int32),
         strides: (Int32, Int32, Int32, Int32),
-        padding: Padding
+        padding: PaddingV1
     ) -> (Tensor, (Tensor) -> Tensor) {
         // TODO: Currently this is not higher order differentiable. Redefine in
         // closed form.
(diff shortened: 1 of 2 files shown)


Sources/DeepLearning/Operators.swift
```diff
@@ -133,6 +133,35 @@ public extension Padding {
     }
 }
 
+/// An older padding scheme. Used by padding, convolution, and pooling ops.
+// @_frozen // SR-9739
+public enum PaddingV1 {
```
rxwei: We should not introduce versioned types in the deep learning library. Why is this necessary? All ops should use the general `Padding` type.

Also, I don't think the extension below compiles. Could you verify things compile and work locally?

eaplatanios: The pooling ops do not support explicit paddings. We could either not support explicit paddings at all for now (i.e., not even for convolution ops), or have two different padding modes.

I checked and it compiled locally but I may have mixed up my branches. I'll check again.

rxwei: In that case, let’s fall back to not supporting explicit paddings for now.

eaplatanios: I just pushed changes removing the explicit padding mode.

Shashi456:
@eaplatanios if this is set to get merged earlier than #70, could you make the operators directory in this one. I could then get started on #54 to avoid a conflict.

eaplatanios:
This needs a couple more fixes I'll do in a bit.

eaplatanios:
@rxwei this now compiles fine on my machine. Sorry for the delays but I've been super busy with preparing my thesis proposal.

PR #79, opened 2019-04-07
unify raw operator logic naming

brettkoonce:
as far as I can tell, this logic is now identical, and is not compiling here for me.
unified the namespaces and removed the extraneous function parameter.


brettkoonce:
see also #70

rxwei:
The root issue is https://github.com/apple/swift/pull/23700. When that is fixed, this PR is no longer necessary.

brettkoonce:
@rxwei kk!

PR #81, opened 2019-04-09
TF-425: Make learning rates dynamically settable.

saeta:
Modern training regiemes require the ability to change the learning rate during
training according to a particular schedule. By making the Optimizer protocol
allow the learning rate to be scheduled, this makes it easy to implement based
on callbacks.

Fixes [TF-425](https://bugs.swift.org/browse/TF-425)


saeta:
Thanks @rxwei !

Issue #77, opened 2019-04-04
Adding Test Cases for Layers in Layer.swift file


Shashi456:
There are a lot of layers which dont have test cases. I will start with adding test cases for Global Average Pooling and Global Max pooling, yet many are left. I will also try to make a list of which ones do and which ones do not. 

Also i thought this could be a good first issue if someone wanted to contribute.

Shashi456:
@dan-zheng can you add a good first issue tag label to this?

Descartess:
> I will also try to make a list of which ones do and which ones do not.

Hi @Shashi456 Any update on the list of the layers that have tests and those that don't. I wanted to pick this up

Shashi456:
@Descartess, Haven't gotten around to making it yet. But from the top of my head, `TranposedConv2D`, `Dense`, `LSTM Cell`, `BatchNorm`, `LayerNorm` dont have tests yet.

jkrukowski:
hi @Shashi456, I've created tests for `Dense` layer here https://github.com/tensorflow/swift-apis/pull/380

mohdsanadzakirizvi:
I would like to work on this, any update on this?

saeta:
@mohdsanadzakirizvi that's awesome!

In the spirit of [teaching to fish](https://www.td.org/magazines/td-magazine/teaching-learners-to-fish), Layers are currently found in [`Sources/TensorFlow/Layers`](https://github.com/tensorflow/swift-apis/tree/master/Sources/TensorFlow/Layers). The corresponding tests appear to be in [`LayerTests`](https://github.com/tensorflow/swift-apis/blob/master/Tests/TensorFlowTests/LayerTests.swift). After poking through a few files, I found that we define `Dropout` in [Core.swift](https://github.com/tensorflow/swift-apis/blob/master/Sources/TensorFlow/Layers/Core.swift#L44), but a quick ctrl/cmd-f in LayerTests.swift indicates "Dropout" doesn't appear in the file. I think that would be a great layer to add a couple test cases for.

Note: because Dropout is actually a stochastic operation (i.e. it is not deterministic, and instead depends on randomness) you'll want to think carefully about how to test this. (While you can run the operation and verify some properties about the output that are probabilistically almost always guaranteed to be true, it'd be even better to set the random seed to ensure it's fully deterministic.) Also, be sure to test both the training behavior as well as the inference behavior, as they are different! :-) 

I hope that helps, and thank you very much for contributing to S4TF!

[...]

PR #82, opened 2019-04-10
Adding GlobalAvgPool1D, 2D, 3D Test

Shashi456:
Some tests for #77

### DIFF
--- a/Tests/DeepLearningTests/LayerTests.swift
+++ b/Tests/DeepLearningTests/LayerTests.swift
@@ -46,8 +46,8 @@ final class LayerTests: XCTestCase {
         let layer = GlobalAveragePooling1D<Float>()
         let input = Tensor<Float>([[0, 1, 2, 3, 4], [10, 11, 12, 13, 14]])
         let output = layer.inferring(from: input)
-        let expected = Tensor<Float>([[2, 10]])
-        XCTAssertEqual(round(output, expected))
+        let expected = Tensor<Float>([2, 12])
+        XCTAssertEqual(round(output), expected)
     }
 
     static var allTests = [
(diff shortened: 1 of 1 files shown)


Tests/DeepLearningTests/LayerTests.swift
```diff
@@ -42,7 +42,15 @@ final class LayerTests: XCTestCase {
         XCTAssertEqual(round(output), expected)
     }
 
+    func testGlobalAvgPool1D() {
+        let layer = GlobalAveragePooling1D<Float>()
+        let input = Tensor<Float>([[0, 1, 2, 3, 4], [10, 11, 12, 13, 14]])
+        let output = layer.inferring(from : input)
```
rxwei: Remove the whitespace before the colon.

Tests/DeepLearningTests/LayerTests.swift
```diff
@@ -42,7 +42,15 @@ final class LayerTests: XCTestCase {
...
+        let expected = Tensor<Float>([2, 10])
+        XCTAssertEqual(round(output, expected))
+    }
+
     static var allTests = [
-        ("testConv1D", testConv1D), ("testMaxPool1D", testMaxPool1D), ("testAvgPool1D", testAvgPool1D)
+        ("testConv1D", testConv1D), ("testMaxPool1D", testMaxPool1D), ("testAvgPool1D", testAvgPool1D), ("testGlobalAvgPool1D", testGlobalAvgPool1D)
```
rxwei: Break this into multiple lines.

Shashi456:
@rxwei yeah I will add tests for the rest of the global average pooling layers too, I was just checking if I had the idea of the input and output right.

Shashi456:
@rxwei so the GlobalAveragePooling1D function is throwing an error because of the `.reshaped(to: [input.shape[0], input.shape[2]]` , it makes sense, since in the test I initialized `let input = Tensor<Float>([[0, 1, 2, 3, 4], [10, 11, 12, 13, 14]])` whose shape is [2,5]. It doesnt have a shape[2]. How do we rewrite that?

Shashi456:
@rxwei I think this can be merged now.

rxwei:
Thanks!

PR #83, opened 2019-04-11
One more math fix for BatchNorm

jekbradbury:
Batch normalization canonically normalizes across the batch and any spatial dimensions (i.e., everything _except_ the channel/feature dimension).


rxwei:
```console
/swift-apis/Sources/DeepLearning/Layer.swift:763:35: error: cannot convert value of type 'Int32' to expected argument type 'Int'
        normalizedAxes.remove(at: positiveAxis)
                                  ^~~~~~~~~~~~
                                  Int(        )
```

PR #86, opened 2019-04-13
Improve RNN cell abstraction.

rxwei:
The existing `RNNCell` protocol and `RNNInput` type are not flexible in that each time step has the take both the previous output and the hidden state. This PR lifts that restriction.

* Rename `RNNInput` to `RNNCellInput` so that it's more accurate.
* Add a new `RNNCellOutput` generic structure type that stores an output and a state.
* Add associated type `State` in `RNNCell`.
* Make the `Output` type of `RNNCell` be `RNNOutput<TimeStepOutput, State>`.

Thanks @superbobry for the suggestions.

### DIFF
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -1318,26 +1318,44 @@ public struct Reshape<Scalar: TensorFlowFloatingPoint>: Layer {
 }
 
 /// An input to a recurrent neural network.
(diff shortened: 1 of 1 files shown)


rxwei:
@superbobry @tanmayb123 @eaplatanios Please review when you get a chance.

eaplatanios:
Thanks Richard! This looks good to me.

PR #87, opened 2019-04-14
[Layer] Remove the 'context' argument from 'Layer.applied(to:in:)'.

rxwei:
This PR removes the `context` argument from `Layer`'s `applied(to:in:)` method to simplify the call sites of layer application. Instead of passing contexts as an argument, we make contexts available as thread-local information like how our device placement APIs (`withDevice(_:execute:)`) work.

We made this decision based on user feedback and API design principles, and would like our APIs to be clear and fluent. The team is aware of potential slight concerns in the context of concurrency, but has decided to start from first principles to make this APIs easier to use for everyone.

#### Internal changes

Add `ContextManager` class, each of whose instance manages a stack of contexts for one thread. Each thread-local singleton can be accessed through thread-safe type property `local`. The most recent context can be accessed through computed property `currentContext`. Pushing and popping can be done via `push(_:)` and `popContext()`.

I chose not to reuse the thread-local stack for device placement in the compiler runtime code base because it is engineered differently and is tied to the TensorFlow backend.

#### User-visible changes

* Remove the `context` argument. Change `applied(to:in:)` to `applied(to:)` in the `Layer` protocol and all layers.

* Change `Context` from a `class` to a `struct`, since access to thread-local contexts is already by reference and involves no copies.

* Add thread-safe computed property `.local` to `Context` for retrieving the current context.
  For example, here's how you can check the current learning phase as part of your layer.
    ```swift
    switch Context.local.learningPhase {
    case .training:
        ...
    case .inference:
        ...
    }
    ```
  There is a default `Context`, where `learningPhase == .inference`. To train a model, instead of creating a `Context` and passing it to `applied(to:in:)`, users can now set the local learning phase and call `applied(to:)` directly.
    ```swift
    let grad = model.gradient { model in
        let ŷ = model.applied(to: x)
        return softmaxCrossEntropy(logits: ŷ, labels: y)
    }
    ```

* Add APIs `withContext(_:_:)`, which calls a closure under a temporary context, and `withLearningPhase(_:_:)`, which calls a closure under a temporary learning phase.
 For example, `Layer.inferring(from:)` is now implemented as
    ```swift
    func inferring(from input: Input) -> Output {
        return withLearningPhase(.inference) { applied(to: input) }
    }
    ```

Future PRs will migrate all existing models and tutorials to the new API.

### DIFF
--- a/Sources/DeepLearning/Context.swift
+++ b/Sources/DeepLearning/Context.swift
@@ -1,0 +1,148 @@
+// Copyright 2018 The TensorFlow Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#if !COMPILING_TENSORFLOW_MODULE
+import TensorFlow
+#endif
+
+#if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
+import Darwin
+#else
+import Glibc
+#endif
+
+/// A value that indicates the phase of using a machine learning model.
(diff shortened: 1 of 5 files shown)


lattner:
yaaaaay! :-)

rxwei:
I'll merge and begin migrating models/tutorials/notebooks now, which I anticipate will be a lot more work than this PR. PR feedback will be welcome.

PR #76, opened 2019-04-04
Adding GlobalMaxPooling 1D, 2D, 3D

Shashi456:
I've added the max operation to operators, since it wasnt defined earlier.

PR #90, opened 2019-04-17
Rename GlobalAveragePooling to GlobalAvgPool

jekbradbury:
For consistency with the other pooling APIs. Also updates to use `squeezingAxes` (should be NFC).

### DIFF
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -943,7 +943,7 @@ public struct AvgPool2D<Scalar: TensorFlowFloatingPoint>: Layer {
 
 /// A global average pooling layer for temporal data.
 @_fixed_layout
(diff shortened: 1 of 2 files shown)


Shashi456:
@jekbradbury why is squeezing axes a better approach than the earlier one?

rxwei:
> @jekbradbury why is squeezing axes a better approach than the earlier one?

It calls fewer TensorFlow ops. `mean(squeezingAxis:)` is equivalent to `tf.reduce_mean(..., keep_dims=false)`. Though eventually we'd expect the compiler to optimize all of this away and make the performance be equivalent, the reality today is that the fewer op dispatches the better.

Shashi456:
@rxwei would this wouldn't solve the non-differentiable error that pops up in #76 right?

rxwei:
The non-differentiable error there is because `max(squeezingAxes:)` does not have a derivative defined in the standard library.

PR #93, opened 2019-04-17
Make layers callable.

dan-zheng:
Add `call func(_ input: Input) -> Output` as an extension method on `Layer`.
The requirement on `Layer` is still named `applied(to:)`.

Depends on https://github.com/apple/swift/pull/24094, please wait to merge.


dan-zheng:
Added a minimal test for now.
More tests and user code to be updated.

PR #95, opened 2019-04-17
Update optimizers.

saeta:
Previously, optimizers were incompatible with
[LayerList](https://github.com/fastai/fastai_docs/commit/a6dc98490fbc2f97cc5f4bac7c1a5d0170ebc5b1)
due to an interaction between KeyPath iterables and the instantiation of
`AllDifferentiableVariables.zero` (which naturally is `[]`). Concretely, when
running the optimizers, you would encounter array out-of-bounds errors. This
change updates the optimizers to take an instance of the model, and instantiate
all the "optimizer slots" (in TensorFlow optimizer parlance) as copies of the
Model's AllDifferentiableVariable's, just with their Tensor values set to zero.

Additionally, this change simplifies Optimizers to remove the Scalar type. For
now, this is handled by iterating over the 2 valid Scalar types (`Float`, and
`Double`), but in the future, support for generic math, the ability to leverage
`VectorNumeric`, and maybe a couple other features should allow for fully
general mixed-precision optimizers without requiring KeyPath iteration over the
respective scalar types.

Finally, this change makes a number of optimizer variables public to allow for
experimentation.

### DIFF
--- a/Sources/DeepLearning/Optimizer.swift
+++ b/Sources/DeepLearning/Optimizer.swift
@@ -23,12 +23,20 @@ import TensorFlow
 public protocol Optimizer {
     /// The type of the model whose parameters are optimized.
     associatedtype Model: Layer
+    /// The scalar parameter type.
+    associatedtype Scalar: FloatingPoint
     /// The learning rate.
-    var learningRate: Float { get set }
+    var learningRate: Scalar { get set }
     /// Updates the specified differentiable variables along the specified
     /// direction.
     mutating func update(_ variables: inout Model.AllDifferentiableVariables,
                          along direction: Model.CotangentVector)
+}
+
+fileprivate extension Tensor where Scalar: Numeric {
+    mutating func resetToZero() {
+        self = Tensor(zeros: shape)
+    }
 }
 
 // MARK: - Key-path based optimizers
@@ -47,11 +55,11 @@ public class Adam<Model: Layer>: Optimizer
     /// A coefficient used to calculate the first and second moments of
     /// gradients.
     public var beta2: Float
-    /// A small Float added to the denominator to improve numerical stability.
+    /// A small scalar added to the denominator to improve numerical stability.
     public var epsilon: Float
     /// The weight decay.
     public var decay: Float
-    /// The current step
+    /// The current step.
     public var step: Int = 0
     /// The first moments of the weights.
     public var firstMoments: Model.AllDifferentiableVariables
@@ -83,12 +91,12 @@ public class Adam<Model: Layer>: Optimizer
         firstMoments = model.allDifferentiableVariables
         secondMoments = model.allDifferentiableVariables
         for kp in firstMoments.recursivelyAllWritableKeyPaths(to: Tensor<Float>.self) {
-          firstMoments[keyPath: kp] = Tensor(zeros: firstMoments[keyPath: kp].shape)
-          secondMoments[keyPath: kp] = Tensor(zeros: secondMoments[keyPath: kp].shape)
+            firstMoments[keyPath: kp].resetToZero()
+            secondMoments[keyPath: kp].resetToZero()
         }
         for kp in firstMoments.recursivelyAllWritableKeyPaths(to: Tensor<Double>.self) {
-          firstMoments[keyPath: kp] = Tensor(zeros: firstMoments[keyPath: kp].shape)
-          secondMoments[keyPath: kp] = Tensor(zeros: secondMoments[keyPath: kp].shape)
+            firstMoments[keyPath: kp].resetToZero()
+            secondMoments[keyPath: kp].resetToZero()
         }
     }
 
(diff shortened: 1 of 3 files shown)


Sources/DeepLearning/Optimizer.swift
```diff
@@ -108,6 +107,15 @@ public class Adam<Model: Layer, Scalar: TensorFlowFloatingPoint>: Optimizer
...
                 stepSize * firstMoments[keyPath: kp] / (sqrt(secondMoments[keyPath: kp]) + epsilon)
         }
+        for kp in model.recursivelyAllWritableKeyPaths(to: Tensor<Double>.self) {
+            firstMoments[keyPath: kp] =
+                firstMoments[keyPath: kp] * Double(beta1) + Double((1 - beta1)) * direction[keyPath: kp]
+            secondMoments[keyPath: kp] =
+                secondMoments[keyPath: kp] * Double(beta2) + Double(1 - beta2) *
+                     direction[keyPath: kp] * direction[keyPath: kp]
+            model[keyPath: kp] -=
+                Double(stepSize) * firstMoments[keyPath: kp] / (sqrt(secondMoments[keyPath: kp]) + Double(epsilon))
```
rxwei: Fit all lines within 100 columns.

Sources/DeepLearning/Optimizer.swift
```diff
@@ -71,35 +76,29 @@ public class Adam<Model: Layer, Scalar: TensorFlowFloatingPoint>: Optimizer
...
-            beta2: beta2,
-            epsilon: epsilon,
-            decay: decay)
+        // Initialize first & second moments to be zeros of the same shape.
+        // We can't use `Model.AllDifferentiableVariables.zero` due to the
+        // interaction between Key Paths and Differentiable Arrays.
+        firstMoments = model.allDifferentiableVariables
+        secondMoments = model.allDifferentiableVariables
+        for kp in firstMoments.recursivelyAllWritableKeyPaths(to: Tensor<Float>.self) {
+          firstMoments[keyPath: kp] = Tensor(zeros: firstMoments[keyPath: kp].shape)
```
rxwei: To reduce boilerplace, you can make a `fileprivate` extension method `formZeros()`.

```swift
fileprivate extension Tensor {
    mutating func formZeros() {
        self = Tensor(zeros: shape)
    }
}
```

Then you can just use 
```swift
 firstMoments[keyPath: kp].formZeros()
```

Sources/DeepLearning/Optimizer.swift
```diff
@@ -233,25 +243,34 @@ public class SGD<Model: Layer, Scalar: TensorFlowFloatingPoint>: Optimizer
...
+                model[keyPath: kp] += velocity[keyPath: kp]
+            }
+        }
     }
 }
 
 // MARK: - Manifold optimizers
 
 /// A Riemann manifold stochastic gradient descent (SGD) optimizer.
-public class RiemannSGD<Model: Layer, Scalar: FloatingPoint>: Optimizer
```
rxwei: Let's not change the `Riemann` one for now since it's not `Tensor`-specific and no one is using it.

Sources/DeepLearning/Optimizer.swift
```diff
@@ -119,22 +127,26 @@ public class Adam<Model: Layer, Scalar: TensorFlowFloatingPoint>: Optimizer
...
-    public var rho: Scalar
-    /// A small scalar added to the denominator to improve numerical stability.
-    public var epsilon: Scalar
+    public var rho: Float
+    /// A small Float added to the denominator to improve numerical stability.
+    public var epsilon: Float
     /// The weight decay.
-    public var decay: Scalar
+    public var decay: Float
+    /// The step count
```
rxwei: End all doc comments with a period.

Sources/DeepLearning/Optimizer.swift
```diff
@@ -119,22 +127,26 @@ public class Adam<Model: Layer, Scalar: TensorFlowFloatingPoint>: Optimizer
...
     where Model.AllDifferentiableVariables == Model.CotangentVector {
     /// The learning rate.
-    public var learningRate: Scalar
+    public var learningRate: Float
     // TODO: Document `rho`. Keras doesn't document `rho`.
-    public var rho: Scalar
-    /// A small scalar added to the denominator to improve numerical stability.
-    public var epsilon: Scalar
+    public var rho: Float
+    /// A small Float added to the denominator to improve numerical stability.
```
rxwei: I don't think the comment needs to change.

Sources/DeepLearning/Optimizer.swift
rxwei: Indent by 4 spaces.

Sources/DeepLearning/Optimizer.swift
```diff
@@ -39,27 +37,34 @@ public protocol Optimizer {
...
-    public var beta2: Scalar
-    /// A small scalar added to the denominator to improve numerical stability.
-    public var epsilon: Scalar
+    public var beta2: Float
+    /// A small Float added to the denominator to improve numerical stability.
+    public var epsilon: Float
     /// The weight decay.
-    public var decay: Scalar
+    public var decay: Float
+    /// The current step
```
rxwei:
```suggestion
    /// The current step.
```

Sources/DeepLearning/Optimizer.swift
```diff
@@ -39,27 +37,34 @@ public protocol Optimizer {
...
     /// gradients.
-    public var beta1: Scalar
+    public var beta1: Float
     /// A coefficient used to calculate the first and second moments of
     /// gradients.
-    public var beta2: Scalar
-    /// A small scalar added to the denominator to improve numerical stability.
-    public var epsilon: Scalar
+    public var beta2: Float
+    /// A small Float added to the denominator to improve numerical stability.
```
rxwei: Same. The comment doesn't need to be changed.

Sources/DeepLearning/Optimizer.swift
```diff
@@ -23,10 +23,8 @@ import TensorFlow
 public protocol Optimizer {
     /// The type of the model whose parameters are optimized.
     associatedtype Model: Layer
-    /// The scalar parameter type.
-    associatedtype Scalar: FloatingPoint
     /// The learning rate.
-    var learningRate: Scalar { get set }
+    var learningRate: Float { get set }
```
rxwei: I don't think we should change the optimizer protocol, since it's generic over not just tensor optimizers. Changing just the concrete non-Riemann optimizers would do what we need.

Sources/DeepLearning/Optimizer.swift
saeta: Fixed.

Sources/DeepLearning/Optimizer.swift
saeta: Fixed (picked a slightly different name that I thought was clearer).

Sources/DeepLearning/Optimizer.swift
saeta: Fixed.

Sources/DeepLearning/Optimizer.swift
saeta: Reverted.

Sources/DeepLearning/Optimizer.swift
saeta: Fixed.

Sources/DeepLearning/Optimizer.swift
saeta: Fixed.

Sources/DeepLearning/Optimizer.swift
saeta: Fixed.

Sources/DeepLearning/Optimizer.swift
saeta: Reverted.

Sources/DeepLearning/Optimizer.swift
saeta: Reverted changes.

Sources/DeepLearning/Optimizer.swift
```diff
@@ -23,14 +23,22 @@ import TensorFlow
...
-    var learningRate: Float { get set }
+    var learningRate: Scalar { get set }
     /// Updates the specified differentiable variables along the specified
     /// direction.
     mutating func update(_ variables: inout Model.AllDifferentiableVariables,
                          along direction: Model.CotangentVector)
 }
 
+fileprivate extension Tensor where Scalar: Numeric {
+    mutating func resetToZero() {
```
rxwei: Nit: I'd say `formZero` is a more Swifty name here. `reset` is a bit unprecedented.

https://swift.org/documentation/api-design-guidelines/#naming

PR #96, opened 2019-04-18
Switch to 'call func'.

rxwei:
Before:
`func applied(to input: Input) -> Output`
After:
`call func(_ input: Input) -> Output`

### DIFF
--- a/Tests/DeepLearningTests/TrivialModelTests.swift
+++ b/Tests/DeepLearningTests/TrivialModelTests.swift
@@ -35,8 +35,8 @@ final class TrivialModelTests: XCTestCase {
             }
             @differentiable
             call func(_ input: Tensor<Float>) -> Tensor<Float> {
-                let h1 = l1.applied(to: input)
-                return l2.applied(to: h1)
+                let h1 = l1(input)
+                return l2(h1)
             }
         }
         var classifier = Classifier(hiddenSize: 4)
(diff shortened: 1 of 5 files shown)


Tests/DeepLearningTests/TrivialModelTests.swift
```diff
@@ -34,7 +34,7 @@ final class TrivialModelTests: XCTestCase {
                 )
             }
             @differentiable
-            func applied(to input: Tensor<Float>) -> Tensor<Float> {
+            call func(_ input: Tensor<Float>) -> Tensor<Float> {
                 let h1 = l1.applied(to: input)
```
dan-zheng: What about call-sites?
```suggestion
                let h1 = l1(input)
```

rxwei: Sure. Still building a toolchain.

rxwei:
The existing implementation of callables seems to have a name lookup issue cross-module. If you run `swift test` using the latest toolchain, you'll get these lookup errors:

```console
/Users/rxwei/Development/ML/swift-apis/Tests/DeepLearningTests/ContextTests.swift:24:31: error: cannot call value of non-function type 'Dropout<Float>'
        XCTAssertEqual(dropout(x), x)
                       ~~~~~~~^
```

dan-zheng:
Verified that all tests pass after https://github.com/apple/swift/pull/24121 is merged.
(I built a toolchain based on https://github.com/apple/swift/pull/24121 and ran `swift test`).

PR #97, opened 2019-04-18
Revert "Revert "Add Recurrent Layers""

dan-zheng:
Reverts tensorflow/swift-apis#94.

---

The major blocking crash is fixed in https://github.com/apple/swift/pull/24114.
`@differentiable` must also be removed from `var zeroState` to prevent another error/crash.

### DIFF
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -1293,3 +1293,123 @@ public extension RNNCell {
         return applied(to: RNNCellInput(input: input, state: state))
     }
 }
+
+/// A Simple RNN Cell.
+public struct SimpleRNNCell<Scalar: TensorFlowFloatingPoint>: RNNCell {
+    public var weight: Tensor<Scalar>
+    public var bias: Tensor<Scalar>
+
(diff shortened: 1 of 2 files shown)


dan-zheng:
Confirmed that this test passes (adapted from `Tests/DeepLearningTests/LayerTests.swift`):
```swift
import TensorFlow

let weight = Tensor<Float>(ones: [7, 5]) * Tensor<Float>([0.3333, 1, 0.3333, 1, 0.3333])
let bias = Tensor<Float>(ones: [5])
var cell = SimpleRNNCell<Float>(inputSize: 2, hiddenSize: 5)
cell.weight = weight
cell.bias = bias
let state = Tensor<Float>(ones: [1, 5]) * Tensor<Float>([1, 0.2, 0.5, 2, 0.6])
let input = Tensor<Float>(ones: [1, 2]) * Tensor<Float>([0.3, 0.7])
let output = cell.applied(to: input, state: state).state
let expected = Tensor<Float>([[2.76649, 6.2999997, 2.76649, 6.2999997, 2.76649]])
assert(output == expected)
```

PR #106, opened 2019-04-21
Minor fix for AD.

eaplatanios:
@rxwei This is based on what you mentioned in [this email thread](https://groups.google.com/a/tensorflow.org/forum/#!topic/swift/a5vaC2GTY_w).

### DIFF
--- a/Sources/DeepLearning/Operators/Math.swift
+++ b/Sources/DeepLearning/Operators/Math.swift
@@ -875,7 +875,7 @@ public extension Tensor where Scalar == Bool {
     @inlinable
     func all() -> Bool {
         let axes = Tensor<Int32>(rangeFrom: 0, to: Int32(rank), stride: 1)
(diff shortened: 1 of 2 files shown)


rxwei:
Thanks!

rxwei:
It's not a bug though. We just need a API that allows you to use any function/property while forcing the derivative to be zero.

eaplatanios:
> It's not a bug though. We just need a API that allows you to use any function/property while forcing the derivative to be zero.

Sorry, the "bug fix" commit was referring to the use of `_TFGetScalarOrDie` in `Math.swift`, which was missing from the previous PR moving stuff over from stdlib. :)

rxwei:
Got it!

rxwei:
Since this doesn't compile with SwiftPM, I'll have to revert it along with #70 to iterate further. Happy to help with resolving the compilation issues!

PR #109, opened 2019-04-21
[TF] TensorFlow/TensorFlowCore Refactoring

eaplatanios:
Moves all of the `TensorFlow` module here. All tests pass locally.

Friend PR: apple/swift#24452.


rxwei:
Could you create a PR that contains all the changes? I reverted #70 and #106 in order to maintain a working build in this repo.

rxwei:
Sorry for the inconvenience!

eaplatanios:
@rxwei No worries. I hope I did the merge correctly :) The only error that remains when I try to compile this now is the ambiguous subscript.

eaplatanios:
@rxwei I tried removing the subscript by putting it in a `#if COMPILING_TENSORFLOW_MODULE`/`#endif` block but now we get a crash from `swift::SILFunctionBuilder::addFunctionAttributes(swift::SILFunction*, swift::DeclAttributes&, swift::SILModule&, swift::SILDeclRef)`, which I've also seen before due to ambiguous functions. I believe this is because all of the extension functions are ambiguous in this case.

eaplatanios:
Also, shouldn't we add `@_semantics("autodiff.nonvarying")` to all functions that we know to be non-varying? E.g., `.>`, `.<`, `.==`, etc. It would also be nice to expose this to users using a nicer name. Is that possible?

rxwei:
> Also, shouldn't we add `@_semantics("autodiff.nonvarying")` to all functions that we know to be non-varying? E.g., `.>`, `.<`, `.==`, etc. It would also be nice to expose this to users using a nicer name. Is that possible?

I think it's better not to do that because we want non-differentiable functions to produce an error instead of silently giving a zero gradient at runtime. This will help sure a predictable programming model. The `@_semantics("autodiff.nonvarying")` attribute can be considered as equivalent to `@noDerivative`, and I think they should only apply to properties, not functions. Also, adding an attribute on everything often implies that it should be defaulted, but we've tried default zero-gradient-by-default and some initial users didn't like it.

My preference is that we explore adding a good API first before considering changing language defaults or library behavior. I see no problem adding a `NoDerivative` type today, because eventually we can just add a `@propertyDelegate` attribute on it when that's available.

eaplatanios:
I see your point and I agree. I'll look into the GPU error as soon as I can.

In the meantime, it turns out it's quite hard to debug TF errors in S4TF because no stack trace is every printed. Oftentimes I get a `fatal error` about mismatching shapes, etc, but no other information and it's very hard to debug. Is there a way to have S4TF print the whole stack trace whenever a fatal error occurs?

rxwei:
Yeah, that is a known pain point. I think that @pschuh has more context on the current status of pretty-printing TensorFlow runtime errors.

eaplatanios:
Even just a plain stack trace dump would be super helpful at this point to be honest. Because right now it's hard to know which part of the user's code is causing the error.

rxwei:
You can step through the code using LLDB.

eaplatanios:
Yes, that is what I'm doing currently, but I feel that won't be an option for a lot of ML researchers using S4TF.

Shashi456:
I'm sorry to divert the thread, @eaplatanios but generally vjps for a function can be thought of returning the dual for the automatic differentiation? 
so if i wanted to write a vjp for a function which squares a number, i would write the vjp such that it returns 2*x? 
I was just looking over your min max vjp function definitions. I'm still trying to wrap my head around writing vjps.

rxwei:
I and the team really agree, and it's definitely on our TODO list.

rxwei:
> I'm sorry to divert the thread, @eaplatanios but generally vjps for a function can be thought of returning the dual for the automatic differentiation?

A VJP function represents an efficient derivative function that takes original arguments and computes both
* the original result, and
* a closure (pullback) that takes a vector and chains it together with the function’s Jacobian, producing vector-Jacobian products, i.e. gradient along the vector.

> so if i wanted to write a vjp for a function which squares a number, i would write the vjp such that it returns 2*x?

Say if you have a `square(_:)` function:

```swift
func square(_ x: Float) -> Float {
    return x * x
}
```

You can define a derivative in the form of a function that has a `@differentiating` attribute.

```swift
@differentiating(square)
func squareDerivative(_ x: Float) -> (value: Float, pullback: (Float) -> Float) {
    return (x, { v in v * 2 * x })
}
```

Note here that it's not just `2x`, but a linear combination function `{ v in v * 2 * x }`. `v` is the "vector" which you can interpret as whatever's chaining with the current derivative. For scalar operations, it's multiplication or element-wise multiplication. For matrix operations (e.g. TensorFlow's `matmul(_:_:)`), it's matrix multiplication.

Shashi456:
@rxwei thanks a lot for the explanation.

Sources/DeepLearning/DifferentialOperators.swift
```diff
@@ -0,0 +1,176 @@
...
+    at x: T,
+    _ y: U,
+    in f: @differentiable (T, U) -> Tensor<R>
+) -> (T.CotangentVector, U.CotangentVector)
+    where T: Differentiable, U: Differentiable, R: TensorFlowFloatingPoint {
+    return pullback(at: x, y, in: f)(Tensor<R>(1))
+}
+
+// @inlinable
+// public func gradient<T, U, V, R>(
```
rxwei: Any reasons this is commented out?

Sources/DeepLearning/Layer.swift
```diff
@@ -558,7 +557,7 @@ public struct TransposedConv2D: Layer {
         self.activation = activation
         self.strides = strides
         self.padding = padding
-        self.paddingIndex = padding == .same ? 0 : 1
+        self.paddingIndex = padding == .same ? 0: 1
```
rxwei: Seems like an accidental change :)
```suggestion
        self.paddingIndex = padding == .same ? 0 : 1
```

Sources/DeepLearning/Tensor.swift
```diff
@@ -0,0 +1,106 @@
...
+    /// true and the element count exceeds twice the `edgeElementCount`.
+    ///
+    /// - Parameters:
+    ///   - lineWidth: The max line width for printing. Used to determine number of scalars to print
+    ///     per line.
+    ///   - edgeElementCount: The maximum number of elements to print before and after summarization
+    ///     via ellipses (`...`).
+    ///   - summarizing: If true, summarize description if element count exceeds twice
+    ///     `edgeElementCount`.
+    @_semantics("autodiff.nonvarying")
```
rxwei: I'd suggest removing this from functions for now and only apply them to common properties.

Sources/DeepLearning/Random.swift
```diff
@@ -536,7 +536,7 @@ public struct BetaDistribution: RandomDistribution {
         } while r + alpha * (log(alpha) - log(b + w)) < t
 
         w = min(w, Float.greatestFiniteMagnitude)
-        return a == alpha0 ? w / (b + w) : b / (b + w)
+        return a == alpha0 ? w / (b + w): b / (b + w)
```
rxwei:
```suggestion
        return a == alpha0 ? w / (b + w) : b / (b + w)
```

Sources/DeepLearning/PythonConversion.swift
```diff
@@ -0,0 +1,169 @@
...
+        dummyPointer.deallocate()
+        scalars.withUnsafeMutableBufferPointer { buffPtr in
+            buffPtr.baseAddress!.assign(from: ptr, count: scalarCount)
+        }
+        self.init(shape: shape, scalars: scalars)
+    }
+}
+
+extension Tensor: ConvertibleFromNumpyArray
+    where Scalar: NumpyScalarCompatible {
```
rxwei: You can move this to the end of the previous line.

Sources/DeepLearning/Operators/Dataset.swift
```diff
@@ -0,0 +1,222 @@
+// Copyright 2018 The TensorFlow Authors. All Rights Reserved.
```
rxwei: This file hasn't adopted 4-space indentation.

Sources/DeepLearning/Operators/Comparison.swift
```diff
@@ -0,0 +1,232 @@
...
+    func elementsApproximatelyEqual(
+        _ other: Tensor,
+        tolerance: Double = 0.00001
+    ) -> Tensor<Bool> {
+        return Raw.approximateEqual(self, other, tolerance: tolerance)
+    }
+}
+
+public extension StringTensor {
+  /// Computes `self == other` element-wise.
```
rxwei: Indent by 4 here.

Sources/DeepLearning/Operators/Comparison.swift
```diff
@@ -0,0 +1,232 @@
...
+    static func >= (lhs: Tensor, rhs: Scalar) -> Bool {
+        return (lhs .>= rhs).all()
+    }
+}
+
+public extension Tensor where Scalar: Equatable {
+    /// Computes `lhs != rhs` element-wise and returns a `Tensor` of Boolean scalars.
+    /// - Note: `.==` supports broadcasting.
+    @inlinable @inline(__always)
+    static func .==(lhs: Tensor, rhs: Tensor) -> Tensor<Bool> {
```
rxwei:
```suggestion
    static func .== (lhs: Tensor, rhs: Tensor) -> Tensor<Bool> {
```
Could you make it so that every operator function has a space between the operator and the left parenthesis?

Sources/DeepLearning/Operators/Basic.swift
```diff
@@ -0,0 +1,710 @@
...
+    ///   - axis: Dimension along which to split this tensor. Negative values wrap around.
+    ///
+    /// - Precondition: `numSplits` must divide the size of dimension `axis` evenly.
+    /// - Precondition: `axis` must be in the range `[-rank, rank)`, where `rank` is the rank of the
+    ///   provided tensors.
+    ///
+    /// - Returns: Array containing the tensors parts.
+    @inlinable
+    @differentiable(vjp: _vjpSplit(numSplits:alongAxis:) where Scalar: TensorFlowFloatingPoint)
+    func split(numSplits: Int, alongAxis axis: Int = 0) -> [Tensor] {
```
rxwei: Swift APIs generally do not use a "num" prefix.
```suggestion
    func split(count: Int, alongAxis axis: Int = 0) -> [Tensor] {
```

Sources/DeepLearning/Operators/Basic.swift
```diff
@@ -0,0 +1,710 @@
...
+    /// 
+    /// - Returns: Array containing the unstacked tensors.
+    @inlinable
+    @differentiable(vjp: _vjpUnstack(alongAxis:) where Scalar: TensorFlowFloatingPoint)
+    func unstack(alongAxis axis: Int = 0) -> [Tensor] {
+        return Raw.unpack(value: self, num: Int64(shape[axis]), axis: Int64(axis))
+    }
+
+    /// Splits a tensor into multiple tensors. The tensor is split along dimension `axis` into
+    /// `numSplits` smaller tensors. This requires that `numSplits` evenly divides `shape[axis]`.
```
rxwei:
```suggestion
    /// `count` smaller tensors. This requires that `count` evenly divides `shape[axis]`.
```

Sources/DeepLearning/Operators/Basic.swift
```diff
@@ -0,0 +1,710 @@
...
+    /// - Parameters:
+    ///   - axis: Dimension along which to unstack. Negative values wrap around.
+    /// 
+    /// - Precondition: `axis` must be in the range `[-rank, rank)`, where `rank` is the rank of the
+    ///   provided tensors.
+    /// 
+    /// - Returns: Array containing the unstacked tensors.
+    @inlinable
+    @differentiable(vjp: _vjpUnstack(alongAxis:) where Scalar: TensorFlowFloatingPoint)
+    func unstack(alongAxis axis: Int = 0) -> [Tensor] {
```
rxwei: Non-mutating methods should read like a noun instead of an imperative verb phrase.
```suggestion
    func unstacked(alongAxis axis: Int = 0) -> [Tensor] {
```

Sources/DeepLearning/Operators/Basic.swift
```diff
@@ -0,0 +1,710 @@
...
+// limitations under the License.
+
+import TensorFlowCore
+
+//===------------------------------------------------------------------------------------------===//
+// Shape Transformations
+//===------------------------------------------------------------------------------------------===//
+
+public extension TensorFlowScalar {
+    /// Convert to a tensor with the specified rank, with all dimensions equal to 1.
```
rxwei:
```suggestion
    /// Convert to a tensor with the specified rank, with all dimensions equal to `1`.
```

Sources/DeepLearning/Operators/Basic.swift
```diff
@@ -0,0 +1,710 @@
...
+    /// // 'value' is a tensor with shape [5, 30]
+    /// // Split 'value' into 3 tensors along dimension 1:
+    /// let parts = value.split(numSplits: 3, alongAxis: 1)
+    /// parts[0] // has shape [5, 10]
+    /// parts[1] // has shape [5, 10]
+    /// parts[2] // has shape [5, 10]
+    /// ```
+    ///
+    /// - Parameters:
+    ///   - numSplits: Number of splits to create.
```
rxwei:
```suggestion
    ///   - count: The number of splits to create.
```

Sources/DeepLearning/Operators/Basic.swift
```diff
@@ -0,0 +1,710 @@
...
+    ///   - axis: Dimension along which to split this tensor. Negative values wrap around.
+    ///
+    /// - Precondition: `numSplits` must divide the size of dimension `axis` evenly.
```
rxwei:
```suggestion
    /// - Precondition: `count` must divide the size of dimension `axis` evenly.
```

Sources/DeepLearning/Operators/Basic.swift
```diff
@@ -0,0 +1,710 @@
...
+    /// // Split 'value' into 3 tensors along dimension 1:
+    /// let parts = value.split(numSplits: 3, alongAxis: 1)
+    /// parts[0] // has shape [5, 10]
+    /// parts[1] // has shape [5, 10]
+    /// parts[2] // has shape [5, 10]
+    /// ```
+    ///
+    /// - Parameters:
+    ///   - numSplits: Number of splits to create.
+    ///   - axis: Dimension along which to split this tensor. Negative values wrap around.
```
rxwei:
```suggestion
    ///   - axis: The dimension along which to split this tensor. Negative values wrap around.
```

Sources/DeepLearning/Operators/Basic.swift
```diff
@@ -0,0 +1,710 @@
...
+    ///
+    /// - Precondition: `numSplits` must divide the size of dimension `axis` evenly.
+    /// - Precondition: `axis` must be in the range `[-rank, rank)`, where `rank` is the rank of the
+    ///   provided tensors.
+    ///
+    /// - Returns: Array containing the tensors parts.
```
rxwei:
```suggestion
    /// - Returns: An array containing split tensors.
```

Sources/DeepLearning/Operators/Basic.swift
```diff
@@ -0,0 +1,710 @@
...
+    @inlinable
+    @differentiable(vjp: _vjpSplit(numSplits:alongAxis:) where Scalar: TensorFlowFloatingPoint)
```
rxwei:
```suggestion
    @differentiable(vjp: _vjpSplit(count:alongAxis:) where Scalar: TensorFlowFloatingPoint)
```

Sources/DeepLearning/Operators/Basic.swift
```diff
@@ -0,0 +1,710 @@
...
+    }
+
+    /// Returns a rank-lifted `Tensor` with a leading dimension of 1.
+    @inlinable
+    @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
+    func rankLifted() -> Tensor {
+        return expandingShape(at: 0)
+    }
+
+    /// Remove the specified dimensions of size 1 from the shape of a tensor. If no dimensions are
```
rxwei:
```suggestion
    /// Removes the specified dimensions of size 1 from the shape of a tensor. If no dimensions are
```

Sources/DeepLearning/Operators/Basic.swift
```diff
@@ -0,0 +1,710 @@
...
+    /// specified, then all dimensions of size 1 will be removed.
+    @inlinable
+    @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
+    func squeezingShape(at axes: Int...) -> Tensor {
+        return squeezingShape(at: axes)
+    }
+
+    /// Remove the specified dimensions of size 1 from the shape of a tensor. If no dimensions are
```
rxwei:
```suggestion
    /// Removes the specified dimensions of size 1 from the shape of a tensor. If no dimensions are
```

Sources/DeepLearning/Operators/Basic.swift
```diff
@@ -0,0 +1,710 @@
...
+    func _vjpUnstack(
+        alongAxis axis: Int = 0
+    ) -> ([Tensor], (Array<Tensor>.CotangentVector) -> Tensor) {
+        let result = unstack(alongAxis: axis)
+        return (result, { v in Tensor(stacking: v.base, alongAxis: axis) })
+    }
+
+    @inlinable
+    func _vjpSplit(
+        numSplits: Int,
```
rxwei:
```suggestion
        count: Int,
```

Sources/DeepLearning/Operators/Basic.swift
```diff
@@ -0,0 +1,710 @@
...
+//===------------------------------------------------------------------------------------------===//
+// Broadcasting
+//===------------------------------------------------------------------------------------------===//
+
+// TODO: What about precedence? Why is this operator used for broadcasting?
+infix operator .=
+
+public extension Tensor {
+    @inlinable
+    func broadcast(toShape shape: Tensor<Int32>) -> Tensor {
```
rxwei: BTW, here's a related TODO: [TF-430](https://bugs.swift.org/browse/TF-430). Need not be done in this patch though!

Tests/DeepLearningTests/InitializerTests.swift
```diff
@@ -0,0 +1,108 @@
...
+@testable import DeepLearning
+
+final class InitializerTests: XCTestCase {
+    func testInitializers() {
+        let scalar = Tensor<Float>(1)
+        let matrix: Tensor<Float> = [[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]]
+        let broadcastScalar = Tensor<Float>(broadcasting: 10, rank: 3)
+        let some4d = Tensor<Float>(
+            shape: [2, 1, 2, 1],
+            scalars: AnyRandomAccessCollection([2, 3, 4, 5]))
```
rxwei: I don't think `AnyRandomAccessCollection` is necessary since `init(shape:scalars:)` is generic.
```suggestion
            scalars: [2, 3, 4, 5])
```

Tests/DeepLearningTests/InitializerTests.swift
```diff
@@ -0,0 +1,108 @@
...
+
+    func testArrayConversion() {
+        let array3D = ShapedArray(repeating: 1.0, shape: [2, 3, 4])
+        let tensor3D = Tensor(array3D)
+        XCTAssertEqual(array3D, tensor3D.array)
+    }
+
+    func testNonTPUDataTypeCast() {
+        // TPU does not support Int8 or 16 casting.
+        guard !_RuntimeConfig.executionMode.isTPU else { return }
```
rxwei: Feel free to remove any check for `_RuntimeConfig`.

Sources/DeepLearning/DifferentialOperators.swift
eaplatanios: Sorry I forgot about this earlier. All the functions in this file that have 4 generic type arguments result in the following compiler crash:
```
1.      While verifying SIL function "@$s12DeepLearning17valueWithGradient2at__2in14TensorFlowCore0H0Vyq1_G0C0_15CotangentVectorQz_AJQy_AJQy0_t8gradienttx_q_q0_AHx_q_q0_tXEts14DifferentiableRzsAOR_sAOR0_AE0hI13FloatingPointR1_r2_lF".
 for <<debugloc at "<compiler-generated>":0:0>>0  swift                    0x0000000109049675 llvm::sys::PrintStackTrace(llvm::raw_ostream&) + 37
1  swift                    0x0000000109048915 llvm::sys::RunSignalHandlers() + 85
2  swift                    0x0000000109049c62 SignalHandler(int) + 258
3  libsystem_platform.dylib 0x00007fff60157b5d _sigtramp + 29
4  swift                    0x000000010a377052 cmark_strbuf__initbuf + 133885
5  libsystem_c.dylib        0x00007fff600176a6 abort + 127
6  libsystem_c.dylib        0x00007fff5ffe020d basename_r + 0
7  swift                    0x000000010626a268 swift::AutoDiffFunctionInst::getAssociatedFunctionPair(unsigned int) const + 136
8  swift                    0x00000001062d4228 swift::SILInstructionVisitor<(anonymous namespace)::SILVerifier, void>::visit(swift::SILInstruction*) + 47832
9  swift                    0x00000001062c73ec (anonymous namespace)::SILVerifier::visitSILBasicBlock(swift::SILBasicBlock*) + 1196
10 swift                    0x00000001062c15b7 swift::SILFunction::verify(bool) const + 8471
11 swift                    0x00000001062c4553 swift::SILModule::verify() const + 211
12 swift                    0x000000010537fadf performCompile(swift::CompilerInstance&, swift::CompilerInvocation&, llvm::ArrayRef<char const*>, int&, swift::FrontendObserver*, swift::UnifiedStatsReporter*) + 10895
13 swift                    0x000000010537bfe8 swift::performFrontend(llvm::ArrayRef<char const*>, char const*, void*, swift::FrontendObserver*) + 3016
14 swift                    0x000000010532dd12 main + 690
15 libdyld.dylib            0x00007fff5ff723d5 start + 1
```
I temporarily commented out as they're not super critical but forgot to report the bug.

eaplatanios:
Thanks @rxwei! :) I have addressed all the review comments.

Shashi456:
@eaplatanios could we try making a small list as to changes and additions in this commit? Would definitely help track what's there and what's not there

eaplatanios:
@Shashi456 A summary of the changes can be found in apple/swift#24261 . A couple new ops have been added as well to support some missing VJPs, but only a very small number of them. I plan to add new ops that were part of #70 as a separate PR after this one is merged.

Shashi456:
Oh alright, Thanks a lot. Absolutely amazing work btw.

eaplatanios:
Thanks! :)

Shashi456:
@eaplatanios I just saw that you added a dataset file as well, mind briefly telling me what that enables us to do for now?

eaplatanios:
This was actually part of the stdlib before but it was in the compiler repo. It adds initial support for the [TF data API](https://www.tensorflow.org/api_docs/python/tf/data/Dataset).

Sources/DeepLearning/Operators/Basic.swift
rxwei: Could you apply this to `unbroadcast(to:)` and `unbroadcast(toShape:)` as well?

Sources/DeepLearning/DifferentialOperators.swift
rxwei: Interesting. I suspect it's a small typo somewhere. It's fine to remove it since no one seems to be using the ternary version.

eaplatanios: Sounds good. Is it ok to keep them commented out or should I completely remove them for now?

rxwei: Commenting out is fine.

eaplatanios: Ok cool. In this case then this PR should be ready along with apple/swift#24261 and TensorFlow/swift-bindings#26 as everything compiles and all tests pass on my machine.

Shashi456:
@eaplatanios @rxwei will this be merged soon? Because i've been working on #102 and i think this introduces some key changes in that, i wouldn't want to make it a conflict for this PR.

rxwei:
> @eaplatanios @rxwei will this be merged soon? Because i've been working on #102 and i think this introduces some key changes in that, i wouldn't want to make it a conflict for this PR.

We are currently putting a hold on this until Wednesday this week.

Sources/DeepLearning/Operators/Basic.swift
```diff
@@ -578,12 +578,12 @@ public extension PartialRangeThrough where Bound == Int {
 
 public extension Tensor {
     @_fixed_layout @usableFromInline
-    internal struct IndexPath {
+    internal struct IndexPath: Differentiable {
```
rxwei: Hmm why is this necessary? It doesn’t make much mathematical sense.

eaplatanios: It's a hack to make `subscript(_ ranges: TensorRangeExpression...) -> Tensor` differentiable. We could avoid it by allowing something like `.withoutDerivative()` on `IndexPath` maybe.

rxwei: What's the error message if you just mark the original function as `@differentiable(wrt: self)` without making `IndexPath` differentiable? I don't think the workaround is a good direction, and i think there's a way to fix it with smaller workarounds.

eaplatanios: I managed to work around this by using thunks like this:
```swift
self[{IndexPath({ranges.map { $0.tensorRange }}())}()]
```
However, this is pretty ugly and counterintuitive. We should probably find a better way to deal with non-differentiable parts.

eaplatanios:
@pschuh @rxwei @saeta All of the TensorFlow module has now moved here and all tests pass locally.

Friend PR: apple/swift#24452.

eaplatanios:
Minor thing: I haven't fixed the formatting of the `Runtime.swift` file but will do so soon.

Sources/TensorFlow/Core/Utilities.swift
```diff
@@ -0,0 +1,179 @@
...
+
+//===------------------------------------------------------------------------------------------===//
+// Type Aliases
+//===------------------------------------------------------------------------------------------===//
+
+// Before assigning a C pointer to one of the pointer type aliases below, caller should check that 
+// the pointer is not NULL.
+
+/// The `TF_Session *` type.
+public typealias CTFSession = OpaquePointer
```
rxwei: All utility and runtime types and functions should be internal.

eaplatanios: Done.

eaplatanios:
@rxwei when do you think we may be able to make this transition? I have a few follow-up PRs after this that improve performance of some VJPs (mainly related to broadcasting and matrix multiplications), and also add support for several new ops. I'm waiting to first merge this before making new PRs for the other changes so we don't mix too many code changes together, but let me know if you feel this move should happen later and I should go ahead and open other PRs in the current form of the repository.

rxwei:
Since @pschuh is currently working on a merge from master and that refactoring is not blocking the merge, I think it's better to do the merge first before refactoring. We expect to finish the merge this week or early next week. Feel free to open more pending PRs on API improvements! One improvement that is super important for performance is https://github.com/apple/swift/pull/24408, but I've been swamped with other things lately and haven't been able to make progress. It'd be really great if you could help with that!

eaplatanios:
Oh interesting. I actually have made some similar changes in my local copy that are passing all tests and have also tested they improve performance (also using the same raw op). One extra optimization I did was to only do the "unbroadcasting" if the shapes are mismatching. I can make a separate PR with these changes, but if the merge is finished early next week and we can do this refactoring then, I can also add it right after so it also reflects my local changes.

rxwei:
These API changes need not be blocked on merging from upstream or refactoring. You can start a PR to apple/swift now if you'd like.

Shashi456:
@eaplatanios I saw that you added a vjp for max function in this PR. Although I took a look at it, could you tell me if #76 is conforming to the function format, If the VJP is added then that PR can be merged right after this.

So a sample of a pooling layer is : 
```
@_fixed_layout
public struct GlobalMaxPooling1D<Scalar: TensorFlowFloatingPoint>: Layer {
    public init() {}

    ///
    /// - Parameters:
    ///   - input: The input to the layer.
    ///   - context: The contextual information for the layer application, e.g. the current learning
    ///     phase.
    /// - Returns: The output.
    @differentiable
        return input.max(squeezingAxes: 1)
    }
}
```

Sources/TensorFlow/Core/Utilities.swift
```diff
@@ -158,32 +158,32 @@ internal func writeContents(of buffer: UnsafePointer<TF_Buffer>, toFile path: St
 /// This is a generic host-only op that hides the details of its impl in the SIL code. This makes 
 /// reading/writing SIL based compiler unit tests simple.
 @usableFromInline @inline(never)
-internal func _hostOp<T>(_ x: T) {
+public func _hostOp<T>(_ x: T) {
```
rxwei: Why make it public?

eaplatanios:
It’s used by some of the stdlib tests and the only way to make it available
seemed to be by making it public.

rxwei:
We should move all TensorFlowRuntime tests to swift-apis. Also, things like _hostOp are GPE-specific and now they amount to dead code. We can do a full cleanup in apple/swift before making the move to swift-apis.

eaplatanios:
I have already moved all the tests over so i can just remove them from the
stdlib tests. I can also clean up the code as part of this PR.

rxwei:
> I have already moved all the tests over so i can just remove them from the
stdlib tests.

That sounds good!

> I can also clean up the code as part of this PR.

In the interest of separation of concerns, I think it's better to remove them in a separate PR to apple/swift. That'll also make the move smaller.

rxwei:
I busted started cleaning up some of our legacy dead code such as `_TFHoistable`. Will send a PR out later today.

rxwei:
https://github.com/apple/swift/pull/24719 is the first cleanup. The next cleanup will be removing things like `_hostOp`.

eaplatanios:
That's great! Thanks Richard! I integrated the changes from both these PRs in tensorflow/swift-apis#109. I'm currently recompiling and running tests locally.

eaplatanios:
@rxwei all tests pass locally.

rxwei:
To minimize the amount of changes, we start by making tensorflow/swift-bindings a SwiftPM package and have tensorflow/swift-apis depend on it?

eaplatanios:
Sounds good. I’ll do that tomorrow morning.

rxwei:
Thanks Anthony!

eaplatanios:
@rxwei there is an issue with making tensorflow/swift-bindings a SwiftPM package. It depends on the TensorFlow module and there is no simple way to add this dependency without introducing a cycle. In this PR I have currently copied all of the tensorflow/swift-bindings sources inside swift-apis, in a `Bindings` directory under `Sources` and that compiles fine.

rxwei:
It doesn’t have to depend on TensorFlow—it only needs CTensorFlow. In swift-apis, we can use ‘@_exported import TensorFlowBindings’.

eaplatanios:
Currently it does due to `EagerExecution.swift` which uses stuff from the runtime.

eaplatanios:
Specifically, `_TFCOpSetDeviceFromScope` and `_TFCEagerExecute` are used in `EagerExecution.swift`.

rxwei:
Oh I see. It's fine to move them to `TensorFlowBindings`.

eaplatanios:
@rxwei that would involve moving a lot of the runtime code to `TensorFlowBindings`. For example, `_TFCEagerExecute` depends on `_RuntimeConfig` which uses `TraceContext`, etc. Eventually we end up with most of the runtime code being replicated in `TensorFlowBindings `. Another option is to simply move all of the runtime code there, but would we really want to do that?

rxwei:
I see. That would be unfortunate. @pschuh and I discussed and decided to try setting up SwiftPM to run script to pull `RawOpsGenerated.swift` to `Sources/` before building swift-apis.

eaplatanios:
That would be nice, but can we actually set up SwiftPM to do that? If so, how? If not, what is the disadvantage of just having a `Sources/Bindings` directory that contains all the bindings generation code, as it's done in this PR (i.e., without a `swift-bindings` repository)?

rxwei:
`Package.swift` is a normal Swift file that gets run when SwiftPM is run, so we can write arbitrary code in it.

rxwei:
> what is the disadvantage of just having a `Sources/Bindings` directory that contains all the bindings generation code, as it's done in this PR (i.e., without a `swift-bindings` repository)?

The disadvantage is that it would make this PR and future PRs have a really large diff, and the diff is not coming from human-written code.  It would be cleaner not to include generated code but just include a script that generates/downloads it.

eaplatanios:
> The disadvantage is that it would make this PR and future PRs have a really large diff, and the diff is not coming from human-written code. It would be cleaner not to include generated code but just include a script that generates/downloads it.

I see, that's a good point.

> `Package.swift` is a normal Swift file that gets run when SwiftPM is run, so we can write arbitrary code in it.

That's cool, even though kind of hacky. I'll try to write a simple implementation sometime in the next few hours.

pschuh:
It appears you can just place  `print(CommandLine.arguments[0])` at the end of the file and it'll give you the current name of the file. (I don't know how much this can be relied on into the future). I would prefer a swiftpm feature where we can make a particular swift-executable be a pre-requisite rule. I haven't tried this with a swiftpm downloaded from github yet, but it seems ok.

I think this is fine because we can always have a prerequisite script that downloads the needed file (and eventually invokes bazel) as a separate thing if we continue to ship with the toolchain.

eaplatanios:
That's interesting. It is quite tacky and I agree a SwiftPM feature would be better. In general, SwiftPM has very limited functionality when it comes to combining with external build tools. Regarding this PR though, when do we want to perform the download? It's weird to have SwiftPM always do it. Should we only do it if `RawOpsGenerated.swift` cannot be found? In this case, the user would have to delete it manually if they want it updated.

pschuh:
The best solution is to always download it unless the file's hash equals a particular hash. Always download, or download only if not present are fine as initial steps. I think this sort of thing can get more sophisticated later if it is really necessary.

eaplatanios:
I just started looking into this and a couple of issues came up:
1. Are we sure that the code I add in Package.swift will also be run when adding this package as a dependency to another package? If so, what would the result of a call to get the current directory be?
2. Processes created in Swift seem to be sandboxed, not allowing git to create a directory. I can only work around this in MacOS 10.13. Is there another way to resolve this?

rxwei:
> Are we sure that the code I add in Package.swift will also be run when adding this package as a dependency to another package? If so, what would the result of a call to get the current directory be?

Package.swift defines a global variable `package` that is initialized by calling the `Package` initializer. Global variables are all lazy in Swift, but we know that the `package` symbol must be evaluated when SwiftPM runs. So we can wrap `package`'s initialization expression in a custom closure that calls custom logic before returning `Package(...)`.

```swift
func fetchTensorFlowBindings() {
    ...
}

let package = {
    fetchTensorFlowBindings()
    return Package(...)
}()
```

pschuh:
This is not necessary Package.swift is the main module. You do not need global lets.

eaplatanios:
@pschuh This is what I have implemented but I am running into issues with the Swift process not being permitted to create a directory. This is what my code looks like:
```swift
@discardableResult
func cloneTensorFlowBindings() -> Int32 {
    let cwd = FileManager.default.currentDirectoryPath
    let task = Process()
    task.launchPath = "/bin/bash"
    task.arguments = [
        "-c",
        "git clone git@github.com:tensorflow/swift-bindings.git \(cwd)/Sources/TensorFlow/Bindings"]
    task.launch()
    task.waitUntilExit()
    return task.terminationStatus
}

cloneTensorFlowBindings()
```

rxwei:
Does calling `system()` work?

pschuh:
@eaplatanios That seems like it works for me. I think you're working too hard on this for not a lot of value. A simple prerequisite .sh script that downloads the file would work just fine.  I'm also perfectly fine checking this into the repo. I think once we commit this, we can sort out these sort of polish details.

eaplatanios:
Yes sorry. I've been too occupied with three paper deadlines next week and only work on this sparsely when I take breaks. The problem with the above is it does not run on MacOS due to sandboxing, and because it runs on every invocation of SwiftPM it means that you cannot currently use this at all on Mac. @rxwei `system()` also does not work on mac: `error: 'system' is unavailable in Swift: Use posix_spawn APIs or NSTask instead.`.

I agree with @pschuh and feel we should just find a temporary solution that works for both mac and linux and improve this later. The bindings should not be updated too regularly and so this should not be too important. An elegant solution would definitely require improvements to SwiftPM directly though.

eaplatanios:
@pschuh @rxwei I've updated this to reflect all changes in apple/swift and added a separate script for cloning the swift bindings. This means that this cannot currently be added as a dependency to another SwiftPM project but can be compiled independently of apple/swift if the `clone_tensorflow_bindings.swift` script is run first.

Are there plans to merge this PR anytime soon? If not maybe we should just close it and completely revisit the move when it's time, because it's quite tedious to keep updating this PR to reflect the apple/swift changes to the stdlib. I believe it is in a state that can be merged with no side effects.

rxwei:
Now that there's no good way of downloading bindings on the fly, I'm okay with going forward with the suboptional solution—including the bindings file in swift-apis. @pschuh what do you think?

It's better to make this happen in multiple steps though. Could we start a PR that **only** adds all the bindings and make sure they build with `swift build`, and then move all the APIs?

pschuh:
Hi Anthony,
Sorry for the hassle. I think it might make sense to transition towards making incremental progress on this. I think there are some process problems (like not having CI on this repo) that prevent immediate merging. Here is a proposed plan for breaking this process up into steps:
- Pull off your copy of the tests into another PR and get that submitted.
- Move Tensor extensions and their gradients.
- Move Differentiable conditional conformance and the types.
--- All differentiation stuff and most APIs have been moved ---
- Move RawOps.
- Move Dataset Etc.
- Move Tensor<...>
- Move Tracing + eager tensorflow wrappers.
- Move CTensorFlow
I'm more than happy to help with / do any or all of these steps as you see fit.

eaplatanios:
@rxwei @pschuh I agree about the "piecewise" approach to this PR and Parker, it would be great if we could organize this and split it among us. Thanks!

Regarding the proposed breakdown, everything sounds good to me except for one important piece: until we move stuff over from apple/swift there is no way to run `swift build` and `swift test` on this project and so there is no way to test whether what we're doing has bugs/issues (that is due to the resulting duplicate symbols). That is why I took a monolithic approach with this. What do you recommend with respect to this issue?

Also, regarding `CTensorFlow`, how do we go about including the TF binaries? I believe it's convenient that it's now part of apple/swift because the more flexible build configuration there allows us to add and link to the binaries. Also, given that this should rarely be updated, we could also put it in a separate repository that is pulled by apple/swift. A better but more expensive approach would be to modify SwiftPM directly to add support for many of the features that appear frequently when trying to work with other native libraries.

eaplatanios:
> It's better to make this happen in multiple steps though. Could we start a PR that **only** adds all the bindings and make sure they build with `swift build`, and then move all the APIs?

I can make such a PR but it won't just build with `swift build`. You need to first manually run the `clone_tensorflow_bindings.swift` script. I can only make that script run with `swift build` on Linux, but not on MacOS due to sandboxing.

pschuh:
The symbols shouldn't be duplicate if we don't change the module name to TensorFlow.

I plan on doing a few initial PRs (eg: https://github.com/tensorflow/swift-apis/pull/126 and one that moves something) to make sure that everything is set up properly. (Currently the tests are failing as denoted by the x). Once the process has been verified, then we can actually divvy up the work. I believe that by doing an end-to-end move of something from apple/swift to tensorflow/swift-apis, I should end up addressing your concerns about testing / duplicate symbols etc.

CTensorFlow can continue to be built into the toolchain. Possibly this could be baked into the tensorflow/tensorflow repository itself. I believe it will possible to build CTensorFlow out of tree and insert it into the swift toolchain as a final build step. This can all be sorted out later after the Tensorflow migration is complete.

eaplatanios:
@pschuh Thanks for the update! Is there a way to gain access to the test log? I can check why it's failing. Also, I haven't updated this over the last 3 days and some things changed in apple/swift. I can update it and make sure all tests pass locally.

Regarding the duplicate symbols, you will still get these errors if you try compiling the current swift-apis (not the version of this PR) independently using `swift build` because symbols defined in extensions are not scoped by the module. At least that's the error I ran into a while ago. #126 is a good way to test for that, as I imagine you'll probably run into that problem.

And yes, I agree about CTensorFlow.

pschuh:
You click on the x and then select details. The old test for this PR was an internal link, so I re-triggered it. The errors on the simple PR look unrelated. (swift-apis doesn't appear to pass at head since CI was just set up). This PR looks like it is triggering a lurking toolchain bug. I suspect we'll trigger that same bug at a later point when we get more things committed.

Here are some logs from PRs:
https://source.cloud.google.com/results/invocations/3a8fc401-c7d1-4d05-937a-a78422a49b79/targets/s4tf%2Fapis%2Fubuntu%2Fpresubmit/log
https://source.cloud.google.com/results/invocations/8c61465e-f4dd-4d0f-bc4a-261c91379a56/targets/s4tf%2Fapis%2Fubuntu%2Fpresubmit/log

eaplatanios:
Ok let me update this PR with the latest changes from apple/swift and we can try rerunning, because I think some of these failures are due to the recent autodiff changes.

eaplatanios:
@pschuh I updated this PR and the corresponding apple/swift PR and I'm currently compiling locally. I'll let you know if tests pass locally and then we can maybe try the CI again.

pschuh:
Same failure. I think we need someone to look at this from the compiler side. It has to do with cross-module AD which is currently incomplete.

eaplatanios:
How about we work on getting this PR to compile and pass tests first and then maybe breaking it down to parts? For now, I'll try to reproduce the failure locally and see if I can fix it.

pschuh:
Sounds good. Making small reproducers and publishing those somehow is probably a useful first step.

eaplatanios:
The tests on apple/swift just finished on my machine and everything passes there when using the changes introduced in this PR. I'll try to build a toolchain now and just test this PR independently to see if I can get that same failure.

eaplatanios:
@pschuh So this kind of error goes away for me locally if I rename the package to something other than `TensorFlow`. I think this may be due a conflict with the `TensorFlow` module that is compiled as part of stdlib. However, if I rename it then I get some errors related to `TensorGroup` conformance derivations because these only apply to the stdlib `TensorFlow` module `TensorGroup` and not to the `TensorGroup` defined in this package. Not sure if this description is helpful, but does this make sense to you? If so, how could we avoid this name collision? One way would be to not have `TensorFlow` in stdlib but rather require users to add it as a dependency using SwiftPM, but that may raise other issues for e.g. users that use Jupyter, etc.

pschuh:
Ah, that has to do with TensorGroup derived conformances. The other problem (the Jupyter import) can be solved by including the necessary Tensorflow .so file directly with the installation. It does not actually matter if it is built by the swift repository itself.

eaplatanios:
Sorry I'm not sure if I described the issue correctly. I can only avoid the error you get on the CI server if I rename this module to something other than `TensorFlow` so that it doesn't clash with the stdlib `TensorFlow`, or if we simply do not include the `TensorFlow` module in stdlib. I personally prefer the latter approach but that makes it harder for people to use S4TF as they have to both install and use the right toolchain, but they also have to make sure they declare a SwiftPM dependency to the module defined in this repository.

eaplatanios:
The latter approach also avoids the derived conformance errors because the module can remain named `TensorFlow`.

rxwei:
@eaplatanios How about starting with moving all the high-level stuff and not touching derived conformances, runtime, or bindings for now?

pschuh:
Yes, there will be issues due to the duplicate name. Lets get as far as we can before handling those.

eaplatanios:
@rxwei That would also require keeping this module as `DeepLearning` and not renaming to `TensorFlow`. Yes, I think it's a good idea. I'll give it a shot.

eaplatanios:
Actually not really. `TensorGroup` requires `Tensor` and `ShapedArray` among other things and so it propagates to quite a lot of stuff. Maybe we can just make the derived conformance for `TensorGroup` and `TensorArrayProtocol` support two modules (i.e., `TensorFlow` and `DeepLearning`). In that case, I believe I can just rename the module in this PR back to `DeepLearning` and everything should work. Is there a way we can do that for the derived conformances?

eaplatanios:
@rxwei @pschuh I think I have a working version now with the modification in the conformance derivation that I mentioned in my previous post. I'm testing locally and if all works out fine I'll push changes later tonight.

rxwei:
Great!

Shashi456:
@eaplatanios, so I saw that this PR has vjps for min,max functions. When you are breaking this PR down, can I push those vjps from #76, that's the only bottleneck in that PR currently. Only if you are okay with it.

pschuh:
@eaplatanios https://github.com/tensorflow/swift-apis/pull/126 is in, maybe make a PR for all your other copied tests?

eaplatanios:
@Shashi456 Yes, that's totally fine. :)

eaplatanios:
@rxwei So after incorporating the latest changes, I only get one error when trying to compile this independently with `swift build` that as to do with `AutoDiffParameterIndices` (we were also getting something like this in the past). More specifically, I get `EXC_BAD_ACCESS` when calling `SmallBitVector::isSmall` from within `AutoDiffParameterIndices::getLowered` in `SILFunctionBuilder::addFunctionAttributes`. You should be able to reproduce this by building the toolchain from apple/swift#24452 and then using `swift build` for this PR. If we resolve this, then we can either merge this directly, or I can make separate PRs for its parts. However, it would be good to resolve this first so I can then use this PR as a working reference.

rxwei:
I’ll try to reproduce tonight.

eaplatanios:
@rxwei I think this was fixed in my previous PR for moving the stdlib [here](https://github.com/eaplatanios/swift-language/blob/23c3ad71d875be4060aa9ed1bd14190f8794650c/lib/SIL/SILFunctionBuilder.cpp#L90). I remember us discussing this before. I'll go ahead and test if things work out when I add this if statement.

eaplatanios:
@rxwei I can confirm that now apple/swift#24452 compiles and all tests pass, and that building a toolchain and using it to do `swift build` also runs successfully for this PR. I need to make a few minor updates to some of the tests, but other than that everything looks fine. Do we still want to try and break this down in this case or should we just go ahead and merge this PR? I could do either one, although other than having a separate PR for copying the test and then moving everything over, I'm not sure how much more it makes sense to break it down.

eaplatanios:
@rxwei @pschuh I may actually also be able to remove the `TensorGroup` hack in stdlib, but I'd prefer to do that as a separate PR after we merge the changes introduced here.

eaplatanios:
@rxwei The only remaining error when I run `swift test -Xlinker -ltensorflow` is that `_InitTensorFlowRuntime` is missing. Where is that symbol defined? `swift build` runs successfully.

rxwei:
> @rxwei @pschuh I may actually also be able to remove the `TensorGroup` hack in stdlib, but I'd prefer to do that as a separate PR after we merge the changes introduced here.

Sounds good. Let's try to break the changes into small incremental pieces.

rxwei:
> @rxwei The only remaining error when I run `swift test -Xlinker -ltensorflow` is that `_InitTensorFlowRuntime` is missing. Where is that symbol defined? `swift build` runs successfully.

Where is that error coming from? Which function is referring to that symbol?

eaplatanios:
> Where is that error coming from? Which function is referring to that symbol?

It's being called from [here](https://github.com/apple/swift/blob/8adfbd3e53dd6e75e0858e9854a8e2886d797b2b/stdlib/public/TensorFlow/CompilerRuntime.swift#L583).

eaplatanios:
@rxwei I added a couple separate PRs, #136 for the tests, #137 for the bindings fetching script, and apple/swift#25091 for replacing `InitTensorFlowRuntime` with a Swift implementation. I believe that once these are merged, then we may be able to merge this one directly since it mostly consists of moving over everything from `apple/swift`. I'm not sure how a partial move would work without adding a significant amount of work to properly separate things for each PR. Please let me know what you think.

rxwei:
Thanks! @bgogul and I will help you merge these PRs this week.

eaplatanios:
@rxwei @bgogul All tests pass locally for me and I have started the CI tests. Tests will probably run successfully with the current toolchain but I tested locally with one built from the corresponding PR apple/swift#24452.

eaplatanios:
The CI tests seem to fail and I can't tell why. Is there a way to test wit the toolchain from apple/swift#24452 directly?

rxwei:
> The CI tests seem to fail and I can't tell why. Is there a way to test wit the toolchain from [apple/swift#24452](https://github.com/apple/swift/pull/24452) directly?

There's no way to do that until the PRs have been merged. Let me build a toolchain locally and test things.

eaplatanios:
Sounds good, thanks! One minor thing I just noticed. There is a test in `LayerTests` that fails due to a difference in some numbers. I'm not sure which one is correct and why the numbers are different now, but I pushed the changed numbers (lines 203-212 in `Tests/TensorFlowTests/LayerTests.swift`) so you can take a look.

rxwei:
> Sounds good, thanks! One minor thing I just noticed. There is a test in `LayerTests` that fails due to a difference in some numbers. I'm not sure which one is correct and why the numbers are different now, but I pushed the changed numbers (lines 203-212 in `Tests/TensorFlowTests/LayerTests.swift`) so you can take a look.

 I think verifying the numeric results against Python TF is the best way to know.

rxwei:
I'm getting a bunch of undefined identifier errors when building a toolchain using this PR and the corresponding apple/swift PR.
```console
/usr/local/src/swift-build/tensorflow-swift-apis/Sources/DeepLearning/Optimizer.swift:175:60: error: use of unresolved identifier 'Tensor'
        for kp in alpha.recursivelyAllWritableKeyPaths(to: Tensor<Float>.self) {
                                                           ^~~~~~
```

Shashi456:
@eaplatanios  @rxwei Sorry to suddenly jump into the convo. So the RNN layer tests seems to have some sorta variability in output and with #130 we did change those values again. Could we just comment it out for now and look into it later?

eaplatanios:
@rxwei that’s because you’re trying to build it with the old checkout of swift-apis. You’d need to checkout this PR in the tensorflow-swift-apis directory.

eaplatanios:
@Shashi456 I don’t see why there should be any variability. Thanks for that information. I’ll try to look into this tomorrow.

rxwei:
> @rxwei that’s because you’re trying to build it with the old checkout of swift-apis. You’d need to checkout this PR in the tensorflow-swift-apis directory.

Oops, just figured that. I fetched the PR but didn't `checkout`.

rxwei:
All tests are passing. I'm about to merge this PR, update the commit hash in the other PR, merge it, and start a new draft PR to test CI once a new toolchain is built.

Shashi456:
@rxwei I'm sorry for asking this again, but what does this PR exactly do? and why was it needed in the long run?

eaplatanios:
@Shashi456 This PR mainly moves all of the `TensorFlow` module in stdlib from apple/swift (i.e., the compiler repository) to this repository. There is also some other minor changes and support for a couple new ops and VJPs, as well as a restructuring of the `Operators.swift` file to a directory.

This move allows us to makes changes to the stdlib and test them without requiring us to recompile the Swift compiler (which is expensive) and it also separates the compiler code from the stdlib code more clearly.

PR #110, opened 2019-04-22
[Layer] Add GRUCell.

rxwei:
* Add `GRUCell`.
* Add `@_fixed_layout` to RNN types.
* Improve doc comments on RNN types.

Issue:
* `GRUCell` is currently triggering a SIL verification failure ([TF-452](https://bugs.swift.org/browse/TF-452)) when built as part of the Swift package.

Todo before merging:
* Add tests.

### DIFF
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -1470,13 +1470,14 @@ public struct GRUCell<Scalar: TensorFlowFloatingPoint>: RNNCell {
         return Output(output: newState, state: newState)
     }
 }
-=======
+
+// TODO: Doc comment.
 public struct RNN<Cell: RNNCell>: Layer {
     public typealias Input = [Cell.TimeStepInput]
     public typealias Output = [Cell.TimeStepOutput]
 
     public var cell: Cell
-    
+
     public init(_ cell: @autoclosure () -> Cell) {
         self.cell = cell()
     }
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -1464,8 +1464,8 @@ public struct GRUCell<Scalar: TensorFlowFloatingPoint>: RNNCell {
         let gateInput = input.input.concatenated(with: input.state, alongAxis: 1)
         let 𝒓 = sigmoid(matmul(resetWeight, gateInput) + resetBias)
         let 𝒛 = sigmoid(matmul(updateWeight, gateInput) + updateBias)
-        let 𝒏 = matmul(inputHiddenWeight, input.input) + inputHiddenBias
-            + 𝒓 * (matmul(hiddenHiddenWeight, input.state) + hiddenHiddenBias)
+        let 𝒏 = tanh(matmul(inputHiddenWeight, input.input) + inputHiddenBias
+            + 𝒓 * (matmul(hiddenHiddenWeight, input.state) + hiddenHiddenBias))
         let newState = (1 - 𝒛) * 𝒏 + 𝒛 * input.state
         return Output(output: newState, state: newState)
     }
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -237,29 +237,7 @@ public extension Dense {
     ///   - activation: The activation function to use. The default value is `identity(_:)`.
     ///   - generator: The random number generator for initialization.
     ///
-    /// - Note: Use `init(inputSize:outputSize:activation:seed:)` for faster random initialization.
-    init<G: RandomNumberGenerator>(
-        inputSize: Int,
-        outputSize: Int,
-        activation: @escaping Activation = identity,
-        generator: inout G
-    ) {
-        self.init(weight: Tensor(glorotUniform: [inputSize, outputSize],
-                                 generator: &generator),
-                  bias: Tensor(zeros: [outputSize]),
-                  activation: activation)
-    }
-
-    init(inputSize: Int, outputSize: Int, activation: @escaping Activation = identity) {
-      self.init(inputSize: inputSize, outputSize: outputSize, activation: activation,
-                generator: &PhiloxRandomNumberGenerator.global)
-    }
-}
-
-public extension Dense {
-    /// Creates a `Dense` layer with the specified input size, output size, and element-wise
-    /// activation function. The weight matrix is created with shape `[inputSize, outputSize]` and
-    /// is initialized using Glorot uniform initialization with the specified seed. The bias vector
+    /// - Note: Use `init(inputSize:outputSize:activation:seed. The bias vector
     /// is created with shape `[outputSize]` and is initialized with zeros.
     ///
     /// - Parameters:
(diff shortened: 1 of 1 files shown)


rxwei:
cc @tanmayb123 @eaplatanios

Shashi456:
@rxwei could we do something about the naming of variables in recurrent layers? They make sense. But take a hit as far as readability is concerned.

rxwei:
I absolutely agree that the property names are too long and that they are not fluent when used, and [striving for fluent usage](https://swift.org/documentation/api-design-guidelines/#strive-for-fluent-usage) is super important to Swift APIs. Do you have some suggestions? I was thinking of naming weights along the lines of `wx`, `bx`, etc. I plan to rename them in a future PR.

Shashi456:
@rxwei Yeah, I think naming the variable in term of their mathematical notations could make sense, and could also provide a direction for documentation, 

For example, if we wrote `inputHiddenWeight` as `W_ih` and `hiddenHiddenWeight` as `W_hh`, the documentation could include a short mention of what the variables mean.

 We need not mention the equations themselves. What i'm sorta loving about these implementations and s4tf is we can sorta see the maths translate to code neatly.

rxwei:
Agreed! We are on the same page.

Shashi456:
@rxwei should i put in a PR for LSTM and RNN cell renaming?

Sources/DeepLearning/Layer.swift
```diff
@@ -1404,12 +1409,75 @@ public struct LSTMCell<Scalar: TensorFlowFloatingPoint>: RNNCell, VectorNumeric
...
+        self.inputHiddenWeight = Tensor(glorotUniform: [inputSize, hiddenSize], seed: seed)
+        self.inputHiddenBias = Tensor(zeros: biasShape)
+        self.hiddenHiddenWeight = Tensor(glorotUniform: [hiddenSize, hiddenSize], seed: seed)
+        self.hiddenHiddenBias = Tensor(zeros: biasShape)
+    }
+
+    /// Returns the output obtained from applying the layer to the given input.
+    ///
+    /// - Parameters:
+    ///   - input: The input to the layer.
```
dan-zheng: Minor note (totally need not be addressed here): `- Parameter input: ` and `- Parameters: - input:` are used inconsistently in `Layer.swift`.

Sources/DeepLearning/Layer.swift
```diff
@@ -1404,12 +1409,75 @@ public struct LSTMCell<Scalar: TensorFlowFloatingPoint>: RNNCell, VectorNumeric
...
+    /// - Returns: The hidden state.
+    @differentiable
+    public func call(_ input: Input) -> Output {
+        let gateInput = input.input.concatenated(with: input.state, alongAxis: 1)
+        let 𝒓 = sigmoid(matmul(resetWeight, gateInput) + resetBias)
+        let 𝒛 = sigmoid(matmul(updateWeight, gateInput) + updateBias)
+        let 𝒏 = matmul(inputHiddenWeight, input.input) + inputHiddenBias
```
dan-zheng: Does `𝒏` need an outer `tanh`?

[Reference:](https://towardsdatascience.com/understanding-gru-networks-2ef37df6c9be)
image

rxwei:
> @rxwei should i put in a PR for LSTM and RNN cell renaming?

That'd be great!

Sources/DeepLearning/Layer.swift
rxwei: Most layers are using a separate line for the `input` parameter. This is legacy left by the fact that we had a context before. Will change later.

Sources/DeepLearning/Layer.swift
rxwei: Good catch. Fixed.

Sources/DeepLearning/Layer.swift
```diff
@@ -1404,12 +1409,75 @@ public struct LSTMCell<Scalar: TensorFlowFloatingPoint>: RNNCell, VectorNumeric
...
+    ///   - inputSize: The number of features in 2-D input tensors.
+    ///   - hiddenSize: The number of features in 2-D hidden states.
+    ///   - seed: The random seed for initialization. The default value is random.
+    public init(inputSize: Int, hiddenSize: Int,
+                seed: (Int64, Int64) = (Int64.random(in: Int64.min..<Int64.max),
+                                        Int64.random(in: Int64.min..<Int64.max))) {
+        let concatenatedHiddenSize = inputSize + hiddenSize
+        let weightShape: TensorShape = [concatenatedHiddenSize, hiddenSize]
+        let biasShape: TensorShape = [hiddenSize]
+        self.resetWeight = Tensor(glorotUniform: weightShape, seed: seed)
```
eaplatanios: It would be nice if we find ways to expose the initialization of the model parameters. Maybe by having additional initializers that allow you to provide initial values for all parameters. Or even by adding these as arguments to this initializer (e.g., `initialResetWeight: Tensor<Scalar>? = nil`).

eaplatanios: Also, I find it awkward that sometimes we use a single `UInt64` for seeds and sometimes pairs of two `Int64`s (that represent TensorFlow graph- and op-level seeds I assume). Why don't we just stick to a single `UInt64` everywhere to be consistent and simplify the API?

Sources/DeepLearning/Layer.swift
```diff
@@ -1404,12 +1409,75 @@ public struct LSTMCell<Scalar: TensorFlowFloatingPoint>: RNNCell, VectorNumeric
     }
 }
 
+/// A gated recurrent unit (GRU) cell.
+///
+/// Reference: [Gated recurrent unit](https://en.wikipedia.org/wiki/Gated_recurrent_unit).
+@_fixed_layout
+public struct GRUCell<Scalar: TensorFlowFloatingPoint>: RNNCell {
+    public var resetWeight, updateWeight, inputHiddenWeight, hiddenHiddenWeight: Tensor<Scalar>
```
Shashi456:
```suggestion
    public var W_r, W_u, W_ih, W_hh: Tensor<Scalar>
```
In accordance with #113, we could just modify these here.

Sources/DeepLearning/Layer.swift
```diff
@@ -1404,12 +1409,75 @@ public struct LSTMCell<Scalar: TensorFlowFloatingPoint>: RNNCell, VectorNumeric
...
+    public var resetBias, updateBias, inputHiddenBias, hiddenHiddenBias: Tensor<Scalar>
```
Shashi456:
```suggestion
    public var b_r, b_u, b_ih, b_hh: Tensor<Scalar>
```
and similarly, in the rest of the code, what do you think?

Sources/DeepLearning/Layer.swift
Shashi456: @eaplatanios i agree with the initializers and having more options, i think that's not just in this layer but something to be done for all layers. but as of now only glorot has been written. I also opened #111 in that regard.

Shashi456: @eaplatanios i agree with the initializers and having more options, i think that's not just in this layer but something to be done for all layers. but as of now only glorot has been written. I also opened #111 in that regard.

Shashi456:
@rxwei is this PR still stuck with the bug you mentioned?

Shashi456:
@rxwei does the blocker SIL failure still exist on this?

Shashi456:
@rxwei should we close this since #417 is working on it as  well?

dhasl002:
@Shashi456 @rxwei I vote to close this. #417 is close to being finished, I am confident that PR will be merged in soon.

burmako:
After a team discussion today, we decided that we'll be closing this PR in favor of #417. Thank you very much for your work!

PR #113, opened 2019-04-23
Modifying recurrent layer for readability

Shashi456:
As discussed [here](https://github.com/tensorflow/swift-apis/pull/110#issuecomment-485325859).

This is a WIP. I'd like to discuss about a few things before changing them.

In the RNN, `h_W`, `h_b` correspond to `Weight` and `bias`. The `h_` prefix for hidden. 

I was unsure about what to write `concatenatedInputSize`. 

Finally, I wanted to write `newState` as `c_t`, but didn't because the naming convention in that case was for cell state at time t. But since we are talking about a single cell, any ideas as to what we can change that to. 

In the LSTM, It was much easier, All gates were made into short forms with weight changed to `_W` and bias changed to `_b`.
Finally, I was wondering how to italicize the text for `u`,`o`,`f` like it was done in #110 .

`newHiddenState` and `newCellState` can we written as `h_t` and `c_t`. I just wanted to confirm if that was okay. 

Should this convention be added to the comments as well?

I'm sorry for so many question.


Shashi456:
cc: @rxwei , @dan-zheng.

Sources/DeepLearning/Layer.swift
```diff
@@ -1282,8 +1282,8 @@ public extension RNNCell {
 
 /// A Simple RNN Cell.
 public struct SimpleRNNCell<Scalar: TensorFlowFloatingPoint>: RNNCell, VectorNumeric {
-    public var weight: Tensor<Scalar>
-    public var bias: Tensor<Scalar>
+    public var h_W: Tensor<Scalar>
```
rxwei: I think it would be easier to read if this was named “W_h” since this is how they are defined in math.

Sources/DeepLearning/Layer.swift
```diff
@@ -1324,15 +1324,15 @@ public struct SimpleRNNCell<Scalar: TensorFlowFloatingPoint>: RNNCell, VectorNum
...
+        let newState = tanh(matmul(concatenatedInput, h_W) + h_b)
         return Output(output: newState, state: newState)
     }
 }
 
 /// An LSTM Cell.
 public struct LSTMCell<Scalar: TensorFlowFloatingPoint>: RNNCell, VectorNumeric {
-    public var inputWeight, updateWeight, forgetWeight, outputWeight: Tensor<Scalar>
-    public var inputBias, updateBias, forgetBias, outputBias: Tensor<Scalar>
+    public var i_W, u_W, f_W, o_W: Tensor<Scalar>
```
rxwei: Same here.

Shashi456:
@rxwei any comments on,
> newHiddenState and newCellState can we written as h_t and c_t. I just wanted to confirm if that was okay.

and this 
>I was unsure about what to write concatenatedInputSize.

Sources/DeepLearning/Layer.swift
```diff
@@ -1309,9 +1309,9 @@ public struct SimpleRNNCell<Scalar: TensorFlowFloatingPoint>: RNNCell, VectorNum
                 seed: (Int64, Int64) = (Int64.random(in: Int64.min..<Int64.max),
                                         Int64.random(in: Int64.min..<Int64.max))) {
         let concatenatedInputSize = inputSize + hiddenSize
-        self.weight = Tensor(glorotUniform: [concatenatedInputSize, hiddenSize],
+        self.W_h = Tensor(glorotUniform: [concatenatedInputSize, hiddenSize],
                              seed: seed)
```
rxwei: Could you move `seed: seed)` to the end of the previous line?

Shashi456:
@rxwei any other notable changes in this? Should we wait for @dan-zheng to review and comment on this?

Shashi456:
@rxwei can this be merged?

rxwei:
There are four test failures. Could you please look into fixing them?

```swift
Test Case '-[DeepLearningTests.LayerTests testRNN]' started.
/Users/rxwei/Development/ML/swift-apis/Tests/DeepLearningTests/LayerTests.swift:106: error: -[DeepLearningTests.LayerTests testRNN] : XCTAssertEqual failed: ("[[[ -0.00262943, -0.005866742,  0.044919778,   0.20036437]], [[0.066890605, 0.049586136, 0.024610005,  0.09341654]], [[0.065792546, 0.009325638,  0.06439907, 0.114802904]], [[  0.055909205, 0.00035158166,   0.054020774,    0.09812111]]]") is not equal to ("[[[-0.0026294366, -0.0058668107,    0.04495003,    0.20311214]], [[ 0.06788494, 0.050665878,  0.02415526,  0.09249911]], [[ 0.06621192, 0.009049267, 0.065047316,  0.11534518]], [[   0.05612204, 0.00022032857,    0.05407162,    0.09784105]]]")
/Users/rxwei/Development/ML/swift-apis/Tests/DeepLearningTests/LayerTests.swift:111: error: -[DeepLearningTests.LayerTests testRNN] : XCTAssertEqual failed: ("[[          0.0,           0.0,           0.0,           0.0],
 [-0.0051169936,  0.0014167001,  0.0074189613,   0.017496519],
 [ -0.010233987,  0.0028334002,  0.0148379225,   0.034993038],
 [ -0.015350982,  0.0042501003,   0.022256885,    0.05248956],
 [          0.0,           0.0,           0.0,           0.0],
 [          0.0,           0.0,           0.0,           0.0],
 [          0.0,           0.0,           0.0,           0.0],
 [          0.0,           0.0,           0.0,           0.0]]") is not equal to ("[[          0.0,           0.0,           0.0,           0.0],
 [-0.0051278225,  0.0013102926,    0.00740262,   0.018119661],
 [ -0.010255645,  0.0026205853,    0.01480524,   0.036239322],
 [ -0.015383467,   0.003930878,    0.02220786,   0.054358985],
 [          0.0,           0.0,           0.0,           0.0],
 [          0.0,           0.0,           0.0,           0.0],
 [          0.0,           0.0,           0.0,           0.0],
 [          0.0,           0.0,           0.0,           0.0]]")
/Users/rxwei/Development/ML/swift-apis/Tests/DeepLearningTests/LayerTests.swift:120: error: -[DeepLearningTests.LayerTests testRNN] : XCTAssertEqual failed: ("[-0.051169936,  0.014167001,   0.07418961,   0.17496519]") is not equal to ("[-0.051278222,  0.013102926,    0.0740262,   0.18119662]")
Test Case '-[DeepLearningTests.LayerTests testRNN]' failed (0.016 seconds).
Test Case '-[DeepLearningTests.LayerTests testSimpleRNNCell]' started.
/Users/rxwei/Development/ML/swift-apis/Tests/DeepLearningTests/LayerTests.swift:95: error: -[DeepLearningTests.LayerTests testSimpleRNNCell] : XCTAssertEqual failed: ("[[0.9921227, 0.9999934, 0.9921227, 0.9999934, 0.9921227]]") is not equal to ("[[  2.76649, 6.2999997,   2.76649, 6.2999997,   2.76649]]")
Test Case '-[DeepLearningTests.LayerTests testSimpleRNNCell]' failed (0.001 seconds).
Test Suite 'LayerTests' failed at 2019-04-25 18:52:18.315.
	 Executed 10 tests, with 4 failures (0 unexpected) in 0.057 (0.057) seconds
```

Shashi456:
Oh I'm sorry I had no idea. I had been relying on docker for that. I'll fix them.

rxwei:
Thanks. The docker CI is currently down, so we need to manually run `swift test`.

Shashi456:
@rxwei I think that should solve it, there was a mismatch since i didnt reflect the change in the test file.

Sources/DeepLearning/Layer.swift
```diff
@@ -1292,8 +1292,8 @@ public extension RNNCell {
 
 /// A simple RNN cell.
 public struct SimpleRNNCell<Scalar: TensorFlowFloatingPoint>: RNNCell, VectorNumeric {
-    public var weight: Tensor<Scalar>
-    public var bias: Tensor<Scalar>
+    public var W_h: Tensor<Scalar>
+    public var b_h: Tensor<Scalar>
 
     @noDerivative public var stateShape: TensorShape {
         return TensorShape([1, weight.shape[1]])
```
saeta: I think this line doesn't compile, and should be changed to:

```suggestion
        return TensorShape([1, W_h.shape[1]])
```

Sources/DeepLearning/Layer.swift
```diff
@@ -1331,15 +1330,15 @@ public struct SimpleRNNCell<Scalar: TensorFlowFloatingPoint>: RNNCell, VectorNum
...
 
 /// An LSTM cell.
 public struct LSTMCell<Scalar: TensorFlowFloatingPoint>: RNNCell, VectorNumeric {
-    public var inputWeight, updateWeight, forgetWeight, outputWeight: Tensor<Scalar>
-    public var inputBias, updateBias, forgetBias, outputBias: Tensor<Scalar>
+    public var W_i, W_u, W_f, W_o: Tensor<Scalar>
+    public var b_i, b_u, b_f, b_o: Tensor<Scalar>
 
     @noDerivative public var stateShape: TensorShape {
         return TensorShape([1, inputWeight.shape[1]])
```
saeta: I think this line doesn't compile, and should be changed to:

```suggestion
        return TensorShape([1, W_i.shape[1]])
```

Shashi456:
@saeta I think this works now.

Shashi456:
@rxwei althought i've made all the required changes, i seem to be getting a mismatch test error where the output of my layer is
```
[[0.9921227, 0.9999934, 0.9921227, 0.9999934, 0.9921227]
```
while the one in the test is : `[[2.76649, 6.2999997, 2.76649, 6.2999997, 2.76649]])`

MostafaGazar:
I fail to see how this improves the readability of this Layer. If the goal is to build Swift friendly APIs then it would make sense to follow the common naming and code styling conventions adapted by the Swift community. 

For upper level (exposed) APIs consistency with Keras would make sense but for implementations writing swifty code would be easier to maintain in the long run in my opinion.

rxwei:
> I fail to see how this improves the readability of this Layer. If the goal is to build Swift friendly APIs then it would make sense to follow the common naming and code styling conventions adapted by the Swift community.
> 
> For upper level (exposed) APIs consistency with Keras would make sense but for implementations writing swifty code would be easier to maintain in the long run in my opinion.

Thanks for raising your concerns. As a Swift API designer who absolutely enjoys Swift naming conventions, I don't quite agree with your view. A key part of the [Swift API Design Guidelines](https://swift.org/documentation/api-design-guidelines) is "strive for fluent usage". The [Use Terminology Well](https://swift.org/documentation/api-design-guidelines/#use-terminology-well) section has a great explanation on how to strike a balance between general Swift names and established terms of art.

While avoiding abbreviations is generally the code style for Swift application projects and a _lot_ of our carefully curated APIs, the goal there is clarity and fluency, not merely consistency. Deep learning layers often have a canonical definition on papers, and their Wikipedia pages often use short mathematical symbols. When code looks like their canonical definition in paper, it is immediately readable and fluent.

To quote the Swift API Design Guidelines:
> * Stick to the established meaning if you do use a term of art.
>
>     The only reason to use a technical term rather than a more common word is that it precisely expresses something that would otherwise be ambiguous or unclear. Therefore, an API should use the term strictly in accordance with its accepted meaning.
>     - **Don’t surprise an expert:** anyone already familiar with the term will be surprised and probably angered if we appear to have invented a new meaning for it.
>     - **Don’t confuse a beginner:** anyone trying to learn the term is likely to do a web search and find its traditional meaning.
> * Embrace precedent. Don’t optimize terms for the total beginner at the expense of conformance to existing culture.
>
>     It is better to name a contiguous data structure Array than to use a simplified term such as List, even though a beginner might grasp of the meaning of List more easily. Arrays are fundamental in modern computing, so every programmer knows—or will soon learn—what an array is. Use a term that most programmers are familiar with, and their web searches and questions will be rewarded.
>
>     Within a particular programming domain, such as mathematics, a widely precedented term such as sin(x) is preferable to an explanatory phrase such as verticalPositionOnUnitCircleAtOriginOfEndOfRadiusWithAngle(x). Note that in this case, precedent outweighs the guideline to avoid abbreviations: although the complete word is sine, “sin(x)” has been in common use among programmers for decades, and among mathematicians for centuries.

These recommendations from the Swift API Design Guidelines heavily influenced some of the important decisions made in Swift Evolution reviews of standard library numerics. To quote some proposal review decision rational for [SE-0251](https://forums.swift.org/t/accepted-with-modifications-se-0251-simd-additions/22801):

> It is a primary goal of Swift to support fluent and expressive libraries. The API design guidelines 1 are explicit that the highest priority in library design is clarity at the point of use. Consistency is only a means to that end, not a goal in itself. In many cases, using a uniform language construct like method calls to spell an API is good design because the consistency contributes to clarity: programmers already understand how method calls work, and method calls combine naturally with features like optional chaining instead of requiring circumlocutions that would harm the overall clarity of the code. But any and all will often be used directly on the result of a binary comparison, making a method call awkward and harder to read, and they will often be used directly as if conditions and rarely in conjunction with optional chaining. In this case, the Core Team feels that a more "DSL-like" approach like that proposed for any and all creates an overall more fluent API. 

Specifically about this PR that changes RNNs to use terms of art, the use of non-camel-case variable names are definitely weird in the context of most other Swift application APIs, but the code **is** more fluent and more mathematically relatable in this domain. As Swift has just take off in the space of machine learning, we should absolutely strive to find the right balance between these naming options while always remembering that our goal is to design expressive and fluent APIs for machine learning.

MostafaGazar:
Thank you @rxwei for your detailed response, it was informative. I am not 100% sold but I get your point.

rxwei:
Thanks! I'm not 100% sold on `W_i` either because there are other options such as `wh`, `Wh`. It would be very useful to start a conversation about "naming mathematical variables in numerical Swift code" at some point with the Swift community and the Swift for TensorFlow community and standardize things.

Shashi456:
@rxwei @dan-zheng can you maybe run the swift test once and tell me what values you get for the tests, i seem to have some discrepancies when i do it locally, for example : 

image

and
image


The latter though, feels like the system is rounding off somewhere where it's unwanted.

liberium:
@rxwei I have concern against these changes. They've made the code harder to read.
It's not obvious for a new reader what do these variables represent:
```swift
public var W_i, W_u, W_f, W_o: Tensor<Scalar>
public var b_i, b_u, b_f, b_o: Tensor<Scalar>
```
There is no convention on them in the literature. The RNN Wikipedia article doesn't state them at all. We do not want the reader to guess variables meaning, right? So, a comment is needed. And then we need to maintain that comment to keep it in sync with the code. And the reader should keep variable meanings he learned from the comment in short-term memory while reading the code. And learn again each time he forgot them once he has not read the code for a while.
All this doesn't sound like good DX to me and definitely doesn't make reading easier. It doesn't even make writing easier because of autocompletion.
I propose to keep the code self-documented and not to bind names to unconventional notation that need to be explained. I'm for rejecting these changes.

Shashi456:
@liberium I will write comments for these once they are approved and i go update the respective changes.
Just one point, I would like to mention is anyone researching in Deep learning would have a fair idea of the four gates within an LSTM and the respective mathematical equations. The earlier names, although good were too long, So in my opinion it would be better if we wrote the code in terms of the mathematical equations they represented and this has been an attempt at that. When i read this code, since i come from a research perspective, this makes a lot of sense to me. So I will either write the equations themselves in the comments or link to the paper.

rxwei:
While I agree mostly with @Shashi456, I'm not fully sold on either end. Notations in publications are not unified either. Let's keep the current state as is for now, and find the most productive way to find a consensus. I plan to start a serious investigation about naming some time next month, and post on the mailing list and Swift forums.

Shashi456:
@rxwei should I close this PR then?
It's just that whatever notations ive seen across various framework in deep learning, have a short-form notation like this and they have just explained what they stood for in the comments.

rxwei:
> @rxwei should I close this PR then?

Sounds good!

rxwei:
Also, thank you @Shashi456 for driving this.  We'll get back to it in a month or two.

liberium:
@Shashi456 I agree on using `W`, `x`, and `b` as arguments to a linear classifier because they are absolutely conventional across the literature.
But it's not the case here and some kind of compromise should be accepted.
It could be a hybrid naming scheme:
```swift
public var W_in, W_update, W_forget, W_out: Tensor<Scalar>
public var b_in, b_update, b_forget, b_out: Tensor<Scalar>
```

liberium:
Thank you for the detailed answer.
> This is from the wikipedia for LSTMs :
> image
> 
> and this is from the wikepedia for RNNs:
> 
> image
> 
> So in a very straightforward manner, W_f, W_o correspond to the weights for forget, and output gates, so on so forth.

I don't see `W_u` in LSTM formulas. But it could be renamed in the code.
To get the best of two worlds, I propose to follow the hybrid scheme above, refer to the Wikipedia article in the code (`///` above the function) and leave a comment on the meaning of the variables (`//` near the variables declaration).

PR #114, opened 2019-04-23
Improve API documentation.

dan-zheng:
- Standardize documentation for `func call(_:)`.
- Minor edits to RNN docs.

### DIFF
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -31,8 +31,7 @@ public protocol Layer: Differentiable & KeyPathIterable
 
     /// Returns the output obtained from applying the layer to the given input.
     ///
-    /// - Parameters:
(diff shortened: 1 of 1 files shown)


Sources/DeepLearning/Layer.swift
```diff
@@ -1329,7 +1307,7 @@ public struct SimpleRNNCell<Scalar: TensorFlowFloatingPoint>: RNNCell, VectorNum
     }
 }
 
-/// An LSTM Cell.
+/// An LSTM cell.
```
Shashi456: Shouldn't this be `A LSTM cell` ?

rxwei: Since people tend to pronounce the acronym direct, the pronunciation starts with a vowel, so "An" would be most appropriate.

PR #117, opened 2019-04-25
Add Avgpool3d and Maxpool3d Layers & Tests

Shashi456:
Before adding tests i wanted to ask, @rxwei can we overload init? Because what we currently support in the max pooling and avg pooling layers is non-square windows, for square windows someone would have to write the layer as 
`MaxPool3d(poolsize:(3,3,3), stride:(2,2,2)`

Is there any way to work around that?

### DIFF
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -889,6 +889,18 @@ public struct MaxPool3D<Scalar: TensorFlowFloatingPoint>: Layer {
     }
 }
 
+public extension MaxPool3D {
+  /// Creates a MaxPool3D Layer with the specified poolsize and strides. The poolsize
+  /// and strides are square windows.
+  init(poolSize: Int, stride: Int, padding: Padding = .valid)
+  { let poolsize = poolSize
+    let stride = stride
+    self.init(
+      poolsize: (poolsize, poolsize, poolsize), strides: (stride, stride, stride), padding: padding
+    )
+  }
+}
+
 /// An average pooling layer for temporal data.
 @_fixed_layout
 public struct AvgPool1D<Scalar: TensorFlowFloatingPoint>: Layer {
@@ -1013,6 +1025,18 @@ public struct AvgPool3D<Scalar: TensorFlowFloatingPoint>: Layer {
     public func call(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
         return input.averagePooled(kernelSize: poolSize, strides: strides, padding: padding)
     }
+}
+
+public extension AvgPool3D {
+  /// Creates a AvgPool3D Layer with the specified poolsize and strides. The poolsize
+  /// and strides are square windows.
+  init(poolSize: Int, stride: Int, padding: Padding = .valid)
+  { let poolsize = poolSize
+    let stride = stride
+    self.init(
+      poolsize: (poolsize, poolsize, poolsize), strides: (stride, stride, stride), padding: padding
+    )
+  }
 }
 
 
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -845,6 +845,50 @@ public struct MaxPool2D<Scalar: TensorFlowFloatingPoint>: Layer {
     }
 }
 
+/// A max pooling layer for spatial or spatio-temporal data.
+@_fixed_layout
+public struct MaxPool3D<Scalar: TensorFlowFloatingPoint>: Layer {
+    /// The size of the sliding reduction window for pooling.
+    @noDerivative let poolSize: (Int, Int, Int, Int, Int)
+    /// The strides of the sliding window for each dimension of a 5-D input.
+    /// Strides in non-spatial dimensions must be `1`.
+    @noDerivative let strides: (Int, Int, Int, Int, Int)
+    /// The padding algorithm for pooling.
+    @noDerivative let padding: Padding
+
+    /// Creates a max pooling layer.
+    public init(
+        poolSize: (Int, Int, Int, Int, Int),
+        strides: (Int, Int, Int, Int, Int),
+        padding: Padding
+    ) {
+        self.poolSize = poolSize
+        self.strides = strides
+        self.padding = padding
+    }
+
+    /// Creates a max pooling layer.
+    ///
+    /// - Parameters:
+    ///   - poolSize: Vertical and horizontal factors by which to downscale.
+    ///   - strides: The strides.
+    ///   - padding: The padding.
+    public init(poolSize: (Int, Int, Int), strides: (Int, Int, Int), padding: Padding = .valid) {
+        self.poolSize = (1, poolSize.0, poolSize.1, poolSize.2, 1)
+        self.strides = (1, strides.0, strides.1, strides.2, 1)
+        self.padding = padding
+    }
+
+    /// Returns the output obtained from applying the layer to the given input.
+    ///
+    /// - Parameter input: The input to the layer.
+    /// - Returns: The output.
+    @differentiable
+    public func call(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+        return input.maxPooled(kernelSize: poolSize, strides: strides, padding: padding)
+    }
+}
+
 /// An average pooling layer for temporal data.
 @_fixed_layout
 public struct AvgPool1D<Scalar: TensorFlowFloatingPoint>: Layer {
@@ -894,7 +938,7 @@ public struct AvgPool2D<Scalar: TensorFlowFloatingPoint>: Layer {
     /// The padding algorithm for pooling.
     @noDerivative let padding: Padding
 
-    /// Creates a average pooling layer.
+    /// Creates an average pooling layer.
     public init(
         poolSize: (Int, Int, Int, Int),
         strides: (Int, Int, Int, Int),
@@ -905,7 +949,7 @@ public struct AvgPool2D<Scalar: TensorFlowFloatingPoint>: Layer {
         self.padding = padding
     }
 
-    /// Creates a average pooling layer.
+    /// Creates an average pooling layer.
     ///
     /// - Parameters:
     ///   - poolSize: Vertical and horizontal factors by which to downscale.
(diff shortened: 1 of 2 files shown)


saeta:
@Shashi456 yes you can overload init, and you can have one init delegate to another init as well. Having convenience initializers to simplify common cases is a good idea.

Shashi456:
Can you explain what you mean by delegate? @saeta

saeta:
Great question! In the Dense layer, we have one initializer call a different initializer to reduce duplication of code: https://github.com/tensorflow/swift-apis/blob/master/Sources/DeepLearning/Layer.swift#L245 Does that make sense?

Shashi456:
Oh that's pretty cool. So you mean in this case, something like: 
public init(poolSize: (Int), strides: (Int), padding: Padding = .valid) {
        self.poolsize = poolsize
        self.stride = stride
        self.padding = padding
    }
{ self.init(poolsize : (poolsize,poolsize,poolsize) ....}

In similar fashion?

saeta:
Yes, and you can even refactor the 2 initializers you currently have in the PR into one delegating (calling) the other. :-)

Shashi456:
You mean initialization delegation is possible across functions?

Sources/DeepLearning/Layer.swift
```diff
@@ -845,6 +845,50 @@ public struct MaxPool2D<Scalar: TensorFlowFloatingPoint>: Layer {
...
+    }
+
+    /// Creates a max pooling layer.
+    ///
+    /// - Parameters:
+    ///   - poolSize: Vertical and horizontal factors by which to downscale.
+    ///   - strides: The strides.
+    ///   - padding: The padding.
+    public init(poolSize: (Int, Int, Int), strides: (Int, Int, Int), padding: Padding = .valid) {
+        self.poolSize = (1, poolSize.0, poolSize.1, poolSize.2, 1)
```
rxwei: As @saeta suggested, if you already have a member wise initializer (`(init(poolSize:strides:padding:)`), calling that one via `self.init` is slightly better than initializing every member manually.

Sources/DeepLearning/Layer.swift
```diff
@@ -927,6 +971,50 @@ public struct AvgPool2D<Scalar: TensorFlowFloatingPoint>: Layer {
...
+    /// Returns the output obtained from applying the layer to the given input.
+    ///
+    /// - Parameter input: The input to the layer.
+    /// - Returns: The output.
+    @differentiable
+    public func call(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+        return input.averagePooled(kernelSize: poolSize, strides: strides, padding: padding)
+    }
+}
+
```
rxwei: Remove extra line.

Sources/DeepLearning/Layer.swift
```diff
@@ -845,6 +845,50 @@ public struct MaxPool2D<Scalar: TensorFlowFloatingPoint>: Layer {
...
+    /// The strides of the sliding window for each dimension of a 5-D input.
+    /// Strides in non-spatial dimensions must be `1`.
+    @noDerivative let strides: (Int, Int, Int, Int, Int)
+    /// The padding algorithm for pooling.
+    @noDerivative let padding: Padding
+
+    /// Creates a max pooling layer.
+    public init(
+        poolSize: (Int, Int, Int, Int, Int),
+        strides: (Int, Int, Int, Int, Int),
```
jekbradbury: I think I would suggest shifting the stored sizes and strides to be the tuples of three spatial sizes/strides rather than all 5; that way you don't need two `init`s and can avoid exposing the strangeness of the 5-tuple API to the user at all. (I recognize that you've done it this way because that's how our 2D pooling layers work, but those should ideally be changed too!)

dan-zheng: @jekbradbury: would it make sense to change operators (e.g. `convolved2D`, `maxPooled`) to not take extra spatial sizes/strides?

```swift
// Current:
func maxPooled(
    kernelSize: (Int, Int, Int, Int),
    strides: (Int, Int, Int, Int),
    padding: Padding
) -> Tensor

// After (implementation uses 1 for first and last size/stride):
func maxPooled(
    kernelSize: (Int, Int),
    strides: (Int, Int),
    padding: Padding
) -> Tensor

// Same for `convolved2D` and `averagePooled`.
```

rxwei: If I recall correctly, the fact it took a full tuple was because of historical GPE reasons.

Shashi456:
@rxwei so the things i have to incorporate is :
- Tests for average and max pooling layers 
- Add another init which reference to the original init

would that be it?

Shashi456:
@dan-zheng @saeta @rxwei please review!

Sources/DeepLearning/Layer.swift
```diff
@@ -845,6 +845,62 @@ public struct MaxPool2D<Scalar: TensorFlowFloatingPoint>: Layer {
...
+    }
+
+    /// Creates a max pooling layer.
+    ///
+    /// - Parameters:
+    ///   - poolSize: Vertical and horizontal factors by which to downscale.
+    ///   - strides: The strides.
+    ///   - padding: The padding.
+    public init(poolSize: (Int, Int, Int), strides: (Int, Int, Int), padding: Padding = .valid) {
+        self.poolSize = (1, poolSize.0, poolSize.1, poolSize.2, 1)
```
dan-zheng: Please rewrite this initializer using a `self.init` call.
```suggestion
        self.init(poolSize: (1, poolSize.0, poolSize.1, poolSize.2, 1),
                  strides: (1, strides.0, strides.1, strides.2, 1),
                  padding: padding)
```

Please do the same for initializers of `MaxPool2D`, `AvgPool2D`, `MaxPool3D`, and `AvgPool3D`.

Sources/DeepLearning/Layer.swift
```diff
@@ -927,6 +983,62 @@ public struct AvgPool2D<Scalar: TensorFlowFloatingPoint>: Layer {
...
+    @differentiable
+    public func call(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+        return input.averagePooled(kernelSize: poolSize, strides: strides, padding: padding)
+    }
+}
+
+public extension AvgPool3D {
+  /// Creates a AvgPool3D Layer with the specified poolsize and strides. The poolsize
+  /// and strides are square windows.
+  init(poolSize: Int, stride: Int, padding: Padding = .valid)
```
dan-zheng: Please be mindful of the following:
- Use 4-space indentation.
- Wrap `{` onto the previous line.
- Local variable bindings like `let poolsize = poolSize` are not meaningful, use the argument `poolSize` directly.

I hope those points make sense! Being mindful of them helps us merge your PR more quickly.

Please use the following:
```swift
public extension AvgPool3D {
    /// Creates an average pooling layer with the specified pooling window size and stride. The
    /// pooling window is square.
    init(poolSize: Int, stride: Int, padding: Padding = .valid) {
        self.init(poolSize: (poolSize, poolSize, poolSize),
                  strides: (stride, stride, stride),
                  padding: padding)
    }
}
```

Tests/DeepLearningTests/LayerTests.swift
```diff
@@ -34,6 +34,22 @@ final class LayerTests: XCTestCase {
...
+    func testMaxPool2D() {
+        let layer = MaxPool2D<Float>(poolSize:(2, 2), strides:(1, 1), padding:.valid)
+        let input = Tensor(shape: [1, 2, 2, 1], scalars: (0..<4).map(Float.init))
+        let output = layer.inferring(from: input)
+        let expected = Tensor<Float>([[[[3]]]])
+        XCTAssertEqual(round(output), expected)
+    }
+
+    func testMaxPool3D() {
+        let layer = MaxPool3D<Float>(poolSize:(2 ,2, 2), strides:(1, 1, 1), padding:.valid)
```
dan-zheng: Please use correct spacing.
```suggestion
        let layer = MaxPool3D<Float>(poolSize: (2, 2, 2), strides: (1, 1, 1), padding: .valid)
```

Please also fix spacing in other functions (`testMaxPool2D`, `testMaxPool3D`, `testAvgPool2D`).

Sources/DeepLearning/Layer.swift
Shashi456: Yes, it does. I will pay heed from the next time, Sorry!

Shashi456: Also @dan-zheng can you think of any other variations of inits and should i extend an init for 2d max and average pooling as well?

dan-zheng: Note that this API is not comprehensive: both pooling window sizes and strides can either be specified as "multiple integers" or "one integer, implicitly duplicated".

In other words, all of the following may be useful:
- `init(poolSize: (Int, Int, Int), strides: (Int, Int, Int), padding: Padding)`
- `init(poolSize: Int, strides: Int, padding: Padding)`
- `init(poolSize: (Int, Int, Int), strides: Int, padding: Padding)`
- `init(poolSize: Int, strides: (Int, Int, Int) padding: Padding)`

Something to think about and visit after this PR.
Here's Keras (note "integer or tuple of two integers"):

image

Shashi456: While writing the code for this, i thought about exactly this case, but had no idea how to address it exactly. In python, It is easier to handle. I'm not very well versed with init overloading in swift

dan-zheng: > Yes, it does. I will pay heed from the next time, Sorry!

Thanks a ton! That would be a huge help. 🙂

> Also @dan-zheng can you think of any other variations of inits and should i extend an init for 2d max and average pooling as well?

I just commented above about this.
To match Keras, we could have the following combinations:
- `poolSize`: `(Int, Int, Int)` or `Int`
- `strides`: `(Int, Int, Int)` or `Int?` (if `nil`, then use `poolSize` as strides)

Shashi456:
@dan-zheng thank you for being so patient with these reviews.

Shashi456:
@rxwei can you review this once more?

rxwei:
This should not have been merged because it does not compile. The `self.init` statement cannot appear on top level. The code does not compile.

@Shashi456 Could you make sure everything compiles with the latest toolchain before you commit?

rxwei:
I reverted it in #118. Feel free to start a new PR when you have a fix.

Shashi456:
@rxwei when i just try to simply add a layer in colab, to get an idea of how the layer works. It often throws me this error, could you tell me how i can solve this ?

```
error: <Cell 8>:2:15: error: type 'MaxPool3D<Scalar>' does not conform to protocol 'Layer'
public struct MaxPool3D<Scalar: TensorFlowFloatingPoint>: Layer {
              ^

TensorFlow.Layer:2:20: note: protocol requires nested type 'Input'; do you want to add it?
    associatedtype Input : Differentiable
                   ^

TensorFlow.Layer:3:20: note: protocol requires nested type 'Output'; do you want to add it?
    associatedtype Output : Differentiable
```

Shashi456:
Is there a temporary fix I could do @rxwei?

rxwei:
Could you share a code example that I use to reproduce your error?

Shashi456:
@rxwei for example, if i use this code, I know this could also be because max is not differentiable, but this input output error is common with any layer i try to define. 
```
@_fixed_layout
public struct GlobalMaxPooling3D<Scalar: TensorFlowFloatingPoint>: Layer {
    public init() {}

    ///
    /// - Parameters:
    ///   - input: The input to the layer.
    ///   - context: The contextual information for the layer application, e.g. the current learning
    ///     phase.
    /// - Returns: The output.
    @differentiable
        return input.max(squeezingAxes: [1, 2, 3])
    }
}

```
image

rxwei:
Colab is not yet updated with the latest APIs (v0.3). You can follow the instructions [here](https://github.com/tensorflow/swift/blob/master/Usage.md#getting-newer-swift-builds) to install a new build to Colab.

Shashi456:
Thanks a lot ! I will try and see if this works.

rxwei:
Actually, it should be updated now if you restart Colab.

Shashi456:
@rxwei the error majorly seems to be :
```
error: cannot convert value of type '(Int, Int, Int, Int, Int)' to expected argument type '(Int, Int, Int, Int)'
        return input.maxPooled(kernelSize: poolSize, strides: strides, padding: padding)
```

PR #123, opened 2019-05-16
Workaround for SR-10697.

dan-zheng:
Add `State` wrapper struct to `SimpleRNNCell` to work around IRGen crash.
[TF-507](https://bugs.swift.org/browse/TF-507) tracks reverting the change after [SR-10697](https://bugs.swift.org/browse/SR-10697) is fixed.

[Workaround](https://gist.github.com/marcrasi/5495924a7d07bb755ac2dc50d7abbe9a) provided by @marcrasi.

### DIFF
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -1300,13 +1300,16 @@ public struct SimpleRNNCell<Scalar: TensorFlowFloatingPoint>: RNNCell, VectorNum
     }
 
     public var zeroState: State {
-        return State(state: Tensor(zeros: stateShape))
+        return State(Tensor(zeros: stateShape))
     }
 
     // TODO(TF-507): Revert to `typealias State = Tensor<Scalar>` after
     // SR-10697 is fixed.
     public struct State: Differentiable {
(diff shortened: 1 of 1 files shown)


Sources/DeepLearning/Layer.swift
```diff
@@ -1299,11 +1299,16 @@ public struct SimpleRNNCell<Scalar: TensorFlowFloatingPoint>: RNNCell, VectorNum
...
-    public var zeroState: Tensor<Scalar> {
-        return Tensor(zeros: stateShape)
+    public var zeroState: State {
+        return State(state: Tensor(zeros: stateShape))
+    }
+
+    // TODO(TF-507): Revert to `typealias State = Tensor<Scalar>` after
+    // SR-10697 is fixed.
+    public struct State: Differentiable {
+        let state: Tensor<Scalar>
```
rxwei: 1. Could you call this `value` so that it doesn't repeat the struct name?
2. Since this is public API, please make the property public and add a memberwise initializer that has no argument labels.

dan-zheng: Done.

PR #129, opened 2019-05-22
Add Complex Numbers

bartchr808:
This PR adds a prototype version of complex numbers. 

**NOTE:** this will not be a part of the library, as this is an experimental project that the @tensorflow/swift-team is using. However, feel free to fork the project and play around with it! 😄

## Status of Complex Numbers in Swift
At the time of this PR, there is discussion about adding complex numbers to the Swift standard library, but it will be a post 5.1 feature. You can read more about it in this [forum post](https://forums.swift.org/t/introducing-complex-numbers-update/24598). 

As such, the @tensorflow/swift-team decided to add a basic, prototype version of complex numbers to the library. This implementation uses a modified implementation from the [xwu/NumericAnnex](https://github.com/xwu/NumericAnnex/blob/master/Sources/Complex.swift) Swift numeric library repo.

## Why We Added it: Differentiation
As part of further exploring the capabilities of the automatic differentiation system the team and the rest of the community has helped create so far, we wanted to see how versatile it is. Currently, we have no problem of differentiating anything in the real number space (like real vectors/tensors/matrices/etc.).

Complex numbers are different in that basic operators behave differently than in the reals. And in a concrete example, the dot product of a complex vector is the following:
```swift
func dot(lhs: Vector<Complex>, rhs: Vector<Complex>) -> Complex {
    var result = Complex(real: 0, imaginary: 0)
    for i in lhs.length {
        resultVector = resultVector + (lhs[i].complexConjugate() * rhs[i])
    }
    return result
}
```
While in the real number space, conjugates are not used.

You can see an example [here](https://www.mathworks.com/help/matlab/ref/dot.html#bt9p8vi-1) and more about the complex conjugate [here](https://en.wikipedia.org/wiki/Complex_conjugate). For an actual implementation example, you can see one of our test cases [here](https://github.com/tensorflow/swift-apis/pull/129/files#diff-9fcb50d9435193fbcdd30464d76cacd2R254). Note that at the time of this PR, control flow is not 100% complete, so the dot product is only for a fixed size vector at the moment.

## What We Learned So Far
* Our system works well with differentiating complex numbers and complex vectors
* ~~We need to update the `VectorNumeric` protocol to add a `complexConjugate` function that will be automatically synthesized for everything (returning self) other than when the `Scalar` type is `Complex`, where a default implementation will be written.~~ We need to figure out where to add `complexConjugate`, and whether we need a `VectorNumeric` protocol. Currently being discussed in the [Swift Forums](https://forums.swift.org/t/vectornumeric-type-in-general-complex-numbers/24875).


Shashi456:
@bartchr808 Please don't mind me asking, But is this something required within some part of deep learning or functionality we are adding to swift-apis? Just plainly curious :P.

rxwei:
> @bartchr808 Please don't mind me asking, But is this something required within some part of deep learning or functionality we are adding to swift-apis? Just plainly curious :P.

This is a purely experimental and internal feature we use to explore the automatic differentiation system.

marcrasi:
Some licensing stuff: Google policy for importing code from third party libraries is that it must go inside a directory named `third_party/<something>`, and that directory must contain the LICENSE file from the original project. This is applicable even if the third-party code is very modified. (https://opensource.google.com/docs/releasing/preparing/#third-party-components)

Therefore, you should put this code under `Sources/third_party/Complex`, and include the LICENSE file from https://github.com/xwu/NumericAnnex/blob/master/LICENSE.

Also it would be nice to include a comment at the top of `Complex.swift` crediting the NumericAnnex project and explaining that this file includes lots of code from it.

You'll have to create a new SwiftPM target so that SwiftPM compiles files in this new location.

Tests/DeepLearningTests/ComplexAutoDiffTests.swift
```diff
@@ -0,0 +1,98 @@
+// RUN: %target-run-simple-swift
+// REQUIRES: executable_test
+//
+// Complex API tests.
+
+// TODO: remove import
+import TensorFlow
+
+import StdlibUnittest
```
rxwei: We should not use `StdlibUnittest` in swift-apis since it's a SwiftPM package. Could you please refactor this to use XCTest like other test files in this directory?

And could you please merge the two test files into one?

Tests/DeepLearningTests/ComplexAutoDiffTests.swift
```diff
@@ -0,0 +1,98 @@
+// RUN: %target-run-simple-swift
```
rxwei: Need a license header here.

Tests/DeepLearningTests/ComplexAutoDiffTests.swift
bartchr808: Yes I noticed the error when I was looking as to why the tests were failing. Let me update the tests so it's in one file!

bartchr808:
Based on how the tests are failing (can't find `Complex`) I'm guessing I have no modify some file to add the path to `Sources/third_party/Complex/`?

Package.swift
```diff
@@ -22,13 +22,16 @@ let package = Package(
     products: [
         .library(
             name: "DeepLearning",
-            targets: ["DeepLearning"]),
+            targets: ["DeepLearning"/*, "third_party"*/]),
```
rxwei: `third_party` should not be part of the `DeepLearning` target, since we are not shipping it as part of `DeepLearning`.

Package.swift
```diff
@@ -22,13 +22,16 @@ let package = Package(
...
     ],
     dependencies: [],
     targets: [
         .target(
             name: "DeepLearning",
             dependencies: []),
+        // .target(
```
rxwei: You can specify paths for individual targets. See example [here in the swift-models repo](https://github.com/tensorflow/swift-models/blob/4275c36b8ea229c761a3d5d339baf674dbb148cd/Package.swift#L23).

Tests/DeepLearningTests/ComplexTests.swift
```diff
@@ -0,0 +1,230 @@
+// RUN: %target-run-simple-swift
```
rxwei: 1. Could you please add a license header here?
2. All source files should be 4-space indented.

Package.swift
bartchr808: Yup! Put it here temporarily as at the time I couldn’t create an Xcode project using swift package manager, and just wanted to verify the tests were passing through our CI since I was having trouble running them locally.

Tests/DeepLearningTests/ComplexTests.swift
bartchr808: Ahh yes keep forgetting the spacing difference between the code here and in apple/swift

Sources/third_party/Experimental/Complex.swift
```diff
@@ -0,0 +1,325 @@
...
+
+
+    static func /= (lhs: inout Complex, rhs: Complex) {
+        lhs = lhs / rhs
+    }
+}
+
+extension Complex {
+    @differentiable(vjp: _vjpComplexConjugate where T : Differentiable)
+    func complexConjugate() -> Complex {
```
bartchr808: Should this be a function or a computed variable? I talked to @dan-zheng and I agree in that this should be a computed property.

rxwei: I think that it's debatable. There are precedents of making unary math operations be methods instead of computed properties. If you look at [`FloatingPoint`](https://developer.apple.com/documentation/swift/floatingpoint), computed properties defined in that protocol are often things that are component-like (e.g. [`significand`](https://developer.apple.com/documentation/swift/floatingpoint/1847298-significand), [`nextUp`](https://developer.apple.com/documentation/swift/floatingpoint/1848104-nextup)) or predicate-like (e.g. [`isSubnormal`](https://developer.apple.com/documentation/swift/floatingpoint/1641181-issubnormal)). Actual unary math operations like square root are actually defined as a nullary method ([`squareRoot()`](https://developer.apple.com/documentation/swift/floatingpoint/3017991-squareroot)).

Since complex-conjugation is more like an operation than a component, I think a method makes most sense.

bartchr808: Oh okay that's a fair argument. With that then ya I'll leave it as a method 😄

Sources/third_party/Experimental/Complex.swift
```diff
@@ -0,0 +1,325 @@
+// A major part of the Complex number implementation comes from
+// [@xwu](https://github.com/xwu)'s project, [NumericAnnex](https://github.com/xwu/NumericAnnex).
+
+struct Complex<T : FloatingPoint> {
```
rxwei: In the swift-apis repository, we adopt the [Google Swift Style Guide](https://google.github.io/swift/) with the exception that indentation is 4 spaces long. One things that is different from the style we use in the standard library is that we leave no space before colons in conformance constraints. That is, `T : P` should be changed to `T: P`.
```suggestion
struct Complex<T: FloatingPoint> {
```
Same for other occurrences of conformance constraints in these files.

Sources/third_party/Experimental/Complex.swift
```diff
@@ -0,0 +1,325 @@
...
+    @usableFromInline
+    static func _vjpInit(real: T, imaginary: T) -> (Complex, (Complex) -> (T, T)) {
+        return (Complex(real: real, imaginary: imaginary), { ($0.real, $0.imaginary) })
+    }
+}
+
+extension Complex where T : Differentiable {
+    @usableFromInline
+    static func _vjpAdd(lhs: Complex, rhs: Complex)
+    -> (Complex, (Complex) -> (Complex, Complex)) {
```
rxwei:
```suggestion
        -> (Complex, (Complex) -> (Complex, Complex)) {
```

Same for other occurrences.

Sources/third_party/Experimental/Complex.swift
```diff
@@ -0,0 +1,325 @@
...
+
+extension Complex where T : Differentiable, T.TangentVector == T {
+    @usableFromInline
+    static func _vjpInit(real: T, imaginary: T) -> (Complex, (Complex) -> (T, T)) {
+        return (Complex(real: real, imaginary: imaginary), { ($0.real, $0.imaginary) })
+    }
+}
+
+extension Complex where T : Differentiable {
+    @usableFromInline
```
rxwei: No need for `@usableFromInline` if no APIs are `public`.

Sources/third_party/Experimental/Complex.swift
```diff
@@ -0,0 +1,325 @@
...
+    }
+}
+
+func abs<T>(_ z: Complex<T>) -> Complex<T> {
+    return Complex(real: z.magnitude)
+}
+
+extension Complex {
+    @differentiable(vjp: _vjpAdding(real:)
+        where T : Differentiable,
```
rxwei: These two constraints can be on the same line since the column limit here is 100.

Sources/third_party/Experimental/Complex.swift
```diff
@@ -0,0 +1,325 @@
...
+                (ac.isInfinite || bd.isInfinite || ad.isInfinite || bc.isInfinite) {
+                if a.isNaN { a = T(signOf: a, magnitudeOf: 0) }
+                if b.isNaN { b = T(signOf: b, magnitudeOf: 0) }
+                if c.isNaN { c = T(signOf: c, magnitudeOf: 0) }
+                if d.isNaN { d = T(signOf: d, magnitudeOf: 0) }
+                recalculate = true
+            }
+            if recalculate {
+                return Complex(
+                real: .infinity * (a * c - b * d),
```
rxwei:
```suggestion
                    real: .infinity * (a * c - b * d),
```

Sources/third_party/Experimental/Complex.swift
```diff
@@ -0,0 +1,325 @@
...
+                imaginary: .infinity * (a * d + b * c)
```
rxwei:
```suggestion
                    imaginary: .infinity * (a * d + b * c)
```

Sources/third_party/Experimental/Complex.swift
```diff
@@ -0,0 +1,325 @@
...
+        real.negate()
+        imaginary.negate()
+    }
+}
+
+extension Complex {
+    @differentiable(vjp: _vjpDivide(lhs:rhs:) where T : Differentiable)
+    static func / (lhs: Complex, rhs: Complex) -> Complex {
+        var a = lhs.real, b = lhs.imaginary, c = rhs.real, d = rhs.imaginary
+        var x: T
```
rxwei: In the conditionals below, since assignments to `x` and `y` on all paths of execution are initialization instead of mutation, you can declare `x` and `y` as `let`.

Tests/ExperimentalTests/ComplexTests.swift
```diff
@@ -0,0 +1,288 @@
...
+            y: Complex(real: -1, imaginary: -3),
+            z: Complex(real: 1, imaginary: -7))
+
+        let (result, pbComplex) = valueWithPullback(at: atVector) { x in
+            return dot(lhs: x, rhs: rhsVector)
+        }
+
+        XCTAssertEqual(Complex(real: 1, imaginary: -5), result)
+        XCTAssertEqual(expectedVector, pbComplex(Complex(real: 1, imaginary: 1)))
+    }
```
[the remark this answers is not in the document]
bartchr808: Ahh okay, I noticed that property when I was making the tests, but when I was running them it had no issue, but I was on Mac so of course I didn't notice, I should have asked earlier!

Sources/third_party/Experimental/Complex.swift
bartchr808: Will make sure to reference that next time I'm submitting a PR in one of our repos! I will have to update my [Protocol Oriented](https://github.com/tensorflow/swift/pull/187) walkthrough as well.

Sources/third_party/Experimental/Complex.swift
```diff
@@ -0,0 +1,325 @@
+// A major part of the Complex number implementation comes from
```
rxwei: From open source policies:
> For files in the third_party dir, leave the existing headers in place. If you’ve modified a file, append “Google LLC” to the copyright line.

Sources/third_party/Experimental/Complex.swift
bartchr808: Ahh true

Sources/third_party/Experimental/Complex.swift
bartchr808: Let me try, but I don't think it will as we have:
```swift
if c.magnitude >= d.magnitude {
    x = (a + b * ratio) / denominator
    y = (b - a * ratio) / denominator
} else {
    x = (a * ratio + b) / denominator
    y = (b * ratio - a) / denominator
}
if x.isNaN && y.isNaN {
  // More if logic
  x = T(signOf: c, magnitudeOf: .infinity) * a
  y = T(signOf: c, magnitudeOf: .infinity) * b
}
```
So there is a path of execution that may set it again, specifically when we doing something equivalent to `0.0/0.0` which would be nan

rxwei: Oh you're right, I missed it!

Sources/third_party/Experimental/Complex.swift
bartchr808: for the "append Google LLC" part I added the following:
This is the original one in the original header:
```swift
//  Code in libc++ is dual-licensed under the MIT and UIUC/NCSA licenses.
//  Copyright © 2009-2017 contributors to the LLVM/libc++ project.
```
Added Google LLC:
```swift
//  Code in libc++ is dual-licensed under the MIT and UIUC/NCSA licenses.
//  Copyright © 2009-2017 contributors to the LLVM/libc++ project, Google LLC.
```

rxwei: Hmm, this is a bit different. We did not contribute to LLVM/libc++ project, so we probably shouldn't add Google LLC to it.

rxwei: I'm writing a new suggestion in a different comment.

Sources/third_party/Experimental/Complex.swift
```diff
@@ -1,5 +1,55 @@
-// A major part of the Complex number implementation comes from
-// [@xwu](https://github.com/xwu)'s project, [NumericAnnex](https://github.com/xwu/NumericAnnex).
+//
+//  Complex.swift
+//  NumericAnnex
+//
+//  Created by Xiaodi Wu on 3/25/17.
+//
```
rxwei: Maybe replace lines 1-5 with

```swift
// Copyright 2017-2019 Xiaodi Wu and The TensorFlow Authors. All Rights Reserved.
//
// Licensed under the MIT License (the "License");
// you may not use this file except in compliance with the License.
// You may obtain a copy of the License at
//
//     https://opensource.org/licenses/MIT
//
// Unless required by applicable law or agreed to in writing, software
// distributed under the License is distributed on an "AS IS" BASIS,
// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
```

bartchr808: So we are including TensorFlow Authors instead of Google LLC? I was even thinking maybe `Copyright 2017-2019 Xiaodi Wu, The TensorFlow Authors, and Google LLC`

bartchr808: And would I then remove lines 14-16:
```swift
//  Code in libc++ is dual-licensed under the MIT and UIUC/NCSA licenses.
//  Copyright © 2009-2017 contributors to the LLVM/libc++ project.
/// A type to represent a complex value in Cartesian form.
```

rxwei: Since all official TensorFlow source files use `The TensorFlow Authors` and because we are a TensorFlow repo, I feel we should use `The TensorFlow Authors` instead.

Sources/third_party/Experimental/Complex.swift
```diff
@@ -0,0 +1,356 @@
...
+                c = T(signOf: c, magnitudeOf: c.isInfinite ? 1 : 0)
+                d = T(signOf: d, magnitudeOf: d.isInfinite ? 1 : 0)
+                x = 0 * (a * c + b * d)
+                y = 0 * (b * c - a * d)
+            }
+        }
+        return Complex(real: x, imaginary: y)
+    }
+
+
```
rxwei: Remove extra empty line.

Tests/ExperimentalTests/ComplexTests.swift
```diff
@@ -0,0 +1,317 @@
...
+        var expected = Complex<Float>(real: 1, imaginary: -4)
+        XCTAssertEqual(expected, input.subtracting(real: 1))
+        
+        input = Complex<Float>(real: 2, imaginary: -4)
+        expected = Complex<Float>(real: 2, imaginary: -5)
+        XCTAssertEqual(expected, input.subtracting(imaginary: 1))
+    }
+    
+    func testVjpAdd() {
+        let pb: (Complex<Float>) -> Complex<Float> = pullback(at: Complex<Float>(real: 2, imaginary: 3)) { x in
```
rxwei: Could you please make sure all source files fit within 100 columns?

bartchr808: I indented it to:
```
let pb: (Complex<Float>) -> Complex<Float> = pullback(at:
            Complex<Float>(real: 20, imaginary: -4)) { x in
            return x.adding(real: 5)
        }
```

rxwei: I'd suggest taking a look at [Google Swift Style Guide - Line Wrapping](https://google.github.io/swift/#line-wrapping) which describes breakable units and unbreakable units in detail.

In this particular case, an argument label + an argument is not a breakable unit.

Therefore, this statement should be formatted as:

```swift
let pb: (Complex<Float>) -> Complex<Float> =
    pullback(at: Complex<Float>(real: 20, imaginary: -4)) { x in
    return x.adding(real: 5)
}
```

Sources/third_party/Experimental/Complex.swift
```diff
@@ -0,0 +1,363 @@
...
+        return real.isFinite && imaginary.isFinite
+    }
+
+    var isInfinite: Bool {
+        return real.isInfinite || imaginary.isInfinite
+    }
+
+    var isNaN: Bool {
+        return (real.isNaN && !imaginary.isInfinite) ||
+        (imaginary.isNaN && !real.isInfinite)
```
rxwei:
```suggestion
            (imaginary.isNaN && !real.isInfinite)
```

bartchr808: I can actually even make this one line, so I'll do that

Tests/ExperimentalTests/ComplexTests.swift
bartchr808: Done! I'll try taking a look closer at the guide this weekend.

rxwei:
One last request: Could you revise the PR comment to describe what specific operations are introduced in this patch, and what this prototype will be used for? This would make it easy for open source followers/contributors to understand what's going on. Thanks!

bartchr808:
> One last request: Could you revise the PR comment to describe what specific operations are introduced in this patch, and what this prototype will be used for? This would make it easy for open source followers/contributors to understand what's going on. Thanks!

Sure! I'll ping you once I add them.

Sources/third_party/Experimental/Complex.swift
```diff
@@ -0,0 +1,362 @@
...
+/// A type to represent a complex value in Cartesian form.
+///
+/// - Note: `Complex64` is a type alias for `Complex<Float>` and `Complex128` is
+///   a type alias for `Complex<Double>`.
+///
+/// Create new instances of `Complex<T>` using integer or floating-point
+/// literals and the imaginary unit `Complex<T>.i`. For example:
+///
+/// ```swift
+/// let x = 2 + 4 * .i // `x` is of type `Complex<Double>`
```
lattner: I don't think this will type infer correctly given that there is no contextual complex type

Sources/third_party/Experimental/Complex.swift
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    @differentiable(vjp: _vjpMultiply(lhs:rhs:) where T: Differentiable)
+    static func * (lhs: Complex, rhs: Complex) -> Complex {
+        var a = lhs.real, b = lhs.imaginary, c = rhs.real, d = rhs.imaginary
+        let ac = a * c, bd = b * d, ad = a * d, bc = b * c
+        let x = ac - bd
+        let y = ad + bc
+
+        if x.isNaN && y.isNaN {
```
lattner: I think this logic would read a lot nicer if the body of this 'if' were pulled out of line into its own function

Sources/third_party/Experimental/Complex.swift
```diff
@@ -0,0 +1,362 @@
...
+            var recalculate = false
+            if a.isInfinite || b.isInfinite {
+                a = T(signOf: a, magnitudeOf: a.isInfinite ? 1 : 0)
+                b = T(signOf: b, magnitudeOf: b.isInfinite ? 1 : 0)
+                if c.isNaN { c = T(signOf: c, magnitudeOf: 0) }
+                if d.isNaN { d = T(signOf: d, magnitudeOf: 0) }
+                recalculate = true
+            }
+            if c.isInfinite || d.isInfinite {
```
lattner: I'm pretty sure this only need to run if 'recalculate' is false.

The code structure could be cleaned up (and recalculate eliminated?) if pulled out of line I think.

Sources/third_party/Experimental/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright © 2017 Xiaodi Wu (https://github.com/xwu).
```
lattner: Please contact Xiaodi about this to see if he is willing to contribute it to the s4tf project.  We don't want to mix licenses in the swift-apis repos.

lattner:
This is super excited @bartchr808 great work!

Sources/third_party/Experimental/Complex.swift
bartchr808: Ahh yes you're right I would need to add more context like `let x: Complex<Double> = 2 + 3 + .i`

I added this documentation back when @rxwei pointed out that I should keep the original header by Xiaodi. However, I think I skimmed over to quickly when I read the following:

> For files in the third_party dir, leave the existing headers in place. If you’ve modified a file, append “Google LLC” to the copyright line.

in the [google third-party info](https://opensource.google.com/docs/releasing/preparing/#third-party-components) (and also I didn't closely read the original header and missed this typo you caught). 

Since the third-party use is only about the license part, and not the entire original documentation provided by the author at the top of the file, I could remove this all, but I think it is best to keep it all and fix this typo.

Sources/third_party/Experimental/LICENSE
bartchr808: Ya that sounds good I sent him an email about this! 😄

Sources/third_party/Experimental/Complex.swift
bartchr808: True it could explain the logic behind this part of the function better. I'll give this a shot Sunday night!

Sources/third_party/Experimental/Complex.swift
bartchr808: Similar to the comment above, I'll take a look at this logic more closely and see what I can pull out, what code I can simplify, and also explain with comments Sunday night.

As background, I kept this code the way it was just like in the NumericAnnex library. It had written [a lot of tests](https://github.com/xwu/NumericAnnex/blob/master/Tests/NumericAnnexTests/ComplexTests.swift), so I did a bit of a blind trust on the implementation of this logic. Additionally, since I was mainly looking at the automatic differentiation aspect of complex numbers, I didn't spend much time on the implementation of the basic operators. 

However, since this will be a part of our repo, I see that it's good to make sure all our code is well documented and explained.

bartchr808:
Thanks @lattner ! 😄

eaplatanios:
@bartchr808 This is really cool! It would also be nice if you could make `Complex<T>` conform to `TensorDataType` when `T` is `Float32` or `Float64`, as that will allow creating complex-valued tensors.

bartchr808:
Thanks @eaplatanios ! Since we are only using this as an experiment in the team while Steve Canon is [adding it Swift](https://forums.swift.org/t/introducing-complex-numbers-update/24598), we will hold off on adding the ability to create complex tensors. However, once complex numbers are officially landed in a future version of Swift, we will definitely add the ability to create complex tensors! 😄

For context, we wanted to push this experiment to the swift-apis repo in case anyone wanted to copy the implementation and use it in their work/projects and to see what the team has been working on. However, we didn't want to expose this implementation to any S4TF users since this isn't a part of what the deep learning library part of the S4TF project is supposed to provide.

Sources/third_party/Experimental/Complex.swift
```diff
@@ -0,0 +1,339 @@
...
+/// A type to represent a complex value in Cartesian form.
+///
+/// Create new instances of `Complex<T>` using integer or floating-point
+/// literals and the imaginary unit `Complex<T>.i`. For example:
+///
+/// ```swift
+/// let x: Complex<Double> = 2 + 4 * .i
+/// ```
+///
+/// Additional Considerations
```
marcrasi: I think you should add an additional Additional Consideration linking to https://github.com/HIPS/autograd/blob/master/docs/tutorial.md#complex-numbers and explaining that we are using that convention.

It's a very non-obvious choice about how to do things. The first time I thought about complex differentiation, I thought that we should only define derivatives for holomorphic functions, but that convention lets us define derivatives for all functions and that link explains why that's a good thing to do.

Sources/third_party/Experimental/Complex.swift
```diff
@@ -0,0 +1,339 @@
...
+/// -------------------------
+///
+/// Floating-point types have special values that represent infinity or NaN
+/// ("not a number"). Complex functions in different languages may return
+/// different results when working with special values.
+
+struct Complex<T: FloatingPoint> {
+    var real: T
+    var imaginary: T
```
marcrasi: Have you tried making `real` and `imaginary` differentiable? If you do that, then the compiler will in principle be able to automatically compute derivatives for all the complex operations because they just take `real` and `imaginary` parts, then perform differentiable operations on those, then construct a new complex number (which is also a differentiable operation)!

In practice, the compiler can't quite derive these yet because most of the complex operations are implemented in terms of mutation and control flow that the compiler can't handle. But soon it should be able to handle it!

It would be very interesting to add some tests that define a few operations in ways that autodiff can handle, and then see if autodiff computes the expected derivatives for those. e.g.

```
func simplifiedAdd(_ lhs: Complex, _ rhs: Complex) -> Complex {
  return Complex(lhs.real + rhs.real, lhs.imaginary + rhs.imaginary)
}
```

eaplatanios:
@bartchr808 Thanks for the explanation! That makes sense. I guess I was a bit confused as to why this PR was made in this repo but your explanation cleared things out.

Sources/third_party/Experimental/Complex.swift
bartchr808: Agreed I added the following:
```swift
/// Our implementation of complex number differentiation follows the same
/// convention as Autograd. In short, we can get the derivative of a
/// holomorphic function, functions whose codomain are the Reals, and
/// functions whose codomain and domain are the Reals. You can read more about
/// Autograd at
///
///   https://github.com/HIPS/autograd/blob/master/docs/tutorial.md#complex-numbers
```

Sources/third_party/Experimental/Complex.swift
rxwei: I believe derived conformances already made stored properties differentiable.

bartchr808: > Have you tried making real and imaginary differentiable?

I did in that when I mark a method as differentiable, I put the requirement that `@differentiable(vjp: _vjpFunc where T: Differentiable)`. I didn't require `real` and `imaginary` to be differentiable where I define the `Complex` struct, because when we are to differentiate the standard library's version of complex numbers, we won't be able to change it, and additionally it will restrict Complex numbers to what they can be more than what the standard library will likely have.

> It would be very interesting to add some tests that define a few operations in ways that autodiff can handle, and then see if autodiff computes the expected derivatives for those.

I'll give it a shot! 😄

bartchr808:
Currently PR is blocked by the following bug which doesn't let me differentiate the initializer: https://bugs.swift.org/browse/TF-546

bartchr808:
Thanks @dan-zheng for the fix in [TF-546](https://bugs.swift.org/browse/TF-546), everything should work now! 😄

PR #130, opened 2019-05-22
Fix tests.

dan-zheng:
WIP, one remaining test to fix: `testUpSampling3D`.

### DIFF
--- a/Sources/DeepLearning/Layer.swift
+++ b/Sources/DeepLearning/Layer.swift
@@ -1340,7 +1340,7 @@ public struct UpSampling1D<Scalar: TensorFlowFloatingPoint>: Layer {
     ///
     /// - Parameter size: The upsampling factor for timesteps.
     public init(size: Int) {
-       self.size = size
+        self.size = size
     }
 
     /// Returns the output obtained from applying the layer to the given input.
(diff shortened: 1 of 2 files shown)


Shashi456:
@dan-zheng is there no simple fix for this? Doesn't that mean upsampling 3d doesn't work effectively?

dan-zheng:
There's no trivial fix, unfortunately. Please see [TF-525](https://bugs.swift.org/browse/TF-525) for more info, including a crash reproducer and a link to Keras's implementation. We can copy Keras to avoid the unsupported broadcasting error.

Shashi456:
@dan-zheng so for using resize-volumes, would we have to write a vjp again?

dan-zheng:
> @dan-zheng so for using resize-volumes, would we have to write a vjp again? I see that the keras implementation uses resize_volumes, which again uses repeat elements, It might not be hard to fix, just looking at a way to do it properly.

Yes, it will be necessary to define some kind of "repeat elements" helper function that is differentiable.
Are you interested in taking this on? Let me share a quick prototype with you soon.

Shashi456:
@dan-zheng yes I would be willing to, but I'm not accustomed to writing vjps for custom functions so that might act as some kind of bottleneck for me there. Still, have a lot more to learn.

PR #131, opened 2019-05-23
Fix UpSampling3D implementation.

dan-zheng:
Broadcasting does not support 8-D tensors.
Copy Keras implementation: simulating broadcasting via splitting, duplication, and concatenation.

Resolves [TF-525](https://bugs.swift.org/browse/TF-525).

### DIFF
--- a/Tests/DeepLearningTests/LayerTests.swift
+++ b/Tests/DeepLearningTests/LayerTests.swift
@@ -134,15 +134,12 @@ final class LayerTests: XCTestCase {
       let size = 6
       let layer = UpSampling3D<Float>(size: size)
       let input = Tensor<Float>(shape: [1, 4, 3, 2, 1], scalars: (0..<24).map(Float.init))
-      // TODO(TF-525): Fix `UpSampling3D.call`.
-      // Broadcasting does not support tensors with high rank:
-      // Broadcast between [1,4,1,3,1,2,1,1] and [1,1,6,1,6,1,6,1] is not supported yet.
-      /*
+      // TODO: Fix bug in `UpSampling3D.repeatingElements`.
+      // Result has shape `[1, 96, 54, 24, 1]` instead of `[1, 24, 18, 12, 1]`.
       let output = layer.inferring(from: input)
       let expected = TensorShape([1, input.shape[1] * size, input.shape[2] * size, input.shape[3] * size, 1])
       XCTAssertEqual(output.shape, expected)
       XCTAssertEqual(output.shape, expected)
-      */
     }
 
     func testReshape() {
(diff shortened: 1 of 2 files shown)


Tests/DeepLearningTests/LayerTests.swift
```diff
@@ -134,15 +134,12 @@ final class LayerTests: XCTestCase {
       let size = 6
       let layer = UpSampling3D<Float>(size: size)
       let input = Tensor<Float>(shape: [1, 4, 3, 2, 1], scalars: (0..<24).map(Float.init))
-      // TODO(TF-525): Fix `UpSampling3D.call`.
-      // Broadcasting does not support tensors with high rank:
-      // Broadcast between [1,4,1,3,1,2,1,1] and [1,1,6,1,6,1,6,1] is not supported yet.
-      /*
+      // TODO: Fix bug in `UpSampling3D.repeatingElements`.
+      // Result has shape `[1, 96, 54, 24, 1]` instead of `[1, 24, 18, 12, 1]`.
```
dan-zheng: Todo: fix the bug in `UpSampling3D.repeatingElements`. I used [`repeat_elements`](https://github.com/tensorflow/tensorflow/blob/0c3ebf695028e8a233cd82b6ba7066a0a131d9f8/tensorflow/python/keras/backend.py#L2634) from Keras as my reference, but the shapes don't work out.

@Shashi456: would you like to continue looking into this?

Shashi456: @dan-zheng It's exact height, length and width more times.

Shashi456:
@dan-zheng you and i came to the same fix xD.

Tests/DeepLearningTests/LayerTests.swift
dan-zheng: @Shashi456 Thanks for the catch! The shape mismatch is fixed in https://github.com/tensorflow/swift-apis/pull/131/commits/c6e817c2c3b3d47d7535492fbc957ee4d21c4128. Upsampling tests should be updated to also check correctness of output values, not just output shapes.

Shashi456: @dan-zheng Actually that way there a lot more tests to be written for many layers, checking different kinds of functionalities :P. I'll keep that in mind though. Great job, being so quick to fix this.

dan-zheng: More tests would be awesome!

Shashi456: I'll make a comprehensive list in #77, might help newcomers.

PR #132, opened 2019-05-23
Test for conv2d

Shashi456:
Passes locally


Shashi456:
@dan-zheng I wonder why Kokoro doesn't get triggered on my PRs.

dan-zheng:
Kokoro CI runs automatically only on PRs created by people with push access, sorry!
Let's see if we can manually trigger CI.

dan-zheng:
Thanks @Shashi456!

Shashi456:
@dan-zheng, now comes the difficult part of choosing something to work on xD.

PR #133, opened 2019-05-23
Move Dataset.swift over from TensorFlow.

pschuh:
@eaplatanios A test of moving dataset over.

### DIFF
--- a/Tests/DeepLearningTests/DatasetTests.swift
+++ b/Tests/DeepLearningTests/DatasetTests.swift
@@ -1,0 +1,156 @@
+// Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


Tests/DeepLearningTests/DatasetTests.swift
```diff
@@ -0,0 +1,154 @@
...
+        ("testSingleValueManualIterator", testSingleValueManualIterator),
+        ("testDatasetIteration", testDatasetIteration),
+        ("testSingleValueTransformations", testSingleValueTransformations),
+        ("testSingleValueHOFs", testSingleValueHOFs),
+        ("testParallelMap", testParallelMap),
+        ("testMapToDifferentType", testMapToDifferentType),
+        ("testSingleValueBatched", testSingleValueBatched),
+        // Currently broken even in TensorFlow ...
+        // This will be easier to fix once everything is moved ...
+        // ("testDoubleValueDatasetIteration", testDoubleValueDatasetIteration),
```
dan-zheng: I think commenting out the test here won't prevent the test from being from on macOS, so I'd recommend commenting out `func testDoubleValueDatasetIteration`. IIRC, XCTest on macOS doesn't use the `allTests` mechanism.

dan-zheng:
Confirmed that `swift test` passes on macOS.

rxwei:
This file should have been adapted to use the tensorflow.org license header and use 4-space indentation.

PR #140, opened 2019-05-30
Addressed comments from apple/swift#25147.

eaplatanios:
@bgogul @rxwei

### DIFF
--- a/Sources/TensorFlow/Core/Runtime.swift
+++ b/Sources/TensorFlow/Core/Runtime.swift
@@ -35,7 +35,7 @@ import Darwin
 import Glibc
 #endif
 import CTensorFlow
-import Foundation
+import Dispatch
 
 /// TraceContext contains the state needed to build a trace graph function (TF_Function). As eager 
 /// ops are executed in tracing mode, their corresponding nodes are added to the trace graph (via
--- a/Sources/TensorFlow/Core/Runtime.swift
+++ b/Sources/TensorFlow/Core/Runtime.swift
@@ -35,7 +35,6 @@ import Darwin
 import Glibc
 #endif
 import CTensorFlow
-import Dispatch
 
 /// TraceContext contains the state needed to build a trace graph function (TF_Function). As eager 
 /// ops are executed in tracing mode, their corresponding nodes are added to the trace graph (via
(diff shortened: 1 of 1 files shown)


eaplatanios:
The tests are currently failing because we need a new toolchain, built after apple/swift#25147 is merged.

Sources/TensorFlow/Core/Runtime.swift
```diff
@@ -599,12 +609,10 @@ public final class _ExecutionContext {
 
                 stringAddrs.withUnsafeMutableBufferPointer { stringAddrsBuffer in
                     var cArgs = [stringAddrsBuffer.baseAddress.map(UnsafeMutablePointer.init)]
```
rxwei: This can be simplified in the same way: Instead of creating a zero-element array, create a `var` and pass it as an inout reference.

eaplatanios: Good point. Done in the last commit.

Sources/TensorFlow/Core/Runtime.swift
```diff
@@ -35,6 +35,7 @@ import Darwin
 import Glibc
 #endif
 import CTensorFlow
+import Foundation
```
rxwei: Is `Foundation` or Grand Central Dispatch required? Note that the the build on the apple/swift side isn't configured to depend on Foundation or Dispatch, which is also why we used pthread instead of `Dispatch` in the compiler runtime.

eaplatanios: Interesting. It's only used in this block:
```swift
signal(SIGINT, SIG_IGN)
let sigintSrc = DispatchSource.makeSignalSource(signal: SIGINT, queue: .main)
sigintSrc.setEventHandler {
    print("Caught interrupt signal, exiting...")
    exit(1)
}
sigintSrc.resume()
```
I switched to using `import Dispatch` in the last commit. Is that fine?

rxwei: I don't think apple/swift can depend on `Dispatch` either, unfortunately. Is there a way to use neither of them? 

A long term solution for our toolchain build is to follow how swift-corelibs-foundation is getting included in Swift toolchain builds, in which case we should be able to rely on `Dispatch` and `Foundation`.

eaplatanios: This is only really necessary for the `SIGINT` handler. Is there a way to provide a closure for the handler directly in `signal(SIGINT, ...)` (in place of `...`)?

rxwei: You can use Swift's closure syntax directly. A closure can be implicitly converted to a C function pointer (aka. a `@convention(c)` function) if the closure 1) does not capture anything and 2) uses C-compatible types in the signature.

rxwei: You can try 
```swift
signal(SIGINT) { _ in
    print("Caught interrupt signal, exiting...")
    exit(1)
}
```

eaplatanios: Thanks, that's cool and it works :) Just pushed it.

PR #141, opened 2019-05-30
Re-implement tensor shape/rank operators using TFE runtime APIs.

eaplatanios:
@rxwei Ported from apple/swift#24949.

Gives me about 98% speedup for `Tensor.rank` and 85% speedup for `Tensor.shape` (for a 2x3 tensor) on a MacBook Pro and all tests pass.

### DIFF
--- a/Sources/TensorFlow/Core/Runtime.swift
+++ b/Sources/TensorFlow/Core/Runtime.swift
@@ -551,7 +551,7 @@ public final class _ExecutionContext {
     @usableFromInline let eagerContext: CTFEContext
 
     /// The status for checking TensorFlow errors.
-    private let status: CTFStatus = TF_NewStatus()
+    @usableFromInline let status: CTFStatus = TF_NewStatus()
 
     /// The mutex for preventing potential concurrent access.
     private var mutex: pthread_mutex_t = pthread_mutex_t()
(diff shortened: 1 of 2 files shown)


rxwei:
Thanks Anthony! In my earlier experiment it crashed on macOS only. Glad to hear you didn't have this issue, but I'll build and verify this locally.

bgogul:
Nice @eaplatanios !  One thing to keep in mind is that using eager API directly will affect the performance of Lazy Tensor as it triggers materialization of symbolic tensors and break traces into smaller ones.

rxwei:
> Nice @eaplatanios ! One thing to keep in mind is that using eager API directly will affect the performance of Lazy Tensor as it triggers materialization of symbolic tensors and break traces into smaller ones.

Would it be possible for you to profile your lazy tensor implementation against changes made by this PR?

eaplatanios:
@bgogul @rxwei I think that's fine because if you want to use the lazy tensor implementation you can still use `rankTensor` and `shapeTensor`.

rxwei:
> @bgogul @rxwei I think that's fine because if you want to use the lazy tensor implementation you can still use `rankTensor` and `shapeTensor`.

Ah yep, totally.

rxwei:
A new CI toolchain will be ready in an hour.

rxwei:
All tests passed. Thank you!

PR #142, opened 2019-05-30
Improved derivative performance for broadcasted operations.

eaplatanios:
Re-implementation of apple/swift#24408.

@rxwei the reason you were getting errors is because you were differentiating a tensor-valued function (as opposed to scalar-valued) and you were assuming that the gradient seed has the same shape as the operation result. I fixed that by broadcasting the gradient seeds before computing the gradients.

### DIFF
--- a/Tests/TensorFlowTests/OperatorTests/MathTests.swift
+++ b/Tests/TensorFlowTests/OperatorTests/MathTests.swift
@@ -199,6 +199,17 @@ final class MathOperatorTests: XCTestCase {
         XCTAssertEqual(0.816997, Double(prediction.scalars[0]), accuracy: 0.0001)
     }
 
+    func testBroadcastedAddGradient() {
+	  func foo(_ x: Tensor<Float>, _ y: Tensor<Float>) -> Tensor<Float> {
+	    return x + y
+	  }
+	  let x = Tensor<Float>(ones: [1, 2, 1, 4])
+	  let y = Tensor<Float>(ones: [4, 1, 3, 1])
+	  let (dx, dy) = gradient(at: x, y, in: foo)
+	  XCTAssertEqual(x.shape, dx.shape)
+	  XCTAssertEqual(y.shape, dy.shape)
+	}
+
     static var allTests = [
         ("testReduction", testReduction),
         ("testArgmax", testArgmax),
@@ -209,6 +220,7 @@ final class MathOperatorTests: XCTestCase {
         ("testMultiOpMath", testMultiOpMath),
         ("testXWPlusB", testXWPlusB),
         ("testXORInference", testXORInference),
-        ("testMLPClassifierStruct", testMLPClassifierStruct)
+        ("testMLPClassifierStruct", testMLPClassifierStruct),
+        ("testBroadcastedAddGradient", testBroadcastedAddGradient)
     ]
 }
--- a/Tests/TensorFlowTests/OperatorTests/MathTests.swift
+++ b/Tests/TensorFlowTests/OperatorTests/MathTests.swift
@@ -201,7 +201,7 @@ final class MathOperatorTests: XCTestCase {
 
     func testBroadcastedAddGradient() {
 	  func foo(_ x: Tensor<Float>, _ y: Tensor<Float>) -> Tensor<Float> {
-	    return x + y
+	    return (x + y).sum()
 	  }
 	  let x = Tensor<Float>(ones: [1, 2, 1, 4])
 	  let y = Tensor<Float>(ones: [4, 1, 3, 1])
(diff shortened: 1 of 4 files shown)


rxwei:
> the reason you were getting errors is because you were differentiating a tensor-valued function (as opposed to scalar-valued) and you were assuming that the gradient seed has the same shape as the operation result. I fixed that by broadcasting the gradient seeds before computing the gradients.

We have been assuming that seeds are always shape-compatible (if not same-shape) as the original result, and existing implementations have not been doing broadcasting to the shape of the original result. I think there are other reasons for it, which may lead to a different solution.

eaplatanios:
That's interesting! I actually traced the test failure and saw that the seed shape was not matching the result shape when the add VJP was being called. Given that this was being directly invoked by the `gradient` function, could there be a bug somewhere more core to AD support?

Tests/TensorFlowTests/OperatorTests/MathTests.swift
```diff
@@ -199,6 +199,17 @@ final class MathOperatorTests: XCTestCase {
         XCTAssertEqual(0.816997, Double(prediction.scalars[0]), accuracy: 0.0001)
     }
 
+    func testBroadcastedAddGradient() {
+	  func foo(_ x: Tensor<Float>, _ y: Tensor<Float>) -> Tensor<Float> {
+	    return x + y
+	  }
+	  let x = Tensor<Float>(ones: [1, 2, 1, 4])
+	  let y = Tensor<Float>(ones: [4, 1, 3, 1])
+	  let (dx, dy) = gradient(at: x, y, in: foo)
```
rxwei: In this test case, the use of `gradient(at:in:)` is not valid because gradient is only mathematically defined for functions that return a scalar. We can either make `foo(_:_:)` to a `sum()` or use `pullback(at:in:)`.

eaplatanios: That's what I thought. This was actually the failing case in the other PR and that's why I copied it here but wasn't sure of the semantics of `gradient`. I'll add a call to `sum()` and remove the seed broadcasts then. :)

rxwei:
> Given that this was being directly invoked by the `gradient` function, could there be a bug somewhere more core to AD support?

AD is fully shape-agnostic, and there's some mathematical consistency to differential operators like `gradient(at:in:)` in that it should fail when the function's result is not semantically a scalar. I think a proper solution is to change the definition of `gradient(at:in:)`, `gradient(at:_:in:)`, `Tensor.gradient(in:)`, `valueWithGradient(at:in:)`, `valueWithGradient(at:_:in:)`and `Tensor.valueWithGradient(in:)` (all defined in [DifferentialOperators.swift](https://github.com/tensorflow/swift-apis/blob/master/Sources/TensorFlow/Core/DifferentialOperators.swift)) to call [`precondition(_:_:file:line:)`](https://developer.apple.com/documentation/swift/1540960-precondition) on the original function's result, for example:

```swift

@inlinable
public func valueWithGradient<T, R>(
    at x: T,
    in f: @differentiable (T) -> Tensor<R>
) -> (value: Tensor<R>, gradient: T.TangentVector)
    where T: Differentiable, R: TensorFlowFloatingPoint {
    let (y, pullback) = valueWithPullback(at: x, in: f)
    precondition(y.rank == 0)
    return (y, pullback(Tensor<R>(1)))
}
```

eaplatanios:
I also prefer that. What's the cost of preconditions? Are they removed when compiling with optimizations enabled?

rxwei:
> I also prefer that. What's the cost of preconditions? Are they removed when compiling with optimizations enabled?


|  | debug | release | release 
| -- | -- | -- | -- 
| assert() | YES | NO | NO 
| assertionFailure() | YES | NO | NO** 
| precondition() | YES | YES | NO
| preconditionFailure() | YES | YES | YES**
| fatalError()* | YES | YES | YES

From [this blog post](https://blog.krzyzanowskim.com/2015/03/09/swift-asserts-the-missing-manual/).

eaplatanios:
Awesome, thanks! :) I'll be away for ~30' but will make those changes once I get back.

eaplatanios:
I'm looking into this now. I can add the precondition for `valueWithGradient`, but not for `gradient` as I don't think it ever evaluates `f` directly.

eaplatanios:
I made all necessary changes and all tests pass locally.

rxwei:
> I'm looking into this now. I can add the precondition for `valueWithGradient`, but not for `gradient` as I don't think it ever evaluates `f` directly.

It actually does because it calls `pullback(at:in:)` which calls `valueWithPullback(at:in:)`. You can change `gradient(at:in:)` to call `valueWithGradient(at:in:)` and drop the value after `precondition`. Don't worry about performance since we can optimize those things away later.

eaplatanios:
Sounds good. Done in the latest commit. :)

rxwei:
All tests pass. Thank you!

PR #143, opened 2019-05-30
Enhanced the 'matmul' wrapper.

eaplatanios:
The `matmul` op now matches the behavior of the Python op (i.e., using `batchMatmul` whenever appropriate, and it also supports transposing either or both of its arguments.


Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1505,25 +1505,59 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
...
-    _ rhs: Tensor<Scalar>
+    _ rhs: Tensor<Scalar>,
+    transposeA: Bool = false,
+    transposeB: Bool = false
 ) -> Tensor<Scalar> {
-    // Default arguments specified explicitly to avoid "external declarations of SILFunctions with
-    // shared visibility is not allowed" SILVerifier error in
-    // "tests/AutoDiff/tensor_autodiff_runtime.swift".
-    return Raw.matMul(lhs, rhs, transposeA: false, transposeB: false)
+    if lhs.rank > 2 && rhs.rank > 2 {
```
rxwei: How about

```swift
switch (lhs.rank, rhs.rank) {
case (3..., 3...):
  ...
case (2, 3...):
  ...
case (3..., 2):
  ...
default:
  ...
}
```

eaplatanios: I didn't know you could pattern match on integers with ranges. That's cool! :)

eaplatanios:
@rxwei I addressed your feedback and also added a helper for adding tests. After we merge this I'll update all the other PRs to add tests using this helper.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1505,25 +1505,59 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
 
 /// Performs matrix multiplication with another tensor and produces the result.
 @inlinable
-@differentiable(vjp: _vjpMatmul(_:_:) where Scalar: TensorFlowFloatingPoint)
+@differentiable(vjp: _vjpMatmul(_:_:transposeA:transposeB:) where Scalar: TensorFlowFloatingPoint)
 public func matmul<Scalar: Numeric>(
```
rxwei: How about calling it `matmul(_:transposed:_:transposed)` where the two `transposed`s have a default value?

eaplatanios: Can you call this like `matmul(x, y)` or do the arguments need to be ordered?

rxwei: Yep

eaplatanios: That's neat. I just pushed this change.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1525,30 +1525,30 @@ public func matmul<Scalar: Numeric>(
...
     _ lhs: Tensor<Scalar>,
+    transposed transposeA: Bool = false,
     _ rhs: Tensor<Scalar>,
-    transposeA: Bool = false,
-    transposeB: Bool = false
+    transposed transposeB: Bool = false
 ) -> (Tensor<Scalar>, (Tensor<Scalar>) -> (Tensor<Scalar>, Tensor<Scalar>)) {
-    let value = matmul(lhs, rhs, transposeA: transposeA, transposeB: transposeB)
+    let value = matmul(lhs, transposed: transposeA, rhs, transposed: transposeB)
     return (value, { v in
```
rxwei: Use 4-space indentation.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1505,25 +1505,57 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+    transposed transposeA: Bool = false,
+    _ rhs: Tensor<Scalar>,
+    transposed transposeB: Bool = false
 ) -> (Tensor<Scalar>, (Tensor<Scalar>) -> (Tensor<Scalar>, Tensor<Scalar>)) {
-    let value = matmul(lhs, rhs)
-    return (value, { v in 
-        (matmul(v, rhs.transposed()), matmul(lhs.transposed(), v))
+    let value = matmul(lhs, transposed: transposeA, rhs, transposed: transposeB)
+    return (value, { v in
+      let (lhsGrad, rhsGrad) = { () -> (Tensor<Scalar>, Tensor<Scalar>) in
```
rxwei: Why does this need to be a closure?

You can declare `let (lhsGrad, rhsGrad): (Tensor<Scalar>, Tensor<Scalar>)` here and assign to it in the switch statement.

eaplatanios:
I updated this. I'll update the other PRs that depend on this after it's merged so they use the final version.

eaplatanios:
The tests will also not pass for this until the new toolchain is ready.

rxwei:
A new toolchain is ready, but it looks like Kokoro is no longer responding.

eaplatanios:
@rxwei yeah I also tried to enable it on #141 and it doesn't seem to work.

eaplatanios:
Could the `kokoro:force-run` flag possibly work?

rxwei:
Looks like it doesn't work. I'll be the CI machine manually tonight.

eaplatanios:
Thanks a lot Richard! :)

I think the best way is to go through them one by one and after each is merged I'll go ahead and merge changes with the next and notify you (because there may be some conflicts).

rxwei:
Yeah, I'm starting with this one. Building a toolchain right now.

eaplatanios:
Sounds good. This, #141, and #142 should be independent of each other and can thus be tested and merged separately. They all pass locally for me.

rxwei:
All tests passed. Thank you!

PR #144, opened 2019-05-30
Added support for the 'sign' op and its VJP.

eaplatanios:


rxwei:
Could you add a test for each of these operators?

eaplatanios:
I can do that yes. I just noticed that there were no tests for other wrappers of TF ops. It would be nice if we added a little helper that executes `XCTAssertEqual(_:_:accuracy:)` element-wise on tensors. I could look into that later tonight.

rxwei:
Yeah, other ops are mostly not tested. The migration made it so much easier to write tests now.

eaplatanios:
I merged the `matmul` PR here (so I can get the test helpers) and added a test for `sign`. The diff once the `matmul` PR is merged should be very small. I'll do the same for the other new ops.

eaplatanios:
@rxwei This has been updated with the merged PR changes and is ready to be tested. #142 is also ready to be tested.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -613,6 +613,19 @@ internal func _vjpFloor<T: TensorFlowFloatingPoint>(
     return (floor(x), { _ in Tensor(0).broadcasted(like: x) })
 }
 
+@inlinable
+@differentiable(vjp: _vjpSign(_:))
+public func sign<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
```
rxwei: Does this have to be `TensorFlowFloatingPoint`? `tf.sign` seems to support all dtypes.

> x: A Tensor. Must be one of the following types: bfloat16, half, float32, float64, int32, int64, complex64, complex128.

eaplatanios: Good catch! I just updated it.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -613,6 +613,19 @@ internal func _vjpFloor<T: TensorFlowFloatingPoint>(
     return (floor(x), { _ in Tensor(0).broadcasted(like: x) })
 }
 
+@inlinable
+@differentiable(vjp: _vjpSign(_:) where T: TensorFlowFloatingPoint)
+public func sign<T: TensorFlowNumeric>(_ x: Tensor<T>) -> Tensor<T> {
```
rxwei: Actually just `T: Numeric` is fine. The use site `Tensor<T>` implies the `T: TensorFlowScalar` constraint ;)

Tests/TensorFlowTests/OperatorTests/MathTests.swift
```diff
@@ -16,6 +16,12 @@ import XCTest
 @testable import TensorFlow
 
 final class MathOperatorTests: XCTestCase {
+    func testSign() {
+        let x = Tensor<Float>([[1, 2, -3, 4, 5], [1, 2, 3, 4, -5]])
+        let y = sign(x)
+        assertEqual(y, Tensor<Float>([[1, 1, -1, 1, 1], [1, 1, 1, 1, -1]]))
```
rxwei:
```suggestion
        XCTAssertEqual(y, Tensor<Float>([[1, 1, -1, 1, 1], [1, 1, 1, 1, -1]]))
```

rxwei:
Kokoro seems back online. Thanks @bgogul for fixing it!

rxwei:
> Kokoro seems back online. Thanks @bgogul for fixing it!

@bgogul said he didn't do anything. Not sure what happened, Kokoro :)

eaplatanios:
Haha yeah CI servers do that sometimes and it's so frustrating when they just don't trigger :)

eaplatanios:
Hmm I'm not sure what's wrong in Kokoro but earlier it was showing that same log while tests were passing locally.

rxwei:
All tests pass locally on my machine. Merging.

PR #145, opened 2019-05-30
Added support for the 'log1p' op and its VJP.

eaplatanios:
@rxwei I decided to break down PRs that add support for new ops, but please let me know if you'd prefer I add them as a single PR.


Sources/TensorFlow/Operators/Math.swift
```diff
@@ -424,6 +424,20 @@ internal func _vjpLog<T: TensorFlowFloatingPoint>(
     return (log(x), { v in v / x })
 }
 
+/// Computes the logarithm of `1 + x` element-wise.
+@inlinable
+@differentiable(vjp: _vjpLog1p)
+public func log1p<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
+    return Raw.log1p(x)
```
rxwei: FYI, Swift supports implicit returns for single line functions now so you can just write `Raw.log1p(x)`.

eaplatanios: That's cool to know thanks! I'll use that going forward.

eaplatanios:
Added a test similar to #144.

rxwei:
Which one's next?

eaplatanios:
I just updated this one so you can go ahead and test.

eaplatanios:
I'll be out for about ~15-30' but will come back and update the next one after this is merged.

eaplatanios:
#151 and #152 are also independent and can be tested and merged in the meantime.

rxwei:
All tests passed. Thank you!

rxwei:
> #151 and #152 are also independent and can be tested and merged in the meantime.

Those two actually need pulling from master and had some comments. Let me know when they are ready!

PR #146, opened 2019-05-30
Added support for the 'expm1' op and its VJP.

eaplatanios:


eaplatanios:
Added a test similar to #144.

eaplatanios:
@rxwei this should be ready now.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1537,7 +1542,6 @@ internal func _vjpMatmul<Scalar: TensorFlowFloatingPoint>(
     transposed transposeB: Bool = false
 ) -> (Tensor<Scalar>, (Tensor<Scalar>) -> (Tensor<Scalar>, Tensor<Scalar>)) {
     let value = matmul(lhs, transposed: transposeA, rhs, transposed: transposeB)
-    return (value, { v in
```
rxwei: Looks like an unintended deletion.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1517,16 +1532,6 @@ public func matmul<Scalar: Numeric>(
     _ rhs: Tensor<Scalar>,
     transposed transposeB: Bool = false
 ) -> Tensor<Scalar> {
-    switch (lhs.rank, rhs.rank) {
```
rxwei: Here too.

eaplatanios:
I'll need to go now, but will continue with this tomorrow morning. This along with #151, #156, and #159 should be ready to test now.

rxwei:
All tests passed.

PR #147, opened 2019-05-30
Added support for the 'log1mexp' op and its VJP.

eaplatanios:
This requires #145 and #146.


Sources/TensorFlow/Core/Tensor.swift
```diff
@@ -559,3 +559,9 @@ extension Tensor: Differentiable where Scalar: TensorFlowFloatingPoint {
     public typealias TangentVector = Tensor
     public typealias AllDifferentiableVariables = Tensor
 }
+
+extension Tensor {
```
rxwei: `withoutDerivative` is available in an extension to `Differentiable`. I'm not sure why this is necessary.

rxwei: If you meant the this is for when `Scalar` is not differentiable, I I think there are other solutions to this. I made a suggestion in a different comment.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -424,6 +424,20 @@ internal func _vjpLog<T: TensorFlowFloatingPoint>(
...
 }
 
+/// Computes `log(1 - exp(x))` using a numerically stable approach.
+///
+/// The approach is shown in Equation 7 of:
+/// https://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf.
+@inlinable
+@differentiable
+public func log1mexp<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
+    let isTooSmall = (-x .< T(log(2.0))).withoutDerivative()
```
```suggestion
    let isTooSmall = (-x.withoutDerivative() .< T(log(2.0)))
```

Sources/TensorFlow/Core/Tensor.swift
eaplatanios: It's required because if you have a `Tensor<Bool>`, for example (e.g., [here](https://github.com/tensorflow/swift-apis/blob/9e75132b54dcc93dca6dee49466e92b184fc83b2/Sources/TensorFlow/Operators/Math.swift#L434)), then you'd need to wrap it into a thunk to allow the function itself to be differentiable. Is there a nicer way to do this?

eaplatanios: Nvm, I saw your other comment. :)

Sources/TensorFlow/Operators/Math.swift
eaplatanios: This fix still does not work actually because the resulting tensor is still boolean.

eaplatanios:
Added a test similar to #144, but we still need to figure out what to do about the `.withoutDerivative()`. Let's revisit this after #145 and #146 are merged because this depends on them.

eaplatanios:
@rxwei this is ready other than the function not being differentiable. If we figure out a `withoutDerivative` API I could make this differentiable, by using it on `let isTooSmall = -x .< T(log(2.0))`.

rxwei:
Let's add a global `withoutDerivative(x)` function.

eaplatanios:
@rxwei I implemented that, but I get the following error:
```
Assertion failed: (Index < Length && "Invalid index!"), function operator[], file /Users/eaplatanios/Development/GitHub/swift-source-temp/llvm/include/llvm/ADT/ArrayRef.h, line 241.
Stack dump:
0.      Program arguments: /Library/Developer/Toolchains/swift-tensorflow-LOCAL-2019-05-30-a.xctoolchain/usr/bin/swift -frontend -c /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Bindings/EagerExecution.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Bindings/RawOpsGenerated.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Bindings/TFTensorOperation.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Context.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/ArrayOps.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/DataTypes.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/DifferentialOperators.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/Execution.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/PythonConversion.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/Runtime.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/ShapedArray.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/StringTensor.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/Tensor.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/TensorGroup.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/TensorHandle.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/TensorProtocol.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/TensorShape.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/Threading.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/Utilities.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Initializers.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layer.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Convolutional.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Core.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Normalization.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Pooling.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Recurrent.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Upsampling.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Loss.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Operators/Basic.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Operators/Comparison.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Operators/Dataset.swift -primary-file /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Operators/Math.swift -primary-file /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Operators/NN.swift -primary-file /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Optimizer.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Random.swift -emit-module-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Operators/Math~partial.swiftmodule -emit-module-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Operators/NN~partial.swiftmodule -emit-module-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Optimizer~partial.swiftmodule -emit-module-doc-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Operators/Math~partial.swiftdoc -emit-module-doc-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Operators/NN~partial.swiftdoc -emit-module-doc-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Optimizer~partial.swiftdoc -emit-dependencies-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Operators/Math.d -emit-dependencies-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Operators/NN.d -emit-dependencies-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Optimizer.d -emit-reference-dependencies-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Operators/Math.swiftdeps -emit-reference-dependencies-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Operators/NN.swiftdeps -emit-reference-dependencies-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Optimizer.swiftdeps -target x86_64-apple-macosx10.10 -enable-objc-interop -sdk /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk -I /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug -F /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/Library/Frameworks -enable-testing -g -module-cache-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/ModuleCache -swift-version 4.2 -Onone -D SWIFT_PACKAGE -D DEBUG -color-diagnostics -enable-anonymous-context-mangled-names -parse-as-library -module-name TensorFlow -o /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Operators/Math.swift.o -o /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Operators/NN.swift.o -o /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Optimizer.swift.o -index-store-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/index/store -index-system-modules 
1.      Swift version 5.1-dev (LLVM 082dec2e22, Swift 5a6c347496)
2.      While running pass #1956 SILModuleTransform "Differentiation".
0  swift                    0x000000010f22fd35 llvm::sys::PrintStackTrace(llvm::raw_ostream&) + 37
1  swift                    0x000000010f22eff5 llvm::sys::RunSignalHandlers() + 85
2  swift                    0x000000010f230318 SignalHandler(int) + 264
4  libsystem_platform.dylib 0x0000000118064938 _sigtramp + 2724888056
7  swift                    0x000000010bcd9690 (anonymous namespace)::ADContext::getOrCreateSubsetParametersThunkForAssociatedFunction(swift::SILValue, swift::SILValue, swift::AutoDiffAssociatedFunctionKind, swift::SILAutoDiffIndices, swift::SILAutoDiffIndices) + 14160
8  swift                    0x000000010bcd517a (anonymous namespace)::ADContext::promoteToDifferentiableFunction(swift::AutoDiffFunctionInst*, swift::SILBuilder&, swift::SILLocation, (anonymous namespace)::DifferentiationInvoker) + 12122
9  swift                    0x000000010bc95038 (anonymous namespace)::ADContext::processAutoDiffFunctionInst(swift::AutoDiffFunctionInst*) + 184
10 swift                    0x000000010bc92371 (anonymous namespace)::Differentiation::run() + 2209
11 swift                    0x000000010bd0e18e swift::SILPassManager::runModulePass(unsigned int) + 734
12 swift                    0x000000010bd0ec64 swift::SILPassManager::execute() + 692
13 swift                    0x000000010b8645eb swift::SILPassManager::executePassPipelinePlan(swift::SILPassPipelinePlan const&) + 187
14 swift                    0x000000010bd14324 swift::runSILDiagnosticPasses(swift::SILModule&) + 132
15 swift                    0x000000010b92c29e swift::CompilerInstance::performSILProcessing(swift::SILModule*, swift::UnifiedStatsReporter*) + 62
16 swift                    0x000000010b6f10e9 performCompile(swift::CompilerInstance&, swift::CompilerInvocation&, llvm::ArrayRef<char const*>, int&, swift::FrontendObserver*, swift::UnifiedStatsReporter*) + 11273
17 swift                    0x000000010b6ed471 swift::performFrontend(llvm::ArrayRef<char const*>, char const*, void*, swift::FrontendObserver*) + 3025
18 swift                    0x000000010b6934a9 main + 729
19 libdyld.dylib            0x00007fff757d13d5 start + 1
```

This appears when I make the following function differentiable:
```swift
@inlinable
@differentiable
public func log1mexp<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
    let isTooSmall = withoutDerivative(at: -x .< T(log(2.0)))
    // This `replacing` will ultimately be a no-op because we will not select this code-path 
    // whenever we use the surrogate `-Tensor(onesLike: x)`.
    let xSafe = x.replacing(with: -Tensor(onesLike: x), where: isTooSmall)
    return log1p(-exp(xSafe)).replacing(with: log(-expm1(x)), where: isTooSmall)
}
```

rxwei:
How is it defined?

eaplatanios:
@rxwei It's just this:
```swift
@_semantics("autodiff.nonvarying")
public func withoutDerivative<T>(at x: T) -> T {
    return x
}
```

I just pushed the changes so you can see them.

eaplatanios:
@rxwei I made a small reproducible example. I cannot make the following function differentiable:
```swift
@inlinable
@differentiable
public func example<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
    let ones = withoutDerivative(at: Tensor<T>(ones: x.shape))
    return x.replacing(with: ones, where: Tensor<Bool>(true))
}
```
I get a similar stacktrace as above and I believe that this error does not have to do with `withoutDerivative` as I also get it with this:
```swift
@inlinable
@differentiable
public func example<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
    return x.replacing(with: Tensor<T>(ones: x.shape), where: Tensor<Bool>(true))
}
```

rxwei:
Hmm that crasher is indeed annoying. I'll try debug this tonight!

eaplatanios:
@rxwei did you manage to figure out what the problem is?

eaplatanios:
@rxwei I updated this PR to also use the new `withoutDerivative(at:in:)` function, but I still get the same auto-diff crash. Did you get a chance to look into it?

Also, I made some small cleanup and also fixed [TF-543](https://bugs.swift.org/browse/TF-543). One thing I noticed is that many new functions were added in random places in the `Math.swift` file (e.g., the `cosineSimilarity` and `cosineDistance` functions were placed in the unary ops region instead of the binary ops region). I'll try to look more for that when reviewing future PRs, but I thought you guys might also want to be aware. On the same note, it'd be nice to have the order of the tests be aligned with the order in which the functions being tested are defined in the `Math.swift` file as that makes it easier to navigate the test files. I made some small changes but we should probably also keep that in mind when reviewing future PRs.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -576,6 +576,21 @@ func _vjpLog1p<T: TensorFlowFloatingPoint>(
     (log1p(x), { v in Raw.xdivy(v, 1 + x) })
 }
 
+/// Computes `log(1 - exp(x))` using a numerically stable approach.
```
```suggestion
/// Returns `log(1 - exp(x))` using a numerically stable approach.
```

#160

rxwei:
> @rxwei I updated this PR to also use the new `withoutDerivative(at:in:)` function, but I still get the same auto-diff crash. Did you get a chance to look into it?
> 
> Also, I made some small cleanup and also fixed [TF-543](https://bugs.swift.org/browse/TF-543). One thing I noticed is that many new functions were added in random places in the `Math.swift` file (e.g., the `cosineSimilarity` and `cosineDistance` functions were placed in the unary ops region instead of the binary ops region). I'll try to look more for that when reviewing future PRs, but I thought you guys might also want to be aware. On the same note, it'd be nice to have the order of the tests be aligned with the order in which the functions being tested are defined in the `Math.swift` file as that makes it easier to navigate the test files. I made some small changes but we should probably also keep that in mind when reviewing future PRs.

Agreed, proper placement is definitely important. Thanks for bringing this up! cc @Shashi456

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -923,6 +921,14 @@ internal func _vjpSigmoid<T: TensorFlowFloatingPoint>(
     (sigmoid(x), { v in Raw.sigmoidGrad(x, dy: v) })
 }
 
+/// Returns the log-sigmoid of the specified tensor element-wise. Specifically,
+/// `y = log(1 / (1 + exp(-x)))`. For numerical stability, we use `y = -softplus(-x)`.
```
```suggestion
/// `log(1 / (1 + exp(-x)))`. For numerical stability, we use `-softplus(-x)`.
```

eaplatanios:
I just updated this, but the auto-diff bug persists on my machine. I'll try to run the CI tests to see if it also comes up in Linux.

eaplatanios:
@rxwei I can now confirm that the same crash happens on the CI.

eaplatanios:
@rxwei I tried to dig a bit into where the error is coming from and I know the following so far:

- The crash happens [here](https://github.com/apple/swift/blob/def2be969bb8212f6cfc388698a4a58d4c4283b5/lib/SILOptimizer/Mandatory/Differentiation.cpp#L5975).
- For the example from this PR, `i` in that loop starts with value `0` and then in the second iteration it has value `2`. However, `linearMapType->getResults()` has length `2` and so this results in a crash on the second iteration.

Do you know what could be causing this from this description? Could it be possible that  `linearMapType->getResults()` should be indexed by the loop iteration number rather than by `I`? I don't know anything really about this part of the codebase and I can't easily tell.

rxwei:
Looks like a bug in the differentiation transform. I'm debugging now.

eaplatanios:
I don't know if this helps but I notice a bit of confusion wrt to parameters vs results here:
```c++
    // Iterate over actual indices.
    for (unsigned i : actualIndices.parameters->getIndices()) {
      auto resultInfo = linearMapType->getResults()[i];
      // Skip direct results. Only indirect results are relevant as arguments.
      if (resultInfo.isFormalDirect())
        continue;
      // If index is desired, use next indirect result.
      if (desiredIndices.isWrtParameter(i)) {
        useNextResult();
        continue;
      }
      // Otherwise, construct and use a zero indirect result.
      buildZeroArgument(resultInfo.getSILStorageType());
    }
```
We iterate over parameter indices but we assume there always exist corresponding results in the original function type? My understanding is limited but I believe something may be inconsistent here.

rxwei:
You can use `swiftc -Xllvm -debug-only=differentiation` to print all differentiation log.

eaplatanios:
I actually attached CLion LLDB debugger and I'm able to step through this part which I find convenient. My issue is that I don't exactly understand why we use parameter indices to index over results of the original function here.

rxwei:
Yeah, I'm also running through this in a debugger. I'm trying to see why `ADContext::getOrCreateSubsetParametersThunkForAssociatedFunction` is trying to access invalid indices. 


> I actually attached CLion LLDB debugger and I'm able to step through this part which I find convenient. My issue is that I don't exactly understand why we use parameter indices to index over results of the original function here.

This is expected, because a pullback's results correspond to each original parameter that's being differentiated with repsect to.

eaplatanios:
> This is expected, because a pullback's results correspond to each original parameter that's being differentiated with repsect to.

I understand that part, but it seems like we're indexing over result types of the original function. I'm wondering if `linearMapType->getResults()[i]` should be replaced by something like `thunk->getResults()[i]`.

rxwei:
```console
[AD] Getting a subset parameters thunk for @callee_guaranteed (@guaranteed Tensor<τ_0_0>) -> (@owned Tensor<τ_0_0>, @owned Tensor<τ_0_0>) from (source=0 parameters=(0 2)) to (source=0 parameters=(2))
```
I don't think there's any problem with `ADContext::getOrCreateSubsetParametersThunkForAssociatedFunction`. The problem should be with the caller.

eaplatanios:
My understanding is that `actualIndices` contains `0` and `2` due to [this line](https://github.com/apple/swift/blob/def2be969bb8212f6cfc388698a4a58d4c4283b5/lib/SILOptimizer/Mandatory/Differentiation.cpp#L1986).

eaplatanios:
I see. I'm really familiar with the semantics of these index sets so it's hard for me to debug this. Please let me know if you find anything or if there is any way in which I can help.

By the way, are you planning to release any documentation for auto-diff internals or something like that which includes details about how the auto-diff mechanics are designed as part of the Swift compiler?

rxwei:
> By the way, are you planning to release any documentation for auto-diff internals or something like that which includes details about how the auto-diff mechanics are designed as part of the Swift compiler?

Definitely, though today is not the best time to do it. The existing implement is still quite distant from [the design overview](bit.ly/swift-autodiff), so if we document the internals today, a lot of things will be obsolete within a month or two.

eaplatanios:
I see that makes sense. I'm looking forward to it in either case. In the meantime I hope I'll gain some basic understanding by trying to implement `inout` support. :)

Regarding this crasher, it seems that `2` is included in the parameter indices from [much earlier in the code](https://github.com/apple/swift/blob/70d7cc79f160b49106580f6a795cfeda6b34b426/lib/SILOptimizer/Mandatory/Differentiation.cpp#L6194).

rxwei:
Still debugging. I don't think the index `2` is wrong here because the original function has a third parameter:
```console
[AD] Processing AutoDiffFunctionInst:
     %36 = partial_apply [callee_guaranteed] %35<τ_0_0>() : $@convention(method) <τ_0_0 where τ_0_0 : TensorFlowScalar> (@guaranteed Tensor<τ_0_0>, @guaranteed Tensor<Bool>, @guaranteed Tensor<τ_0_0>) -> @owned Tensor<τ_0_0> // user: %37
->   %37 = autodiff_function [wrt 2] [order 1] %36 : $@callee_guaranteed (@guaranteed Tensor<τ_0_0>, @guaranteed Tensor<Bool>, @guaranteed Tensor<τ_0_0>) -> @owned Tensor<τ_0_0> // user: %38
     %38 = autodiff_function_extract [vjp] [order 1] %37 : $@differentiable @callee_guaranteed (@nondiff @guaranteed Tensor<τ_0_0>, @nondiff @guaranteed Tensor<Bool>, @guaranteed Tensor<τ_0_0>) -> @owned Tensor<τ_0_0> // user: %39
[AD] Getting a subset parameters thunk for associated function   %38 = partial_apply [callee_guaranteed] %37<τ_0_0>() : $@convention(method) <τ_0_0 where τ_0_0 : TensorFlowFloatingPoint> (@guaranteed Tensor<τ_0_0>, @guaranteed Tensor<Bool>, @guaranteed Tensor<τ_0_0>) -> (@owned Tensor<τ_0_0>, @owned @callee_guaranteed (@guaranteed Tensor<τ_0_0>, @guaranteed Tensor<τ_0_0>) -> @owned Tensor<τ_0_0>)
 of the original function   %36 = partial_apply [callee_guaranteed] %35<τ_0_0>() : $@convention(method) <τ_0_0 where τ_0_0 : TensorFlowScalar> (@guaranteed Tensor<τ_0_0>, @guaranteed Tensor<Bool>, @guaranteed Tensor<τ_0_0>) -> @owned Tensor<τ_0_0> // user: %39
 from (source=0 parameters=(0 2)) to (source=0 parameters=(2))
[AD] Getting a subset parameters thunk for @callee_guaranteed (@guaranteed Tensor<τ_0_0>, @guaranteed Tensor<τ_0_0>) -> @owned Tensor<τ_0_0> from (source=0 parameters=(0 2)) to (source=0 parameters=(2))
[AD] Getting a subset parameters thunk for associated function   %43 = partial_apply [callee_guaranteed] %42<τ_0_0>() : $@convention(method) <τ_0_0 where τ_0_0 : TensorFlowFloatingPoint> (@guaranteed Tensor<τ_0_0>, @guaranteed Tensor<Bool>, @guaranteed Tensor<τ_0_0>) -> (@owned Tensor<τ_0_0>, @owned @callee_guaranteed (@guaranteed Tensor<τ_0_0>) -> (@owned Tensor<τ_0_0>, @owned Tensor<τ_0_0>))
 of the original function   %36 = partial_apply [callee_guaranteed] %35<τ_0_0>() : $@convention(method) <τ_0_0 where τ_0_0 : TensorFlowScalar> (@guaranteed Tensor<τ_0_0>, @guaranteed Tensor<Bool>, @guaranteed Tensor<τ_0_0>) -> @owned Tensor<τ_0_0> // user: %44
 from (source=0 parameters=(0 2)) to (source=0 parameters=(2))
[AD] Getting a subset parameters thunk for @callee_guaranteed (@guaranteed Tensor<τ_0_0>) -> (@owned Tensor<τ_0_0>, @owned Tensor<τ_0_0>) from (source=0 parameters=(0 2)) to (source=0 parameters=(2))
Assertion failed: (Index < Length && "Invalid index!"), function operator[], file /Users/rxwei/Development/Swift/swift-tf-source/llvm/include/llvm/ADT/ArrayRef.h, line 241.
```

eaplatanios:
Just throwing another kind of educated guess out there. What if we change `linearMapType->getResults()[i]` to `targetType->getResults()[i]`? I'm saying this because I noticed that in `buildThunkType` the result types are obtained from `targetType`.

rxwei:
Using `targetType` didn't work. 

```swift
// FIXME: The logic for resolving `assocRef` does not reapply function
  // conversions, which is problematic if `assocFn` is a `partial_apply`
  // instruction.
  SILValue assocRef;
  if (auto *assocFnRef =
          peerThroughFunctionConversions<FunctionRefInst>(assocFn)) {
    auto *assoc = assocFnRef->getReferencedFunctionOrNull();
    assocRef = builder.createFunctionRef(loc, assoc);
  } else if (auto *assocMethodInst =
                 peerThroughFunctionConversions<WitnessMethodInst>(assocFn)) {
    assocRef = builder.createWitnessMethod(
        loc, assocMethodInst->getLookupType(),
        assocMethodInst->getConformance(), assocMethodInst->getMember(),
        thunk->mapTypeIntoContext(assocMethodInst->getType()));
  }
  assert(assocRef && "Expected associated function to be resolved");
```

Confirmed this is related to this FIXME.

rxwei:
Disregard my previous response. I think what is going on here is that the subset parameters thunk generation logic is not handling cases where the original JVP/JVP is already producing a subset-parameters linear map. A proper fix would be to pass in a index subset **with respect to the linear map** `getOrCreateSubsetParametersThunkForAssociatedFunction`, not one with respect to the original function.
Do this:

eaplatanios:
I see. Could you please explain a bit more detail what the index subsets represent and what the linear map is exactly? That will help me a lot in understanding what’s going on and attempting a fix.

rxwei:
`actualIndices` and `desiredIndices` are with respect to the original function. However, the differential parameters and pullback results may already be w.r.t. a subset, for example:
- Original: `(T0, T1, T2) -> R`
- Actual indices: `0, 2`
- Original differential: `(T0, T2) -> R`
- Original pullback: `R -> (T0, T2)`
- Desired indices w.r.t. original: `2`
- Desired indices w.r.t. linear map: `1`

I have a PR that will fix this.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -576,6 +576,21 @@ func _vjpLog1p<T: TensorFlowFloatingPoint>(
     (log1p(x), { v in Raw.xdivy(v, 1 + x) })
 }
 
+/// Returns `log(1 - exp(x))` using a numerically stable approach.
+///
+/// The approach is shown in Equation 7 of:
```
```suggestion
/// - Note: The approach is shown in Equation 7 of:
```

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -576,6 +576,21 @@ func _vjpLog1p<T: TensorFlowFloatingPoint>(
...
+/// https://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf.
```
```suggestion
///   https://cran.r-project.org/web/packages/Rmpfr/vignettes/log1mexp-note.pdf.
```

rxwei:
Fixed in https://github.com/apple/swift/pull/25699.

eaplatanios:
Richard thanks a lot for debugging this and helping me also understand what's going on.

eaplatanios:
It looks like kokoro does not re-run tests even though I added the `kokoro:run` label, probably because it has ran them earlier for the same commit. Is there an easy way to force re-run with the new toolchain?

eaplatanios:
Nvm, it just started running them now.

eaplatanios:
@rxwei the CI seems to still be using a toolchain before the auto-diff fix. Could you please re-run tests once the CI toolchain is updated?

rxwei:
@bgogul, how long does a new toolchain take to get to Kokoro? A new one was built last night so I assumed that it had propagated to CI.

eaplatanios:
Sorry I believe the toolchain was updated after all but this PR was not. I will merge with master and re-run tests.

PR #149, opened 2019-05-31
 Added support for 'isFinite', 'isInfinite', and 'isNaN'.

eaplatanios:
The names are not very swifty. @rxwei do you have any better suggestions? We could make them functions of `Tensor`.


Sources/TensorFlow/Operators/Math.swift
```diff
@@ -686,6 +686,24 @@ func _vjpRelu<T: TensorFlowFloatingPoint>(
     return (relu(x), { v in Tensor(x .> 0) * v })
 }
 
+/// Checks which elements of `x` are finite.
+@inlinable
+public func isFinite<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<Bool> {
```
rxwei: I think they should be defined as computed properties on `Tensor`. The names sound good as properties.

eaplatanios: Sounds good. Done in the last commit.

eaplatanios:
@rxwei I just updated this and all math tests pass locally, so it should be ready to merge. There are some failing tests related to lazy tensor that are not related to this PR.

eaplatanios:
Also verified that all CI tests pass.

eaplatanios:
@rxwei This is also merged as part of #157. I just did a minor cleanup of the tests here. Should I just merge this when I confirm CI tests pass?

eaplatanios:
@rxwei Is it ok to directly merge minor fixes like this one as long as CI tests pass?

rxwei:
Yes, thank you!

PR #150, opened 2019-05-31
Added support for 'logSumExp'.

eaplatanios:
Depends on #149. @rxwei I was also relying on a general `Tensor.withoutDerivative()` for this. It'd be nice to find a nice way to deal with this situation more generally.


rxwei:
Could you use the global `Swift.withoutDerivative(at:in:)` for now?

eaplatanios:
I addressed that. It remains to add documentation and tests which I'll do later today.

eaplatanios:
@rxwei I added documentation and tests. The CI tests are running now and so this only requires an approving review.

eaplatanios:
CI tests pass so this is only pending a review.

rxwei:
> @rxwei I was also relying on a general `Tensor.withoutDerivative()` for this. It'd be nice to find a nice way to deal with this situation more generally.

Yeah I saw your use case. I think this justifies having a top-level `withoutDerivative<T>(_ x: T) -> T` identity function.

PR #151, opened 2019-05-31
Added support for 'gathering' and its VJP.

eaplatanios:


Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -310,6 +310,57 @@ public extension Tensor {
...
+    /// where `N = self.rank` and `M = indices.rank`.
+    ///
+    /// The shape of the resulting tensor is:
+    /// `self.shape[..<axis] + indices.shape + self.shape[(axis + 1)...]`.
+    ///
+    /// - Note: On CPU, if an out-of-range index is found, an error is thrown. On GPU, if an
+    /// out-of-range index is found, a 0 is stored in the corresponding output values.
+    ///
+    /// - Parameters:
+    ///   - indices: Contains the indices to gather.
```
```suggestion
    ///   - indices: Contains the indices to gather at.
```

rxwei:
I think this one needs to pull from master.

eaplatanios:
@rxwei this should be ready now.

Tests/TensorFlowTests/OperatorTests/BasicTests.swift
```diff
@@ -26,6 +26,12 @@ precedencegroup StridedRangeFormationPrecedence {
 }
 
 final class BasicOperatorTests: XCTestCase {
+    func testGathering() {
+        let x = Tensor<Float>([[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]])
+        let y = x.gathering(atIndices: Tensor<Int32>(2), alongAxis: 1)
+        assertEqual(y, Tensor<Float>([3.0, 6.0]))
```
rxwei:
```suggestion
        XCTAssertEqual(y, Tensor<Float>([3.0, 6.0]))
```

rxwei: I wonder how this passed tests. Where is `assertEqual` defined?

eaplatanios: Oh right. I created a `Helpers.swift` file. This is redundant, but the one that has the accuracy argument is needed. I'll remove this one.

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -375,6 +426,103 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+    /// // 'input' has 5 true values and so the output has 5 rows.
+    /// // 'input' has rank 3, and so the second dimension of the output has size 3.
+    /// input.nonZeroIndices() // is [[0, 0, 0],
+    ///                        //     [0, 1, 0],
+    ///                        //     [1, 0, 1],
+    ///                        //     [1, 1, 1],
+    ///                        //     [2, 1, 1]]
+    /// ```
+    ///
+    /// - Returns: A tensor with shape `(num_true, rank(condition))`.
```
rxwei: `num_true` isn't defined anywhere. Maybe reword this?

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -375,6 +426,103 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+            let inversePermutations = Tensor<Int32>(concatenating: [
+                outerIndices + 1,
+                Tensor<Int32>([0]),
+                innerIndices])
+            return gradient.transposed(withPermutations: inversePermutations)
+        })
+    }
+}
+
+public extension Tensor {
```
rxwei: I think that this API should be kept internal until we find a good spelling/semantics for it. The fact that it applies to `Tensor<Bool>` makes the `nonZero` name weird.

eaplatanios: We could split it in two functions: `nonZeroIndices` and `falseIndices` or something like that. How does that sound?

rxwei: Linguistically those adjectives are attached to "indices", which is not the intended meaning and could cause confusion. Also, which Python TF operator does it correspond to?

eaplatanios: It's part of the `tf.where` functionality when no `x` and `y` are provided (i.e., only `condition` is provided).

rxwei: It seems this is not being used by the implementation of `gathering`. Could we defer this to later PRs instead?

eaplatanios: That's true. I think it's used by one of the other PRs so I'll just move it there.

rxwei:
@eaplatanios Is it correct that this PR isn't blocked on any bugs?

rxwei:
Looks like #156 has already added everything!

PR #152, opened 2019-05-31
Added support for a 'tiled(multiples:)' and its VJP.

eaplatanios:


Sources/TensorFlow/Initializers.swift
```diff
@@ -180,6 +180,20 @@ public extension Tensor {
...
+    ///
+    /// This constructor creates a new tensor by replicating `tensor` `multiples` times. The
+    /// constructed tensor's `i`'th dimension has `tensor.shape[i] * multiples[i]` elements, and the
+    /// values of `tensor` are replicated `multiples[i]` times along the `i`'th dimension. For
+    /// example, tiling `[a b c d]` by `[2]` produces `[a b c d a b c d]`.
+    ///
+    /// - Precondition: The shape of `multiples` must be `[tensor.rank]`.
+    @inlinable
+    @differentiable(wrt: tensor, vjp: _vjpTiling where Scalar: TensorFlowFloatingPoint)
+    init(tiling tensor: Tensor, multiples: Tensor<Int32>) {
```
rxwei: I feel that a method would be better:

```swift
extension Tensor {
    func tiled(multiples: Tensor<Int32>) -> Tensor
}
```

WDYT?

Shashi456:
xD Thats a load of PRs in a few span of hours. @eaplatanios @rxwei if you need any help in porting anything do tell, I'd be happy to help :P

Sources/TensorFlow/Initializers.swift
eaplatanios: That's a good point. I debated this actually and leaned towards a constructor due to the similar `Tensor(stacking:)` and `Tensor(concatenating:)` constructors. We could also have both given that both may feel natural to users, but I typically lean towards avoiding multiple ways to do the same thing.

eaplatanios:
Thanks a lot @Shashi456! Most of them were long time coming, waiting for the migration to be completed first. Now that it's done all of this is much easier because testing is much faster.

Sources/TensorFlow/Initializers.swift
rxwei: I see you point. I thought about this for a while also, but I think that tiling is quite a different case in that it is not operating on multiple tensors like `init(stacking:)` and `init(concatenating:)`, but on one tensor instead (like `split`).

I too prefer not introducing multiple ways to do the same thing, and so far these operators are the only way to do so. However, the choice between a method and an initializer should depend on their actual semantics.

In general, here are some of my personal design principles that I've been following:
- Operators that take multiple tensors and return a single one should be initializers, e.g. `init(stacking:)`, `init(concatenating:)`.
- Operators that take a single tensor and produce multiple tensors should be methods, e.g. `split`.
- Operators that take a single tensor and produce a tensor should usually be a method or a top-level function. When there's no significant prior art for this operator as a top-level function, a method is preferred.

WDYT?

eaplatanios: I like the way you put it and it's nice that you laid out a few design principles. I'll go ahead and modify that. At some point we could put such principles in a doc somewhere as they might also help users figure out where to find desired functionality.

eaplatanios: Done.

rxwei: > At some point we could put such principles in a doc somewhere as they might also help users figure out where to find desired functionality.

Totally! I have some work-in-progress on such a doc and plan to finish it some time in July.

eaplatanios: That's awesome! :) Let me know if you'd like any feedback at any point.

rxwei:
All tests passed! Thanks!

PR #157, opened 2019-05-31
Added support for 'Tensor.batchGathering(atIndices:)'.

eaplatanios:
@rxwei this also makes use of `Tensor.withoutDerivative` where `Scalar` is not differentiable. How do you think we should avoid these uses here?

I also have a more general version of this that requires control flow AD so instead of committing it as non-differentiable but more general function now, I'll leave that one for later.


Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -310,6 +310,100 @@ public extension Tensor {
...
+    ///
+    /// - Precondition: `batchDims` must be less than `indices.rank`.
+    ///
+    /// - Returns: The gathered tensor.
+    @inlinable
+    @differentiable(wrt: self where Scalar : TensorFlowFloatingPoint)
+    func batchGathering(atIndices indices: Tensor<Int32>) -> Tensor {
+        var batchIndices = indices
+        var accumulated = Tensor<Int32>(ones: [])
+        accumulated *= shapeTensor.withoutDerivative()[1]
```
rxwei: How about using `shape` instead of `shapeTensor`? That way you won't need `withoutDerivative()`.

eaplatanios: We would need to eventually convert all these shapes to tensors and so that would generate a potentially significant performance penalty. I also think the same differentiability issue will come up when converting them to tensors that are used further down the line.

rxwei: I see. I also see the problem with `withoutDerivative()` being unavailable now: In conditionally differentiable functions, the tensor type does not conform to `Differentiable`.

Since the issue happens to all conditionally differentiable functions, it's necessary to introduce an global API for any type. I have two options in mind:
1. Introduce an identity function API.
```swift
@_semantics("autodiff.nonvarying")
public func withoutDerivative<T>(at x: T) -> T
```
2. Introduce a trailing closure API.
```swift
@_semantics("autodiff.nonvarying")
public func withoutDerivative<T, R>(at x: T, in f: (T) -> R) -> R
```

I'm a little torn between these.

eaplatanios: I see. I like the explanation on conditional differentiability. That makes sense.

In terms of usability I'd go with option 1 because I believe option 2 would often result in lots of nesting. Also, in most cases you'd like to specify that a particular tensor is not differentiable.

eaplatanios: Regarding the identity API, that could also be defined as a function of `Tensor`, to be "compatible" with the existing function on `Differentiable`. This could look like:
```swift
public extension Tensor {
  @inlinable
  @_semantics("autodiff.nonvarying")
  func withoutDerivative() -> Tensor { return self }
}
```

eaplatanios:
@rxwei I removed `Tensor.nonZeroIndices()`, as discussed and added a `noDerivative` global function helper. I could not name it `withoutDerivative` because that causes compilation errors due to a conflicts with the existing `withoutDerivative`. This passes all tests now.

eaplatanios:
Nvm the `Tensor.nonZeroIndices()` edit. It's actually used by `Tensor.gathering(where:)` and has already been merged into master.

eaplatanios:
@rxwei I updated this PR so that all tests pass, but there is one issue that came up. I had to use `Swift.withoutDerivative(at:in:)` because if I don't use the module to disambiguate the compiler tries to use the instance method of `Differentiable` and fails. Not sure why though since the signatures are different. This seems like a compiler bug.

rxwei:
`Swift.withoutDerivative(at:in:)` sounds good. This is a bug which I was trying to fix with https://github.com/apple/swift/pull/25316, but it needs a more work to get there.

eaplatanios:
Right, I also noticed the same problem with the new elementary function support. Thanks for the update! This should be ready to merge then, along with the few other PRs I updated today. I also ran CI tests on them.

rxwei:
Just got back from a conference. Will get to these PRs tonight!

eaplatanios:
Thanks Richard!

Issue #161, opened 2019-05-31
Fix failing tests on Linux.


rxwei:
`TrivialModelTests.testXOR` has been failing on the Linux CI machine. It does not fail on macOS. This also seems related to the case where `-O` is enabled on the `TensorFlow` module in https://github.com/apple/swift/pull/25166.

```console
Test Case 'TrivialModelTests.testXOR' started at 2019-05-31 07:16:34.843
931
Exited with signal code 11
932
The command '/bin/sh -c /swift-tensorflow-toolchain/usr/bin/swift test' returned a non-zero code: 1
```

rxwei:
cc @bgogul

bgogul:
Here is the stack trace of the crash:
```console
* thread #1, name = 'TensorFlowPacka', stop reason = signal SIGSEGV: invalid address (fault address: 0x0)
  * frame #0: 0x00007ffff7c64a9b libswiftCore.so`swift_getAssociatedTypeWitnessSlowImpl(swift::MetadataRequest, swift::TargetWitnessTable<swift::InProcess>*, swift::TargetMetadata<swift::InProcess> const*, swift::TargetProtocolRequirement<swift::InProcess> const*, swift::TargetProtocolRequirement<swift::InProcess> const*) + 203
    frame #1: 0x00007ffff7c62f2e libswiftCore.so`swift_getAssociatedTypeWitness + 142
    frame #2: TensorFlowPackageTests.xctest`$s10TensorFlow20_ShapedArrayProtocolPA2aB7ElementRpzAD_ADQZAERSrlE11description33_DBF849D2A9DA24E23B4FF3FA4D3704F0LL11indentLevel04edgeF5Count15maxScalarLength0vwU7PerLine11summarizingSSSi_S3iSbtF at <compiler-generated>:0
    frame #3: TensorFlowPackageTests.xctest`$s10TensorFlow20_ShapedArrayProtocolPA2aB7ElementRpzAD_ADQZAERSrlE11description33_DBF849D2A9DA24E23B4FF3FA4D3704F0LL11indentLevel04edgeF5Count15maxScalarLength0vwU7PerLine11summarizingSSSi_S3iSbtF18elementDescriptionL_ySSAeaBRzAabERQAfERSlF at ShapedArray.swift:356
[...]

PR #162, opened 2019-05-31
Temporarily disable trivial model tests that are failing on Linux.

rxwei:
#161


eaplatanios:
I believe all layer-related tests cause this issue.

PR #163, opened 2019-05-31
Update DockerFile to use the TEST tool chain

bgogul:


bgogul:
That is because the test toolchain is currently building. Should be
available in an hour or so.

bgogul:
Ah, the test toolchain build is failing for some reason. I am looking into it. The other toolchains are building fine.

bgogul:
@saeta I fixed the issues and merged it in to unblock other PRs.

PR #169, opened 2019-06-03
Make seed type to be Int32 so that it is compatible with TPUs.

bgogul:
Otherwise, @burmako and I encountered the following error when running MNIST on TPUs:

```
 (OpKernel was found, but attributes didn't match) Requested Attributes: T=DT_INT32, Tseed=DT_INT64, dtype=DT_FLOAT
	.  Registered:  device='TPU'; Tseed in [DT_INT32]; T in [DT_INT32, DT_INT64]; dtype in [DT_FLOAT, DT_BFLOAT16]
```

### DIFF
--- a/Sources/TensorFlow/Initializers.swift
+++ b/Sources/TensorFlow/Initializers.swift
@@ -376,12 +376,12 @@ public extension Tensor where Scalar: BinaryFloatingPoint {
     ///
     init(
         randomUniform shape: TensorShape,
-        seed: (Int64, Int64) = (Int64.random(in: Int64.min..<Int64.max),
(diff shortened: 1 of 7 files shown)


Shashi456:
@bgogul just purely curious here, you were trying to use s4tf to train a model on a tpu?

bgogul:
> @bgogul just purely curious here, you were trying to use s4tf to train a model on a tpu?
> 
> That's so cool :o.

@Shashi456, yes.  :) now that GPE is deprecated, we are experimenting with alternative ways of extracting graphs.  I will have more PRs on this shortly.

PR #171, opened 2019-06-03
Make Element requirement in TensorArrayProtocol.init a conformance.

bgogul:
Related to https://bugs.swift.org/browse/TF-542

### DIFF
--- a/Sources/TensorFlow/Core/TensorGroup.swift
+++ b/Sources/TensorFlow/Core/TensorGroup.swift
@@ -35,7 +35,7 @@ public protocol TensorArrayProtocol {
     var _tensorHandles: [_AnyTensorHandle] { get }
 
     init(_owning tensorHandles: UnsafePointer<CTensorHandle>?, count: Int)
-    init<C: RandomAccessCollection>(_handles: C) where C.Element : _AnyTensorHandle
+    init<C: RandomAccessCollection>(_handles: C) where C.Element: _AnyTensorHandle
 }
 
 /// A protocol representing types that can be mapped to and from `Array<CTensorHandle>`.
@@ -102,7 +102,7 @@ extension TensorHandle: TensorGroup {
 
     public init<C: RandomAccessCollection>(
         _handles: C
-    ) where C.Element : _AnyTensorHandle {
+    ) where C.Element: _AnyTensorHandle {
         precondition(_handles.count == 1)
         self.init(handle: _handles[_handles.startIndex])
     }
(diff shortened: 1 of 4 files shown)


rxwei:
Do you ever need to call `init(_handles:)` on an array of type-erased `_AnyTensorHandle`s? If so, this change would make that impossible.

Sources/TensorFlow/Core/TensorGroup.swift
```diff
@@ -35,7 +35,7 @@ public protocol TensorArrayProtocol {
     var _tensorHandles: [_AnyTensorHandle] { get }
 
     init(_owning tensorHandles: UnsafePointer<CTensorHandle>?, count: Int)
-    init<C: RandomAccessCollection>(_handles: C) where C.Element == _AnyTensorHandle
+    init<C: RandomAccessCollection>(_handles: C) where C.Element : _AnyTensorHandle
```
rxwei: The Google Swift Style Guide recommends no whitespaces before colons in a conformance constraint.

```suggestion
    init<C: RandomAccessCollection>(_handles: C) where C.Element: _AnyTensorHandle
```

bgogul:
> Do you ever need to call `init(_handles:)` on an array of type-erased `_AnyTensorHandle`s? If so, this change would make that impossible.

There is one use case in the test. That is why I made the following change: 
```
+copy(of: TensorHandle<T>) -> TFETensorHandle
-copy(of: TensorHandle<T>) -> _AnyTensorHandle
```

Without this change, I was not able to initialize a TensorGroup with an array of specific type, e.g.,  `[TFETensorHandle] or `[LazyTensor]`. Is there a better way to achieve this?

rxwei:
A `[TFETensorHandle]` should be directly convertible to a `[_AnyTensorHandle]` because it's a subtype.

bgogul:
> A `[TFETensorHandle]` should be directly convertible to a `[_AnyTensorHandle]` because it's a subtype.

Ah, that is right. The problem that I faced was with using  an `ArraySlice<TFETensorHandle>` as argument to `init(_handles:_)`.

Sources/TensorFlow/Core/TensorGroup.swift
```diff
@@ -290,7 +290,8 @@ extension Array: TensorArrayProtocol where Element: TensorGroup {
 
     public init<C: RandomAccessCollection>(
         _handles: C
-    ) where C.Element == _AnyTensorHandle {
+    ) where C.Element: _AnyTensorHandle {
+
```
rxwei:
```suggestion
```

rxwei:
I see. Yeah, variance only works with a few stdlib types in Swift such as `Array`.

Sources/TensorFlow/Core/TensorGroup.swift
bgogul: Done.

PR #172, opened 2019-06-04
[Layer] 'func call' -> 'func callAsFunction'.

rxwei:
Apply the final naming decision in [SE-0253](https://github.com/apple/swift-evolution/blob/master/proposals/0253-callable.md) to the TensorFlow library. This depends on [apple/swift#25235](https://github.com/apple/swift/pull/25235).

Shall be merged in time for the Swift for TensorFlow 0.4 release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ struct Model: Layer {
     var layer3 = Dense<Float>(inputSize: hiddenSize, outputSize: 3, activation: identity)
     
     @differentiable
-    func call(_ input: Tensor<Float>) -> Tensor<Float> {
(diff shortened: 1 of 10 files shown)


pschuh:
I think we're approaching the point where changes like this should compile before and after the breaking compiler change.

rxwei:
How would you make that work? This depends on a compiler change, but a toolchain with new new changes cannot be built without this change. It's a circular dependency.

pschuh:
Here are some approaches:
If the feature is orthogonal (a rename), make both names work.
A more general approach that will always work is to protect with a flag inside the compiler or a flag in the swift-apis repo.

rxwei:
Sounds good.

PR #177, opened 2019-06-04
Temporarily add back 'func call' to unblock CI.

pschuh:
This allows toolchains built with call or callAsFunction to work on swift-apis. This is to smooth over the transition period.


rxwei:
CI is not responding

Issue #160, opened 2019-05-31
Update doc comments on non-mutating API to use proper wording.


rxwei:
Doc comments for non-mutating methods often starts with `Returns a ... by <verb>ing ...` instead of `<Verb>s ...`. Quite a few doc comments in the library do not follow this guideline yet and should be updated.

PR #183, opened 2019-06-08
Update doc comments on non-mutating API to use proper wording.

Shashi456:
Fixes a part of #160.

@rxwei could you comment after seeing this once and i will follows suit for the rest of the files in `Operators/`

### DIFF
--- a/Sources/TensorFlow/Operators/Basic.swift
+++ b/Sources/TensorFlow/Operators/Basic.swift
@@ -34,15 +34,15 @@ public extension TensorFlowScalar {
 }
 
 public extension Tensor {
-    /// Unpacks the given dimension of a rank-`R` tensor into multiple rank-`(R-1)` tensors. Unpacks 
+    /// Unpacks the given dimension of a rank-`R` tensor into multiple rank-`(R-1)` tensors. Unpacks
     /// `N` tensors from this tensor by chipping it along the `axis` dimension, where `N` is
     /// inferred from this tensor's shape. For example, given a tensor with shape `[A, B, C, D]`:
-    /// 
-    ///   - If `axis == 0` then the `i`-th tensor in the returned array is the slice 
-    ///     `self[i, :, :, :]` and each tensor in that array will have shape `[B, C, D]`. 
+    ///
+    ///   - If `axis == 0` then the `i`-th tensor in the returned array is the slice
+    ///     `self[i, :, :, :]` and each tensor in that array will have shape `[B, C, D]`.
     ///     (Note that the dimension unpacked along is gone, unlike
     ///     `Tensor.split(numSplits:alongAxis)`, or `Tensor.split(sizes:alongAxis)`).
-    ///   - If `axis == 1` then the `i`-th tensor in the returned array is the slice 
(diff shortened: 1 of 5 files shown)


Shashi456:
Update: `Math.swift` is remaining, It's huge. So if the current changes are right, I'll start on that.

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -315,7 +315,7 @@ public extension Tensor {
         return transposed(withPermutations: Tensor<Int32>(defaultPermutations))
     }
 
-    /// Concatenates tensors along the specified axis.
+    /// Returns a concatenated tensors along the specified axis.
```
```suggestion
    /// Returns a concatenated tensor along the specified axis.
```

Sources/TensorFlow/Operators/Comparison.swift
```diff
@@ -20,80 +20,80 @@ infix operator .==: ComparisonPrecedence
 infix operator .!=: ComparisonPrecedence
 
 public extension Tensor where Scalar: Numeric & Comparable {
-    /// Computes `lhs < rhs` element-wise and returns a `Tensor` of Boolean /// scalars.
+    /// Returns a `Tensor` of Boolean scalars by computing `lhs < rhs` element-wise.
```
rxwei: Other comments mention "tensor" instead of backquoted `Tensor`. How about making it consistent here?
```suggestion
    /// Returns a tensor of Boolean scalars by computing `lhs < rhs` element-wise.
```

Same for other changes in this file.

Sources/TensorFlow/Operators/Dataset.swift
```diff
@@ -62,7 +62,7 @@ public extension Dataset {
 }
 
 public extension Dataset {
-    /// Creates a dataset from a batch of elements as a tensor.
+    /// Returns a tensor by creating a dataset from a batch of elements.
```
rxwei: The wording described in #160 only applies to non-mutating instance methods, not initializers. The original doc comment here is already okay: All initializer doc comments should start with `Creates a ...`.

Sources/TensorFlow/Operators/NN.swift
```diff
@@ -77,7 +77,7 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
 }
 
 public extension Tensor where Scalar: BinaryFloatingPoint {
-    /// Computes the batch normalized tensor along the specified axis.
+    /// Returns the batch normalized tensor by computing it along the specified axis.
```
```suggestion
    /// Returns a tensor computed from batch-normalizing the input along the specified axis.
```

Sources/TensorFlow/Operators/Comparison.swift
Shashi456: @rxwei I thought the same, but the commenting style is extremely uneven, some tensors have `Tensor` and some don't.

Sources/TensorFlow/Operators/Dataset.swift
Shashi456: Yes, Sorry. I wasn't sure of this change. I'll change it.

Shashi456:
@rxwei, sorry for the trivial doubt, but does `Math.swift` have a lot of mutating or non-mutating methods?

Issue #127, opened 2019-05-21
Adding more Optimizers and Losses


Shashi456:
Just a small-ish roadmap to different Optimizers and losses we can look at to add :

Optimizers:- 
- [ ] Adam
- [ ]
- [ ]
- [ ]

Losses :-

rxwei:
Thanks for listing these!

Shashi456:
@rxwei just a small roadmap, so that we have in mind what we need to add.

liberium:
Maybe it would be better to organise this as a project? It's not a pull-request by nature.

Shashi456:
@liberium. This is not priority right now. I wrote this task list so that we know what we have and what we don't. We are currently implementing all the layers. This is just a simple checklist so that we know where we are, and if someone wants to implement they can help us. Issues are a good way to get open source contributions.

Shashi456:
@dan-zheng @rxwei so this Google group thread [here](https://groups.google.com/a/tensorflow.org/forum/m/#!topic/swift/nIVNOJ1RDYA) . Talks about regularization. Although we are a little far from that stage right now. I think one functionality we should look at adding is retrieving model weights. Is there an easy way to add this ? Could we start looking at this I would definitely like to help in adding this functionality.

eaplatanios:
@Shashi456 I believe that regularization schemes are heavily intertwined with the training abstractions, which do not yet exist in this repo. It would be nice to start a discussion around them at some point though.

jon-tow:
@Shashi456 Categorical Cross-Entropy seems to already be implemented through Softmax Cross-Entropy with Logits. Maybe we can cross it off the list?

Shashi456:
@rxwei So I'm almost done with adding all the losses. Thanks to @Descartess and @jon-tow for helping out in that matter. One of the only listed losses remaining is #31 which is the triplet loss, If you merge it I can add a test for it. 

cc : @dan-zheng @eaplatanios

Shashi456:
@eaplatanios is there any optimizer which hasn't been refactored yet?

eaplatanios:
All that can be refactored currently, have been. We are missing some functionality that would allow us to simplify AdaMax that is not straightforward to add so that’s the only one using key paths right now.

iyaja:
Have you considered adding the [LAMB optimizer](https://arxiv.org/abs/1904.00962) to the list? It seems to be one of the fastest so far (especially with large batches).

[...]

PR #185, opened 2019-06-08
Add hinge loss function

shingt:
Added hinge loss function.
Ref: https://github.com/tensorflow/swift-apis/issues/127

### DIFF
--- a/Sources/TensorFlow/Loss.swift
+++ b/Sources/TensorFlow/Loss.swift
@@ -36,13 +36,13 @@ public func meanAbsoluteError<Scalar: TensorFlowFloatingPoint>(
     return abs(expected - predicted).mean()
 }
 
-/// Computes the hinge loss between predictions and expectations.
+/// Returns the hinge loss between predictions and expectations.
 ///
 /// - Parameters:
 ///   - predicted: Predicted outputs from a neural network.
 ///   - expected: Expected values, i.e. targets, that correspond to the correct output.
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/Loss.swift
```diff
@@ -36,6 +36,18 @@ public func meanAbsoluteError<Scalar: TensorFlowFloatingPoint>(
     return abs(expected - predicted).mean()
 }
 
+/// Computes the hinge loss between predictions and expectations.
```
```suggestion
/// Returns the hinge loss between predictions and expectations.
```

#160 has some more context.

Sources/TensorFlow/Loss.swift
```diff
@@ -36,6 +36,18 @@ public func meanAbsoluteError<Scalar: TensorFlowFloatingPoint>(
...
+///
+/// - Parameters:
+///   - predicted: Predicted outputs from a neural network.
+///   - expected: Expected values, i.e. targets, that correspond to the correct output.
+@differentiable(wrt: predicted)
+public func hinge<Scalar: TensorFlowFloatingPoint>(
```
rxwei: I'd suggest renaming this to `hingeLoss(predicted:expected:)`, since "hinge loss" is the canonical name and would be more clear at the call site.

shingt: Got it, thanks for your feedback!

PR #187, opened 2019-06-09
Adding Squared Hinge Loss

Shashi456:
Build and test passes locally. 
Reference : #127


Shashi456:
@rxwei Is there a clip function? Or should I add it from [here](https://raw.githubusercontent.com/tensorflow/swift-bindings/master/RawOpsGenerated.swift)? The function I was thinking of adding was `clipbyvalue` would I have to define a vjp for it, given I will use it in a loss fn?

rxwei:
Not yet. Feel free to add one. I'd suggest naming it `clipped(min:max:)`.

Shashi456:
Yeah I'll add that in a while. Also, @rxwei this PR is good to go, The kokoro run was a success, the latest commit is just merging into master.

PR #191, opened 2019-06-10
Make some LazyTensor properties @usableFromInline

bgogul:


bgogul:
> Is there a reason more things need to be `@usableFromInline`? Public linkage was only necessary for GPE, transitively required for code serialization. Instead of adding more public symbols, I think we should clean up the `@inlinable` and `@usableFromInline`'s from our library.

Unfortunately, this is a consequence of already having @inlinable in some parts of the TensorFlow library. In a downstream PR, I am using these in a function in `Runtime.swift` that is marked as @inlinable. We have https://bugs.swift.org/browse/TF-513 to clean up inlining-related annotations.

PR #193, opened 2019-06-11
Add support for Mean Squared Logarthmic Error

jon-tow:
Adding support for mean squared logarithmic error along with a test for correctness. 

#127

### DIFF
--- a/Sources/TensorFlow/Loss.swift
+++ b/Sources/TensorFlow/Loss.swift
@@ -26,7 +26,7 @@ public func meanSquaredError<Scalar: TensorFlowFloatingPoint>(
 
 /// Computes the mean squared logarithmic error between predictions and expectations.
 ///
-/// Note that negative Tensor entries will be clamped at 0 to avoid undefined
+/// - Note: Negative Tensor entries will be clamped at 0 to avoid undefined
 /// logarithmic behavior, as log() is undefined for negative reals.
 ///
 /// - Parameters:
--- a/Sources/TensorFlow/Loss.swift
+++ b/Sources/TensorFlow/Loss.swift
@@ -27,7 +27,7 @@ public func meanSquaredError<Scalar: TensorFlowFloatingPoint>(
 /// Returns the mean squared logarithmic error between predictions and expectations.
 ///
 /// - Note: Negative tensor entries will be clamped at `0` to avoid undefined
-///   logarithmic behavior, as log() is undefined for negative reals.
+///   logarithmic behavior, as `log(_:)` is undefined for negative reals.
 ///
 /// - Parameters:
 ///   - predicted: Predicted outputs from a neural network.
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/Loss.swift
```diff
@@ -24,6 +24,23 @@ public func meanSquaredError<Scalar: TensorFlowFloatingPoint>(
     return (expected - predicted).squared().mean()
 }
 
+/// Computes the mean squared logarithmic error between predictions and expectations.
+///
+/// Note that negative Tensor entries will be clamped at 0 to avoid undefined
```
```suggestion
/// - Note: Negative Tensor entries will be clamped at 0 to avoid undefined
```

We can use the markup syntax for notes in documentation comments. These will be properly rendered in API documentation.

Sources/TensorFlow/Loss.swift
```diff
@@ -24,6 +24,23 @@ public func meanSquaredError<Scalar: TensorFlowFloatingPoint>(
...
+/// logarithmic behavior, as log() is undefined for negative reals.
```
```suggestion
///   logarithmic behavior, as log() is undefined for negative reals.
```

Sources/TensorFlow/Loss.swift
```diff
@@ -24,6 +24,23 @@ public func meanSquaredError<Scalar: TensorFlowFloatingPoint>(
     return (expected - predicted).squared().mean()
 }
 
+/// Computes the mean squared logarithmic error between predictions and expectations.
```
```suggestion
/// Returns the mean squared logarithmic error between predictions and expectations.
```

See #160 for more context.

Sources/TensorFlow/Loss.swift
jon-tow: Wow I did not think of even looking that up. Thank you for teaching me that!

Sources/TensorFlow/Loss.swift
```diff
@@ -24,6 +24,23 @@ public func meanSquaredError<Scalar: TensorFlowFloatingPoint>(
     return (expected - predicted).squared().mean()
 }
 
+/// Returns the mean squared logarithmic error between predictions and expectations.
+///
+/// - Note: Negative Tensor entries will be clamped at 0 to avoid undefined
```
rxwei: In documentation comments, we usually wrap numeric values in backquotes.

```suggestion
/// - Note: Negative Tensor entries will be clamped at `0` to avoid undefined
```

Sources/TensorFlow/Loss.swift
```diff
@@ -24,6 +24,23 @@ public func meanSquaredError<Scalar: TensorFlowFloatingPoint>(
...
+///   logarithmic behavior, as log() is undefined for negative reals.
```
rxwei: Swift functions are identified by their full name (base name + argument labels), so we should refer to them by their full name in documentation comments and wrap such references in backquotes. All of this is for having the best quality API documentation :)
```suggestion
///   logarithmic behavior, as `log(_:)` is undefined for negative reals.
```

Sources/TensorFlow/Loss.swift
rxwei: Glad it helped!

jon-tow:
I'm all for proper API documentation :smile:. Thanks for the the guidance.

Shashi456:
@rxwei Is the swift log natural log or log base 10?

rxwei:
`log(_:)` is natural log, same as TensorFlow's.

rxwei:
Please disregard the test failure—it's a compiler regression introduced in https://github.com/apple/swift/pull/25293. It will be fixed soon!

PR #196, opened 2019-06-11
Handle concrete values when pretty printing a `LazyTensorOperation`

bgogul:


Tests/TensorFlowTests/LazyTensorOperationTests.swift
```diff
@@ -33,17 +33,17 @@ final class LazyTensorOperationTests: XCTestCase {
         let op0 = LazyTensorOperation(
             _id: "0", name: "Identity", outputCount: 1)
         op0.addInput(zeroTFEHandle)
-        XCTAssertEqual("\(op0)", "Identity_0[](conc):1")
+        XCTAssertEqual("\(op0)", "Identity_0[](0.0):1")
```
rxwei:
```suggestion
        XCTAssertEqual(op0.description, "Identity_0[](0.0):1")
```

Usually, instead of using string interpolation for things that do not need interpolation `"\(x)"` we'd prefer calling the `x.description` property directly or using `String(reflecting: x)` when the type does not conform to `CustomStringConvertible`.

Sources/TensorFlow/LazyTensor/LazyTensorOperation.swift
```diff
@@ -525,31 +525,46 @@ extension LazyTensorOperation: TFTensorOperation {
...
-        case .BoolArray(let values): return arrayAsString("", values)
-        case .IntArray(let values): return arrayAsString("Int", values)
-        case .FloatArray(let values): return arrayAsString("Float", values)
-        case .DoubleArray(let values): return arrayAsString("Double", values)
-        case .StringArray(let values): return arrayAsString("String", values)
-        case .ConstTensor(let v): return "Const(\(v))"
-        case .TensorDataTypeValue(let v): return dataTypeAsString(v)
+extension TFETensorHandle: CustomStringConvertible {
+    public var description: String {
+        let dtype = TFE_TensorHandleDataType(self._cTensorHandle)
```
pschuh: I think this switch-on-type-and make Tensor / TensorHandle is already somewhere else. Hopefully those would be consolidated to exist only once.

Sources/TensorFlow/LazyTensor/LazyTensorOperation.swift
```diff
@@ -525,31 +525,46 @@ extension LazyTensorOperation: TFTensorOperation {
...
+        switch dtype {
+        case TF_FLOAT:
+            return "\(Tensor(handle: TensorHandle<Float>(handle: self)))"
```
rxwei: Similarly, it is better for this to call `Tensor(handle: TensorHandle<Float>(handle: self)).description` instead of using a trivial interpolation.

Sources/TensorFlow/LazyTensor/LazyTensorOperation.swift
```diff
@@ -573,17 +588,46 @@ extension LazyTensorOperation.Attribute: CustomStringConvertible {
...
+
+    private func arrayAsString<T>(_ desc: String, _ values: [T]) -> String {
+        let arrayDesc = (values.map { "\($0)" }).joined(separator: ", ")
+        return "\(desc)[\(arrayDesc)]"
+    }
+
+    private func dataTypeAsString(_ dataType: TensorDataType) -> String {
+        return TFETensorHandle.tfDataTypeAsString(dataType._cDataType)
+    }
+
```
rxwei: Remove extra line.

Sources/TensorFlow/LazyTensor/LazyTensorOperation.swift
```diff
@@ -525,31 +525,46 @@ extension LazyTensorOperation: TFTensorOperation {
...
-        case .DoubleValue(let v): return "Double(\(v))"
-        case .StringValue(let v): return "\"\(v)\""
-        case .BoolArray(let values): return arrayAsString("", values)
-        case .IntArray(let values): return arrayAsString("Int", values)
-        case .FloatArray(let values): return arrayAsString("Float", values)
-        case .DoubleArray(let values): return arrayAsString("Double", values)
-        case .StringArray(let values): return arrayAsString("String", values)
-        case .ConstTensor(let v): return "Const(\(v))"
-        case .TensorDataTypeValue(let v): return dataTypeAsString(v)
+extension TFETensorHandle: CustomStringConvertible {
```
rxwei: I'm not sure if a handle should be printed like a value. Perhaps some extra metadata (e.g. handle identifier) should be printed solely or along with the value.

bgogul:
Addressed the review comments.

Sources/TensorFlow/LazyTensor/LazyTensorOperation.swift
```diff
@@ -525,31 +525,46 @@ extension LazyTensorOperation: TFTensorOperation {
...
-        case .StringValue(let v): return "\"\(v)\""
-        case .BoolArray(let values): return arrayAsString("", values)
-        case .IntArray(let values): return arrayAsString("Int", values)
-        case .FloatArray(let values): return arrayAsString("Float", values)
-        case .DoubleArray(let values): return arrayAsString("Double", values)
-        case .StringArray(let values): return arrayAsString("String", values)
-        case .ConstTensor(let v): return "Const(\(v))"
-        case .TensorDataTypeValue(let v): return dataTypeAsString(v)
+extension TFETensorHandle {
+    func valueAsString() -> String {
```
```suggestion
    var valueDescription: String {
```

PR #197, opened 2019-06-11
Glorot Normal Initializer 

Shashi456:
I overloaded the function with `fromStandardNormal`, afaik swift function can be overloaded this way. Let me know if i'm wrong. 

Also @rxwei could you tell me why we are naming these glorot and not Xavier? Xavier initialization is a much more popular usage of this initializer. 

Can we atleast mention at the bottom of docs that glorot is also called as xavier?


Shashi456:
@rxwei mind putting a build tag on this till then?

Sources/TensorFlow/Initializers.swift
```diff
@@ -499,3 +499,54 @@ public extension Tensor where Scalar: BinaryFloatingPoint,
         self = Tensor.glorot(fromStandardUniform: uniform, shape: shape)
     }
 }
+
+fileprivate extension Tensor where Scalar: BinaryFloatingPoint {
+    private static func glorot(
+        fromStandardNormal normalUniform: __shared Tensor<Scalar>,
```
eaplatanios: I don't think this argument tag and name represent what this argument is. Maybe something like this is better:
```swift
fromStandardNormalScaledBy varianceScalingFactor: __shared Tensor<Scalar>,
```
Even though it's long I believe it better represents what the argument is.

Shashi456: done

Shashi456:
Requesting Review on this PR. Once this goes, I'll add kaiming normal and kaiming uniform.

Sources/TensorFlow/Initializers.swift
```diff
@@ -499,3 +499,54 @@ public extension Tensor where Scalar: BinaryFloatingPoint,
         self = Tensor.glorot(fromStandardUniform: uniform, shape: shape)
     }
 }
+
+fileprivate extension Tensor where Scalar: BinaryFloatingPoint {
```
rxwei: It seems that this function is exactly the same as the one defined on line 452. Why do you need to defined this? Also, the `varianceScalingFactor` parameter is completely unused in the function body.

Shashi456: @rxwei The function on line 452 does it by uniform distribution, this function does it by normal distribution, I read parameters could be overloaded in swift as long as the names are swift, Am i wrong?
 And `varianceScalingFactor` is something proposed in the above review, Just for better variable naming.

rxwei: > The function on line 452 does it by uniform distribution, this function does it by normal distribution,

The function is almost exactly the same as the one defined on line 452, character by character, except for the last line. Maybe there is a way to unify the identical logic to a separate method.

> I read parameters could be overloaded in swift as long as the names are swift, Am i wrong?

Yep, things can be overloaded and I'm not worrying about this.

> And `varianceScalingFactor` is something proposed in the above review, Just for better variable naming.

However, the point I brought up was that this variable is unused _at all_ in the body of this function, and compilation failed.

Shashi456: @rxwei how should I move forward on this now? Primarily I'll fix the variable error.

rxwei: Please fix the variable error for now.

Shashi456: @rxwei fixed the variable errors and other minor things, This passes locally now.

PR #198, opened 2019-06-11
Add a 'clipped(min:max:)' method.

Shashi456:
Refer discussion [here](https://github.com/tensorflow/swift-apis/pull/187#issuecomment-500205653).

I need some help here though, I'm sure this needs a vjp as well but what would a vjp for this look like? 

Currently getting this error from the build : 
```
      error: cannot convert value of type 'Tensor<Scalar>' to expected argument type 'Tensor<_>'

      return Raw.clipByValue(t: self, clipValueMin: minimum, clipValueMax: maximum)
                                ^~~~
                                     as! Tensor<_>
```

FInally, min and max have a default value of -infinity and +infinity, do we have those identifiers in swift?

### DIFF
--- a/Sources/TensorFlow/Operators/Basic.swift
+++ b/Sources/TensorFlow/Operators/Basic.swift
@@ -213,6 +213,13 @@ public extension Tensor {
     func squeezingShape(at axes: [Int]) -> Tensor {
         return Raw.squeeze(self, squeezeDims: axes.map(Int32.init))
     }
+
+    /// Returns a `Tensor` by clipping tensor values to a specified min and max.
+    @inlinable
+    @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
+    func clipped(min minimum: Tensor<Int32>, max maximum: Tensor<Int32>) -> Tensor{
+      return Raw.clipByValye(self, clipValueMin: minimum, clipValueMax: maximum)
+    }
 }
 
 internal extension Tensor where Scalar: TensorFlowFloatingPoint {
--- a/Sources/TensorFlow/Operators/Basic.swift
+++ b/Sources/TensorFlow/Operators/Basic.swift
@@ -217,7 +217,7 @@ public extension Tensor {
     /// Returns a `Tensor` by clipping tensor values to a specified min and max.
     @inlinable
     @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
-    func clipped(min minimum: Tensor<Int32>, max maximum: Tensor<Int32>) -> Tensor{
+    func clipped<T>(min minimum: Tensor<T>, max maximum: Tensor<T>) -> Tensor{
       return Raw.clipByValue(t: self, clipValueMin: minimum, clipValueMax: maximum)
     }
 }
--- a/Sources/TensorFlow/Operators/Basic.swift
+++ b/Sources/TensorFlow/Operators/Basic.swift
@@ -576,11 +576,11 @@ public extension Tensor {
 }
 
 public extension Tensor where Scalar: Numeric {
-    /// Returns a `Tensor` by clipping tensor values to a specified min and max.
+    /// Returns a tensor by clipping scalars to a specified minimum and maximum.
+    // FIXME: Define a derivative function.
     @inlinable
-    @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
-    func clipped(min minimum: Tensor, max maximum: Tensor) -> Tensor{
-      return Raw.clipByValue(t: self, clipValueMin: minimum, clipValueMax: maximum)
+    func clipped(min: Tensor, max: Tensor) -> Tensor {
+      return Raw.clipByValue(t: self, clipValueMin: min, clipValueMax: max)
     }
 }
 
--- a/Sources/TensorFlow/Operators/Basic.swift
+++ b/Sources/TensorFlow/Operators/Basic.swift
@@ -218,7 +218,7 @@ public extension Tensor {
     @inlinable
     @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
     func clipped(min minimum: Tensor<Int32>, max maximum: Tensor<Int32>) -> Tensor{
-      return Raw.clipByValue(self, clipValueMin: minimum, clipValueMax: maximum)
+      return Raw.clipByValue(t: self, clipValueMin: minimum, clipValueMax: maximum)
     }
 }
 
--- a/Sources/TensorFlow/Operators/Basic.swift
+++ b/Sources/TensorFlow/Operators/Basic.swift
@@ -218,7 +218,7 @@ public extension Tensor {
     @inlinable
     @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
     func clipped(min minimum: Tensor<Int32>, max maximum: Tensor<Int32>) -> Tensor{
-      return Raw.clipByValye(self, clipValueMin: minimum, clipValueMax: maximum)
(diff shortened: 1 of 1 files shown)


Shashi456:
cc: @eaplatanios

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -213,6 +213,13 @@ public extension Tensor {
     func squeezingShape(at axes: [Int]) -> Tensor {
         return Raw.squeeze(self, squeezeDims: axes.map(Int32.init))
     }
+
+    /// Returns a `Tensor` by clipping tensor values to a specified min and max.
+    @inlinable
+    @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
+    func clipped<T>(min minimum: Tensor<T>, max maximum: Tensor<T>) -> Tensor{
```
rxwei: This won't work because you defined `clipped(min:max:)` in a general extension where `Scalar` is not constrained as it should be. `Raw.clipByValue(t:clipValueMin:clipValueMax:)` is defined as the following:
```swift
public static func clipByValue<T: Numeric & TensorFlowScalar>(
    t: Tensor<T>,
    clipValueMin: Tensor<T>,
    clipValueMax: Tensor<T>
) -> Tensor<T>
```
Namely, `T` is constrained to `Numeric` in addition to `TensorFlowScalar`.  Therefore, you should define method `clipped(min:max:)` in an extension that constrains `Scalar` to `Numeric`.  And the `clipped(min:max:)` method itself should not have any generic parameter (i.e. no `<T>`) because its genericity comes from the extension.

```swift
public extension Tensor where Scalar: Numeric {
    func clipped(min minimum: Tensor, max maximum: Tensor) -> Tensor {
        ....
    }
}
```

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -610,6 +611,14 @@ public extension Tensor {
 }
 
 public extension Tensor where Scalar: Numeric {
+
+    /// Returns a `Tensor` by clipping tensor values to a specified min and max.
+    @inlinable
+    @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
+    func clipped(min minimum: Tensor, max maximum: Tensor) -> Tensor{
```
eaplatanios:
```suggestion
    func clipped(min minimum: Tensor, max maximum: Tensor) -> Tensor {
```

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -610,6 +611,14 @@ public extension Tensor {
 }
 
 public extension Tensor where Scalar: Numeric {
+
```
eaplatanios:
```suggestion
```

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -610,6 +611,13 @@ public extension Tensor {
 }
 
 public extension Tensor where Scalar: Numeric {
+    /// Returns a `Tensor` by clipping tensor values to a specified min and max.
+    @inlinable
+    @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
+    func clipped(min minimum: Tensor, max maximum: Tensor) -> Tensor {
```
rxwei: Line 580 indicates that this block of code should be about broadcasting. However, `clipped(min:max:)` is not about broadcasting. It should be moved to an extension in an appropriate place.

Shashi456:
Sorry for the trivial errors.

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -213,6 +213,7 @@ public extension Tensor {
     func squeezingShape(at axes: [Int]) -> Tensor {
         return Raw.squeeze(self, squeezeDims: axes.map(Int32.init))
     }
+
```
saeta: Could you revert this additional blank line addition?

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -575,6 +576,15 @@ public extension Tensor {
     }
 }
 
+public extension Tensor where Scalar: Numeric {
+  /// Returns a `Tensor` by clipping tensor values to a specified min and max.
```
eaplatanios: 4-space indentation.

Sources/TensorFlow/Operators/Basic.swift
Shashi456: done

Sources/TensorFlow/Operators/Basic.swift
Shashi456: done

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -575,6 +575,15 @@ public extension Tensor {
     }
 }
 
+public extension Tensor where Scalar: Numeric {
+    /// Returns a `Tensor` by clipping tensor values to a specified min and max.
+    @inlinable
+    @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
```
rxwei: This will not compile because `Raw` functions are not differentiable and you haven't defined a VJP for this yet.  Please comment out the `@differentiable` attribute and add a FIXME.
```suggestion
    // FIXME: Define a derivative function.
    // @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
```

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -575,6 +575,15 @@ public extension Tensor {
     }
 }
 
+public extension Tensor where Scalar: Numeric {
+    /// Returns a `Tensor` by clipping tensor values to a specified min and max.
```
```suggestion
    /// Returns a tensor by clipping scalars to a specified minimum and maximum.
```

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -575,6 +575,15 @@ public extension Tensor {
...
+    @inlinable
+    @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
+    func clipped(min minimum: Tensor, max maximum: Tensor) -> Tensor{
```
rxwei: 1. Just using `min` and `max` as variable names is fine.
2. Add a space before `{`.
```suggestion
    func clipped(min: Tensor, max: Tensor) -> Tensor {
```

Sources/TensorFlow/Operators/Basic.swift
Shashi456: @rxwei I'll do this, but how does one define a vjp for clipped function?

rxwei: Please file an issue for this and move the discussion there. Thanks!

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -575,6 +575,16 @@ public extension Tensor {
     }
 }
 
+public extension Tensor where Scalar: Numeric {
+    /// Returns a tensor by clipping scalars to a specified minimum and maximum.
+    // FIXME: Define a derivative function.
+    // @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
+    @inlinable
+    func clipped(min: Tensor, max: Tensor) -> Tensor {
+      return Raw.clipByValue(t: self, clipValueMin: min, clipValueMax: max)
```
rxwei: Use 4-space indentation.

```suggestion
        Raw.clipByValue(t: self, clipValueMin: min, clipValueMax: max)
```

rxwei:
If you pull from master, failures regarding `move(along:)` should be fixed.

Shashi456:
@rxwei this can be merged now.

PR #200, opened 2019-06-11
Conform `Tensor` to `ElementaryFunctions`.

rxwei:


rxwei:
This is also semi-blocked on https://github.com/apple/swift/pull/25316.

Sources/TensorFlow/Initializers.swift
```diff
@@ -459,7 +459,12 @@ fileprivate extension Tensor where Scalar: BinaryFloatingPoint {
         let fanIn = shape[shape.count - 2] * receptiveField
         let fanOut = shape[shape.count - 1] * receptiveField
         let minusOneToOne = 2 * randomUniform - 1
-        return sqrt(Scalar(6) / Scalar(fanIn + fanOut)) * minusOneToOne
+#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
+        let _sqrt = Darwin.sqrt as (Scalar) -> Scalar
+#else
+        let _sqrt = Glibc.sqrt as (Scalar) -> Scalar
+#endif
+        return _sqrt(Scalar(6) / Scalar(fanIn + fanOut)) * minusOneToOne
```
dan-zheng: This is necessary to avoid an unfortunate type-checking ambiguity:
```
/Users/danielzheng/swift-apis/Sources/TensorFlow/Initializers.swift:462:16: error: referencing static method 'sqrt' on 'Tensor' requires that 'Scalar' conform to 'TensorFlowFloatingPoint'
        return sqrt(Scalar(6) / Scalar(fanIn + fanOut)) * minusOneToOne
               ^
/Users/danielzheng/swift-apis/Sources/TensorFlow/Operators/Math.swift:26:1: note: where 'Scalar' = 'Scalar'
extension Tensor: ElementaryFunctions where Scalar: TensorFlowFloatingPoint {
^
```

dan-zheng:
Ready for review. cc @rxwei

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -15,18 +15,148 @@
 infix operator .>: ComparisonPrecedence
 infix operator .==: ComparisonPrecedence
 
-// `pow` is defined in Darwin/Glibc on `Float` and `Double`, but there doesn't exist a generic
-// version for `FloatingPoint`.
-// This is a manual definition.
-@inlinable
-func pow<T: BinaryFloatingPoint>(_ x: T, _ y: T) -> T {
```
dan-zheng: This function is no longer necessary.

rxwei:
Will look soon!

Sources/TensorFlow/Initializers.swift
rxwei: Instead of using the global libc function, you should use the static ElementaryFunction method instead. That shouldn’t have any ambiguity. cc @stephentyrone

dan-zheng: Done in https://github.com/tensorflow/swift-apis/pull/200/commits/0c1ff4142130ac7a226e882a7f39ee5cefc13bdb. Changed some constraints to `Scalar: TensorFlowFloatingPoint`.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -15,18 +15,148 @@
...
+    public static func log10(_ x: Self) -> Self {
+        TensorFlow.log10(x)
+    }
+
+    /// `log(1 + x)` evaluated so as to preserve accuracy close to zero.
+    public static func log1p(_ x: Self) -> Self {
+        TensorFlow.log1p(x)
+    }
+
+    /// `x**y` interpreted as `exp(y * log(x))`
```
rxwei: I wonder where this comment is from. It doesn't look like a Swift/Apple-official comment because it does not have a period.

dan-zheng: Sorry, that comment (along with all other comments) was copied from https://github.com/apple/swift-evolution/blob/master/proposals/0246-mathable.md.

Rewrote this comment using what is defined in `stdlib/public/core/MathFunctions.swift.gyb`:
```swift
/// `exp(y log(x))` computed without loss of intermediate precision.
```

Functions like `Float.exp` currently have no comment, but perhaps something will be added before SE-0246 is merged.

PR #203, opened 2019-06-11
Add a `LazyTensor.forEachOperation` method.

bgogul:
The function is useful to iterate over all the `LazyTensorOperation` instances that have been created.

### DIFF
--- a/Sources/TensorFlow/LazyTensor/LazyTensorOperation.swift
+++ b/Sources/TensorFlow/LazyTensor/LazyTensorOperation.swift
@@ -105,6 +105,10 @@ class LazyTensor: _AnyTensorHandle {
         }
     }
 
+    static func onAllOperations(_ perform: (LazyTensorOperation) -> ()) {
+        for (_, counts) in operationRefCounts { perform(counts.op) }
+    }
+
     @usableFromInline
     static var _materializationCallback: (String) -> () = { _ in }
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/LazyTensor/LazyTensorOperation.swift
```diff
@@ -105,6 +105,10 @@ class LazyTensor: _AnyTensorHandle {
         }
     }
 
+    static func onAllOperations(_ perform: (LazyTensorOperation) -> ()) {
```
```suggestion
    static func forEachOperation(_ perform: (LazyTensorOperation) -> Void) {
```

1. It is more common for such methods to be called `forEachXXX`, based on the precedent [`forEach(_:)`](https://developer.apple.com/documentation/swift/array/1689783-foreach) in the Swift standard library.
2. It is more common (throughout the standard library) to write `-> Void` for void returns instead of `-> ()`.
3. Would you like to make this method `throws`-`rethrows`? This would make it easier to put things that throw inside the body closure and work nicely with XCTest.

dan-zheng: Please also rename `onLiveOperations` to `forEachLiveOperation`.

bgogul: Thanks! Done.

PR #206, opened 2019-06-11
Print `LazyTensorOperation` in the form "%id = op[...](...)".

bgogul:


Sources/TensorFlow/LazyTensor/LazyTensorOperation.swift
```diff
@@ -147,6 +147,34 @@ class LazyTensorOperation: TensorOperation {
         }
     }
 
+    func outputName(_ index: Int) -> String {
```
```suggestion
    func outputName(at index: Int) -> String {
```

bgogul: Done.

Sources/TensorFlow/LazyTensor/LazyTensorOperation.swift
```diff
@@ -147,6 +147,36 @@ class LazyTensorOperation: TensorOperation {
...
+        }
+        return ssaName
+    }
+
+    var outputName: String {
+        switch outputCount {
+        case 0: return ""
+        case 1: return "\(outputName(at: 0))"
+        default:
+            do {
```
rxwei: Why use a `do` block here?

Sources/TensorFlow/LazyTensor/LazyTensorOperation.swift
```diff
@@ -147,6 +147,36 @@ class LazyTensorOperation: TensorOperation {
         }
     }
 
+    func outputName(at index: Int) -> String {
+        precondition(index < outputCount,
+            "Output index out of bounds when getting outputName.")
+        var ssaName = ""
+        if let id = self.id {
```
```suggestion
        ssaName = "%\(id ?? ObjectIdentifier(self))"
```

Sources/TensorFlow/LazyTensor/LazyTensorOperation.swift
```diff
@@ -147,6 +147,36 @@ class LazyTensorOperation: TensorOperation {
...
+        return ssaName
+    }
+
+    var outputName: String {
+        switch outputCount {
+        case 0: return ""
+        case 1: return "\(outputName(at: 0))"
+        default:
+            do {
+                let outputNames = (0..<outputCount).map {
```
rxwei: Use `.lazy` so it won't create an array first.

```suggestion
                let outputNames = (0..<outputCount).lazy.map {
```

Sources/TensorFlow/LazyTensor/LazyTensorOperation.swift
```diff
@@ -630,8 +660,8 @@ extension LazyTensor: CustomStringConvertible {
                 : "\(h.valueDescription)"
         case LazyTensor.Handle.symbolic(let op, let index, let isLive):
             return isLive
```
```suggestion
            return op.outputName(at: index) + isLive ? "*" : ""
```

Sources/TensorFlow/LazyTensor/LazyTensorOperation.swift
bgogul: ah, habit from C++ to put `case` blocks within {}. Removed it.

Sources/TensorFlow/LazyTensor/LazyTensorOperation.swift
```diff
@@ -147,6 +147,36 @@ class LazyTensorOperation: TensorOperation {
...
+        if outputCount > 1 {
+            ssaName += ".\(index)"
+        }
+        return ssaName
+    }
+
+    var outputName: String {
+        switch outputCount {
+        case 0: return ""
+        case 1: return "\(outputName(at: 0))"
```
```suggestion
        case 1: return outputName(at: 0)
```

bgogul: Ah, done.

Sources/TensorFlow/LazyTensor/LazyTensorOperation.swift
bgogul: Nice.. Thanks! Done.

Sources/TensorFlow/LazyTensor/LazyTensorOperation.swift
bgogul: done

PR #211, opened 2019-06-12
Replace @_fixed_layout with @frozen.

rxwei:
See [SE-0260: Library Evolution for Stable ABIs](https://github.com/apple/swift-evolution/blob/master/proposals/0260-library-evolution.md).


rxwei:
CI will break due to toolchain changes. Everything passes locally.

PR #213, opened 2019-06-12
Clean up VectorProtocol conformance and the scalar-tensor multiplication operator.

rxwei:
Re-order things so that they'll make more sense.

### DIFF
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -33,26 +33,6 @@ func pow<T: BinaryFloatingPoint>(_ x: T, _ y: T) -> T {
 
 extension Tensor: VectorProtocol where Scalar: Numeric {
     public typealias VectorSpaceScalar = Scalar
-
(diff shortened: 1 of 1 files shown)


Sources/TensorFlow/Operators/Math.swift
```diff
@@ -33,26 +33,6 @@ func pow<T: BinaryFloatingPoint>(_ x: T, _ y: T) -> T {
...
-    @inlinable
-    @differentiable(vjp: _vjpMultiply(lhs:rhs:) where Scalar: TensorFlowFloatingPoint)
-    public static func * (lhs: Scalar, rhs: Tensor) -> Tensor {
-        return Tensor(lhs) * rhs
-    }
-}
-
-internal extension Tensor where Scalar: TensorFlowFloatingPoint {
-    @inlinable
-    static func _vjpMultiply(lhs: Tensor, rhs: Tensor) -> (Tensor, (Tensor) -> (Tensor, Tensor)) {
```
rxwei: This shouldn't have been placed here, since it has nothing to do with the scalar-tensor multiplication operator.

PR #214, opened 2019-06-12
Add top-level versions of `ElementaryFunctions` operators for `Tensor`.

dan-zheng:
Add `Tensor` operators:
- `acos`, `asin`, `atan`,  `acosh`, `asinh`, `atanh`,
  `log2`, `log10`, `exp2`, `exp10`, `pow(Tensor, Int)`, `root(Tensor, Int)`

Todos:
- Add operator tests.
- Mark all operators as `@differentiable`. Blocked by some bugs.
- Add derivative tests.

### DIFF
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -406,6 +406,20 @@ internal func _vjpLog<T: TensorFlowFloatingPoint>(
     (log(x), { v in v / x })
 }
 
+/// Computes the base-two logarithm of the specified tensor element-wise.
+@inlinable
+@differentiable
+public func log2<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
+    log(x) / T.log(2)
+}
+
+/// Computes the base-ten logarithm of the specified tensor element-wise.
+@inlinable
+@differentiable
+public func log10<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
+    log(x) / T.log(10)
+}
+
 /// Computes the logarithm of `1 + x` element-wise.
 @inlinable
 @differentiable(vjp: _vjpLog1p)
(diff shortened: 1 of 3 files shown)


Sources/TensorFlow/Operators/Math.swift
```diff
@@ -408,6 +408,20 @@ internal func _vjpLog<T: TensorFlowFloatingPoint>(
     return (log(x), { v in v / x })
 }
 
+/// Computes the base-two logarithm of the specified tensor element-wise.
+@inlinable
+@differentiable(where T: TensorFlowFloatingPoint)
+public func log2<T: FloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
+    log(x) / log(Tensor(2))
```
rxwei: Does this work?

```suggestion
    log(x) / log(2)
```

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -408,6 +408,20 @@ internal func _vjpLog<T: TensorFlowFloatingPoint>(
...
+}
+
+/// Computes the base-ten logarithm of the specified tensor element-wise.
+@inlinable
+@differentiable(where T: TensorFlowFloatingPoint)
+public func log10<T: FloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
+    log(x) / log(Tensor(10))
```
```suggestion
    log(x) / log(10)
```

Sources/TensorFlow/Operators/Math.swift
dan-zheng: It doesn't work. I believe it's because `FloatingPoint` does not conform to `ExpressibleByFloatLiteral`.
```
/Users/danielzheng/swift-apis/Sources/TensorFlow/Operators/Math.swift:415:12: error: binary operator '/' cannot be applied to operands of type 'Tensor<T>' and '_'
    log(x) / log(2)
    ~~~~~~ ^ ~~~~~~
```

Sources/TensorFlow/Operators/Math.swift
dan-zheng: Doesn't work, see [this comment](https://github.com/tensorflow/swift-apis/pull/214#discussion_r292749224).

Sources/TensorFlow/Operators/Math.swift
rxwei: `FloatingPoint` types do have an initializer that takes a literal. You can do `log(Scalar(2))`. 

That said, I think we should change all of these operators to require `Scalar: TensorFlowFloatingPoint` for consistency with other APIs.

dan-zheng: `log(T(2))` doesn't work because `FloatingPoint` (nor `TensorFlowFloatingPoint`) refine `ElementaryFunctions`.

I'll keep the code as is: `log(Tensor(2))`.

rxwei:
Is this ready?

dan-zheng:
> Is this ready?

Yes. Added elementary function tests in https://github.com/tensorflow/swift-apis/pull/214/commits/ddfe8c7036d0f79c442a7a2ee06114d8305556a0.

rxwei:
Tests look great!

PR #216, opened 2019-06-12
Gardening.

dan-zheng:
- Change all `@differentiable` operators constrained on `T: FloatingPoint`
  to be constrained on `T: TensorFlowFloatingPoint` for consistency.
- Omit return for single expression functions in Sources/TensorFlow/Operators/Math.swift for consistency.

### DIFF
--- a/Sources/TensorFlow/Operators/NN.swift
+++ b/Sources/TensorFlow/Operators/NN.swift
@@ -76,36 +76,6 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
     }
 }
 
-public extension Tensor where Scalar: BinaryFloatingPoint {
-    /// Computes the batch normalized tensor along the specified axis.
-    ///
-    /// Specifically, returns `(self - mu)/(var + epsilon) * gamma + beta` where
-    /// `mu` and `var` are respectively the mean and variance of `self` along
-    /// `axis`.
-    ///
-    /// - Parameters:
-    ///     - axis: The batch dimension.
-    ///     - offset: The offset, also known as beta.
-    ///     - scale: The scale, also known as gamma.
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/Operators/NN.swift
```diff
@@ -76,36 +76,6 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
...
-    /// - Parameters:
-    ///     - axis: The batch dimension.
-    ///     - offset: The offset, also known as beta.
-    ///     - scale: The scale, also known as gamma.
-    ///     - epsilon: A small value added to the denominator for numerical stability.
-    @inlinable
-    @differentiable(
-        wrt: (self, offset, scale),
-        vjp: _vjpBatchNormalized where Scalar: TensorFlowFloatingPoint)
-    func batchNormalized(
```
dan-zheng: Removing this is necessary because `Scalar` is constrained by `BinaryFloatingPoint`, not `TensorFlowFloatingPoint`.
```
/Users/danielzheng/swift-apis/Sources/TensorFlow/Operators/NN.swift:104:19: error: global function 'rsqrt' requires that 'Scalar' conform to 'TensorFlowFloatingPoint'
        let inv = rsqrt(variance + epsilon) * scale
                  ^
/Users/danielzheng/swift-apis/Sources/TensorFlow/Operators/Math.swift:543:13: note: where 'T' = 'Scalar'
public func rsqrt<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
            ^
```

This block is duplicated with `func batchNormalized` above for some reason. I took a look at `git blame` but couldn't figure out why.

rxwei: I'm very surprised this was duplicated.

PR #221, opened 2019-06-12
Adding poisson loss

Shashi456:
#127 .
build and test passes locally.


Sources/TensorFlow/Loss.swift
```diff
@@ -86,6 +86,18 @@ public func categoricalHingeLoss<Scalar: TensorFlowFloatingPoint>(
     return max(Tensor(0), negative - positive + Tensor(1))
 }
 
+/// Returns the poisson loss between predictions and expectations.
```
```suggestion
/// Returns the Poisson loss between predictions and expectations.
```

PR #222, opened 2019-06-12
Add support for calling functions directly by name.

pschuh:
This should allow using the TFE_Op code directly when dispatching traced code. This will make tracing and lazy-tensor more symmetric.

### DIFF
--- a/Sources/TensorFlow/Bindings/TFTensorOperation.swift
+++ b/Sources/TensorFlow/Bindings/TFTensorOperation.swift
@@ -12,6 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+/// Opaque reference to a function that has been made callable by loading it
+/// into the runtime.
+public struct _TensorFunctionPointer {
+    var name: String
+}
+
 // A protocol for a tensor operation.
 public protocol TensorOperation {
     // We use functions instead of fields to give freedom in the representation for the conforming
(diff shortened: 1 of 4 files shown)


Sources/TensorFlow/Bindings/TFTensorOperation.swift
```diff
@@ -12,6 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+/// Opaque reference to a function that has been made callable by loading it
+/// into the runtime.
+public struct TensorFunctionPointer {
+  var name: String
```
rxwei: 4-space indentation.

Sources/TensorFlow/Bindings/TFTensorOperation.swift
```diff
@@ -12,6 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+/// Opaque reference to a function that has been made callable by loading it
+/// into the runtime.
+public struct TensorFunctionPointer {
```
rxwei: Why does it need to be public?

pschuh: Alas, the protocol it is used in is public.

rxwei: I see. In that case, the type name should start with an underscore because it's an internal implementation detail.

Sources/TensorFlow/Core/LazyTensorOperation.swift
```diff
@@ -152,6 +152,7 @@ class LazyTensorOperation: TensorOperation {
         case StringArray([String])
         case ConstTensor(TFETensorHandle)
         case TensorDataTypeValue(TensorDataType)
+        case TensorFunctionPointer(_TensorFunctionPointer)
```
bgogul: Can you add a test case to the LazyTensorOperationTests ?

PR #226, opened 2019-06-13
Adding kullback Leibler Divergence

Shashi456:
build and test passes locally.

Reference #127

### DIFF
--- a/Sources/TensorFlow/Loss.swift
+++ b/Sources/TensorFlow/Loss.swift
@@ -98,6 +98,18 @@ public func poissonLoss<Scalar: TensorFlowFloatingPoint>(
     return (predicted - expected * log(predicted)).mean()
 }
 
+/// Returns the Kullback-Leibler divergence between predictions and expectations.
+///
+/// - Parameters:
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/Loss.swift
```diff
@@ -98,6 +98,18 @@ public func poissonLoss<Scalar: TensorFlowFloatingPoint>(
     return (predicted - expected * log(predicted)).mean()
 }
 
+/// Returns the Kullback-Leibler divergence between predictions and expectations.
```
eaplatanios: You should probably say "between expectations and predictions" and also add the actual equation because KL divergence is not symmetric and so this may lead people to thinking the opposite of what's intended.

eaplatanios: @Shashi456 could you possibly make a new PR addressing this?

Shashi456: On it. I'll have it up in a few hours.

eaplatanios: Great, thanks!

PR #228, opened 2019-06-13
Add mean absolute percentage error

Shashi456:
Reference #127

### DIFF
--- a/Sources/TensorFlow/Loss.swift
+++ b/Sources/TensorFlow/Loss.swift
@@ -53,7 +53,7 @@ public func meanAbsoluteError<Scalar: TensorFlowFloatingPoint>(
     return abs(expected - predicted).mean()
 }
 
-/// Computes the mean absolute percentage error between predictions and expectations.
+/// Returns the mean absolute percentage error between predictions and expectations.
 ///
 /// - Parameters:
 ///   - predicted: Predicted outputs from a neural network.
--- a/Sources/TensorFlow/Loss.swift
+++ b/Sources/TensorFlow/Loss.swift
@@ -53,6 +53,19 @@ public func meanAbsoluteError<Scalar: TensorFlowFloatingPoint>(
     return abs(expected - predicted).mean()
 }
 
+/// Computes the mean absolute percentage error between predictions and expectations.
+///
+/// - Parameters:
+///   - predicted: Predicted outputs from a neural network.
+///   - expected: Expected values, i.e. targets, that correspond to the correct output.
+@differentiable(wrt: predicted)
+public func meanAbsolutePercentageError<Scalar: TensorFlowFloatingPoint>(
+    predicted: Tensor<Scalar>, expected: Tensor<Scalar>
+) -> Tensor<Scalar> {
+    let diff = abs((expected - predicted) / abs(expected))
+    return Tensor(100) * diff.mean()
+}
+
 /// Returns the hinge loss between predictions and expectations.
 ///
 /// - Parameters:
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/Loss.swift
```diff
@@ -53,6 +53,19 @@ public func meanAbsoluteError<Scalar: TensorFlowFloatingPoint>(
...
+/// Computes the mean absolute percentage error between predictions and expectations.
+///
+/// - Parameters:
+///   - predicted: Predicted outputs from a neural network.
+///   - expected: Expected values, i.e. targets, that correspond to the correct output.
+@differentiable(wrt: predicted)
+public func meanAbsolutePercentageError<Scalar: TensorFlowFloatingPoint>(
+    predicted: Tensor<Scalar>, expected: Tensor<Scalar>
+) -> Tensor<Scalar> {
+    let diff = abs((expected - predicted) / abs(expected))
```
jon-tow: Just wondering. Could we avoid the call to abs() in the denominator?

Shashi456: @jon-tow I'm afraid not if you don't take the absolute value and you have a negative sign in one of your values then your percentages might no longer be uniform.

jon-tow: If you remove the abs() in the denominator of line 65, then the code is computationally equivalent to the equation you've commented above.

The code currently stands as: 
     abs(num / abs(denom))
The abs(denom) seems redundant. How does this break some uniformity?

Shashi456: A variety of loss functions have to be clipped. I'm waiting for #198 to pass to update the loss functions, in that case an abs function is needed. I guess we might not need the abs in this PR. I'll remove it in a while.

jon-tow: Ah I see. Thanks for the quick response and explanation. 👍

Shashi456: Thanks for pointing it out.

Sources/TensorFlow/Loss.swift
```diff
@@ -53,6 +53,19 @@ public func meanAbsoluteError<Scalar: TensorFlowFloatingPoint>(
     return abs(expected - predicted).mean()
 }
 
+/// Computes the mean absolute percentage error between predictions and expectations.
```
```suggestion
/// Returns the mean absolute percentage error between predictions and expectations.
```

Tests/TensorFlowTests/LossTests.swift
```diff
@@ -49,6 +49,16 @@ final class LossTests: XCTestCase {
...
 
+    func testMeanAbsolutePercentageError() {
+        let predicted = Tensor<Float>([1, 2, 3, 4, 5, 6, 7, 8])
+        let expected = Tensor<Float>([0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8])
+
+        let loss = meanAbsolutePercentageError(predicted: predicted, expected: expected)
+        let expectedLoss: Float = 900.0
+        assertElementsEqual(expected: Tensor(expectedLoss), actual: loss)
+    }
+
```
rxwei: Remove extra line
```suggestion
```

Shashi456:
@rxwei this is ready to go.

Sources/TensorFlow/Loss.swift
```diff
@@ -53,6 +53,19 @@ public func meanAbsoluteError<Scalar: TensorFlowFloatingPoint>(
...
+///
+/// - Parameters:
+///   - predicted: Predicted outputs from a neural network.
+///   - expected: Expected values, i.e. targets, that correspond to the correct output.
+@differentiable(wrt: predicted)
+public func meanAbsolutePercentageError<Scalar: TensorFlowFloatingPoint>(
+    predicted: Tensor<Scalar>, expected: Tensor<Scalar>
+) -> Tensor<Scalar> {
+    let diff = abs((expected - predicted) / abs(expected))
+    return Tensor(100) * diff.mean()
```
```suggestion
    return 100 * diff.mean()
```

Shashi456:
@rxwei this is ready to go as well.

PR #229, opened 2019-06-13
Some Doc errors in Losses and testlosses

Shashi456:
Strangely squared hinge function doesnt have any documentation.


rxwei:
Thanks!

PR #230, opened 2019-06-13
Add negative log likelihood

Descartess:
Reference #127

### DIFF
--- a/Sources/TensorFlow/Loss.swift
+++ b/Sources/TensorFlow/Loss.swift
@@ -53,6 +53,17 @@ public func meanAbsoluteError<Scalar: TensorFlowFloatingPoint>(
     return abs(expected - predicted).mean()
 }
 
+/// Returns the negative logarithmic likelihood of the predictions.
+///
+/// - Parameters:
+///   - predicted: Predicted outputs from a neural network.
+@differentiable(wrt: predicted)
+public func negativeLogLikelihood<Scalar: TensorFlowFloatingPoint>(
+    predicted: Tensor<Scalar>
+    ) -> Tensor<Scalar> {
+    return -log(predicted).mean()
+}
+
 /// Returns the hinge loss between predictions and expectations.
 ///
 /// - Parameters:
--- a/Sources/TensorFlow/Loss.swift
+++ b/Sources/TensorFlow/Loss.swift
@@ -60,7 +60,7 @@ public func meanAbsoluteError<Scalar: TensorFlowFloatingPoint>(
 @differentiable(wrt: predicted)
 public func negativeLogLikelihood<Scalar: TensorFlowFloatingPoint>(
     predicted: Tensor<Scalar>
-    ) -> Tensor<Scalar> {
+) -> Tensor<Scalar> {
     return -log(predicted).mean()
 }
 
--- a/Tests/TensorFlowTests/LossTests.swift
+++ b/Tests/TensorFlowTests/LossTests.swift
@@ -67,6 +67,18 @@ final class LossTests: XCTestCase {
 
         assertElementsEqual(expected: expectedGradients, actual: gradients)
     }
+    
+    func testNegativeLogLikelihood() {
+        let predicted = Tensor<Float>(shape: [2, 4], scalars: [1, 2, 3, 4, 5, 6, 7, 8])
+        let expectedGradients = Tensor<Float>(
+            shape: [2, 4],
+            scalars: [-0.125, -0.0625, -0.0416666, -0.03125, -0.025, -0.0208333, -0.017857, -0.015625])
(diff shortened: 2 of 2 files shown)


Descartess:
I signed it

Tests/TensorFlowTests/LossTests.swift
```diff
@@ -67,6 +67,18 @@ final class LossTests: XCTestCase {
 
         assertElementsEqual(expected: expectedGradients, actual: gradients)
     }
+    
+    func testNegativeLogLikelihood() {
+        let predicted = Tensor<Float>(shape: [2, 4], scalars: [1, 2, 3, 4, 5, 6, 7, 8])
```
Shashi456: Can you also have an expected variable which just checks if the outputs are right? You can add the Assert equal to the same test case.

Tests/TensorFlowTests/LossTests.swift
```diff
@@ -190,6 +202,7 @@ final class LossTests: XCTestCase {
         ("testMeanSquaredErrorGrad", testMeanSquaredErrorGrad),
         ("testMeanSquaredLogarithmicError", testMeanSquaredLogarithmicError),
         ("testMeanAbsoluteError", testMeanAbsoluteError),
+        ("testNegativeLogLikelihood",testNegativeLogLikelihood),
```
Shashi456:
```suggestion
        ("testNegativeLogLikelihood", testNegativeLogLikelihood),
```

Tests/TensorFlowTests/LossTests.swift
```diff
@@ -67,6 +67,18 @@ final class LossTests: XCTestCase {
...
+    
+    func testNegativeLogLikelihood() {
+        let predicted = Tensor<Float>(shape: [2, 4], scalars: [1, 2, 3, 4, 5, 6, 7, 8])
+        let expectedGradients = Tensor<Float>(
+            shape: [2, 4],
+            scalars: [-0.125, -0.0625, -0.0416666, -0.03125, -0.025, -0.0208333, -0.017857, -0.015625])
+        let gradients = gradient(
+            at: predicted,
+            in: { negativeLogLikelihood(predicted: $0) })
+
```
Shashi456:
```suggestion
```
Remove the extra line.

Tests/TensorFlowTests/LossTests.swift
```diff
@@ -67,6 +67,18 @@ final class LossTests: XCTestCase {
...
+    func testNegativeLogLikelihood() {
+        let predicted = Tensor<Float>(shape: [2, 4], scalars: [1, 2, 3, 4, 5, 6, 7, 8])
+        let expectedGradients = Tensor<Float>(
+            shape: [2, 4],
+            scalars: [-0.125, -0.0625, -0.0416666, -0.03125, -0.025, -0.0208333, -0.017857, -0.015625])
+        let gradients = gradient(
+            at: predicted,
+            in: { negativeLogLikelihood(predicted: $0) })
+        
+        assertElementsEqual(expected: expectedGradients, actual: gradients)
```
Descartess: > Can you also have an expected variable which just checks if the outputs are right? You can add the Assert equal to the same test case.

@Shashi456 Correct me if am wrong, but I believe that is being done here.

Shashi456: You are actually checking the gradients here. You could name the loss function `testNegativeLogLikelihoodGrad`.

Shashi456:
@Descartess good job, LGTM. I'm not very sure about the naming though, a variety of frameworks name this NLLloss as well. perhaps @rxwei or @eaplatanios might be of help here.

Sources/TensorFlow/Loss.swift
```diff
@@ -53,6 +53,17 @@ public func meanAbsoluteError<Scalar: TensorFlowFloatingPoint>(
...
 }
 
+/// Returns the negative logarithmic likelihood of the predictions.
+///
+/// - Parameters:
+///   - predicted: Predicted outputs from a neural network.
+@differentiable(wrt: predicted)
+public func negativeLogLikelihood<Scalar: TensorFlowFloatingPoint>(
+    predicted: Tensor<Scalar>
+    ) -> Tensor<Scalar> {
```
eaplatanios:
```suggestion
) -> Tensor<Scalar> {
```

rxwei:
Thanks for putting this together @Descartess.  I agree with @eaplatanios and I think functions like this could complicate our API surface in that it's just a negative log and does not have the same semantics as prior art like [NLLLoss](https://pytorch.org/docs/stable/nn.html#nllloss). So I'd suggest dropping this function for now.

Descartess:
👍

PR #234, opened 2019-06-13
Change `Differentiable.moved(along:)` to `move(along:)`.

dan-zheng:
Friend PR: https://github.com/apple/swift/pull/25437

PR #240, opened 2019-06-14
Refactor cosine similarity and add cosine distance

Shashi456:
Refer to [this](https://github.com/tensorflow/swift-apis/pull/233#discussion_r293516299)

cc: @rxwei. 
Actually if we are removing the argument labels, i think we should do away with predicted and expected and have separate variable names.

### DIFF
--- a/Sources/TensorFlow/Loss.swift
+++ b/Sources/TensorFlow/Loss.swift
@@ -103,10 +103,6 @@ public func hingeLoss<Scalar: TensorFlowFloatingPoint>(
 }
 
 /// Returns the cosine similarity between predictions and expectations.
-///
-/// - Parameters:
-///   - predicted: Predicted outputs from a neural network.
-///   - expected: Expected values, i.e. targets, that correspond to the correct output.
 @differentiable(wrt: (predicted, expected))
 public func cosineSimilarity<Scalar: TensorFlowFloatingPoint>(
     predicted: Tensor<Scalar>, expected: Tensor<Scalar>
--- a/Sources/TensorFlow/Loss.swift
+++ b/Sources/TensorFlow/Loss.swift
@@ -102,19 +102,6 @@ public func hingeLoss<Scalar: TensorFlowFloatingPoint>(
     return max(Tensor(1) - expected * predicted, Tensor(0)).mean()
 }
 
-/// Returns the cosine similarity between predictions and expectations.
-///
-/// - Parameters:
-///   - predicted: Predicted outputs from a neural network.
-///   - expected: Expected values, i.e. targets, that correspond to the correct output.
-@differentiable(wrt: (predicted, expected))
-public func cosineSimilarity<Scalar: TensorFlowFloatingPoint>(
-    _ predicted: Tensor<Scalar>, _ expected: Tensor<Scalar>
-) -> Tensor<Scalar> {
-    return -(expected * predicted).sum() /
-        (sqrt(expected.squared().sum()) * sqrt(predicted.squared().sum()))
-}
-
 /// Returns the squared hinge loss between predictions and expectations.
 ///
 /// - Parameters:
--- a/Sources/TensorFlow/Loss.swift
+++ b/Sources/TensorFlow/Loss.swift
@@ -103,9 +103,13 @@ public func hingeLoss<Scalar: TensorFlowFloatingPoint>(
 }
 
 /// Returns the cosine similarity between predictions and expectations.
+///
+/// - Parameters:
+///   - predicted: Predicted outputs from a neural network.
+///   - expected: Expected values, i.e. targets, that correspond to the correct output.
 @differentiable(wrt: (predicted, expected))
 public func cosineSimilarity<Scalar: TensorFlowFloatingPoint>(
-    predicted: Tensor<Scalar>, expected: Tensor<Scalar>
+    _ predicted: Tensor<Scalar>, _ expected: Tensor<Scalar>
 ) -> Tensor<Scalar> {
     return -(expected * predicted).sum() /
         (sqrt(expected.squared().sum()) * sqrt(predicted.squared().sum()))
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -645,7 +645,7 @@ public func rsqrt<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
 }
 
 /// Returns the cosine similarity between x and y.
-@differentiable(wrt: (predicted, expected))
+@differentiable(wrt: (x, y))
 public func cosineSimilarity<Scalar: TensorFlowFloatingPoint>(
     _ x: Tensor<Scalar>, _ y: Tensor<Scalar>
 ) -> Tensor<Scalar> {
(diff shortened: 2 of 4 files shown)


Shashi456:
@rxwei Ah, I'll fix this immediately. Thanks for taking time and explaining this to me.

Sources/TensorFlow/Loss.swift
```diff
@@ -109,7 +109,7 @@ public func hingeLoss<Scalar: TensorFlowFloatingPoint>(
 ///   - expected: Expected values, i.e. targets, that correspond to the correct output.
 @differentiable(wrt: (predicted, expected))
 public func cosineSimilarity<Scalar: TensorFlowFloatingPoint>(
-    predicted: Tensor<Scalar>, expected: Tensor<Scalar>
+    _ predicted: Tensor<Scalar>, _ expected: Tensor<Scalar>
```
rxwei: Thanks for the change. With this, the call sites no longer need to write `predicted:` or `expected:`, which is good. However, Since this is an API for more general use than loss functions, we should not distinguish between `predicted` and `expected` at the implementation site or in API documentation comments. As such, could you please rename argument names `predicted` and `expected` to `x` and `y` respectively?

Shashi456: @rxwei How about I put this as a Math op and have a higher level function within losses calling the base function? would that count as redundancy of code? Because It's just sort of weird that we are putting a general API usage function in losses.swift

rxwei: Putting it in Math.swift is good. A wrapper in Loss.swift is unnecessary.

Shashi456:
@rxwei well the build passes locally. I cant build tests because i keep getting error since #234. Can you trigger a build?

Shashi456:
@rxwei Why do some kokoro labels get blocked?

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -644,6 +644,15 @@ public func rsqrt<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
     Raw.rsqrt(x)
 }
 
+/// Returns the cosine similarity between x and y.
```
rxwei: Please add backquotes around all parameters (or generally any program identifiers) in API documentation comments.
```suggestion
/// Returns the cosine similarity between `x` and `y`.
```

eaplatanios:
This is not computing cosine similarity but rather the negative cosine similarity. I would suggest you remove the minus sign and also introduce a dual function called `cosineDistance` which is computed as `1 - cosineSimilarity`.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -644,6 +644,23 @@ public func rsqrt<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
     Raw.rsqrt(x)
 }
 
+/// Returns the cosine similarity between `x` and `y`.
+@differentiable(wrt: (x, y))
```
```suggestion
@differentiable
```

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -644,6 +644,23 @@ public func rsqrt<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
...
+public func cosineSimilarity<Scalar: TensorFlowFloatingPoint>(
+    _ x: Tensor<Scalar>, _ y: Tensor<Scalar>
+) -> Tensor<Scalar> {
+    return (x * y).sum() /
+        (sqrt(x.squared().sum()) * sqrt(y.squared().sum()))
+}
+
+/// Returns the cosine distance between `x` and `y`.
+@differentiable(wrt: (x, y))
```
```suggestion
@differentiable
```

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -644,6 +644,23 @@ public func rsqrt<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
     Raw.rsqrt(x)
 }
 
+/// Returns the cosine similarity between `x` and `y`.
+@differentiable(wrt: (x, y))
+public func cosineSimilarity<Scalar: TensorFlowFloatingPoint>(
+    _ x: Tensor<Scalar>, _ y: Tensor<Scalar>
+) -> Tensor<Scalar> {
+    return (x * y).sum() /
```
rxwei: I think if you remove the `return` keyword the whole body can fit in one line.

```suggestion
    (x * y).sum() / (sqrt(x.squared().sum()) * sqrt(y.squared().sum()))
```

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -644,6 +644,23 @@ public func rsqrt<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
...
+        (sqrt(x.squared().sum()) * sqrt(y.squared().sum()))
+}
+
+/// Returns the cosine distance between `x` and `y`.
+@differentiable(wrt: (x, y))
+public func cosineDistance<Scalar: TensorFlowFloatingPoint>(
+    _ x: Tensor<Scalar>, _ y: Tensor<Scalar>
+) -> Tensor<Scalar> {
+    return 1 - cosineSimilarity(x, y)
```
```suggestion
    1 - cosineSimilarity(x, y)
```

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -644,6 +644,23 @@ public func rsqrt<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
...
+/// Returns the cosine similarity between `x` and `y`.
+@differentiable(wrt: (x, y))
+public func cosineSimilarity<Scalar: TensorFlowFloatingPoint>(
+    _ x: Tensor<Scalar>, _ y: Tensor<Scalar>
+) -> Tensor<Scalar> {
+    return (x * y).sum() /
+        (sqrt(x.squared().sum()) * sqrt(y.squared().sum()))
+}
+
+/// Returns the cosine distance between `x` and `y`.
```
```suggestion
/// Returns the cosine distance between `x` and `y`. Cosine distance is defined as
/// `1 - cosineSimilarity(x, y)`.
```

Sources/TensorFlow/Operators/Math.swift
rxwei: Hmm this doesn't seem to have been resolved yet.

Shashi456:
Also @rxwei any idea on passing #31, one of the last losses to be added.

Shashi456:
@rxwei could you retrigger a build?

Tests/TensorFlowTests/OperatorTests/MathTests.swift
```diff
@@ -65,6 +65,14 @@ final class MathOperatorTests: XCTestCase {
...
     }
 
+    func testCosineSimilarity() {
+        let x = Tensor<Float>([1, 2, 3, 4, 5, 6, 7, 8])
+        let y = Tensor<Float>([0.5, 1, 1.5, 2.0, 2.5, 3.0, 3.5, 4.0])
+        let z = cosineSimilarity(x, y)
+        let output: Float = 1.0
+        XCTAssertEqual(z, Tensor(output))
+    }
+
```
saeta: Could you please also add a test for cosine distance?

Shashi456: @saeta, actually cosine distance just returns 1 - cosineSimilarity, so in effect we would be testing the same function. To avoid redundancy I didn't add a test

rxwei: Even though the function is implemented as a trivial wrapper, it's good practice to assume the implementation can change in the future for whatever reasons and have tests be implementation-agnostic.

rxwei:
Merging this to keep things rolling. It is fine to add a test for `cosineDistance(_:_:)` later.

PR #242, opened 2019-06-14
Using .sum() rather than .mean() for Poisson loss

mirca:
For the Poisson loss, we don't actually need to compute the mean of the loss vector, the sum is enough to characterize the negative log likelihood.

### DIFF
--- a/Sources/TensorFlow/Loss.swift
+++ b/Sources/TensorFlow/Loss.swift
@@ -150,7 +150,7 @@ public func categoricalHingeLoss<Scalar: TensorFlowFloatingPoint>(
 public func poissonLoss<Scalar: TensorFlowFloatingPoint>(
     predicted: Tensor<Scalar>, expected: Tensor<Scalar>
 ) -> Tensor<Scalar> {
-    return (predicted - expected * log(predicted)).mean()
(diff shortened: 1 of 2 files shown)


rxwei:
Actually both the [Keras implementation](https://github.com/tensorflow/tensorflow/blob/93dd14dce2e8751bcaab0a0eb363d55eb0cc5813/tensorflow/python/keras/losses.py#L517) and [PyTorch default implementation](https://pytorch.org/docs/stable/nn.html#torch.nn.PoissonNLLLoss) use mean instead of sum. For now, we should keep the default behavior consistent with Keras. In the near future, we can consider adding a default argument `reduction: @differentiable (Tensor<Scalar>) -> Tensor<Scalar> = { $0.mean() }` to most loss functions to make them more flexible with respect to reduction functions.

mirca:
sounds good!

eaplatanios:
@rxwei currently some losses use mean and some use sum in the Swift API which can result to unexpected behavior for users. I suggest sticking to one of the two and being consistent at least. To that end, sum makes more sense in terms of the objective function that's used in empirical risk minimization which underlies most deep learning loss functions. I'm fine with either one as long as we're consistent, but of course adding support for a custom reduction function would be best. :)

Issue #244, opened 2019-06-15
Convolutional layers documentation inconsistency


Shashi456:
#138 refactored the convolutional and pooling layers, There's no self anymore in these layers after the refactoring. The documentation wasn't updated accordingly. `self` should be changed to `input`.

Refer to [this](https://github.com/tensorflow/swift-apis/pull/241#discussion_r293731261).

Shashi456:
Also, I think while we are at it we should solve #181 with this issue.

PR #241, opened 2019-06-14
Adding Depthwise Convolution 2D

Shashi456:
Build passes locally.

cc: @rxwei @dan-zheng - If you both are okay with naming and indentation, I'll go ahead and write the tests.

PR #246, opened 2019-06-15
Updating convolution docs

Shashi456:
Fix: #244

### DIFF
--- a/Sources/TensorFlow/Operators/NN.swift
+++ b/Sources/TensorFlow/Operators/NN.swift
@@ -387,7 +387,7 @@ func _vjpAvgPool3D<Scalar: TensorFlowFloatingPoint>(
 /// Computes a 2-D convolution with the specified input, filter, strides, and padding.
 ///
 /// - Parameters:
-///   - input: The input
+///   - input: The input.
 ///   - filter: The convolution filter.
 ///   - strides: The strides of the sliding filter for each dimension of the input.
 ///   - padding: The padding for the operation.
@@ -411,7 +411,7 @@ public func conv2D<Scalar: TensorFlowFloatingPoint>(
 /// Computes a 3-D convolution with the specified input, filter, strides, and padding.
 ///
 /// - Parameters:
-///   - input: The input
+///   - input: The input.
 ///   - filter: The convolution filter.
 ///   - strides: The strides of the sliding filter for each dimension of the input.
 ///   - padding: The padding for the operation.
@@ -436,7 +436,7 @@ public func conv3D<Scalar: TensorFlowFloatingPoint>(
 /// padding.
 ///
 /// - Parameters:
-///   - input: The input
+///   - input: The input.
 ///   - filterSize: The dimensions of the pooling kernel.
 ///   - strides: The strides of the sliding filter for each dimension of the input.
 ///   - padding: The padding for the operation.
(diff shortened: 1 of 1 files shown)


Sources/TensorFlow/Operators/NN.swift
```diff
@@ -384,15 +384,14 @@ func _vjpAvgPool3D<Scalar: TensorFlowFloatingPoint>(
...
-/// filter, strides, and padding.
+/// Computes a 2-D convolution with the specified input, filter, strides, and padding.
 ///
 /// - Parameters:
 ///   - x: The input.
 ///   - filter: The convolution filter.
 ///   - strides: The strides of the sliding filter for each dimension of the input.
 ///   - padding: The padding for the operation.
-/// - Precondition: `self` must have rank 4.
+/// - Precondition: `input` must have rank `4`.
```
rxwei: Oh, there is no parameter called `input`. It's currently called `x`. Could you please help change the first parameter name to `input` as well as the doc comment on line 390?

Sources/TensorFlow/Operators/NN.swift
```diff
@@ -409,15 +408,14 @@ public func conv2D<Scalar: TensorFlowFloatingPoint>(
...
-/// filter, strides, and padding.
+/// Computes a 3-D convolution with the specified input, filter, strides, and padding.
 ///
 /// - Parameters:
 ///   - x: The input.
 ///   - filter: The convolution filter.
 ///   - strides: The strides of the sliding filter for each dimension of the input.
 ///   - padding: The padding for the operation.
-/// - Precondition: `self` must have rank 5.
+/// - Precondition: `input` must have rank `5`.
```
rxwei: Same issue here.

Shashi456: @rxwei why did we change it to input in #241 then o.O? I'll change it to x there as well. cool.

rxwei: I actually suggested changing all the `x`'s to `input`.

Shashi456: @rxwei what about the comments in the parameters?

rxwei: Yes.

Shashi456: @rxwei one final thing, Should I change all the x's in nn.swift to input?

rxwei: For convolution and pooling layers, yes.

Sources/TensorFlow/Operators/NN.swift
```diff
@@ -384,50 +384,48 @@ func _vjpAvgPool3D<Scalar: TensorFlowFloatingPoint>(
     })
 }
 
-/// Computes a 2-D convolution using `self` as input, with the specified
-/// filter, strides, and padding.
+/// Computes a 2-D convolution with the specified input, filter, strides, and padding.
 ///
 /// - Parameters:
-///   - x: The input.
+///   - input.
```
rxwei: This is not the correct documentation comment syntax. [Swift API Design Guidelines - Fundamentals](https://swift.org/documentation/api-design-guidelines/#fundamentals) which describes the documentation comment syntax.

You should add a colon after the _parameter name_, followed by a _noun phrase_ that describes the parameter, followed by a period.
```suggestion
///   - input: The input.
```

Shashi456: @rxwei alright, i just find this very redundant is all. I'll update it accordingly.

PR #252, opened 2019-06-18
Add Exponential Linear Unit

Shashi456:
Working on leakyrelu, Selu, and a few others. Will add them in another PR. Build and test passes locally.


Sources/TensorFlow/Operators/Math.swift
```diff
@@ -996,6 +996,23 @@ func _vjpLogSoftmax<T: TensorFlowFloatingPoint>(
     return (value, { v in v - v.sum(alongAxes: -1) * exp(value) })
 }
 
+/// Computes `elu` of the specified tensor element-wise.
```
```suggestion
/// Returns a tensor by applying an exponential linear unit element-wise.
```

#160

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -996,6 +996,23 @@ func _vjpLogSoftmax<T: TensorFlowFloatingPoint>(
...
+@differentiable(vjp: _vjpElu(_:))
+public func elu<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
+    Raw.elu(features: x)
+}
+
+@inlinable
+func _vjpElu<T: TensorFlowFloatingPoint>(
+    _ x: Tensor<T>
+) -> (Tensor<T>, (Tensor<T>) -> Tensor<T>) {
+    (elu(x), { v in Raw.eluGrad(gradients: v, outputs: elu(x)) })
```
rxwei: The current implementation is computing `elu(x)` twice, which is not efficient. Instead, the output should be computed first and be returned and captured by the pullback closure.

```suggestion
    let y = elu(x)
    return (y, { v in Raw.eluGrad(gradients: v, outputs: y) })
```

Tests/TensorFlowTests/OperatorTests/MathTests.swift
```diff
@@ -236,6 +236,13 @@ final class MathOperatorTests: XCTestCase {
       XCTAssertEqual(y, expected)
     }
 
+    func testElu() {
+      let x = Tensor<Float>([-1.0, 2.0, 3.0])
```
rxwei: Please use 4-space indentation. Same for `testSoftPlus()` and `testSoftSign()` above.

PR #255, opened 2019-06-18
Handle more attribute kinds in LazyTensorOperation.

bgogul:


Sources/TensorFlow/Core/LazyTensorOperation.swift
```diff
@@ -153,6 +153,9 @@ class LazyTensorOperation: TensorOperation {
         case ConstTensor(TFETensorHandle)
         case TensorDataTypeValue(TensorDataType)
         case TensorFunctionPointer(_TensorFunctionPointer)
+        case TensorDataTypeArray([TensorDataType])
```
rxwei: Enum cases should use camel case. Didn't notice it earlier.

Sources/TensorFlow/Core/LazyTensorOperation.swift
```diff
@@ -659,6 +662,12 @@ extension LazyTensorOperation.Attribute: CustomStringConvertible {
         case .ConstTensor(let v): return v.valueDescription
         case .TensorDataTypeValue(let v): return dataTypeAsString(v)
         case .TensorFunctionPointer(let v): return "TFFunction(\(v.name))"
+        case .TensorDataTypeArray(let values):
+            let descriptions = (values.map { dataTypeAsString($0) })
```
rxwei:
```suggestion
            let descriptions = values.map { dataTypeAsString($0) }
```

Sources/TensorFlow/Core/LazyTensorOperation.swift
```diff
@@ -659,6 +662,12 @@ extension LazyTensorOperation.Attribute: CustomStringConvertible {
...
+            let descString = descriptions.joined(separator: ", ")
+            return "[\(descString)]"
+        case .OptionalTensorShape(let t):  return String(describing: t)
```
rxwei:
```suggestion
        case .OptionalTensorShape(let t): return String(describing: t)
```

Sources/TensorFlow/Core/LazyTensorOperation.swift
```diff
@@ -659,6 +662,12 @@ extension LazyTensorOperation.Attribute: CustomStringConvertible {
...
+        case .OptionalTensorShapeArray(let t):  return "\(t)"
```
rxwei:
```suggestion
        case .OptionalTensorShapeArray(let t): return "\(t)"
```

Sources/TensorFlow/Core/LazyTensorOperation.swift
bgogul: Will send a follow up PR for this.

PR #257, opened 2019-06-18
Add an Embedding layer

jon-tow:
Add support for embeddings.

#54 

It might be a good idea to find a type alias for this layer to avoid having to supply a useless generic type argument, e.g.`Embedding<Float>`, since a `TensorFlowFloatingPoint` is not the true type (`Int`) of input into the layer. Maybe Swift will soon implement default generic parameters?


Sources/TensorFlow/Layers/Embedding.swift
```diff
@@ -0,0 +1,63 @@
...
+/// `Embedding` is effectively a lookup table that maps indices from a fixed vocabulary to fix sized
+/// (dense) vector representations.
+/// E.g. `[[0], [3]] -> [[0.25, 0.1], [0.6, -0.2]]`
+public struct Embedding<Scalar: TensorFlowFloatingPoint>: Layer {
+    /// A helper structure containing the indices to be passed into the `Embedding` layer.
+    ///
+    /// - Note: Often times, `Embedding` is followed by a `Flatten` and a `Dense` layer. When this 
+    ///   is the case, ensure that all input sequences of indices have the same dimension.
+    /// - Note: This structure is needed to conform `Embedding` to the Layer protocol.
+    public struct Input:Differentiable {
```
rxwei:
```suggestion
    public struct Input: Differentiable {
```

Sources/TensorFlow/Layers/Embedding.swift
```diff
@@ -0,0 +1,63 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// An embedding layer.
+///
+/// `Embedding` is effectively a lookup table that maps indices from a fixed vocabulary to fix sized
```
rxwei:
```suggestion
/// `Embedding` is effectively a lookup table that maps indices from a fixed vocabulary to fixed-size
```

Sources/TensorFlow/Layers/Embedding.swift
```diff
@@ -0,0 +1,63 @@
...
+/// (dense) vector representations.
```
rxwei:
```suggestion
/// (dense) vector representations, e.g. `[[0], [3]] -> [[0.25, 0.1], [0.6, -0.2]]`.
```

Sources/TensorFlow/Layers/Embedding.swift
```diff
@@ -0,0 +1,63 @@
...
+/// E.g. `[[0], [3]] -> [[0.25, 0.1], [0.6, -0.2]]`
```
rxwei:
```suggestion
```

Sources/TensorFlow/Layers/Embedding.swift
```diff
@@ -0,0 +1,63 @@
...
+public struct Embedding<Scalar: TensorFlowFloatingPoint>: Layer {
+    /// A helper structure containing the indices to be passed into the `Embedding` layer.
```
rxwei:
```suggestion
    /// A input structure containing the embedding indices.
```

Sources/TensorFlow/Layers/Embedding.swift
```diff
@@ -0,0 +1,63 @@
...
+    ///
+    /// - Note: Often times, `Embedding` is followed by a `Flatten` and a `Dense` layer. When this 
+    ///   is the case, ensure that all input sequences of indices have the same dimension.
+    /// - Note: This structure is needed to conform `Embedding` to the Layer protocol.
```
rxwei: This particular note is an implementation note (not part of API documentation comments), so it should use two slashes.
```suggestion
    // NOTE: This structure is needed to conform `Embedding` to the Layer protocol.
```

Sources/TensorFlow/Layers/Embedding.swift
```diff
@@ -0,0 +1,63 @@
...
+    /// A learnable lookup table that maps vocabulary indices to their dense vector representations.
+    public var embeddings: Tensor<Scalar>
+
+    /// Creates an `Embedding` layer with randomly initialized embeddings of shape 
+    /// `(vocabularySize, embeddingSize)` so that each vocabulary index is given a vector 
+    /// representation.
+    ///
+    /// - Parameters:
+    ///   - vocabularySize: The number of distinct indices (words) in the vocabulary. This number
+    ///     should be the `largest integer index + 1`.
```
rxwei: Since this is mostly verbal (not code), it's better to remove backquotes and just use words.
```suggestion
    ///     should be the largest integer index plus one.
```

Sources/TensorFlow/Layers/Embedding.swift
```diff
@@ -0,0 +1,63 @@
...
+    ///   - embeddingSize: The number of entries in a single embedding vector representation.
+    public init(vocabularySize: Int, embeddingSize: Int) {
+        self.embeddings = Tensor(randomUniform: [vocabularySize, embeddingSize])
+    }
+
+    /// Creates an `Embedding` layer from the provided embeddings. Useful for introducing 
+    /// pretrained embeddings into a model.
+    /// 
+    /// - Parameter embeddings: The pretrained embeddings table.
+    public init(from embeddings: Tensor<Scalar>) {
```
rxwei: According to the Swift API Design Guidelines, argument labels in initializers usually do not form a phrase with the type name. Also, since this initializer is a memberwise initializer, the argument label should be the public property itself.

```suggestion
    public init(embeddings: Tensor<Scalar>) {
```

Sources/TensorFlow/Layers/Embedding.swift
```diff
@@ -0,0 +1,63 @@
...
+        self.embeddings = embeddings
+    }
+
+    /// Replaces each index in the input with corresponding dense vector representations.
```
rxwei:
```suggestion
    /// Returns an output by replacing each index in the input with corresponding dense vector representation.
```
#160 can provide more context.

Sources/TensorFlow/Layers/Embedding.swift
```diff
@@ -0,0 +1,63 @@
...
+    ///
+    /// - Parameter
+    ///   - input: The indices that will be mapped to their vector representations.
+    /// - Returns:
+    ///   - output: The tensor created by replacing input indices with their vector representations.
```
rxwei:
```suggestion
```

Sources/TensorFlow/Layers/Embedding.swift
```diff
@@ -0,0 +1,63 @@
...
+    /// - Parameter embeddings: The pretrained embeddings table.
+    public init(from embeddings: Tensor<Scalar>) {
+        self.embeddings = embeddings
+    }
+
+    /// Replaces each index in the input with corresponding dense vector representations.
+    ///
+    /// - Parameter
+    ///   - input: The indices that will be mapped to their vector representations.
+    /// - Returns:
```
rxwei: The `- Returns:` API documentation item should be followed directly by the output description, since there's no such identifier called `output`.

```suggestion
    /// - Returns: The tensor created by replacing input indices with their vector representations.
```

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -205,6 +205,23 @@ final class LayerTests: XCTestCase {
         XCTAssertEqual(output.shape, expected)
     }
 
+    func testEmbedding() {       
+        var layer = Embedding<Float>(vocabularySize: 3, embeddingSize: 5)       
+        var data = Tensor<Int32>(shape: [2,3], scalars: [0, 1, 2, 1, 2, 2])
+        var input = Embedding<Float>.Input(indices: data)
```
rxwei: This line shows that it's suboptimal to define `Input` as a nested struct inside `Embedding` because it requires a generic parameter (i.e. `Embedding.Scalar`) to create an `Input`. As such, I think it's better to define a top-level struct `EmbeddingInput`.

```swift
@frozen
public struct EmbeddingInput: Differentiable {
    @noDerivative public var indices: Tensor<Int32>
}
```

This way you can create an embedding input directly: `Embedding(indices: data)` and not care about the output scalar type of the embedding layer.

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -205,6 +205,23 @@ final class LayerTests: XCTestCase {
         XCTAssertEqual(output.shape, expected)
     }
 
+    func testEmbedding() {       
+        var layer = Embedding<Float>(vocabularySize: 3, embeddingSize: 5)       
+        var data = Tensor<Int32>(shape: [2,3], scalars: [0, 1, 2, 1, 2, 2])
```
rxwei:
```suggestion
        var data = Tensor<Int32>(shape: [2, 3], scalars: [0, 1, 2, 1, 2, 2])
```

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -205,6 +205,23 @@ final class LayerTests: XCTestCase {
...
+        var input = Embedding<Float>.Input(indices: data)
+        var output = layer.inferring(from: input)
+        let expectedShape = TensorShape([2, 3, 5])
+        XCTAssertEqual(output.shape, expectedShape)
+    
+        let pretrained = Tensor<Float>(shape:[2, 2], scalars: [0.4, 0.3, 0.2, 0.1])
+        layer = Embedding<Float>(from: pretrained)
+        data = Tensor<Int32>(shape: [2,2], scalars: [0, 1, 1, 1])
```
rxwei:
```suggestion
        data = Tensor<Int32>(shape: [2, 2], scalars: [0, 1, 1, 1])
```

rxwei:
> Maybe Swift will soon implement default generic parameters?

Yep, the Swift for TensorFlow team plans to implement default generic arguments at some point if it's not yet done by the Swift community.

jon-tow:
@rxwei Glad to help out! Sorry for all the tiny mistakes. I'll be much more careful on any future PRs I submit.

Sources/TensorFlow/Layers/Embedding.swift
```diff
@@ -0,0 +1,70 @@
...
+
+/// An input structure containing the embedding indices.
+///
+/// - Note: Often times, `Embedding` is followed by a `Flatten` and a `Dense` layer. When this 
+///   is the case, ensure that all input sequences of indices have the same dimension.
+// NOTE: This structure is needed to conform `Embedding` to the Layer protocol.
+@frozen
+public struct EmbeddingInput: Differentiable {
+    /// Sequences of indices that will be passed into the layer.
+    @noDerivative public var indices: Tensor<Int32>
```
jekbradbury: It's absolutely possible to have an embedding layer with more than 2^32 embeddings (and therefore one that requires `Int64` indices), but it's fine to not worry about that for now.

rxwei: `gathering(atIndices:alongAxis:)` is not yet generic over `BinaryInteger`. To fix this, we need to start from there.

rxwei: https://github.com/tensorflow/swift-apis/issues/259

PR #260, opened 2019-06-19
Add Leaky relu activation function

Shashi456:
I was going to add SELU as well, but it has a prerequisite of initializing with variance scaling, which needs flow control. I'll add that later. Test and build passes locally.

Working on Separable layers next and the layer abstraction we spoke about.

### DIFF
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -1026,7 +1026,7 @@ public func leakyRelu<T: TensorFlowFloatingPoint>(_ x: Tensor<T>, alpha: Double 
 func _vjpLeakyRelu<T: TensorFlowFloatingPoint>(
     _ x: Tensor<T>,
     alpha: Double = 0.2
-) -> (Tensor<T>, (Tensor<T>) -> Tensor<T>) {
+) -> (Tensor<T>, Double, (Tensor<T>) -> Tensor<T>) {
     return (leakyRelu(x), { v in Raw.leakyReluGrad(gradients: v, features: x, alpha: alpha) })
 }
 
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -1014,22 +1014,26 @@ func _vjpElu<T: TensorFlowFloatingPoint>(
     return (y, { v in Raw.eluGrad(gradients: v, outputs: y) })
 }
 
-/// Computes `leakyRelu` of the specified tensor element-wise.
+/// Returns a tensor by applying the leaky ReLU activation function to the specified tensor element-wise.
 /// Specifically, computes `max(features, features * alpha)`.
 @inlinable
 @differentiable(wrt: x, vjp: _vjpLeakyRelu)
-public func leakyRelu<T: TensorFlowFloatingPoint>(_ x: Tensor<T>, alpha: Double = 0.2) -> Tensor<T> {
+public func leakyRelu<T: TensorFlowFloatingPoint>(
+    _ x: Tensor<T>,
+    alpha: Double = 0.3
+) -> Tensor<T> {
     Raw.leakyRelu(features: x, alpha: alpha)
 }
 
 @inlinable
 func _vjpLeakyRelu<T: TensorFlowFloatingPoint>(
     _ x: Tensor<T>,
-    alpha: Double = 0.2
+    alpha: Double
 ) -> (Tensor<T>, (Tensor<T>) -> Tensor<T>) {
-    return (leakyRelu(x, alpha: alpha), { v in Raw.leakyReluGrad(gradients: v, features: x, alpha: alpha) })
+    return (leakyRelu(x, alpha: alpha), { v in
+        Raw.leakyReluGrad(gradients: v, features: x, alpha: alpha)
+    })
 }
-
 
 /// Computes `relu` of the specified tensor element-wise.
 /// Specifically, computes `max(0, x)`.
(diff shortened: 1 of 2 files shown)


Shashi456:
Also are there any other activations to be added?

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1014,6 +1014,23 @@ func _vjpElu<T: TensorFlowFloatingPoint>(
     return (y, { v in Raw.eluGrad(gradients: v, outputs: y) })
 }
 
+/// Computes `leakyRelu` of the specified tensor element-wise.
```
rxwei: - The API documentation here is a bit redundant in that it does not explain what `leakyRelu` is. Other documentation like `relu` is also unclear. It should be changed to using the standard terminology.
- Could you please make sure #160 is addressed in future PRs?

```suggestion
/// Returns a tensor by applying the leaky ReLU activation function to the specified tensor element-wise.
```

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1014,6 +1014,23 @@ func _vjpElu<T: TensorFlowFloatingPoint>(
...
+
+@inlinable
+func _vjpLeakyRelu<T: TensorFlowFloatingPoint>(
+    _ x: Tensor<T>,
+    alpha: Double = 0.2
+) -> (Tensor<T>, (Tensor<T>) -> Tensor<T>) {
+    return (leakyRelu(x, alpha: alpha), { v in Raw.leakyReluGrad(gradients: v, features: x, alpha: alpha) })
+}
+
+
```
rxwei: Remove extra line.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1014,6 +1014,23 @@ func _vjpElu<T: TensorFlowFloatingPoint>(
...
+@inlinable
+@differentiable(wrt: x, vjp: _vjpLeakyRelu)
+public func leakyRelu<T: TensorFlowFloatingPoint>(_ x: Tensor<T>, alpha: Double = 0.2) -> Tensor<T> {
+    Raw.leakyRelu(features: x, alpha: alpha)
+}
+
+@inlinable
+func _vjpLeakyRelu<T: TensorFlowFloatingPoint>(
+    _ x: Tensor<T>,
+    alpha: Double = 0.2
```
rxwei: Please remove default values from VJP functions.

```suggestion
    alpha: Double
```

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1014,6 +1014,23 @@ func _vjpElu<T: TensorFlowFloatingPoint>(
...
+) -> (Tensor<T>, (Tensor<T>) -> Tensor<T>) {
+    return (leakyRelu(x, alpha: alpha), { v in Raw.leakyReluGrad(gradients: v, features: x, alpha: alpha) })
```
rxwei: Please make sure all lines fit within 100 columns.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1014,6 +1014,23 @@ func _vjpElu<T: TensorFlowFloatingPoint>(
     return (y, { v in Raw.eluGrad(gradients: v, outputs: y) })
 }
 
+/// Computes `leakyRelu` of the specified tensor element-wise.
+/// Specifically, computes `max(features, features * alpha)`.
+@inlinable
+@differentiable(wrt: x, vjp: _vjpLeakyRelu)
+public func leakyRelu<T: TensorFlowFloatingPoint>(_ x: Tensor<T>, alpha: Double = 0.2) -> Tensor<T> {
```
rxwei: - For consistency with tf.keras, `alpha` should be `0.3`.
- Fit within 100 columns.

Shashi456: @rxwei so 0.2 is the default in the raw c++ apis tho, what would you like to choose?

Sources/TensorFlow/Operators/Math.swift
Shashi456: @rxwei I'll start working on #160 for all the files, I just take the neighborhood functions as references and the documentation is pretty inconsistent so I make these mistakes. I'll work on #160 for math.swift this week.

Sources/TensorFlow/Operators/Math.swift
rxwei: Ok, in that case, `0.2` is good. Thanks!

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1014,6 +1014,27 @@ func _vjpElu<T: TensorFlowFloatingPoint>(
     return (y, { v in Raw.eluGrad(gradients: v, outputs: y) })
 }
 
+/// Returns a tensor by applying the leaky ReLU activation function to the specified tensor element-wise.
+/// Specifically, computes `max(features, features * alpha)`.
```
rxwei: `features` is undefined.

```suggestion
/// Specifically, computes `max(x, x * alpha)`.
```

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1014,6 +1014,27 @@ func _vjpElu<T: TensorFlowFloatingPoint>(
     return (y, { v in Raw.eluGrad(gradients: v, outputs: y) })
 }
 
+/// Returns a tensor by applying the leaky ReLU activation function to the specified tensor element-wise.
```
rxwei: Fit within 100 columns.

PR #261, opened 2019-06-19
Provide proper documentation for non-mutating functions

jon-tow:
A quick clean up to help with #160 to avoid wasting time.

### DIFF
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -619,7 +619,7 @@ internal func _vjpTan<T: TensorFlowFloatingPoint>(
     return (value, { v in v * (1 + value.squared()) })
 }
 
-/// Returns `sinh` of the specified tensor element-wise.
+/// Returns the hyperbolic sine of the specified tensor element-wise.
 @inlinable
 @differentiable(vjp: _vjpSinh(_:))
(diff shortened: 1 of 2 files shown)


Shashi456:
I think [this](https://github.com/tensorflow/swift-apis/pull/260#discussion_r295137460) still needs to be addressed.

jon-tow:
Ah sorry, I thought I'd save you some time with this. Should I close this?

Shashi456:
@jon-tow no, this is good work. We can discuss some things about the variety of functions in Math.swift, this does save time.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -619,7 +619,7 @@ internal func _vjpTan<T: TensorFlowFloatingPoint>(
     return (value, { v in v * (1 + value.squared()) })
 }
 
-/// Computes `sinh` of the specified tensor element-wise.
+/// Returns `sinh` of the specified tensor element-wise.
```
Shashi456: @rxwei what do you think about 
```suggestion
/// Returns the hyperbolic sine of the specified tensor element-wise.
```
and similarly for all hyperbolic functions

jon-tow: I see what your saying! Sounds good. I'll go through those and give a bit more explanation where possible.

jon-tow:
I confirm the co-authoring!

PR #262, opened 2019-06-19
Track the device name in `LazyTensorOperation`

bgogul:


Sources/TensorFlow/Core/LazyTensorOperation.swift
```diff
@@ -162,6 +162,7 @@ class LazyTensorOperation: TensorOperation {
     let outputCount: Int
     var inputs: [Input]
     var attributes: [String: Attribute]
+    var device: String?
```
rxwei: Since this is not a device but a device name, how about calling it `deviceName` instead?

bgogul: Done.

PR #268, opened 2019-06-21
Add support for gelu

jon-tow:
Given its usage in recent state of the art NLP models (XLNet, GPT-2, BERT, etc.), gelu seems to be a useful activation for future users to reach for.

### DIFF
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -1051,6 +1051,24 @@ func _vjpRelu<T: TensorFlowFloatingPoint>(
     (relu(x), { v in Tensor(x .> 0) * v })
 }
 
+/// Returns the Gaussian Error Linear Unit (GELU) of the specified tensor element-wise.
+///
+/// Specifically, `gelu` approximates `xP(X <= x)`, where `P(X <= x)` is the Standard Gaussian
+/// cumulative distribution, by computing: x * [0.5 * (1 + tanh[√(2/π) * (x + 0.044715 * x^3)])].
+///
+/// See [Gaussian Error Linear Units](https://arxiv.org/abs/1606.08415).
+@inlinable
+@differentiable
+public func gelu<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
+    let ratio = Tensor<T>(0.7978845608) // An approximation of √(2/π).
+    // An approximation of the Gauss error function.
+    // NOTE: This is needed because the compiler otherwise gives an "unable to type-check this
+    // in reasonable time" error when the below expressions are written on a single line.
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1051,6 +1051,23 @@ func _vjpRelu<T: TensorFlowFloatingPoint>(
     (relu(x), { v in Tensor(x .> 0) * v })
 }
 
+/// Returns the Gaussian Error Linear Unit (GELU) of the specified tensor element-wise.
```
Shashi456:
```suggestion
/// Returns a tensor by applying the Gaussian Error Linear Unit (GELU) to the specified tensor element-wise.
```

Also, if you are adding activation functions which aren't extremely popular, could you also add a link to the paper this was introduced in?

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1051,6 +1051,23 @@ func _vjpRelu<T: TensorFlowFloatingPoint>(
     (relu(x), { v in Tensor(x .> 0) * v })
 }
 
+/// Returns a tensor by applying the Gaussian Error Linear Unit (GELU) to the specified tensor element-wise.
```
Shashi456: Fit in 100 columns.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1051,6 +1051,23 @@ func _vjpRelu<T: TensorFlowFloatingPoint>(
...
+///
+/// Specifically, `gelu` computes `x * [0.5 * (1 + tanh[√(2/π) * (x + 0.044715 * x^3)])]` 
+/// to approximate `xP(X <= x)`, where `P(X <= x)` is the Standard Gaussian cumulative distribution.
```
Shashi456: Fit this in 100 columns as well.

Sources/TensorFlow/Operators/Math.swift
Shashi456: @jon-tow you seem to have reversed this change

jon-tow: @Shashi456 It brought line past the 100 column limit. I didn't realize it until after the commit.Sorry about that.

Shashi456: @jon-tow, It's absolutely alright. You can even do something like 
```
/// Returns a tensor by applying the Gaussian Error Linear Unit (GELU) 
/// to the specified tensor element-wise.
```
But your current wording is alright as well.

jon-tow: @Shashi456 I thought your suggestion was worded well. I ended up using a different wording that managed to keep the line under 100 columns (most recent commit here). Appreciate the suggestions!

PR #269, opened 2019-06-21
Improving documentation in Math.swift #160

Shashi456:


Sources/TensorFlow/Operators/Math.swift
```diff
@@ -895,6 +895,7 @@ internal func _vjpFloor<T: TensorFlowFloatingPoint>(
     (floor(x), { _ in Tensor(0).broadcasted(like: x) })
 }
 
+///Returns an indication of the sign of the specified tensor element-wise.
```
saeta:
```suggestion
/// Returns an indication of the sign of the specified tensor element-wise.
```

Also, could you talk about what the user should expect to find in the returned Tensor? -1, 0, and 1? Or something else? (Note: from a quick poke around below, looks like it's -1, 0, and 1.)

```
saeta@saeta-v04-jupyter:~$ swift
Welcome to Swift version 5.1-dev (LLVM 7d54a7c711, Swift 899693953e).
Type :help for assistance.
  1> import TensorFlow
  2> let x = Tensor([1, 2, 0, -1])
x: Tensor<Double> = [ 1.0,  2.0,  0.0, -1.0]
  3> sign(x)
$R2: Tensor<Double> = [ 1.0,  1.0,  0.0, -1.0]
  4>  

```

PR #270, opened 2019-06-21
Fix doc of mutating function in Math.swift.

naru-jpn:
Reference #160 

Fix doc of _mutating function_ with no return value.


Shashi456:
You seem to be reverting the changes made in #183.

naru-jpn:
Doc says `This function returns value` but actually function returns no value and store value in left side variable.
#160 is an issue about _non-mutating_ function.
I don't want to revert some PR... But I think that I should improve doc for developers.

PR #271, opened 2019-06-21
Made 'Tensor.gathering(where:alongAxis:)' differentiable.

eaplatanios:


eaplatanios:
Tests pass so this should be ready to merge.

PR #272, opened 2019-06-21
Makes 'Tensor.gathering' generic over 'Int32' and 'Int64'.

### LINKED ISSUE #259: Make 'gathering(atIndices:alongAxis:)' generic over 'Int32' and 'Int64'.
We need to make `Tensor.gathering(atIndices:alongAxis:)` support `Int64` indices, but we cannot do so by making the index tensor's `Scalar` type be generic over `BinaryInteger` because the TensorFlow `Gather` operator (like many other operators that take indices) only supports `Int32` and `Int64` indices.

To resolve this, we should consider adding a `TensorIndexScalar` protocol and conform only `Int32` and `Int64` to it.
```swift
public protocol TensorIndexScalar: BinaryInteger & TenosrFlowScalar {
    ...
}
```

eaplatanios:
Fixes #259.


rxwei:
I’m not sure this is the safest solution, because Int8 and Int16 also conforms to TensorFlowInteger.

eaplatanios:
Oh I see. I can make a new protocol then, as you suggested.

eaplatanios:
It's weird though that the raw ops generation script assigns `TensorFlowScalar & BinaryInteger` as the signature of the gather ops.

eaplatanios:
In either case, I created a separate protocol and local tests pass. Currently running CI tests.

rxwei:
Yeah, I think the wrapper generation script doesn’t pay attention to the specific integer types. We should change the script to use the new protocol for APIs that only take index scalar types.

eaplatanios:
Sure, I can do that in a separate PR after we're done with this one.

rxwei:
Thanks!

PR #273, opened 2019-06-21
Fixed the handling of integer data types in the raw ops generation script.

eaplatanios:
Follow-up to #272.


eaplatanios:
Tests pass locally and currently running on the CI.

Sources/TensorFlow/Bindings/generate_wrappers.py
```diff
@@ -76,19 +76,21 @@
   ({types_pb2.DT_UINT8,
     types_pb2.DT_UINT16,
     types_pb2.DT_UINT32,
-    types_pb2.DT_UINT64}, 'UnsignedInteger & TensorFlowScalar'),
+    types_pb2.DT_UINT64}, 'TensorFlowUnsignedInteger'),
```
rxwei: It's a little weird to have an `Unsigned` protocol because Swift standard library doesn't. Are there any TensorFlow ops that only take unsigned integers?

eaplatanios: Good point. There aren't actually. I had just left that there because it existed before I started modifying the op generation script, but given this, I'll just remove it.

eaplatanios: I guess there is `encodePng` and `decodePng`, but we can remove it since there's only two of them.

eaplatanios: Done.

eaplatanios:
@rxwei for this and future PRs, if approved by at least one reviewer and the CI test pass, should I go ahead and merge?

rxwei:
Yes, please! Thank you for help maintaining the library!

eaplatanios:
Thanks for making it easier to do so! :)

Issue #274, opened 2019-06-21
Add a 'dilations:' parameter to 'conv2D(_:filter:strides:padding:)'


rxwei:
The raw operator `Raw.conv2D(_:filter:strides:useCudnnOnGpu:padding:explicitPaddings:dataFormat:dilations:)` has a dilations parameter. We should add a parameter to  'conv2D(_:filter:strides:padding:)' so that it becomes 'conv2D(_:filter:strides:padding:dilations:)'.

jon-tow:
I'll take a go at this!

rxwei:
Excellent!

Shashi456:
@rxwei I'll take care of the conv1d and conv3d updates for dilations.

rxwei:
Awesome :)

jon-tow:
@Shashi456 I completed the conv1d updates for dilation b/c it depends on a call to conv2d. Hope that's okay with you?

Shashi456:
@jon-tow cool.

eaplatanios:
Resolved in #275. Thanks @jon-tow!

rxwei:
This change broke a lot of existing code because the newly added parameter does not have a default. We need to add a default argument for dilations.

eaplatanios:
@rxwei Did #277 not resolve this?

Shashi456:
@eaplatanios, I think the parameter also needs to be made optional something like `_ dilations` or else update dilations in other repos like swift models and I think tensorflow/swift

rxwei:
> @rxwei Did #277 not resolve this?

#277 added defaults to layers, but not the op itself.

Shashi456:
@rxwei I'll take this up then. Will get the PR up in a few hours? I'm working on #283 as well.

jon-tow:
Ah geez :frowning: sorry folks. @Shashi456 Why would it also need to be optional if it's given a default?

Shashi456:
@jon-tow to keep the current implementations with conv2d intact

Shashi456:
Ah, Looks like you put the PR already. Alright

jon-tow:
@Shashi456 I thought I'd fix my own wrong doing, rather than have you waste your time on my mistake!

Shashi456:
@jon-tow chill xD.

rxwei:
> Ah geez 😦 sorry folks. @Shashi456 Why would it also need to be optional if it's given a default?

[...]

PR #275, opened 2019-06-22
Add support for dilated Conv1D and Conv2D

jon-tow:
#274
 
The documentation for some layers seem to be lacking in detail. I'll try to clarify some things when I get the chance.

### DIFF
--- a/Sources/TensorFlow/Layers/Convolutional.swift
+++ b/Sources/TensorFlow/Layers/Convolutional.swift
@@ -65,11 +65,12 @@ public struct Conv1D<Scalar: TensorFlowFloatingPoint>: Layer {
     /// - Returns: The output `[batchCount, newWidth, outputChannels]`.
     @differentiable
     public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
-        let conv = conv2D(input.expandingShape(at: 1),
-                          filter: filter.expandingShape(at: 0),
-                          strides: (1, 1, stride, 1),
-                          padding: padding,
-                          dilations: (1, 1, dilation, 1))
+        let conv = conv2D(
+            input.expandingShape(at: 1),
+            filter: filter.expandingShape(at: 0),
+            strides: (1, 1, stride, 1),
+            padding: padding,
+            dilations: (1, 1, dilation, 1))
         return activation(conv.squeezingShape(at: 1) + bias)
     }
 }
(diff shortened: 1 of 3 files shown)


Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -60,8 +65,11 @@ public struct Conv1D<Scalar: TensorFlowFloatingPoint>: Layer {
     /// - Returns: The output `[batchCount, newWidth, outputChannels]`.
     @differentiable
     public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
-        let conv = conv2D(input.expandingShape(at: 1), filter: filter.expandingShape(at: 0),
-                          strides: (1, 1, stride, 1), padding: padding)
+        let conv = conv2D(input.expandingShape(at: 1),
```
eaplatanios:
```suggestion
        let conv = conv2D(
            input.expandingShape(at: 1),
```

Can you align the arguments in a new line, like in the suggested edit?

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -178,8 +197,11 @@ public struct Conv2D<Scalar: TensorFlowFloatingPoint>: Layer {
     /// - Returns: The output.
     @differentiable
     public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
-        return activation(conv2D(input, filter: filter, strides: (1, strides.0, strides.1, 1),
-                                 padding: padding) + bias)
+        return activation(conv2D(input,
```
eaplatanios: Same comment as above.

jon-tow:
@eaplatanios Thanks for pointing that out. Wish we had a designated formatter :smile:.

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -178,8 +198,12 @@ public struct Conv2D<Scalar: TensorFlowFloatingPoint>: Layer {
     /// - Returns: The output.
     @differentiable
     public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
-        return activation(conv2D(input, filter: filter, strides: (1, strides.0, strides.1, 1),
-                                 padding: padding) + bias)
+        return activation(conv2D(
+                            input,
```
eaplatanios: The indentation for input should just be 4-spaces further than the `return` statement.

jon-tow: Sorry about that. I was unsure what to do and should've asked beforehand.

Shashi456:
@eaplatanios @rxwei Although I see the extensions have dilation as optional, I'm wondering why dilation wasn't made optional in the main public init, because the dilated convolutions isn't used by everyone. Even within the raw bindings, they are an optional parameter with some default value declared. Wdyt?

eaplatanios:
@Shashi456 good point. I added some defaults along with some minor stylistic changes in #277.

rxwei:
+1 to default arguments.

PR #276, opened 2019-06-22
[WIP] Remove the `Layer.Input: Differentiable` Requirement

eaplatanios:
@rxwei This is work-in-progress for removing the `Differentiable` requirement from `Layer.Input`. I'll work on the recurrent layers later but I first want to sort out an issue that comes up with the current changes alone. When I try to compile this PR I get the same auto-diff error as in #147. I'm not sure what the underlying reason is, but it seems like an auto-diff bug in the compiler.

Regarding this PR, I feel that `Input` should not need to conform to `Differentiable` in layers as in the general use cases we're only differentiating with respect to the layer itself. Regarding layer compositions (e.g., `sequenced`), I added conditional differentiability support.

### DIFF
--- a/Sources/TensorFlow/Layer.swift
+++ b/Sources/TensorFlow/Layer.swift
@@ -22,7 +22,7 @@
 public protocol Layer: Differentiable & KeyPathIterable
     where AllDifferentiableVariables: KeyPathIterable {
     /// The input type of the layer.
-    associatedtype Input: Differentiable
+    associatedtype Input
     /// The output type of the layer.
     associatedtype Output: Differentiable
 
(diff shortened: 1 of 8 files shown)


Sources/TensorFlow/Layer.swift
```diff
@@ -105,7 +106,7 @@ public extension Differentiable {
     ///   - l2: The second layer.
     ///   - l3: The third layer.
     /// - Returns: The final layer's output after sequential application.
-    @differentiable
+    @differentiable(where L2.Input: Differentiable, L3.Input: Differentiable)
     func sequenced<L1: Layer, L2: Layer, L3: Layer>(through l1: L1, _ l2: L2, _ l3: L3) -> L3.Output
         where L1.Input == Self, L1.Output == L2.Input, L2.Output == L3.Input {
         let o1 = l1(self)
```
rxwei: This won't work because `Layer` is not differentiable w.r.t. the input after this change. Even protocol conformances can't make the `callAsFunction(_:)` function differentiable because it is not in the protocol witness table. The proper solution will require some internal compiler work -- happy to talk about that in a separate GitHub issue.

eaplatanios: I see, yes that makes sense. It would be good to talk about this and see how to make it possible. Also, I noticed that `@differentiable` does not imply `@differentiable(wrt: self)` (e.g., when there are more function arguments. I believe this should not be the case.

There is one more way that could allow us to support this, but I have a question first. Can we support having multiple instances of the `@differentiable` attribute per function? E.g., `@differentiable(wrt: self)` and `@differentiable(wrt: (self, input) where Input: Differentiable)`, for a single function.

eaplatanios: @rxwei can we currently define multiple `@differentiable` attributes? For example:
```swift
@differentiable(wrt: self)
@differentiable(wrt: (self, input) where Input: Differentiable)
func callAsFunction(_ input: Input) -> Output {
  ...
}
```

rxwei: Yes.

eaplatanios: When I try to compile the following:
```swift
@differentiable(wrt: (self, input) where Input: Differentiable)
func callAsFunction(_ input: Input) -> Output {
  ...
}
```
I get the following auto-diff crash:
```
1.      Swift version 5.1-dev (LLVM 7d54a7c711, Swift 4ce1aaa2fb)
2.      While emitting IR SIL function "@AD__$s10TensorFlow5LayerPAAE4cally6OutputQz5InputQzF__vjp_src_0_wrt_0_1".
 for 'call(_:)' (at /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layer.swift:41:5)
0  swift                    0x000000010ef69955 llvm::sys::PrintStackTrace(llvm::raw_ostream&) + 37
1  swift                    0x000000010ef68998 llvm::sys::RunSignalHandlers() + 248
2  swift                    0x000000010ef69f48 SignalHandler(int) + 264
4  libsystem_platform.dylib 0x00007ffee4902ef8 _sigtramp + 1861510072
5  swift                    0x000000010bb68866 (anonymous namespace)::SILTypeSubstituter::substSILFunctionType(swift::CanTypeWrapper<swift::SILFunctionType>) + 710
6  swift                    0x000000010bb683e7 swift::SILType::subst(swift::SILModule&, llvm::function_ref<swift::Type (swift::SubstitutableType*)>, llvm::function_ref<llvm::Optional<swift::ProtocolConformanceRef> (swift::CanType, swift::Type, swift::ProtocolDecl*)>, swift::CanGenericSignature, bool) const + 263
7  swift                    0x000000010bb5e16f swift::GenericEnvironment::mapTypeIntoContext(swift::SILModule&, swift::SILType) const + 111
8  swift                    0x000000010b491660 swift::irgen::TypeConverter::getTypeEntry(swift::CanType) + 352
9  swift                    0x000000010b4912cf swift::irgen::IRGenModule::getTypeInfo(swift::SILType) + 47
10 swift                    0x000000010b3ae972 swift::irgen::Signature::getUncached(swift::irgen::IRGenModule&, swift::CanTypeWrapper<swift::SILFunctionType>) + 898
11 swift                    0x000000010b42aa4b swift::irgen::IRGenModule::getSignature(swift::CanTypeWrapper<swift::SILFunctionType>) + 107
12 swift                    0x000000010b46e56b swift::irgen::emitWitnessMethodValue(swift::irgen::IRGenFunction&, llvm::Value*, swift::SILDeclRef) + 347
13 swift                    0x000000010b46e6a1 swift::irgen::emitWitnessMethodValue(swift::irgen::IRGenFunction&, swift::CanType, llvm::Value**, swift::SILDeclRef, swift::ProtocolConformanceRef) + 65
14 swift                    0x000000010b4d393a swift::SILInstructionVisitor<(anonymous namespace)::IRGenSILFunction, void>::visit(swift::SILInstruction*) + 36714
15 swift                    0x000000010b4c77c3 swift::irgen::IRGenModule::emitSILFunction(swift::SILFunction*) + 8051
16 swift                    0x000000010b3e719e swift::irgen::IRGenerator::emitGlobalTopLevel() + 734
17 swift                    0x000000010b4a2b32 performIRGeneration(swift::IRGenOptions&, swift::ModuleDecl*, std::__1::unique_ptr<swift::SILModule, std::__1::default_delete<swift::SILModule> >, llvm::StringRef, swift::PrimarySpecificPaths const&, llvm::LLVMContext&, swift::SourceFile*, llvm::GlobalVariable**) + 1282
18 swift                    0x000000010b4a3022 swift::performIRGeneration(swift::IRGenOptions&, swift::SourceFile&, std::__1::unique_ptr<swift::SILModule, std::__1::default_delete<swift::SILModule> >, llvm::StringRef, swift::PrimarySpecificPaths const&, llvm::LLVMContext&, llvm::GlobalVariable**) + 82
19 swift                    0x000000010b352a1f performCompile(swift::CompilerInstance&, swift::CompilerInvocation&, llvm::ArrayRef<char const*>, int&, swift::FrontendObserver*, swift::UnifiedStatsReporter*) + 13423
20 swift                    0x000000010b34e5d1 swift::performFrontend(llvm::ArrayRef<char const*>, char const*, void*, swift::FrontendObserver*) + 3025
21 swift                    0x000000010b2f7a99 main + 729
22 libdyld.dylib            0x00007fff757d13d5 start + 1
23 libdyld.dylib            0x0000000000000052 start + 2323836030
```
Is this a bug or am I doing something wrong? More specifically, can the `where` clause in `@differentiable` constrain generic types from the enclosing scope (e.g., the struct)?

rxwei: Did you add it to a protocol requirement, or a concrete implementation?

rxwei: `@differentiable` in a protocol requirement cannot have a `where` clause—this level of polymorphism can be done with a separate protocol. `where` clauses are probably not yet diagnosed in when type checking a `@differentiable` on a protocol requirement, so we should fix that.

eaplatanios: I see. Would that be an easier fix that the ongoing `inout` support I'm working on? If so, I could tackle it first with some guidance.

rxwei: Yes, this should be in `TypeCheckAttr.cpp` near the place where you are diagnosing `inout` parameters.

eaplatanios: This is now done as part of apple/swift#25974.

eaplatanios:
Moved to #346.

PR #277, opened 2019-06-22
Added default values to some of the convolution layer initializers.

eaplatanios:
Addresses #275 comment.


Shashi456:
@eaplatanios, yeah this works. Thanks.

PR #280, opened 2019-06-23
Adding dilation factor to Convolution 3D Layer

Shashi456:
WIP : I'll add the tests.


Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -327,19 +332,21 @@ public struct Conv3D<Scalar: TensorFlowFloatingPoint>: Layer {
             input,
             filter: filter,
             strides: (1, strides.0, strides.1, strides.2, 1),
-            padding: padding) + bias)
+            padding: padding,
+            dilations: (1, dilations.0, dilations.1, dilations.2, 1)) + bias)
```
eaplatanios:
```suggestion
            dilations: (1, dilations.0, dilations.1, dilations.2, 1)
        ) + bias)
```

Sources/TensorFlow/Operators/NN.swift
```diff
@@ -337,14 +352,15 @@ func _vjpConv3DBackpropFilter<Scalar: TensorFlowFloatingPoint>(
     _ filterSizes: Tensor<Int32>,
     _ strides: (Int, Int, Int, Int, Int),
     _ padding: Padding
```
eaplatanios:
```suggestion
    _ padding: Padding,
```

Shashi456:
Updated this PR, the build passes locally. 

I've added a dilation test, which passes with its tensorflow equivalent and dimensions : 
```python
```

The error i get locally is :
```
Precondition failed: file /home/paws/swift-apis/Sources/TensorFlow/Core/Tensor.swift, line 246
Current stack trace:
0    libswiftCore.so                    0x00007f221aa7d270 swift_reportError + 50
1    libswiftCore.so                    0x00007f221aaec010 _swift_stdlib_reportFatalErrorInFile + 115
2    libswiftCore.so                    0x00007f221aa144ce <unavailable> + 3736782
3    libswiftCore.so                    0x00007f221aa14647 <unavailable> + 3737159
4    libswiftCore.so                    0x00007f221a7e27fd <unavailable> + 1435645
5    libswiftCore.so                    0x00007f221a9e94d8 <unavailable> + 3560664
6    libswiftCore.so                    0x00007f221a7e1c59 <unavailable> + 1432665
7    TensorFlowPackageTests.xctest      0x00007f221b1870cd <unavailable> + 2388173
8    TensorFlowPackageTests.xctest      0x00007f221b186e30 <unavailable> + 2387504
9    TensorFlowPackageTests.xctest      0x00007f221b1870fa <unavailable> + 2388218
10   TensorFlowPackageTests.xctest      0x00007f221b1871a8 <unavailable> + 2388392
11   libswiftCore.so                    0x00007f221a7ce460 Array.withUnsafeBufferPointer<A>(_:) + 38
12   TensorFlowPackageTests.xctest      0x00007f221b18660c <unavailable> + 2385420
13   TensorFlowPackageTests.xctest      0x00007f221b332ecd <unavailable> + 4140749
14   TensorFlowPackageTests.xctest      0x00007f221b345889 <unavailable> + 4216969
15   TensorFlowPackageTests.xctest      0x00007f221b3269cc <unavailable> + 4090316
16   TensorFlowPackageTests.xctest      0x00007f221b345911 <unavailable> + 4217105
17   libXCTest.so                       0x00007f221ae59b91 <unavailable> + 183185
18   libXCTest.so                       0x00007f221ae59bee <unavailable> + 183278
19   libXCTest.so                       0x00007f221ae59b64 <unavailable> + 183140
20   libXCTest.so                       0x00007f221ae4d737 <unavailable> + 132919
21   libXCTest.so                       0x00007f221ae59190 XCTestCase.invokeTest() + 73
22   libXCTest.so                       0x00007f221ae58df0 XCTestCase.perform(_:) + 172
23   libXCTest.so                       0x00007f221ae5cce0 XCTest.run() + 191
24   libXCTest.so                       0x00007f221ae5a910 XCTestSuite.perform(_:) + 232
25   libXCTest.so                       0x00007f221ae5cce0 XCTest.run() + 191
26   libXCTest.so                       0x00007f221ae5a910 XCTestSuite.perform(_:) + 232
27   libXCTest.so                       0x00007f221ae5cce0 XCTest.run() + 191
28   libXCTest.so                       0x00007f221ae57a00 XCTMain(_:) + 1091
29   TensorFlowPackageTests.xctest      0x00007f221b3242e4 <unavailable> + 4080356
30   libc.so.6                          0x00007f220f6f1ab0 __libc_start_main + 231
31   TensorFlowPackageTests.xctest      0x00007f221af9bbda <unavailable> + 375770
Exited with signal code 4
```

But then even in tensorflow, dilation operations other than 1 aren't allowed on CPU {Should i add a precondition / warning for that?), and it only ran when i was on a gpu also, it does not run when you have multiple dilation dimensions. It says "No algorithm worked"
cc: @rxwei . @eaplatanios

Shashi456:
@rxwei I think this can be tested. Dilation factors other than one don't seem to be supported it seems.

eaplatanios:
@Shashi456 Adding preconditions for expected failures sounds like a good idea to me. Could you please add them?

Shashi456:
@eaplatanios would this precondition suffice, if you have suggestions to write it even more properly, I'll fix it asap.

eaplatanios:
I am not actually familiar with the constraints imposed on dilations and so cannot comment on that. Could you explore what happens with the Python API and decide what should be allowed based on that?

Shashi456:
@eaplatanios, i did try exploring the python api, it doesn't really allow dilations other than one.

eaplatanios:
If it does not and you are confident about that then I believe we should add the precondition.

Shashi456:
Is this precondition okay?

Shashi456:
@eaplatanios Just a small doubt, So the keras docs use the same backend we do and what they mention in docs is this `Currently, specifying any dilation_rate value != 1 is incompatible with specifying any stride value != 1.`, Should I experiment with this and check or is the current precondition okay?

eaplatanios:
@Shashi456 Yes it would be good to verify this constraint as it's not a straightforward one and also does not match the current preconditions. We may also be able to enforce it in this case by providing two constructors; one for which you can only provide dilations and one for which you can only provide strides.

Shashi456:
@eaplatanios is there anyway to check the machine type that the code is running on? So the way this goes is on CPU no other dilations than one are supported. 
In the GPU runtime, dilation in the depth dimension must be one.

Shashi456:
Tested and built locally, although it builds, i cant check the test, since I'm building on a cpu. Requesting review on this @eaplatanios @rxwei.

Shashi456:
@rxwei could you trigger a build on this?

Shashi456:
@eaplatanios so the build error is because the dilation rate is 2,2 and as I mentioned, we check for CPU implementation, not the GPU one. Should I update the test accordingly?

saeta:
Hi @Shashi456 Do you think you can fix the merge conflicts here? Thanks! -Brennan

Shashi456:
@saeta, sorry for inactive on the repo for the past few weeks. Just been looking at how else I can contribute here. I'll take care of this soon.

Shashi456:
@saeta, but this doesn't work currently.

marcrasi:
Is this ready for review and merge? What doesn't work about it?

Shashi456:
so the build error is because the dilation rate is 2,2 and as I mentioned, we check for CPU implementation, not the GPU one. Should I update the test accordingly?

@marcrasi

marcrasi:
Oh, I read the whole history of comments and I think I understand now.

I think the best thing to do is to update the test to work on CPU. There's no need to test that 2,2 works on GPU, because we don't actually run CI on GPU right now. (It would be good to do that, but that's a separate issue, and whoever does that could come back and add a 2,2 test for GPU).

dan-zheng:
CI [reproduced](https://source.cloud.google.com/results/invocations/be38238d-c22b-4074-85d1-ec99024637e1/targets/s4tf%2Fapis%2Fubuntu%2Fpresubmit/log) the error on CPU:
```
Test Case 'LayerTests.testConv3DDilation' started at 2019-11-20 18:31:04.899
Fatal error: CPU implementation of Conv3D currently only supports dilated rates of 1.: file /swift-apis/Sources/TensorFlow/Bindings/EagerExecution.swift, line 300
Current stack trace:
0    libswiftCore.so                    0x00007f3be7285cb0 swift_reportError + 50
1    libswiftCore.so                    0x00007f3be72f75f0 _swift_stdlib_reportFatalErrorInFile + 115
2    libswiftCore.so                    0x00007f3be6ff7e1e <unavailable> + 1478174
3    libswiftCore.so                    0x00007f3be6ff7a27 <unavailable> + 1477159
4    libswiftCore.so                    0x00007f3be6ff8008 <unavailable> + 1478664
5    libswiftCore.so                    0x00007f3be6ff62c0 _assertionFailure(_:_:file:line:flags:) + 517
```

asuhan:
Let's not test with dilation > 1, the TF kernels don't support that.

PR #281, opened 2019-06-23
Fixing Documentation errors

Shashi456:
Looks like #275 didn't update a few places of the documentation to reflect dilation.

### DIFF
--- a/Sources/TensorFlow/Layers/Convolutional.swift
+++ b/Sources/TensorFlow/Layers/Convolutional.swift
@@ -32,9 +32,9 @@ public struct Conv1D<Scalar: TensorFlowFloatingPoint>: Layer {
     @noDerivative public let padding: Padding
     /// The dilation factor for the temporal dimension.
     @noDerivative public let dilation: Int
-    
-    /// Creates a `Conv1D` layer with the specified filter, bias, activation function, stride, and
-    /// padding.
+
+    /// Creates a `Conv1D` layer with the specified filter, bias, activation function, stride,
+    /// dilation and padding.
     ///
     /// - Parameters:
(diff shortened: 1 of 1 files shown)


jon-tow:
Hey @Shashi456 isn't it a bit redundant to summarize the layer by listing its initialization parameters?

Shashi456:
@jon-tow I think so too on some level. But what i've discovered is we have documentation consistently for all the extension inits as well. Although i found it somewhat extensive and redundant initially, reading the docs i've discovered that it helps understand different types of initializations for layers.

eaplatanios:
Personally I would rather not list all parameters in the summary. That's why they're document as arguments of the initializer anyway. Listing them is redundant and also makes the documentation harder to maintain, as also indicated by this PR. :)

Shashi456:
@eaplatanios, I think the current norm as far as layers are concerned is to write description for every function. Since the docs generated from the files, have all functions including init. 

We could look at a format where in we avoid redundancy atleast in the parent layer functions by not describing the init since we already have parameter descriptions.

Shashi456:
The error :
```
[3/39] Compiling TensorFlow Recurrent.swift
/swift-apis/Sources/TensorFlow/Layers/Recurrent.swift:267:58: error: value of type 'Cell.State' has no member 'withoutDerivative'
        return self(inputs, initialState: cell.zeroState.withoutDerivative())
```

Shashi456:
Oh I'll have to merge master. Probably. Will do that in a second.

rxwei:
Could you please do a pull from master?

rxwei:
Yep, you beat me to it :)

Shashi456:
Can this be merged?

PR #283, opened 2019-06-23
Update parameter documentation for TransposedConv2D

jon-tow:
Addresses bug [TF-450](https://bugs.swift.org/browse/TF-540).

### DIFF
--- a/Sources/TensorFlow/Layers/Convolutional.swift
+++ b/Sources/TensorFlow/Layers/Convolutional.swift
@@ -419,7 +419,7 @@ public struct TransposedConv2D: Layer {
     /// - Parameters:
     ///   - filter: A 4-D tensor of shape
     ///     `[width, height, input channel count, output channel count]`.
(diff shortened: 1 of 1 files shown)


Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -417,8 +417,8 @@ public struct TransposedConv2D: Layer {
     /// activation function, strides, and padding.
     ///
     /// - Parameters:
-    ///   - filter: The 4-D convolution kernel.
-    ///   - bias: The bias vector.
+    ///   - filter: A 4-D tensor of shape [width, height, inputChannels, outputChannels].
```
rxwei: Identifiers used in this line are not undefined, so I would suggest not avoiding camel case because they are not code. Just use "input channel count" and "output channels count" verbatim.

jon-tow: Sorry about that. I was following the python documentation where they use their underscore convention to describe the shape.

PR #284, opened 2019-06-23
Removed all uses of 'Differentiable.withoutDerivative()'.

eaplatanios:
Friend PR of apple/swift#25691.

cc @rxwei

### DIFF
--- a/Sources/TensorFlow/Layers/Recurrent.swift
+++ b/Sources/TensorFlow/Layers/Recurrent.swift
@@ -264,7 +264,7 @@ public struct RNN<Cell: RNNCell>: Layer {
 
     @differentiable(wrt: (self, inputs))
     public func callAsFunction(_ inputs: [Cell.TimeStepInput]) -> [Cell.TimeStepOutput] {
-        return self(inputs, initialState: cell.zeroState.withoutDerivative())
+        return self(inputs, initialState: Swift.withoutDerivative(at: cell.zeroState))
     }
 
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -264,7 +264,7 @@ public struct RNN<Cell: RNNCell>: Layer {
 
     @differentiable(wrt: (self, inputs))
     public func callAsFunction(_ inputs: [Cell.TimeStepInput]) -> [Cell.TimeStepOutput] {
-        return self(inputs, initialState: cell.zeroState.withoutDerivative())
+        return self(inputs, initialState: Swift.withoutDerivative(at: cell.zeroState))
```
rxwei: You should be able to drop `Swift.` when a new toolchain lands. But let's keep it so that CI won't be broken.

```suggestion
        return self(inputs, initialState: withoutDerivative(at: cell.zeroState))
```

eaplatanios:
This requires the new toolchain anyway because of using `withoutDerivative(at:)` so we might as well remove `Swift.` and merge them both together. What do you think?

eaplatanios:
Actually nvm, to make this smoother I added `in: identity` for the two uses of `Swift.withoutDerivative(at:)`. This will make it pass current CI tests. Then we can update the commit hash in apple/swift, test there, and after that is merged I'll make a small cleanup here.

rxwei:
That sounds like a better option. A new toolchain will be later this afternoon.

PR #286, opened 2019-06-24
Add default argument for conv2D dilations

jon-tow:
Adds a default dilation argument for the conv2D operator. #274


Shashi456:
@rxwei just for consistency, wouldn't it be good if we made all operators have default values like the layers

rxwei:
> @rxwei just for consistency, wouldn't it be good if we made all operators have default values like the layers

Yes, agreed!

PR #289, opened 2019-06-25
Add support for Cropping1D and Cropping2D

jon-tow:
Adding support for Cropping1D and Cropping2D along with tests for correctness. #54 

Currently only supports cropping based on a channels last data format (Keras's default), until control flow differentiation is added.


Shashi456:
@jon-tow Cropping is a convolution operation, would suggest moving it to convolution.swift instead of making a new file.

jon-tow:
Oops. I assumed since it isn't a type of convolution I'd create a separate file. Will stick it in convolution.swift as I see that's how Keras handles it. Thanks for the heads up!

Sources/TensorFlow/Layers/Cropping.swift
```diff
@@ -0,0 +1,86 @@
...
+    @noDerivative public let cropping: (Int, Int)
+
+    /// Creates a cropping layer to trim the temporal dimension.
+    ///
+    /// - Parameter cropping: A tuple of two integers describing how many units should be trimmed
+    ///   off the beginning and end of the cropping dimension.
+    public init(cropping: (Int, Int) = (1, 1)) {
+        self.cropping = cropping
+    }
+
```
Shashi456: @jon-tow for consistency of api, I think you should add symmetric initialization for the 1D layer as well.

jon-tow: Yup. I wasn't sure about it but I'll add it in!

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -641,3 +641,84 @@ public extension DepthwiseConv2D {
...
+    }
+}
+
+/// A layer for cropping tensors along spatial dimensions, e.g. image cropping.
+///
+/// `Cropping2D` can trim an input at the top, bottom, left, and right sides of
+/// its spatial dimensions.
+public struct Cropping2D<Scalar: TensorFlowFloatingPoint>: Layer {
+    /// The cropping dimensions along the height and width axes.
+    @noDerivative public let cropping: ((Int, Int), (Int, Int))
```
rxwei: We should compensate for weak type information. In this case, a property name that repeats the struct name and a nested tuple of integers do not meaningfully tell us anything about what the value represents, so the user always needs to go to the documentation.

Instead, I would suggest listing out the exact dimensions as _separate properties_ for clarity. Each property should have its own API documentation.

```suggestion
    @noDerivative public let top: Int
    @noDerivative public let bottom: Int
    @noDerivative public let left: Int
    @noDerivative public let right: Int
```

eaplatanios:
@Shashi456 Why do you believe that cropping is a convolution operation? I would just put it in a new file called `Manipulation.swift` or `Image.swift`. `Manipulation.swift` could also contains more tensor manipulation operations (e.g., reshape, tiling, ...).

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -641,3 +641,84 @@ public extension DepthwiseConv2D {
             padding: padding)
     }
 }
+
+/// A layer for cropping tensors along the temporal dimension.
+public struct Cropping1D<Scalar: TensorFlowFloatingPoint>: Layer {
+    /// The cropping dimensions along the temporal axis.
+    @noDerivative public let cropping: (Int, Int)
```
rxwei: Same here.

Shashi456:
@eaplatanios, just mostly following the keras api structure for now.

rxwei:
> @Shashi456 Why do you believe that cropping is a convolution operation? I would just put it in a new file called `Manipulation.swift` or `Image.swift`. `Manipulation.swift` could also contains more tensor manipulation operations (e.g., reshape, tiling, ...).

I agree with Anthony here. `Image.swift` sounds like a good name.

Shashi456:
Can we have a separate PR for refactoring the layers into image.swift? And pass this through?

rxwei:
> Can we have a separate PR for refactoring the layers into image.swift? And pass this through?

This PR introduces a purely additive change, so it would make more sense to place `Cropping1D` and `Cropping2D` directly in `Image.swift` in this PR.

Shashi456:
@eaplatanios, do mention other layers you'd want to be a part of Image.swift, I just personally think that we should avoid over-granularity of files by separating them so much. Image.swift does make sense in this case though.

jon-tow:
@rxwei I'll add this into `Image.swift` with the changes you suggested. Should the `Image.swift` stay in the `Layers` directory?

rxwei:
> Should the `Image.swift` stay in the `Layers` directory?

Yes. Thank you!

Sources/TensorFlow/Layers/Image.swift
```diff
@@ -0,0 +1,100 @@
...
+/// A layer for cropping tensors along the temporal dimension.
+public struct Cropping1D<Scalar: TensorFlowFloatingPoint>: Layer {
+    /// The number of units to crop from the beginning of an input's temporal dimension.
+    @noDerivative public let begin: Int
+    /// The number of units to crop from the end of an input's temporal dimension.
+    @noDerivative public let end: Int
+
+    /// Creates a cropping layer to trim the temporal dimension.
+    ///
+    /// - Parameter
```
rxwei:
```suggestion
    /// - Parameters:
```

eaplatanios:
@Shashi456 I haven't given this much thought other than the comment above, but there is lots of operations that are typically performed on images so I wouldn't this is granular. On the contrary, if we find that `Convolutional.swift` ends up being small we could potentially eventually move the convolution operations in `Image.swift`, as that would probably make more sense in terms of a classification hierarchy.

Sources/TensorFlow/Layers/Image.swift
```diff
@@ -0,0 +1,100 @@
...
+        self.bottom = bottom
+        self.left = left
+        self.right = right
+    }
+
+    /// Creates a cropping layer to trim spatial dimensions, i.e. height and width.
+    ///
+    /// - Parameters
+    ///   - symmetricHeight: The number of units to trim off the top and bottom of an input's height
+    ///   dimension.
```
eaplatanios:
```suggestion
    ///     dimension.
```

Sources/TensorFlow/Layers/Image.swift
```diff
@@ -0,0 +1,100 @@
...
+    ///   - symmetricWidth: The number of units to trim off the left and right of an input's width
+    ///   dimension.
```
eaplatanios:
```suggestion
    ///     dimension.
```

Sources/TensorFlow/Layers/Image.swift
```diff
@@ -0,0 +1,100 @@
...
+    public init(top: Int = 1, bottom: Int = 1, left: Int = 1, right: Int = 1 ) {
+        self.top = top
+        self.bottom = bottom
+        self.left = left
+        self.right = right
+    }
+
+    /// Creates a cropping layer to trim spatial dimensions, i.e. height and width.
+    ///
+    /// - Parameters
```
rxwei:
```suggestion
    /// - Parameters:
```

jon-tow:
I confirm! @googlebot

eaplatanios:
@jon-tow Could you please merge with master?

jon-tow:
@eaplatanios Sorry for the late merge, was a bit busy.

jon-tow:
@eaplatanios This should be good to go! Please let me know if you'd like me to add anything.

eaplatanios:
@rxwei I started tests, but could you please handle the CLA issue?

Shashi456:
@jon-tow if you handle the conflicts, I think this PR might be good to go as well.

jon-tow:
@Shashi456 Thanks for the alert! Forgot about this one :)

saeta:
Hi @jon-tow Thank you very much for this PR, but I think that cropping-as-a-layer is relatively unused, and might be better served as a stateless function on Tensor instead. As a result, I'm going to close out this PR, but if you would like to discuss this further, let's definitely dedicate some time at an upcoming design meeting!

All the best,
-Brennan

PR #290, opened 2019-06-25
Add default implementations of new requirements in TensorArrayProtocol

bgogul:
Note that these implementations access `_cTensorHandle` and therefore, will trigger
materialization in the case of lazy tensor. Therefore, if the user needs to defer materialization, they will need to manually implement these methods. 

Note that this is a stop-gap solution while we continue to work on the DerivedConformances for TensorArrayProtocol.

### DIFF
--- a/Sources/TensorFlow/Core/TensorGroup.swift
+++ b/Sources/TensorFlow/Core/TensorGroup.swift
@@ -42,15 +42,16 @@ public extension TensorArrayProtocol {
     init<C: RandomAccessCollection>(_handles: C) where C.Element: _AnyTensorHandle {
         let status = TF_NewStatus()
         defer { TF_DeleteStatus(status) }
-        let buffer = UnsafeMutablePointer<CTensorHandle>.allocate(capacity: _handles.count)
+        let buffer = UnsafeMutableBufferPointer<CTensorHandle>.allocate(capacity: _handles.count)
         defer { buffer.deallocate() }
-        for handle in _handles {
+        for (i, handle) in _handles.enumerated() {
             // Increment the reference count in TF.
             let handleCopy = TFE_TensorHandleCopySharingTensor(handle._cTensorHandle, status)
             checkOk(status)
-            buffer.initialize(to: handleCopy!)
+            buffer[i] = handleCopy!
         }
-        self.init(_owning: buffer, count: _handles.count)
+        let baseAddress = UnsafeMutablePointer<OpaquePointer>(buffer.baseAddress)
+        self.init(_owning: baseAddress, count: _handles.count)
     }
 
     var _tensorHandles: [_AnyTensorHandle] {
@@ -59,11 +60,9 @@ public extension TensorArrayProtocol {
         let count = Int(_tensorHandleCount)
         var buffer = UnsafeMutableBufferPointer<CTensorHandle>.allocate(capacity: count)
         defer { buffer.deallocate() }
-        let pointer = UnsafeMutablePointer<OpaquePointer?>(buffer.baseAddress)
         self._unpackTensorHandles(into: buffer.baseAddress)
         let result: [TFETensorHandle] = (0..<count).map {
-            let current = pointer!.advanced(by: $0)
-            let cTensorHandle = current.pointee
+            let cTensorHandle = buffer[$0]
             // Increment the reference count in TF.
             let handleCopy = TFE_TensorHandleCopySharingTensor(cTensorHandle, status)
             checkOk(status)
--- a/Sources/TensorFlow/Core/TensorGroup.swift
+++ b/Sources/TensorFlow/Core/TensorGroup.swift
@@ -38,6 +38,42 @@ public protocol TensorArrayProtocol {
     init<C: RandomAccessCollection>(_handles: C) where C.Element: _AnyTensorHandle
 }
 
+extension TensorArrayProtocol {
+    init<C: RandomAccessCollection>(_handles: C) where C.Element: _AnyTensorHandle {
+        let status = TF_NewStatus()
+        defer { TF_DeleteStatus(status) }
+        let buffer: UnsafeMutablePointer<CTensorHandle> =
+            UnsafeMutablePointer.allocate(capacity: _handles.count)
+        defer { buffer.deallocate() }
+        for handle in _handles {
+            // Increment the reference count in TF.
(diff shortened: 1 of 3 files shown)


Sources/TensorFlow/Core/TensorGroup.swift
```diff
@@ -38,6 +38,42 @@ public protocol TensorArrayProtocol {
...
+        for handle in _handles {
+            // Increment the reference count in TF.
+            let handleCopy = TFE_TensorHandleCopySharingTensor(handle._cTensorHandle, status)
+            checkOk(status)
+            buffer.initialize(to: handleCopy!)
+        }
+        self.init(_owning: buffer, count: _handles.count)
+    }
+
+    var _tensorHandles : [_AnyTensorHandle] {
```
rxwei:
```suggestion
    var _tensorHandles: [_AnyTensorHandle] {
```

Sources/TensorFlow/Core/TensorGroup.swift
```diff
@@ -38,6 +38,42 @@ public protocol TensorArrayProtocol {
     init<C: RandomAccessCollection>(_handles: C) where C.Element: _AnyTensorHandle
 }
 
+extension TensorArrayProtocol {
```
rxwei:
```suggestion
public extension TensorArrayProtocol {
```

Sources/TensorFlow/Core/TensorGroup.swift
```diff
@@ -38,6 +38,42 @@ public protocol TensorArrayProtocol {
...
+    init<C: RandomAccessCollection>(_handles: C) where C.Element: _AnyTensorHandle {
+        let status = TF_NewStatus()
+        defer { TF_DeleteStatus(status) }
+        let buffer: UnsafeMutablePointer<CTensorHandle> =
+            UnsafeMutablePointer.allocate(capacity: _handles.count)
```
rxwei:
```suggestion
            UnsafeMutablePointer<CTensorHandle>.allocate(capacity: _handles.count)
```
This can be moved to the end of the previous line once the type signature gets removed.

Sources/TensorFlow/Core/TensorGroup.swift
```diff
@@ -38,6 +38,42 @@ public protocol TensorArrayProtocol {
     init<C: RandomAccessCollection>(_handles: C) where C.Element: _AnyTensorHandle
 }
 
+extension TensorArrayProtocol {
+    init<C: RandomAccessCollection>(_handles: C) where C.Element: _AnyTensorHandle {
+        let status = TF_NewStatus()
+        defer { TF_DeleteStatus(status) }
+        let buffer: UnsafeMutablePointer<CTensorHandle> =
```
rxwei:
```suggestion
        let buffer =
```

rxwei:
Interesting. What led to the decision that we don't need derived conformances?

pschuh:
We still need a derived conformance because this will materialize lazy-tensor for custom TensorGroup code, but it isn't strictly necessary at this time. Submitting this now just makes it so that this is no longer broken.

eaplatanios:
If no one is working on this now, I can look into the derives conformances
and get them ready by the end of the week. It shouldn’t require too many
changes from my last implementation.

bgogul:
> If no one is working on this now, I can look into the derives conformances and get them ready by the end of the week. It shouldn’t require too many changes from my last implementation.
> […](#)

I am working on the deriving `init(_handles: C)`.  If you can work on deriving `var _tensorHandles: [AnyTensorHandle]` that would be super useful. 
 
For `init(_handles: C)`, I started based off your implementation. However, it  got a bit complicated due to generics. :( To make things simple, I am adding some helper structs so that it is much more straightforward to synthesize the body of the function.

pschuh:
For both these conformances, I think it would be best if we had helper constructs that were much simpler to work with inside the derived conformance:

Those would be:
For init(_handles: ...) there should be a:
protocol TensorHandleSource {
  func next() -> AnyTensorHandle
  func hasNext() -> Bool
}
This can be implemented by the collection. (Note that this is not an iterator in order to support TensorArrayProtocol).

Then,
The deriving become easy:
init(... source) {
  field1.init(source)
  field2.init(source)
...
}
and,
func collectAllTensorHandles(collect: (AnyTensorHandle) -> ()) {
  field1.collectAllTensorHandles(collect)
  field2.collectAllTensorHandles(collect)
  field3.collectAllTensorHandles(collect)
}

All of these "sugar" versions should be implemented in terms of this.

bgogul:
> For both these conformances, I think it would be best if we had helper constructs that were much simpler to work with inside the derived conformance:
> 
> Those would be:
> For init(_handles: ...) there should be a:
> protocol TensorHandleSource {
> func next() -> AnyTensorHandle
> func hasNext() -> Bool
> }
> This can be implemented by the collection. (Note that this is not an iterator in order to support TensorArrayProtocol).
> 
> Then,
> The deriving become easy:
> init(... source) {
> field1.init(source)
> field2.init(source)
> ...
> }
> and,
> func collectAllTensorHandles(collect: (AnyTensorHandle) -> ()) {
> field1.collectAllTensorHandles(collect)
> field2.collectAllTensorHandles(collect)
> field3.collectAllTensorHandles(collect)
> }
> 
> All of these "sugar" versions should be implemented in terms of this.

Yes, this is the direction I am going with `init` based off of our discussions.

bgogul:
> I am working on the deriving `init(_handles: C)`. If you can work on deriving `var _tensorHandles: [AnyTensorHandle]` that would be super useful.
> 
> For `init(_handles: C)`, I started based off your implementation. However, it got a bit complicated due to generics. :( To make things simple, I am adding some helper structs so that it is much more straightforward to synthesize the body of the function.

@eaplatanios, hold off working on this. As we are unblocked for the release with this PR, we have been having discussions on whether we should change the requirements as @pschuh mentioned above. That will change how we derive the conformance.

eaplatanios:
Sounds good! I like the approach @pschuh proposed. Let me know if I can
help in any way. In the meantime, I’ll  just wait for updates. :)

rxwei:
When this PR lands, [this model](https://github.com/tensorflow/swift-models/blob/b9920aeebd3816732a5dfbe0b1a7b9d6e8d8b708/CIFAR/Data.swift#L38) can be simplified.

bgogul:
We should probably leave this in the model until we get the derived
conformances. Otherwise, I cannot experiment with lazy tensor.

PR #295, opened 2019-06-25
Eliminate 'assert(false)'

### LINKED ISSUE #291: Remove uses of assert(false)
`assert(false)` should not be used as a terminator because it gets compiled away in release builds. `fatalError` or `preconditionFailure` should be used instead. There are two occurrences of `assert(false)` and they have broken release builds of swift-apis, both in LazyTensorOperations.swift.

1. 
```swift
            assert(false, "TODO: to be send out in a separate PR.")
            // return op.materialized(index: index)
```

2. 
```swift
        default: assert(false, "Unhandled type: \(cDataType)")
```

rxwei:
Remove more `assert(false)`. This is blocking release builds.

This is a follow-up to #292, and fixes #291.

### DIFF
--- a/Sources/TensorFlow/Core/LazyTensorTFFunctionBuilder.swift
+++ b/Sources/TensorFlow/Core/LazyTensorTFFunctionBuilder.swift
@@ -163,7 +163,7 @@ class TFGraph {
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


bgogul:
Thanks, @rxwei

PR #298, opened 2019-06-26
Add Function/Lambda Layer 

Shashi456:
Working on this currently, IMO a major functionality addition to the Layer API we currently have. This would allow for any differentiable function to be wrapped around and added as a layer in a model.

Also avoids API redundancies for layers like reshape, padded etc. 

I'm currently getting the error that the layer doesn't conform to protocol of Layer and that a call function is needed. As far as i understand, for a structure to inherit a protocol, you need to extend and define all the functions in the protocol, something like abstract classes theoretically. Any thoughts on where i might be going or doing it wrong?

Refer to discussion in #54 [here](https://github.com/tensorflow/swift-apis/issues/54#issuecomment-502435415)

### DIFF
--- a/Sources/TensorFlow/Layers/Core.swift
+++ b/Sources/TensorFlow/Layers/Core.swift
@@ -225,19 +225,15 @@ public extension Dense {
 }
 
 
-public struct Function<InputScalar: TensorFlowFloatingPoint, OutputScalar:TensorFlowFloatingPoint>: Layer {
+public struct Function<Input: Differentiable, Output: Differentiable>: Layer {
     /// The Function Layer arbitrarily takes any differentiable function and returns a Layer which can
     /// used as a wrapper around that function.
-    public typealias Input = Tensor<InputScalar>
-
-    public typealias Output = Tensor<OutputScalar>
 
     public typealias Body = @differentiable (Input) -> Output
 
     @noDerivative public let body: Body
 
-    public init(
-        body: @escaping Body) {
+    public init(_ body: @escaping Body) {
         self.body = body
     }
 
--- a/Sources/TensorFlow/Layers/Core.swift
+++ b/Sources/TensorFlow/Layers/Core.swift
@@ -224,7 +224,7 @@ public extension Dense {
     }
 }
 
-/// /// A layer that encloses a custom differentiable function.
+/// A layer that encloses a custom differentiable function.
 public struct Function<Input: Differentiable, Output: Differentiable>: Layer {
     public typealias Body = @differentiable (Input) -> Output
 
--- a/Sources/TensorFlow/Layers/Core.swift
+++ b/Sources/TensorFlow/Layers/Core.swift
@@ -224,10 +224,8 @@ public extension Dense {
     }
 }
 
-/// A `Function` layer arbitrarily stores any differentiable function and returns its
-/// `callAsFunction(_:)` method which can used as a wrapper around that function.
+/// /// A layer that encloses a custom differentiable function.
 public struct Function<Input: Differentiable, Output: Differentiable>: Layer {
-
     public typealias Body = @differentiable (Input) -> Output
 
     @noDerivative public let body: Body
--- a/Sources/TensorFlow/Layers/Core.swift
+++ b/Sources/TensorFlow/Layers/Core.swift
@@ -223,3 +223,26 @@ public extension Dense {
                   activation: activation)
     }
 }
+
+
+public struct Function<InputScalar: TensorFlowFloatingPoint, OutputScalar:TensorFlowFloatingPoint>: Layer {
+    /// The Function Layer arbitrarily takes any differentiable function and returns a Layer which can
+    /// used as a wrapper around that function.
+    public typealias Input = Tensor<InputScalar>
+
+    public typealias Output = Tensor<OutputScalar>
+
+    public typealias Body = @differentiable (Input) -> Output
+
+    @noDerivative public let body: Body
+
+    public init(
+        body: @escaping Body) {
+        self.body = body
+    }
+
+    @differentiable
+    public func callAsFunction(_ input: Input) -> Output {
+        return body(input)
+    }
+}
--- a/Sources/TensorFlow/Layers/Core.swift
+++ b/Sources/TensorFlow/Layers/Core.swift
@@ -224,10 +224,9 @@ public extension Dense {
     }
 }
 
-
+/// A `Function` layer arbitrarily stores any differentiable function and returns its
+/// `callAsFunction(_:)` method which can used as a wrapper around that function.
 public struct Function<Input: Differentiable, Output: Differentiable>: Layer {
-    /// The Function Layer arbitrarily takes any differentiable function and returns a Layer which can
-    /// used as a wrapper around that function.
 
     public typealias Body = @differentiable (Input) -> Output
 
--- a/Tests/TensorFlowTests/LayerTests.swift
+++ b/Tests/TensorFlowTests/LayerTests.swift
@@ -250,7 +250,7 @@ final class LayerTests: XCTestCase {
 
 
     func testFunction() {
-        let tanhLayer = Function<Float, Float>(tanh)
+        let tanhLayer = Function<Tensor<Float>, Tensor<Float>>(tanh)
         let input = Tensor(shape:[5, 1], scalars:(0..<5).map(Float.init))
         let output = tanhLayer.inferring(from: input)
         let expected = Tensor<Float>([[0.0], [0.7615942], [0.9640276], [0.9950547], [0.9993292]])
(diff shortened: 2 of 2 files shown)


jon-tow:
Which toolchain are you using? Seems to compile fine on the June 17 nightly, thought I haven't tested it.

Sources/TensorFlow/Layers/Core.swift
```diff
@@ -223,3 +223,26 @@ public extension Dense {
                   activation: activation)
     }
 }
+
+
+public struct Function<InputScalar: TensorFlowFloatingPoint, OutputScalar:TensorFlowFloatingPoint>: Layer {
```
dan-zheng: Please generalize this layer to wrap a `<T: Differentiable, U: Differentiable> @differentiable (T) -> U` function instead of wrapping a `@differentiable (Tensor<T>) -> Tensor<U>` function.

```suggestion
public struct Function<Input: Differentiable, Output: Differentiable>: Layer {
```

Sources/TensorFlow/Layers/Core.swift
```diff
@@ -223,3 +223,26 @@ public extension Dense {
...
+    /// used as a wrapper around that function.
+    public typealias Input = Tensor<InputScalar>
+
+    public typealias Output = Tensor<OutputScalar>
+
+    public typealias Body = @differentiable (Input) -> Output
+
+    @noDerivative public let body: Body
+
+    public init(
```
dan-zheng: Make initializer label-less and merge line from below.

```suggestion
    public init(_ body: @escaping Body) {
```

A Precedent for `init(_ body:)` is [`AnyIterator.init`](https://developer.apple.com/documentation/swift/anyiterator/1784409-init).

Shashi456:
@dan-zheng Yeah on it.

Sources/TensorFlow/Layers/Core.swift
```diff
@@ -223,3 +223,22 @@ public extension Dense {
                   activation: activation)
     }
 }
+
+
+public struct Function<Input: Differentiable, Output: Differentiable>: Layer {
+    /// The Function Layer arbitrarily takes any differentiable function and returns a Layer which can
```
rxwei: This comment does not accurately describe the layer in that
- A layer does not _take_ a differentiable function, it _stores_ a differentiable function.
- A layer does not _return_ any layer. What returns is its `callAsFunction(_:)` method.

Also, the function layer is not a proper name, and thus should not be spelled as "Function Layer". In API documentation, it should be spelled as "The `Function` layer" with the word "layer" starting with a lower case `l`.

Moreover, this is an API summary. It should be placed on top of the declaration of the `Function` type, not within.

Sources/TensorFlow/Layers/Core.swift
```diff
@@ -223,3 +223,22 @@ public extension Dense {
                   activation: activation)
     }
 }
+
+
```
rxwei: Remove extra line.

Sources/TensorFlow/Layers/Core.swift
Shashi456: @rxwei Yes, I was going to ask some comments on the documentation. I am just getting the code right.

Shashi456:
Alright @rxwei , @dan-zheng I'd like to ask a few things:
- Are there any layers that we need to remove which are just wrapper functions currently?
- Could I write a tutorial about this functionality, and if yes how could i go about doing that?

Also the build and tests pass locally, I'm just getting the documentation right now.

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -331,6 +340,7 @@ final class LayerTests: XCTestCase {
         ("testAvgPool1D", testAvgPool1D),
         ("testAvgPool2D", testAvgPool2D),
         ("testAvgPool3D", testAvgPool3D),
+        ("testFunction", testFunction),
```
rxwei: Please move this to the end of the list.

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -248,6 +248,15 @@ final class LayerTests: XCTestCase {
         XCTAssertEqual(output.shape, expected)
     }
 
+
+    func testFunction() {
```
rxwei: Could you please make this be the last test in this file? We expect the ordering of tests in `allTests` to reflect exactly the ordering of test case function declarations.

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -248,6 +248,15 @@ final class LayerTests: XCTestCase {
         XCTAssertEqual(output.shape, expected)
     }
 
+
```
rxwei: Please remove the extra line.

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -248,6 +248,15 @@ final class LayerTests: XCTestCase {
...
+    func testFunction() {
+        let tanhLayer = Function<Tensor<Float>, Tensor<Float>>(tanh)
+        let input = Tensor(shape:[5, 1], scalars:(0..<5).map(Float.init))
```
rxwei: Add a space after each colon.
```suggestion
        let input = Tensor(shape: [5, 1], scalars: (0..<5).map(Float.init))
```

Tests/TensorFlowTests/LayerTests.swift
Shashi456: @rxwei  I thought we were grouping the tests in a file together, Since i added this to the Core.swift file, i thought it would belong to that class of tests, am I wrong?

rxwei:
> * Are there any layers that we need to remove which are just wrapper functions currently?

We don't need to remove anything at the moment. But now that I think about it, I think that a protocol for function layers could be interesting. Any conforming type can get an automatically-implemented `callAsFunction(_:)` method.

```swift
public protocol FunctionLayer: Layer {
    @noDerivative
    var function: @differentiable (Input) -> Output { get }
}

public extension FunctionLayer {
    func callAsFunction(_ input: Input) -> Output {
        function(input)
    }
}
```

With this, you can redefine a `Tanh` layer like this:

```swift
struct Tanh<Scalar: TensorFlowFloatingPoint>: FunctionLayer {
    @noDerivative var function = tanh
}
```

But yeah, it is still a wrapper, not the best interface. It would be great if Swift supported function conforming to protocols, in which way all differentiable functions can conform to `Layer`. This is one of the directions we can look into later with the Swift community.

> * Could I write a tutorial about this functionality, and if yes how could i go about doing that?

It is great that you are looking to write a tutorial about it! However, as you've seen in the test case, this functionality is a bit cumbersome in that it requires the user to spell out the `Input` and `Output` types. I'd suggest holding off on this until we figure out a better way to make this functionality available, without the cost of requiring explicit generic arguments.

Shashi456:
@rxwei So we discussed that layers like zero padding are just wrapper functions, So I'll mark them as done in the list with this functionality?

rxwei:
I put some more thoughts to this. I believe that it is still valuable to have layers `ZeroPadding1D`, `ZeroPadding2D` and `ZeroPadding3D` as `struct`s defined in the library for API completeness (completing the layer list you presented!). After all layers have been ported, we can think about what to do in the long-term for reducing the boilerplate in these layer definitions.

Sources/TensorFlow/Layers/Core.swift
```diff
@@ -223,3 +223,26 @@ public extension Dense {
...
+    public typealias Input = Tensor<InputScalar>
+
+    public typealias Output = Tensor<OutputScalar>
+
+    public typealias Body = @differentiable (Input) -> Output
+
+    @noDerivative public let body: Body
+
+    public init(
+        body: @escaping Body) {
```
rxwei: Could you please move this line to the end of the previous line?

Shashi456: @rxwei I've already done this xD

Sources/TensorFlow/Layers/Core.swift
```diff
@@ -223,3 +223,21 @@ public extension Dense {
                   activation: activation)
     }
 }
+
+/// A `Function` layer arbitrarily stores any differentiable function and returns its
+/// `callAsFunction(_:)` method which can used as a wrapper around that function.
+public struct Function<Input: Differentiable, Output: Differentiable>: Layer {
+
```
rxwei: Remove extra line.

Sources/TensorFlow/Layers/Core.swift
```diff
@@ -223,3 +223,21 @@ public extension Dense {
                   activation: activation)
     }
 }
+
+/// A `Function` layer arbitrarily stores any differentiable function and returns its
```
rxwei: This comment can be improved for accuracy. The function layer does _not_ "return its `callAsFunction(_:)` method".

I would suggest:
```
/// A layer that encloses a custom differentiable function.
```

Shashi456: Oh that sounds very neat, I shall update it.

PR #299, opened 2019-06-26
Adds a mechanism to manually promote constants to inputs in Lazy Tensor.

bgogul:
Adds a `_concreteInputLazyTensor` property to the tensor-related types in the library. This provides **temporary** mechanism to promote constants to inputs manually in trace. Here is an example use case:
```
import TensorFlow

let a = Tensor<Float>(10.0)
let b = Tensor<Float>(5.0)
let w = a._concreteInputLazyTensor * b
print ("\(w)")
```

This will result in the following trace for `w`:
```
lazyTrace_3(%0: float) -> (%2) {
  %1 = Const[dtype: float, value: 5.0]()
  %2 = Mul[T: float](%0, %1)
} 
```
Note that the slot for `a` got promoted to an input. This is added to experiment with the lazy tensor implementation and is mostly intended to be used within the `TensorFlow` module.  If this turns out to be useful, we should provide a cleaner user-visible API.

This PR also adds a helper file for tests related to lazy tensor.

### DIFF
--- a/Sources/TensorFlow/Core/LazyTensorOperation.swift
+++ b/Sources/TensorFlow/Core/LazyTensorOperation.swift
@@ -272,6 +272,60 @@ class LazyTensorOperation: TensorOperation {
     }
 }
 
+extension _AnyTensorHandle {
+    /// Wraps the `_tfeTensorHandle` as a `LazyTensor`.
+    var lazyTensor: LazyTensor { LazyTensor(self._tfeTensorHandle) }
+
+    /// Wraps the `_tfeTensorHandle` in a `LazyTensor` that will
+    /// be promoted to an input when used in a extracted trace.
+    var inputLazyTensor: LazyTensor { LazyTensor(_materialized: self._tfeTensorHandle) }
+}
+
+extension TensorHandle {
+    /// Wraps this handle in a `LazyTensor`.
+    public var lazyTensor: TensorHandle { TensorHandle(handle: handle.lazyTensor) }
+
+    /// Wraps this handle in a `LazyTensor` that will be promoted
+    /// to an input when used in a extracted trace.
+    public var inputLazyTensor: TensorHandle { TensorHandle(handle: handle.inputLazyTensor) }
+}
+
+extension Tensor {
+    /// Wraps this tensor in a `LazyTensor`.
+    public var lazyTensor: Tensor { Tensor(handle: handle.lazyTensor) }
+
+    /// Wraps this tensor in a `LazyTensor` that will be promoted
+    /// to an input when used in a extracted trace.
+    public var inputLazyTensor: Tensor { Tensor(handle: handle.inputLazyTensor) }
+}
+
+extension StringTensor {
+    /// Wraps this tensor in a `LazyTensor`.
+    public var lazyTensor: StringTensor { StringTensor(handle: handle.lazyTensor) }
+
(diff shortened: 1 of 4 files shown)


saeta:
+1 for needing a motivating example. Also, could we start with this being `_`-prefixed? I'm making a few guesses, but we might want to express these concepts in a slightly different way. I'd like us to carefully think through the API design space here. Does that make sense? Thanks!

Sources/TensorFlow/Core/LazyTensorOperation.swift
```diff
@@ -272,6 +272,77 @@ class LazyTensorOperation: TensorOperation {
...
+            return nil
+        }
+    }
+    var _concreteLazyTensor: LazyTensor { LazyTensor(self._tfeTensorHandle) }
+    var _concreteInputLazyTensor: LazyTensor { LazyTensor(_materialized: self._tfeTensorHandle) }
+}
+
+extension TensorHandle: _LazyTensorCompatible {
+    var _lazyTensor: LazyTensor? { handle._lazyTensor }
+    public var _concreteLazyTensor: TensorHandle { TensorHandle(handle: handle._concreteLazyTensor) }
```
rxwei: The protocol is internal. Why does this requirement need to be public? Same for other occurrences of `public`.

bgogul:
@pschuh @saeta @rxwei I scaled back this PR to publicly expose only what I needed. I also underscored it. 

I moved the `_LazyTensorCompatible` protocol to a test helper.I don't think it is useful outside of the tests  for the time being. If we have a need, we can reassess and design the API properly.

bgogul:
> Constants in the example don't match up.

Thanks for flagging it. Fixed now.

Sources/TensorFlow/Core/LazyTensorOperation.swift
```diff
@@ -130,6 +130,57 @@ class LazyTensor: _AnyTensorHandle {
...
+    /// traces. (Note that this may trigger materialization.)
+    var _concreteInputLazyTensor: LazyTensor {
+        LazyTensor(_materialized: self._tfeTensorHandle)
+    }
+}
+
+extension TensorHandle {
+    /// Returns `Self` that wraps `_concreteInputLazyTensor` of the underlying
+    /// `_AnyTensorHandle`
+    public var _concreteInputLazyTensor: TensorHandle {
```
rxwei: Hmm is there a reason this needs to be public?

bgogul: I expect that it will only be used within the library. However, I wanted it to be available for experimentation in the user models if needed.

rxwei: How and why should a user be expected to use this?

rxwei: Just wanted to be careful exposing public APIs. I think we need to make sure public APIs have at least some justification of use cases before they are exposed.

bgogul: The use case is the one in description where we want to force a constant to be an input to an extracted trace. Ultimately, we may not need this API. However, in the meantime,  I want to have this knob while we develop and improve the lazy tensor implementation.

rxwei: Oh I see. Sorry for missing parts of the PR description! And thanks for explaining :)

PR #302, opened 2019-06-27
Add Adadelta Optimizer

dash2507:


Sources/TensorFlow/Optimizer.swift
```diff
@@ -355,3 +355,91 @@ public class AdaGrad<Model: Layer>: Optimizer
...
+            accDeltaNew += (1 - Double(rho)) * (stepSize * stepSize)
+            accDelta[keyPath: kp] =  accDeltaNew
+        }
+    }
+
+    public func update(_ model: inout Model,
+                       along direction: Model.TangentVector) {
+        update(&model.allDifferentiableVariables, along: direction)
+    }
+}
```
Shashi456: Is this addition here by mistake?

Sources/TensorFlow/Optimizer.swift
```diff
@@ -355,3 +355,91 @@ public class AdaGrad<Model: Layer>: Optimizer
         update(&model.allDifferentiableVariables, along: direction)
     }
 }
+
```
Shashi456: You need to add comments for the optimizer itself and a link to the paper that this is introduced in.

Sources/TensorFlow/Optimizer.swift
```diff
@@ -355,3 +355,91 @@ public class AdaGrad<Model: Layer>: Optimizer
...
+    public var rho: Float
+    /// A small scalar added to the denominator to improve numerical stability.
+    public var epsilon: Float
+    /// The learning rate decay.
+    public var decay: Float
+    /// The current step.
+    public var step: Int = 0
+    public var accumulators: Model.AllDifferentiableVariables
+    public var accDelta: Model.AllDifferentiableVariables
+
```
rxwei: Remove extra line.

Sources/TensorFlow/Optimizer.swift
```diff
@@ -355,3 +355,91 @@ public class AdaGrad<Model: Layer>: Optimizer
...
+    public var learningRate: Float
+    ///decay factor, corresponding to fraction of gradient to keep at each time step.
+    public var rho: Float
+    /// A small scalar added to the denominator to improve numerical stability.
+    public var epsilon: Float
+    /// The learning rate decay.
+    public var decay: Float
+    /// The current step.
+    public var step: Int = 0
+    public var accumulators: Model.AllDifferentiableVariables
```
rxwei: Could you add a doc comment that explains this property?

Sources/TensorFlow/Optimizer.swift
```diff
@@ -355,3 +355,91 @@ public class AdaGrad<Model: Layer>: Optimizer
...
+    public var accDelta: Model.AllDifferentiableVariables
```
rxwei: - Could you add a doc comment that explains this property?
- Avoid abbreviations in this property name since `accDelta` is not a word of art and may be confusing.

Sources/TensorFlow/Optimizer.swift
rxwei: Please leave a trailing empty line in the file.

Sources/TensorFlow/Optimizer.swift
```diff
@@ -355,3 +355,102 @@ public class AdaGrad<Model: Layer>: Optimizer
...
+    public var learningRate: Float
+    ///decay factor, corresponding to fraction of gradient to keep at each time step.
+    public var rho: Float
+    /// A small scalar added to the denominator to improve numerical stability.
+    public var epsilon: Float
+    /// The learning rate decay.
+    public var decay: Float
+    /// The current step.
+    public var step: Int = 0
+    /// Accumulate Gradients
```
rxwei: Two suggestions on documentation comments:
- Properties should be described with a noun phrase ending with a period.
- Documentation comments for mathematical variables should use the term of art. In this case, we should switch to the accurate, proper terminology that's used by the paper.
```suggestion
    /// The accumulated, exponentially decaying average of squared gradients.
```

Sources/TensorFlow/Optimizer.swift
```diff
@@ -355,3 +355,102 @@ public class AdaGrad<Model: Layer>: Optimizer
...
+    public var accumulators: Model.AllDifferentiableVariables
```
rxwei:
```suggestion
    public var averageSquared: Model.TangentVector
```

Sources/TensorFlow/Optimizer.swift
```diff
@@ -355,3 +355,102 @@ public class AdaGrad<Model: Layer>: Optimizer
...
+    /// Accumulate Updates(here StepSizes)
+    public var updates: Model.AllDifferentiableVariables
```
rxwei:
```suggestion
    public var accumulatedDelta: Model.AllDifferentiableVariables
```

Sources/TensorFlow/Optimizer.swift
rxwei: Although Keras uses the term `accumulator`, it is not very descriptive and does not connect with the paper very well. I'd suggest using the exact term mentioned in the paper:
> our methods implements this accumulation as an exponentially decaying **average of the squared gradients**.

This is why I suggest changing the name to `averageSquared`. Changing the type to `Model.TanegentVector` makes it more clearly convey that this is a **gradient** vector.

Sources/TensorFlow/Optimizer.swift
```diff
@@ -355,3 +355,102 @@ public class AdaGrad<Model: Layer>: Optimizer
...
+    public var rho: Float
+    /// A small scalar added to the denominator to improve numerical stability.
+    public var epsilon: Float
+    /// The learning rate decay.
+    public var decay: Float
+    /// The current step.
+    public var step: Int = 0
+    /// Accumulate Gradients
+    public var accumulators: Model.AllDifferentiableVariables
+    /// Accumulate Updates(here StepSizes)
```
rxwei:
```suggestion
    /// The accumulated parameter updates.
```

Sources/TensorFlow/Optimizer.swift
rxwei: Here using the word "delta" would make this variable better correspond to the `𝚫x` term in the paper.

Sources/TensorFlow/Optimizer.swift
```diff
@@ -355,3 +355,102 @@ public class AdaGrad<Model: Layer>: Optimizer
...
+    public var updates: Model.AllDifferentiableVariables
+
+    public init(
+        for model: __shared Model,
+        learningRate: Float = 1,
+        rho: Float = 0.9,
```
rxwei: Keras uses `rho = 0.95` by default.

Sources/TensorFlow/Optimizer.swift
```diff
@@ -355,3 +355,102 @@ public class AdaGrad<Model: Layer>: Optimizer
...
+
+    // TODO: Deprecate this when `Differentiable.AllDifferentiableVariables` is removed.
+    public func update(_ model: inout Model.AllDifferentiableVariables,
+                       along direction: Model.AllDifferentiableVariables) {
+        step += 1
+        let learningRate = self.learningRate * 1 / (1 + decay * Float(step))
+        
+        // Update Float & Double Tensor variables.
+        for kp in model.recursivelyAllWritableKeyPaths(to: Tensor<Float>.self) {
+            var accNew = rho * accumulators[keyPath: kp]
```
rxwei: For clarity, I'd suggest rewriting this step as

```swift
averageSquared[keyPath: kp] *= rho
averageSquared[keyPath: kp] += (1 - rho) * (direction[keyPath: kp] * direction[keyPath: kp])
```

This way, you won't need to define extra variables. Same suggestion for the `Tensor<Double>` handling block.

Sources/TensorFlow/Optimizer.swift
```diff
@@ -355,3 +355,102 @@ public class AdaGrad<Model: Layer>: Optimizer
...
+            accNew += (1 - rho) * (direction[keyPath: kp] * direction[keyPath: kp])
+            accumulators[keyPath: kp] = accNew 
+
+            var stepSize =  direction[keyPath: kp] * sqrt(updates[keyPath: kp] + epsilon)
+            stepSize /=  sqrt(accumulators[keyPath: kp] + epsilon)
+            model[keyPath: kp] -= learningRate * stepSize
+
+            var updatesNew = updates[keyPath: kp] * rho
```
rxwei: I would suggest changing this step to the following:

```swift
accumulatedDelta *= rho
accumulatedDelta += (1 - rho) * stepSize.squared()
```
 Same suggestion for the `Tensor<Double>` handling block.

Sources/TensorFlow/Optimizer.swift
```diff
@@ -355,3 +355,98 @@ public class AdaGrad<Model: Layer>: Optimizer
...
+            averageSquared[keyPath: kp].resetToZero()
+            accumulatedDelta[keyPath: kp].resetToZero()
+        }
+    }
+
+    // TODO: Deprecate this when `Differentiable.AllDifferentiableVariables` is removed.
+    public func update(_ model: inout Model.AllDifferentiableVariables,
+                       along direction: Model.AllDifferentiableVariables) {
+        step += 1
+        let learningRate = self.learningRate * 1 / (1 + decay * Float(step))
```
rxwei: `* 1` is redundant.
```suggestion
        let learningRate = self.learningRate / (1 + decay * Float(step))
```

Sources/TensorFlow/Optimizer.swift
```diff
@@ -355,3 +355,98 @@ public class AdaGrad<Model: Layer>: Optimizer
...
+        
+        // Update Float & Double Tensor variables.
+        for kp in model.recursivelyAllWritableKeyPaths(to: Tensor<Float>.self) {
+            averageSquared[keyPath: kp] *= rho
+            averageSquared[keyPath: kp] +=
+                    (1 - rho) * (direction[keyPath: kp] * direction[keyPath: kp])
```
rxwei: Indent by 4.
```suggestion
                (1 - rho) * (direction[keyPath: kp] * direction[keyPath: kp])
```

Sources/TensorFlow/Optimizer.swift
```diff
@@ -355,3 +355,98 @@ public class AdaGrad<Model: Layer>: Optimizer
...
+            var stepSize = direction[keyPath: kp] *
+                    sqrt(accumulatedDelta[keyPath: kp] + epsilon)
```
rxwei:
```suggestion
                sqrt(accumulatedDelta[keyPath: kp] + epsilon)
```

Sources/TensorFlow/Optimizer.swift
```diff
@@ -355,3 +355,98 @@ public class AdaGrad<Model: Layer>: Optimizer
...
+            stepSize /= sqrt(averageSquared[keyPath: kp] + epsilon)
+            model[keyPath: kp] -= learningRate * stepSize
+            accumulatedDelta[keyPath: kp] *= rho
+            accumulatedDelta[keyPath: kp] += (1 - rho) * stepSize.squared()
+        }
+        for kp in model.recursivelyAllWritableKeyPaths(to: Tensor<Double>.self) {
+            averageSquared[keyPath: kp] *= Double(rho)
+            averageSquared[keyPath: kp] +=
+                    (1 - Double(rho)) * (direction[keyPath: kp] * direction[keyPath: kp])
```
rxwei:
```suggestion
                (1 - Double(rho)) * (direction[keyPath: kp] * direction[keyPath: kp])
```

Sources/TensorFlow/Optimizer.swift
```diff
@@ -355,3 +355,98 @@ public class AdaGrad<Model: Layer>: Optimizer
...
+            var stepSize = direction[keyPath: kp] *
+                    sqrt(accumulatedDelta[keyPath: kp] + Double(epsilon))
```
rxwei:
```suggestion
                sqrt(accumulatedDelta[keyPath: kp] + Double(epsilon))
```

Sources/TensorFlow/Optimizer.swift
```diff
@@ -355,3 +355,98 @@ public class AdaGrad<Model: Layer>: Optimizer
         update(&model.allDifferentiableVariables, along: direction)
     }
 }
+
+/// Adadelta Optimizer
```
rxwei: - Use the canonical name (the name in the paper) to describe this class. 
- End all comments with a period.
- Make the "O" in "Optimizer" be lowercased, for consistency with API documentation on other optimizers in this file.
```suggestion
/// ADADELTA optimizer.
```

Sources/TensorFlow/Optimizer.swift
```diff
@@ -355,3 +355,98 @@ public class AdaGrad<Model: Layer>: Optimizer
...
+///
+/// It is a method that uses the magnitude of recent gradients
```
rxwei: API documentation should fit to size (i.e. to 100 columns).

I'd suggest changing the summary to the following, similar to the Keras documentation.
> ADADELTA is a more robust extension of AdaGrad that adapts learning rates based on a moving window of gradient updates, instead of accumulating all past gradients. This way, Adadelta continues learning even when many updates have been done. Compared to AdaGrad, in the original version of ADADELTA you don't have to set an initial learning rate. In this version, initial learning rate and decay factor can be set, as in most other optimizers.

Sources/TensorFlow/Optimizer.swift
```diff
@@ -355,3 +355,98 @@ public class AdaGrad<Model: Layer>: Optimizer
...
+        }
+    }
+
+    // TODO: Deprecate this when `Differentiable.AllDifferentiableVariables` is removed.
+    public func update(_ model: inout Model.AllDifferentiableVariables,
+                       along direction: Model.AllDifferentiableVariables) {
+        step += 1
+        let learningRate = self.learningRate * 1 / (1 + decay * Float(step))
+        
+        // Update Float & Double Tensor variables.
```
rxwei:
```suggestion
        // Update `Tensor<Float>` and `Tensor<Double>` variables.
```

Sources/TensorFlow/Optimizer.swift
```diff
@@ -355,3 +355,100 @@ public class AdaGrad<Model: Layer>: Optimizer
         update(&model.allDifferentiableVariables, along: direction)
     }
 }
+
+/// ADADELTA optimizer.
+///
+/// ADADELTA is a more robust extension of AdaGrad that adapts learning rates
```
rxwei: This block of API documentation has lines whose length varies significantly from lines to lines. Could you make sure  each line is filled with words until it reaches 100 columns? This is what I meant by suggesting "fit to size". Hope this clarifies things!

rxwei:
Thanks @dan-zheng for reordering the class definitions!

Shashi456:
#127

PR #306, opened 2019-06-28
[Optimizers] Simplified the Adam optimizer.

eaplatanios:
@rxwei Do you know why the typechecker fails even for not too long/complex expressions and we have to break them down?

### DIFF
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-infix operator .>: ComparisonPrecedence
-infix operator .==: ComparisonPrecedence
+infix operator .> : ComparisonPrecedence
+infix operator .== : ComparisonPrecedence
 
 // TODO:
 // - Consider explicit broadcasting for elementwise binary ops when
@@ -22,8 +22,8 @@ infix operator .==: ComparisonPrecedence
 // TODO: Remove the following extension once `./` and `./=` are defined for 
 // `PointwiseMultiplicative`.
 
-infix operator ./: MultiplicationPrecedence
-infix operator ./=: AssignmentPrecedence
+infix operator ./ : MultiplicationPrecedence
+infix operator ./= : AssignmentPrecedence
 
 public extension PointwiseMultiplicative {
     static func ./ (lhs: Self, rhs: Self) -> Self {
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -19,6 +19,22 @@ infix operator .==: ComparisonPrecedence
 // - Consider explicit broadcasting for elementwise binary ops when
 //   scalarization and rank getter are implemented.
 
+// TODO: Remove the following extension once `./` and `./=` are defined for 
+// `PointwiseMultiplicative`.
+
+infix operator ./: MultiplicationPrecedence
+infix operator ./=: AssignmentPrecedence
+
+public extension PointwiseMultiplicative {
+    static func ./ (lhs: Self, rhs: Self) -> Self {
+        lhs .* rhs.reciprocal
+    }
+
+    static func ./= (lhs: inout Self, rhs: Self) {
+        lhs = lhs ./ rhs
+    }
+}
+
 //===------------------------------------------------------------------------------------------===//
 // Generic elementary functions
 //===------------------------------------------------------------------------------------------===//
--- a/Sources/TensorFlow/Optimizer.swift
+++ b/Sources/TensorFlow/Optimizer.swift
@@ -97,8 +97,7 @@ public class Adam<Model: Differentiable>: Optimizer
         secondMoments = secondMoments * beta2
         secondMoments += direction .* direction * (1 - beta2)
         let denominator = Model.TangentVector.sqrt(secondMoments) + epsilon
-        // TODO: Update this when `./` becomes available.
-        model.move(along: -stepSize * firstMoments .* denominator.reciprocal)
+        model.move(along: -stepSize * firstMoments ./ denominator)
     }
 }
 
--- a/Sources/TensorFlow/Optimizer.swift
+++ b/Sources/TensorFlow/Optimizer.swift
@@ -39,7 +39,8 @@ fileprivate extension Tensor where Scalar: Numeric {
 /// https://arxiv.org/abs/1412.6980v8)
 public class Adam<Model: Differentiable>: Optimizer
     where Model.TangentVector: VectorProtocol & PointwiseMultiplicative & ElementaryFunctions,
-          Model.TangentVector.VectorSpaceScalar == Float {
+          Model.TangentVector.VectorSpaceScalar == Float,
+          Model.AllDifferentiableVariables == Model.TangentVector {
     public typealias Model = Model
     /// The learning rate.
     public var learningRate: Float
@@ -91,6 +92,25 @@ public class Adam<Model: Differentiable>: Optimizer
         let denominator = Model.TangentVector.sqrt(secondMoments) + epsilon
         // TODO: Update this when `./` becomes available.
         model.move(along: -stepSize * firstMoments .* denominator.reciprocal)
+    }
+
+    // TODO: Deprecate this when `Differentiable.AllDifferentiableVariables` is removed.
+    public func update(
+        _ model: inout Model.AllDifferentiableVariables,
+        along direction: Model.AllDifferentiableVariables
+    ) {
+        step += 1
+        let learningRate = self.learningRate * 1 / (1 + decay * Float(step))
+        // Note: `stepSize` and `secondMoments` are split into two lines to avoid the "compiler is 
+        // unable to type-check this expression in reasonable time" error.
+        var stepSize = learningRate * sqrt(1 - pow(beta2, Float(step)))
+        stepSize = stepSize / (1 - pow(beta1, Float(step)))
+        firstMoments = firstMoments * beta1 + direction * (1 - beta1)
+        secondMoments = secondMoments * beta2
+        secondMoments += direction .* direction * (1 - beta2)
+        let denominator = Model.TangentVector.sqrt(secondMoments) + epsilon
+        // TODO: Update this when `./` becomes available.
+        model -= stepSize * firstMoments .* denominator.reciprocal
     }
 }
 
--- a/Sources/TensorFlow/Optimizer.swift
+++ b/Sources/TensorFlow/Optimizer.swift
@@ -22,8 +22,7 @@ public protocol Optimizer {
     associatedtype Scalar: FloatingPoint
     /// The learning rate.
     var learningRate: Scalar { get set }
-    /// Updates the specified differentiable variables along the specified
-    /// direction.
+    /// Updates the specified differentiable variables along the specified direction.
     mutating func update(_ variables: inout Model, along direction: Model.TangentVector)
 }
 
--- a/Sources/TensorFlow/Optimizer.swift
+++ b/Sources/TensorFlow/Optimizer.swift
@@ -12,10 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-/// A machine learning optimizer.
+/// A numerical optimizer.
 ///
-/// Optimizers apply an optimization algorithm to update the differentiable variables of a machine
-/// learning model.
+/// Optimizers apply an optimization algorithm to update the differentiable models.
 public protocol Optimizer {
     /// The type of the model whose parameters are optimized.
     associatedtype Model: Differentiable
@@ -38,16 +37,15 @@ fileprivate extension Tensor where Scalar: Numeric {
 ///
 /// Reference: ["Adam - A Method for Stochastic Optimization"](
 /// https://arxiv.org/abs/1412.6980v8)
-public class Adam<Model: Layer>: Optimizer
-    where Model.AllDifferentiableVariables == Model.TangentVector {
+public class Adam<Model: Differentiable>: Optimizer
+    where Model.TangentVector: VectorProtocol & PointwiseMultiplicative & ElementaryFunctions,
+          Model.TangentVector.VectorSpaceScalar == Float {
     public typealias Model = Model
     /// The learning rate.
     public var learningRate: Float
-    /// A coefficient used to calculate the first and second moments of
-    /// gradients.
+    /// A coefficient used to calculate the first and second moments of the gradients.
     public var beta1: Float
-    /// A coefficient used to calculate the first and second moments of
-    /// gradients.
+    /// A coefficient used to calculate the first and second moments of the gradients.
     public var beta2: Float
     /// A small scalar added to the denominator to improve numerical stability.
     public var epsilon: Float
@@ -56,9 +54,9 @@ public class Adam<Model: Layer>: Optimizer
     /// The current step.
     public var step: Int = 0
     /// The first moments of the weights.
-    public var firstMoments: Model.AllDifferentiableVariables
+    public var firstMoments: Model.TangentVector = .zero
     /// The second moments of the weights.
-    public var secondMoments: Model.AllDifferentiableVariables
+    public var secondMoments: Model.TangentVector = .zero
 
     public init(
         for model: __shared Model,
@@ -78,57 +76,21 @@ public class Adam<Model: Layer>: Optimizer
         self.beta2 = beta2
         self.epsilon = epsilon
         self.decay = decay
-
-        // Initialize first & second moments to be zeros of the same shape.
-        // We can't use `Model.AllDifferentiableVariables.zero` due to the
-        // interaction between Key Paths and Differentiable Arrays.
-        firstMoments = model.allDifferentiableVariables
-        secondMoments = model.allDifferentiableVariables
-        for kp in firstMoments.recursivelyAllWritableKeyPaths(to: Tensor<Float>.self) {
-            firstMoments[keyPath: kp].resetToZero()
-            secondMoments[keyPath: kp].resetToZero()
-        }
-        for kp in firstMoments.recursivelyAllWritableKeyPaths(to: Tensor<Double>.self) {
-            firstMoments[keyPath: kp].resetToZero()
-            secondMoments[keyPath: kp].resetToZero()
-        }
     }
 
-    // TODO: Deprecate this when `Differentiable.AllDifferentiableVariables` is removed.
-    public func update(_ model: inout Model.AllDifferentiableVariables,
-                       along direction: Model.AllDifferentiableVariables) {
+    public func update(_ model: inout Model, along direction: Model.TangentVector) {
         step += 1
         let learningRate = self.learningRate * 1 / (1 + decay * Float(step))
-        // Note: `stepSize` is split into two lines to avoid the "compiler is unable to type-check
-        // this expression in reasonable time" error.
+        // Note: `stepSize` and `secondMoments` are split into two lines to avoid the "compiler is 
+        // unable to type-check this expression in reasonable time" error.
         var stepSize = learningRate * sqrt(1 - pow(beta2, Float(step)))
         stepSize = stepSize / (1 - pow(beta1, Float(step)))
-        // Update Float & Double Tensor variables.
-        for kp in model.recursivelyAllWritableKeyPaths(to: Tensor<Float>.self) {
-            firstMoments[keyPath: kp] =
-                firstMoments[keyPath: kp] * beta1 + (1 - beta1) * direction[keyPath: kp]
-            secondMoments[keyPath: kp] =
-                secondMoments[keyPath: kp] * beta2 + (1 - beta2) *
-                direction[keyPath: kp] * direction[keyPath: kp]
-            model[keyPath: kp] -=
-                stepSize * firstMoments[keyPath: kp] / (sqrt(secondMoments[keyPath: kp]) + epsilon)
-        }
-        for kp in model.recursivelyAllWritableKeyPaths(to: Tensor<Double>.self) {
-            firstMoments[keyPath: kp] =
-                firstMoments[keyPath: kp] * Double(beta1) +
-                Double((1 - beta1)) * direction[keyPath: kp]
-            secondMoments[keyPath: kp] =
-                secondMoments[keyPath: kp] * Double(beta2) + Double(1 - beta2) *
-                direction[keyPath: kp] * direction[keyPath: kp]
-            model[keyPath: kp] -=
-                Double(stepSize) * firstMoments[keyPath: kp] /
-                sqrt(secondMoments[keyPath: kp]) + Double(epsilon)
-        }
-    }
-
-    public func update(_ model: inout Model,
-                       along direction: Model.TangentVector) {
-        update(&model.allDifferentiableVariables, along: direction)
+        firstMoments = firstMoments * beta1 + direction * (1 - beta1)
+        secondMoments = secondMoments * beta2
+        secondMoments += direction .* direction * (1 - beta2)
+        let denominator = Model.TangentVector.sqrt(secondMoments) + epsilon
+        // TODO: Update this when `./` becomes available.
+        model.move(along: -stepSize * firstMoments .* denominator.reciprocal)
     }
 }
 
--- a/Sources/TensorFlow/Optimizer.swift
+++ b/Sources/TensorFlow/Optimizer.swift
@@ -80,18 +80,7 @@ public class Adam<Model: Differentiable>: Optimizer
     }
 
     public func update(_ model: inout Model, along direction: Model.TangentVector) {
-        step += 1
-        let learningRate = self.learningRate * 1 / (1 + decay * Float(step))
-        // Note: `stepSize` and `secondMoments` are split into two lines to avoid the "compiler is 
-        // unable to type-check this expression in reasonable time" error.
-        var stepSize = learningRate * sqrt(1 - pow(beta2, Float(step)))
-        stepSize = stepSize / (1 - pow(beta1, Float(step)))
-        firstMoments = firstMoments * beta1 + direction * (1 - beta1)
-        secondMoments = secondMoments * beta2
-        secondMoments += direction .* direction * (1 - beta2)
-        let denominator = Model.TangentVector.sqrt(secondMoments) + epsilon
-        // TODO: Update this when `./` becomes available.
-        model.move(along: -stepSize * firstMoments .* denominator.reciprocal)
+        update(&model.allDifferentiableVariables, along: direction)
     }
 
     // TODO: Deprecate this when `Differentiable.AllDifferentiableVariables` is removed.
--- a/Sources/TensorFlow/Optimizer.swift
+++ b/Sources/TensorFlow/Optimizer.swift
@@ -39,8 +39,7 @@ fileprivate extension Tensor where Scalar: Numeric {
 /// https://arxiv.org/abs/1412.6980v8)
 public class Adam<Model: Differentiable>: Optimizer
     where Model.TangentVector: VectorProtocol & PointwiseMultiplicative & ElementaryFunctions,
-          Model.TangentVector.VectorSpaceScalar == Float,
-          Model.AllDifferentiableVariables == Model.TangentVector {
+          Model.TangentVector.VectorSpaceScalar == Float {
     public typealias Model = Model
     /// The learning rate.
     public var learningRate: Float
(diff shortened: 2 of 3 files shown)


rxwei:
[Operator type checking](https://forums.swift.org/t/pitch-making-expression-type-checking-of-operator-expressions-fast/18037) has long been very problematic in Swift.

Tests/TensorFlowTests/SequentialTests.swift
```diff
@@ -48,7 +48,6 @@ final class SequentialTests: XCTestCase {
             rmsprop.update(&model, along: 𝛁model)
             rmsprop.update(&model.allDifferentiableVariables, along: 𝛁model)
             adam.update(&model, along: 𝛁model)
-            adam.update(&model.allDifferentiableVariables, along: 𝛁model)
```
rxwei: Let's keep the `allDifferentiableVariable` version for now for source compatibility. We'll remove them in `0.4.1` along with `AllDifferentiableVariables`.

eaplatanios: Sounds good. I duplicated the implementation in this case because I can't have it call the new update function and we can just remove the duplicate code then.

Sources/TensorFlow/Optimizer.swift
```diff
@@ -38,16 +37,16 @@ fileprivate extension Tensor where Scalar: Numeric {
 ///
 /// Reference: ["Adam - A Method for Stochastic Optimization"](
 /// https://arxiv.org/abs/1412.6980v8)
-public class Adam<Model: Layer>: Optimizer
-    where Model.AllDifferentiableVariables == Model.TangentVector {
+public class Adam<Model: Differentiable>: Optimizer
+    where Model.TangentVector: VectorProtocol & PointwiseMultiplicative & ElementaryFunctions,
+          Model.TangentVector.VectorSpaceScalar == Float,
+          Model.AllDifferentiableVariables == Model.TangentVector {
```
rxwei: This constraint is no longer necessary.

Sources/TensorFlow/Optimizer.swift
```diff
@@ -78,57 +77,40 @@ public class Adam<Model: Layer>: Optimizer
...
-        for kp in firstMoments.recursivelyAllWritableKeyPaths(to: Tensor<Double>.self) {
-            firstMoments[keyPath: kp].resetToZero()
-            secondMoments[keyPath: kp].resetToZero()
-        }
     }
 
-    // TODO: Deprecate this when `Differentiable.AllDifferentiableVariables` is removed.
-    public func update(_ model: inout Model.AllDifferentiableVariables,
-                       along direction: Model.AllDifferentiableVariables) {
+    public func update(_ model: inout Model, along direction: Model.TangentVector) {
```
rxwei: Instead of duplicating the logic, you should call `update(&model.allDifferentiableVariables, along: direction)` like `SGD`.

eaplatanios: Done.

eaplatanios: Done.

Sources/TensorFlow/Optimizer.swift
```diff
@@ -78,57 +77,29 @@ public class Adam<Model: Layer>: Optimizer
...
-    }
-
-    public func update(_ model: inout Model,
-                       along direction: Model.TangentVector) {
-        update(&model.allDifferentiableVariables, along: direction)
+        firstMoments = firstMoments * beta1 + direction * (1 - beta1)
+        secondMoments = secondMoments * beta2
+        secondMoments += direction .* direction * (1 - beta2)
+        let denominator = Model.TangentVector.sqrt(secondMoments) + epsilon
+        // TODO: Update this when `./` becomes available.
```
rxwei: Feel free to define a `./` in an extension somewhere in `Math.swift`.

Sources/TensorFlow/Optimizer.swift
eaplatanios: It does not compile without it:
```
/Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Optimizer.swift:82:23: error: property 'allDifferentiableVariables' requires the types 'Model' and 'Model.AllDifferentiableVariables' be equivalent
                      ^
Swift.Differentiable:1:11: note: where 'Self' = 'Model', 'Self.AllDifferentiableVariables' = 'Model.AllDifferentiableVariables'
extension Differentiable where Self == Self.AllDifferentiableVariables {
          ^
/Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Optimizer.swift:96:57: error: binary operator '*' cannot be applied to operands of type 'Model.AllDifferentiableVariables' and 'Float'
        firstMoments = firstMoments * beta1 + direction * (1 - beta1)
                                              ~~~~~~~~~ ^ ~~~~~~~~~~~
/Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Optimizer.swift:96:57: note: expected an argument list of type '(Float, Float)'
        firstMoments = firstMoments * beta1 + direction * (1 - beta1)
                                                        ^
/Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Optimizer.swift:98:36: error: binary operator '.*' cannot be applied to two 'Model.AllDifferentiableVariables' operands
        secondMoments += direction .* direction * (1 - beta2)
                         ~~~~~~~~~ ^  ~~~~~~~~~
/Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Optimizer.swift:98:36: note: expected an argument list of type '(Tensor<Scalar>, Tensor<Scalar>)'
        secondMoments += direction .* direction * (1 - beta2)
                                   ^
/Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Optimizer.swift:101:15: error: binary operator '-=' cannot be applied to operands of type 'Model.AllDifferentiableVariables' and 'Model.TangentVector'
        model -= stepSize * firstMoments .* denominator.reciprocal
        ~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Optimizer.swift:101:15: note: expected an argument list of type '(inout Self, Self)'
        model -= stepSize * firstMoments .* denominator.reciprocal
              ^
[1/2] Compiling TensorFlow Optimizer.swift
```

rxwei: Instead of `model -=`, `move(along:)` should be used.

Sources/TensorFlow/Optimizer.swift
```diff
@@ -78,57 +77,29 @@ public class Adam<Model: Layer>: Optimizer
...
+    public func update(_ model: inout Model, along direction: Model.TangentVector) {
+        update(&model.allDifferentiableVariables, along: direction)
     }
 
     // TODO: Deprecate this when `Differentiable.AllDifferentiableVariables` is removed.
-    public func update(_ model: inout Model.AllDifferentiableVariables,
-                       along direction: Model.AllDifferentiableVariables) {
+    public func update(
+        _ model: inout Model.AllDifferentiableVariables,
+        along direction: Model.AllDifferentiableVariables
```
rxwei:
```suggestion
        along direction: Model.TangentVector
```
This will fix the other type error.

Sources/TensorFlow/Optimizer.swift
eaplatanios: Isn't it better to put this in the apple/swift repo where `PointwiseMultiplicative` is defined? Or should I just add it here temporarily.

rxwei: The idea is to define it here temporarily. Changing that in apple/swift will require some changes to derived conformances, because providing a default implementation is tricky.

eaplatanios: Sounds good. I made the necessary changes.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -19,6 +19,22 @@ infix operator .==: ComparisonPrecedence
 // - Consider explicit broadcasting for elementwise binary ops when
 //   scalarization and rank getter are implemented.
 
+// TODO: Remove the following extension once `./` and `./=` are defined for 
+// `PointwiseMultiplicative`.
+
+infix operator ./: MultiplicationPrecedence
```
```suggestion
infix operator ./ : MultiplicationPrecedence
```

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -19,6 +19,22 @@ infix operator .==: ComparisonPrecedence
...
+infix operator ./=: AssignmentPrecedence
```
```suggestion
infix operator ./= : AssignmentPrecedence
```

eaplatanios:
Sounds good! Will do that and add new ones after being done with the refactoring.

eaplatanios:
@rxwei what about the `max(_:_:)` and `abs(_:)` operations? We currently do not have automatically derived protocols for them, right?

rxwei:
We do not, unfortunately.

eaplatanios:
@rxwei Would it be easy to add derived conformances for `Comparable` similar to `ElementaryFunctions`? There are a few optimizers that use `max` accumulators (e.g., `AdaMax` and `AMSGrad`, for which I have an upcoming PR) and this would help a lot with having simplified implementations for them too.

rxwei:
Comparable requires Boolean operators to return a single Bool, so it’s not the right abstraction for our use case.

eaplatanios:
Yes that's right. For now I simplified AdaMax as much as I could and added an implementation of AMSGrad that's also still using key-paths for some part of it. We can think about what protocols we would need to support simplifying these optimizers later on.

rxwei:
We should consider pitching a PointwiseComparable protocol on the Swift forums. It can be applied to SIMD types too.

eaplatanios:
I agree. I also feel that automatically deriving protocol conformances keeps coming up very frequently (it also has come up multiple times in some research projects I haven't published yet) and it is currently very tedious having to modify the compiler to support such cases. Do you think it may be a good time to try and come up with a Swift-side way of doing this and implement it at least as part of S4TF? It will be hard to pitch it early on because there are many opinions on this topic, but if we come up with a clean practical way that works for us and implement it, I think we'll also gain a better understanding of what it takes.

Maybe I'm biased because I'm used to defining type classes to achieve my goals (they're super powerful in Scala), but I do believe that they are a very important feature that we keep bumping into.

dan-zheng:
A better, general conformance derivation mechanism would certainly be amazing. My thoughts:
* A robust design seems difficult, given Swift's limited metaprogramming support. I imagine a full design would involve reflection and metaprogramming in some form. [Scala typeclass derivation](https://dotty.epfl.ch/docs/reference/contextual/derivation.html) could be inspiration.
* Besides derivation, there are other problems, like "actually designing/iterating on the right protocols" (e.g. the newly added `PointwiseMultiplicative`, and maybe `PointwiseComparable`). To me, designing the right protocols is a more pressing issue, especially given Evolution challenges, like source/ABI compatibility. Hardcoding derived conformances isn't too much effort (<1 day). Until a general derivation mechanism exists, we can view "automated conformances" as a detail.

I would love to see solid math protocols land before general conformance derivation! A [fully algebraic design](https://typelevel.org/algebra/typeclasses/overview.html) probably isn't realistic for Swift. I hope we can find some middle ground that unblocks our use cases (e.g. aggregate math) and improves the current status in Swift (e.g. maybe removing [`Magnitude`](https://developer.apple.com/documentation/swift/numeric/2884423-magnitude) from `Numeric`) while remaining suitable for Evolution.

rxwei:
@dan-zheng is right, we should think about the right protocol design first. Currently there are harder problems to deal with: backporting, retroactive refinement without breaking ABI, etc.  This needs us to pitch these designs on Swift forums.

eaplatanios:
Thanks @dan-zheng for the detailed response. I see your points. I guess I was also motivated by other uses of derived conformances in code outside S4TF where it gets inconvenient to have to use a fork of the Swift compiler. But yes, figuring out the right protocols does seem like an appropriate first steps. I haven't looked in too much detail into the ones you currently have, but will go through them and see if I can make any useful suggestions.

One suggestion I have for now is that it would be nice to allow for something like `@noDerivative` that allows us to mark stored properties for which the `ElementaryFunctions` operations act as an identity function, for example. More generally, I believe we should think about what it would mean for being able to opt-out for certain stored properties. Haven't thought this through much but I think it will come up with more complicated layers or layers that store side information (e.g., shape information about their inputs and outputs).

Also, a side comment is that I noticed that `VectorProtocol` is not being derived automatically for structs with no stored properties. Why is that?

rxwei:
> Also, a side comment is that I noticed that `VectorProtocol` is not being derived automatically for structs with no stored properties. Why is that?

This is because `VectorProtocol` cannot infer `VectorSpaceScalar` from a `struct` that has no stored properties. Hard-coding this inference to use something like `Float` seems unprincipled.

eaplatanios:
I see. However it also causes a usability issue. Layers with no parameters (e.g., activation functions if defined as layers) cannot be used to build models that will later be trained using our current optimizers. That is because it won't be possible to derive `VectorProtocol` for the layers that use them. We should probably work around this somehow in a principled manner.

rxwei:
> I see. However it also causes a usability issue. Layers with no parameters (e.g., activation functions if defined as layers) cannot be used to build models that will later be trained using our current optimizers. That is because it won't be possible to derive `VectorProtocol` for the layers that use them. We should probably work around this somehow in a principled manner.

We can add a `typealias VectorSpaceScalar` in those layers for now.

eaplatanios:
We would also have to explicitly declare conformance to `VectorProtocol` though, right?

rxwei:
Yes. Ideally, we should be able to declare all of these in `protocol Layer` and expect derived conformances to work. But I'm not sure whether that will work today.

eaplatanios:
I've been working on a PR that does that but currently the `@noDerivative` example I mentioned above is blocking this. For example, `MaxPool1D` cannot conform to `VectorProtocol` because it has integer stored properties. Do you have any ideas for how to handle this?

eaplatanios:
One possible solution would be to separate `Layer` from its parameters and have all layers have a stored property called `parameters` and a `Parameter` associated type which has to conform to `Differentiable`, `VectorProtocol`, etc.

rxwei:
We shouldn't need layers to conform to `VectorProtocol`. Instead, we want their `TangentVector` to conform. Ideally you should be able to just extend those layers's `TangentVector` to conform, for example:
```swift
extension MaxPool1D.TangentVector: VectorProtocol {
    public typealias VectorSpaceScalar = Float
}
```

However, this is currently blocked on [this bug](https://forums.swift.org/t/synthesized-member-types-and-type-checking/26386) that @dan-zheng is investigating.

By "defining it in the protocol", I meant something like:

```swift
public protocol Layer: Differentiable where TangentVector: VectorProtocol {
    typealias VectorSpaceScalar = Float
}
```

> One possible solution would be to separate `Layer` from its parameters and have all layers have a stored property called `parameters` and a `Parameter` associated type which has to conform to `Differentiable`, `VectorProtocol`, etc.

This is something we experimented with a year ago, called `Parameterized`. The solution was abandoned because any `Parameter` associated type would make a model with multiple parameter types be either impossible or really complicated to define. We concluded that it wasn't a future-proof solution, so we moved away from that model.

eaplatanios:
Thanks for the explanation! That totally makes sense. Regarding the current bug, defining `VectorSpaceScalar` in the protocol itself does not resolve the issue for its `TangentVector`. However, adding a property like `@usableFromInline internal var _vectorSpace = 0.0` to the related structs resolved this. So, after the initializers and the random seed PRs are merged, I'll open a new PR that adds constraints to layers so they can be used with the new optimizers. As a temporary workaround for parameter-less layers, I will define this protocol:
```swift
public protocol ParameterlessLayer: Layer {
    @usableFromInline
    internal var _vectorSpace = 0.0
}
```
and add a TODO along with a bug filing to remove this once Dan fixes the underlying issue you mentioned. How does that sound?

rxwei:
> Thanks for the explanation! That totally makes sense. Regarding the current bug, defining `VectorSpaceScalar` in the protocol itself does not resolve the issue for its `TangentVector`. However, adding a property like `@usableFromInline internal var _vectorSpace = 0.0` to the related structs resolved this. So, after the initializers and the random seed PRs are merged, I'll open a new PR that adds constraints to layers so they can be used with the new optimizers. As a temporary workaround for parameter-less layers, I will define this protocol:
> 
> ```swift
> public protocol ParameterlessLayer: Layer {
>     @usableFromInline
>     internal var _vectorSpace = 0.0
> }
> ```
> 
> and add a TODO along with a bug filing to remove this once Dan fixes the underlying issue you mentioned. How does that sound?

I don't feel this is the best solution. Stored properties cannot be defined in protocols, so even if you add a protocol requirement, every conforming type will have to define such a stored property to satisfy this requirement. We are not yet prioritizing a fix for that bug, so it is likely that this suboptimal pattern will last longer.

I'll suggest a different solution: One can always define a custom `TangentVector`, so it should be trivial to define one for those parameter-less layers. I believe all you need to add in those structs is:
```swift
struct TangentVector: Differentiable, VectorProtocol {}
```

eaplatanios:
Oh right. Good point. Ok I’ll look into this when I get back home tonight.

eaplatanios:
@rxwei one issue that comes up is that we need `AllDifferentiableVariables == TangentVector` which causes a failure when I try what you suggested earlier.

eaplatanios:
I also get the following error when I use the suggested approach:
```
Assertion failed: (paramMember && "Could not find corresponding parameter member"), function operator(), file /Users/eaplatanios/Development/GitHub/swift-source-temp/swift/lib/Sema/DerivedConformanceDifferentiable.cpp, line 279.
Stack dump:
0.      Program arguments: /Library/Developer/Toolchains/swift-tensorflow-LOCAL-2019-06-30-a.xctoolchain/usr/bin/swift -frontend -c /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Bindings/EagerExecution.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Bindings/RawOpsGenerated.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Bindings/TFTensorOperation.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Context.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/ArrayOps.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/DataTypes.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/DifferentialOperators.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/Execution.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/LazyTensorOperation.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/LazyTensorTFFunctionBuilder.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/LazyTensorTrace.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/PythonConversion.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/Runtime.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/ShapedArray.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/StringTensor.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/Tensor.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/TensorGroup.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/TensorHandle.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/TensorProtocol.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/TensorShape.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/Threading.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/Utilities.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Initializers.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layer.swift -primary-file /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Convolutional.swift -primary-file /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Core.swift -primary-file /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Embedding.swift -primary-file /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Initialization.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Normalization.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Pooling.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Recurrent.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Upsampling.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Loss.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Operators/Basic.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Operators/Comparison.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Operators/Dataset.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Operators/Math.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Operators/NN.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Optimizers/MomentumBased.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Optimizers/Optimizer.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Optimizers/SGD.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Random.swift -emit-module-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Convolutional~partial.swiftmodule -emit-module-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Core~partial.swiftmodule -emit-module-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Embedding~partial.swiftmodule -emit-module-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Initialization~partial.swiftmodule -emit-module-doc-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Convolutional~partial.swiftdoc -emit-module-doc-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Core~partial.swiftdoc -emit-module-doc-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Embedding~partial.swiftdoc -emit-module-doc-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Initialization~partial.swiftdoc -emit-dependencies-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Convolutional.d -emit-dependencies-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Core.d -emit-dependencies-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Embedding.d -emit-dependencies-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Initialization.d -emit-reference-dependencies-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Convolutional.swiftdeps -emit-reference-dependencies-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Core.swiftdeps -emit-reference-dependencies-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Embedding.swiftdeps -emit-reference-dependencies-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Initialization.swiftdeps -target x86_64-apple-macosx10.10 -enable-objc-interop -sdk /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk -I /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug -I /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/usr/lib -F /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/Library/Frameworks -enable-testing -g -module-cache-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/ModuleCache -swift-version 4.2 -Onone -D SWIFT_PACKAGE -D DEBUG -color-diagnostics -enable-anonymous-context-mangled-names -parse-as-library -module-name TensorFlow -o /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Convolutional.swift.o -o /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Core.swift.o -o /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Embedding.swift.o -o /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Initialization.swift.o -index-store-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/index/store -index-system-modules 
1.      Swift version 5.1-dev (LLVM 7d54a7c711, Swift f62bd5687d)
0  swift                    0x000000010a86c955 llvm::sys::PrintStackTrace(llvm::raw_ostream&) + 37
1  swift                    0x000000010a86b998 llvm::sys::RunSignalHandlers() + 248
2  swift                    0x000000010a86cf48 SignalHandler(int) + 264
4  libsystem_platform.dylib 0x00000001173a9938 _sigtramp + 2711539192
7  swift                    0x000000010abb0d13 deriveBodyDifferentiable_move(swift::AbstractFunctionDecl*, void*) (.cold.9) + 35
8  swift                    0x0000000107609ff7 deriveBodyDifferentiable_move(swift::AbstractFunctionDecl*, void*) + 2279
9  swift                    0x0000000107708883 swift::TypeChecker::typeCheckAbstractFunctionBody(swift::AbstractFunctionDecl*) + 83
10 swift                    0x0000000107729453 typeCheckFunctionsAndExternalDecls(swift::SourceFile&, swift::TypeChecker&) + 275
11 swift                    0x0000000107729c1d swift::performTypeChecking(swift::SourceFile&, swift::TopLevelContext&, swift::OptionSet<swift::TypeCheckingFlags, unsigned int>, unsigned int, unsigned int, unsigned int, unsigned int, unsigned int) + 1005
12 swift                    0x0000000106e62d71 swift::CompilerInstance::parseAndCheckTypesUpTo(swift::CompilerInstance::ImplicitImports const&, swift::SourceFile::ASTStage_t)::$_1::operator()(swift::SourceFile&) const + 81
13 swift                    0x0000000106e5e74b swift::CompilerInstance::parseAndCheckTypesUpTo(swift::CompilerInstance::ImplicitImports const&, swift::SourceFile::ASTStage_t) + 507
14 swift                    0x0000000106e5e04a swift::CompilerInstance::performSemaUpTo(swift::SourceFile::ASTStage_t) + 618
15 swift                    0x0000000106c4be92 performCompile(swift::CompilerInstance&, swift::CompilerInvocation&, llvm::ArrayRef<char const*>, int&, swift::FrontendObserver*, swift::UnifiedStatsReporter*) + 1730
16 swift                    0x0000000106c4a7f1 swift::performFrontend(llvm::ArrayRef<char const*>, char const*, void*, swift::FrontendObserver*) + 3025
17 swift                    0x0000000106bf2bf9 main + 729
18 libdyld.dylib            0x00007fff757d13d5 start + 1
19 libdyld.dylib            0x0000000000000077 start + 2323836067
```

eaplatanios:
Nevermind, I worked out a solution and will be opening a PR soon. :)

eaplatanios:
Done in #322 .

PR #307, opened 2019-06-28
[Optimizers] Simplified the RMSProp optimizer.

eaplatanios:


Sources/TensorFlow/Optimizer.swift
```diff
@@ -202,14 +202,15 @@ public class AdaMax<Model: Layer>: Optimizer
...
 ///
-/// It is recommended to leave the parameters of this optimizer at their default values (except the
-/// learning rate, which can be freely tuned). This optimizer is usually a good choice for recurrent
-/// neural networks.
+/// It is recommended to leave the parameters of this optimizer at their default values (except for 
+/// the learning rate, which can be freely tuned). This optimizer is usually a good choice for 
+/// recurrent neural networks.
 ///
-/// Reference: ["rmsprop: Divide the gradient by a running average of its recent magnitude"](
+/// Reference: ["RMSProp: Divide the gradient by a running average of its recent magnitude"](
```
rxwei: Since the paper calls it `rmsprop`, I think it's fine to leave it as is. Similarly, when we referred to the AdaDelta paper, we used `ADADELTA` instead of `AdaDelta` because the paper said so.

eaplatanios:
I generally feel we should have better tests for optimizers. Convergence is tricky as numerical differences will add up and may result in different numbers across different machines. I think checks verifying correctness for a few steps should be fine. I can look into that after we're done with the main refactoring of optimizers.

The change here resulted from the switch from `pow(direction, 2)` to `direction .* direction`. The latter is both faster and more accurate since the first is an approximation.

rxwei:
> The change here resulted from the switch from `pow(direction, 2)` to `direction .* direction`. The latter is both faster and more accurate since the first is an approximation.

That is great to hear!

eaplatanios:
Also, would you mind if I create an `Optimizers` directory similar to `Operators` and `Layers` in a separate PR so we split them out to separate files? This file has also started growing and will keep growing as there are quite a few optimizers one could define.

rxwei:
Yes, that's the definitely the right direction.

PR #308, opened 2019-06-28
[Optimizers] Simplified the AdaGrad optimizer.

eaplatanios:


eaplatanios:
I could do that, but why do you expect the performance to be different?

dan-zheng:
> I could do that, but why do you expect the performance to be different?

The code structure of optimizer update functions has changed quite a bit.
Previously, there were two for-loops iterating over `recursivelyAllKeyPaths`, and operations were performed per `Tensor`. Now, aggregate operations are used.

I would expect the new optimizers to be more performant.

eaplatanios:
Yes that makes sense. I believe this will be true for complex `Model` structures (i.e., consisting of multiple tensors that are not too big). I will try to add some benchmarking tests once I get the chance.

It would be great if we could set up some sort of benchmarking suite on top of the CI server so that we can see how the performance of various tests/benchmarks changes across commits. It will also help us detect potential performance regressions.

dan-zheng:
> Yes that makes sense. I believe this will be true for complex `Model` structures (i.e., consisting of multiple tensors that are not too big). I will try to add some benchmarking tests once I get the chance.

That would be awesome!

> It would be great if we could set up some sort of benchmarking suite on top of the CI server so that we can see how the performance of various tests/benchmarks changes across commits. It will also help us detect potential performance regressions.

Yes, a benchmarking suite would be sweet (😎). [`apple/swift` CI has such a suite](https://github.com/apple/swift/pull/24299#issuecomment-506780329). Our team doesn't have bandwidth to set this up though.

PR #310, opened 2019-06-28
[Optimizers] Refactored the optimizers in their own directory.

eaplatanios:


eaplatanios:
How about the current split? I feel "adaptive" is too general in the sense that most optimizers are adaptive in some way, whereas "momentum-based" is a bit more specific to these methods. Also, I used simply "SGD" for the other two, as they are SGD variants. I also feel "gradient descent" is too general as they are all effectively gradient descent methods. Please let me know what you think.

rxwei:
Sounds good!

rxwei:
Maybe rename `MomentumBasedOptimizers.swift` to `MomentumBased.swift`.

eaplatanios:
Done. :)

Sources/TensorFlow/Optimizers/SGD.swift
```diff
@@ -0,0 +1,96 @@
...
+        if nesterov {
+            model.move(along: momentum * velocity - direction * learningRate)
+        } else {
+            model.move(along: velocity)
+        }
+    }
+}
+
+/// A Riemann manifold stochastic gradient descent (SGD) optimizer.
+public class RiemannSGD<Model: Differentiable>: Optimizer
```
rxwei: We can remove `RiemannSGD` at some point since `SGD` is essentially a functionally richer `RiemannSGD` now.

PR #311, opened 2019-06-28
[Optimizers] Removed the 'RiemannSGD' optimizer.

eaplatanios:


rahulbhalley:
@eaplatanios why RiemannSGD optimizer was removed?

dan-zheng:
@eaplatanios: I wonder if you recall why `RiemannSGD` was removed?

[TF-981](https://bugs.swift.org/browse/TF-981) tracks a use case for optimizers that accept `Differentiable`-conforming types whose `VectorProtocol.VectorSpaceScalar` is not `Float`.

Shashi456:
@dan-zheng, you can look at the comment [here](https://github.com/tensorflow/swift-apis/pull/310#discussion_r298761082).

PR #313, opened 2019-06-29
[Optimizers] Simplified the AdaMax optimizer.

eaplatanios:
This PR simplifies the AdaMax optimizer implementation a bit and also fixes a bug (the denominator of the step size was not computed correctly).


eaplatanios:
Thanks Dan! I have that in my todo list but haven't gotten to it yet because I'm busy with some deadlines.

PR #314, opened 2019-06-29
[Optimizers] Added support for the AMSGrad optimizer.

eaplatanios:


Sources/TensorFlow/Optimizers/MomentumBased.swift
```diff
@@ -357,3 +357,93 @@ public class AdaMax<Model: Layer>: Optimizer
         update(&model.allDifferentiableVariables, along: direction)
     }
 }
+
+/// AMSGrad optimizer.
+///
+/// This algorithm is a modification of Adam that allows for better convergence properties when
```
dan-zheng: Minor tweaks:
```suggestion
/// This algorithm is a modification of Adam with better convergence properties when close to local
/// optima.
```

eaplatanios:
Thanks for the reviews Dan!

PR #315, opened 2019-06-30
Fixed one code snippet in README.md.

leoxzhao:


Shashi456:
Why make y an `INT32`?  because imo implicit conversion would promote an int value to float anyway.

README.md
```diff
@@ -38,11 +38,11 @@ struct Model: Layer {
 #### Initialize a model and an optimizer
 
 ```swift
-let optimizer = SGD(for: model, learningRate: 0.02)
 var classifier = Model()
+let optimizer = SGD(for: classifier, learningRate: 0.02)
 let x: Tensor<Float> = ...
-let y: Tensor<Float> = ...
+let y: Tensor<Int32> = ...
```
dan-zheng: I believe `y: Tensor<Float>` makes more sense here.
(It's more general for expected predictions to have type `Tensor<Float>`. Practically, actual predictions will always have type `Tensor<Float>`.)

leoxzhao: The next code snippet after this uses `softmaxCrossEntropy` loss function which requires `Tensor<Int32>` as labels. In general, I agree `Tensor<Float>` makes more sense.

eaplatanios: Both are correct actually and there is a `softmaxCrossEntropy` implementation for both. It depends on whether you want to use a sparse representation for the labels or not. A sparse representation (i.e., `Int32`) would be faster (much faster if the number of possible labels is large) and so that is most often used for multi-class problems (as opposed to multi-label where more than one labels can apply at a time). The MNIST example in `swift-models` uses `Int32` and so we could be consistent with that here too. @dan-zheng what do you think?

rxwei: This code snippet was using `Tensor<Float>` because `softmaxCrossEntropy(logits:labels:)` used to take `Tensor<Float>` labels many months ago.

This change is good as is, because `Tensor<Int32>` labels are more efficient and they correctly correspond to the use of `softmaxCrossEntropy(logits:labels:)` API below.

PR #316, opened 2019-06-30
Fixed the matmul function and its VJP.

eaplatanios:
The current implementation of `matmul` was implicitly reshaping arguments if necessary for the case of batch matrix multiplications. This causes unexpected behavior and it's better to have the user reshape as necessary before calling the function. Also, the VJP implementation was wrong due to these reshapes. The batch matmul op supports broadcasting and so I fixed the VJP, while also accounting for broadcasting.

### DIFF
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -2114,16 +2114,11 @@ public func matmul<Scalar: Numeric>(
     _ rhs: Tensor<Scalar>,
     transposed transposeB: Bool = false
 ) -> Tensor<Scalar> {
-    switch (lhs.rank, rhs.rank) {
-    case (3..., 3...):
-        return Raw.batchMatMulV2(lhs, rhs, adjX: transposeA, adjY: transposeB)
-    case (2, 3...):
-        return Raw.batchMatMulV2(lhs.expandingShape(at: 1), rhs, adjX: transposeA, adjY: transposeB)
-    case (3..., 2):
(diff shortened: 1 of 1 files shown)


eaplatanios:
Is the description I added sufficient? If not, I can add more detail.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -2114,16 +2114,11 @@ public func matmul<Scalar: Numeric>(
...
-    case (3..., 3...):
-        return Raw.batchMatMulV2(lhs, rhs, adjX: transposeA, adjY: transposeB)
-    case (2, 3...):
-        return Raw.batchMatMulV2(lhs.expandingShape(at: 1), rhs, adjX: transposeA, adjY: transposeB)
-    case (3..., 2):
-        return Raw.batchMatMulV2(lhs, rhs.expandingShape(at: 1), adjX: transposeA, adjY: transposeB)
-    default:
-        return Raw.matMul(lhs, rhs, transposeA: transposeA, transposeB: transposeB)
+    if lhs.rank > 2 || rhs.rank > 2 {
+      // TODO: Conjugate to make compatible with the adjoint.
```
rxwei: File a bug for this?

eaplatanios: I'll fix this as part of this PR. Give me a few minutes.

eaplatanios: Actually, nvm this only matters if we support complex number types. Should I clarify this in the TODO alone or also file a bug?

rxwei: Filing a bug sounds good.

PR #318, opened 2019-06-30
Added a random seed helper function.

eaplatanios:


Sources/TensorFlow/Random.swift
```diff
@@ -18,6 +18,10 @@ import Darwin
 import Glibc
 #endif
 
+public func randomSeed() -> (Int32, Int32) {
```
rxwei: The reason I didn't add such a helper earlier was because default argument initializers cannot be `internal`, and I'm not sure a global function called `randomSeed()` is the best API. But for now, the benefit of switch switching to this seems greater than API uncertainties.

eaplatanios: I also had some concerns about the name, but in general I feel we should have a shared mechanism for default seeds.

rxwei: How about `randomSeedForTensorFlow()`?

eaplatanios: Done! :)

Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -148,9 +146,7 @@ public struct LSTMCell<Scalar: TensorFlowFloatingPoint>: RNNCell, VectorProtocol
     /// - Parameters:
     ///   - inputSize: The number of features in 2-D input tensors.
     ///   - hiddenSize: The number of features in 2-D hidden states.
-    public init(inputSize: Int, hiddenSize: Int,
-                seed: (Int32, Int32) = (Int32.random(in: Int32.min..<Int32.max),
-                                        Int32.random(in: Int32.min..<Int32.max))) {
+    public init(inputSize: Int, hiddenSize: Int, seed: (Int32, Int32) = randomSeedForTensorFlow()) {
```
rxwei: 100 columns

eaplatanios: Sorry about this. I'm addressing it in #319 .

eaplatanios:
I gave this a bit more thought and it may be good idea to have a global TF seed variable that is initialized randomly, but can also be set by the user so that it propagates everywhere. We would need to think this through a bit though.

rxwei:
Yeah, it would not be hard to set it up. You can add it to `Context` which is thread-safe.

eaplatanios:
It's not that straightforward because we also need to update the seed whenever it's used. Otherwise, ops like `Raw.statelessRandomUniform` will be deterministic across multiple calls. I'm currently working on a PR for this.

eaplatanios:
@rxwei I just opened a PR for the updated version (#320).

PR #320, opened 2019-06-30
 Added a better mechanism for handling random seeds.

eaplatanios:
The default values used for random seeds and random number generators are now consistently defined as the ones stored in the thread-local context. Note that this guarantees that both the seed and the generator are updated as needed to guarantee that stateless random TensorFlow ops do not output deterministic results.


rxwei:
Thanks Anthony. I'll review the details tonight.

eaplatanios:
Sounds good. Thanks Richard!

Sources/TensorFlow/Context.swift
```diff
@@ -47,6 +47,24 @@ public struct Context {
     /// The learning phase.
     public var learningPhase: LearningPhase = .inference
 
+    /// The random seed.
+    /// Note that whenever obtained, the random seed is also updated so that future stateless
```
rxwei:
```suggestion
    /// - Note: Whenever obtained, the random seed is also updated so that future stateless 
```

Sources/TensorFlow/Context.swift
```diff
@@ -47,6 +47,24 @@ public struct Context {
     /// The learning phase.
     public var learningPhase: LearningPhase = .inference
 
+    /// The random seed.
+    /// Note that whenever obtained, the random seed is also updated so that future stateless 
+    /// random TensorFlow op executions will result in non-deterministic results.
```
rxwei:
```suggestion
    ///   random TensorFlow op executions will result in non-deterministic results.
```

Sources/TensorFlow/Random.swift
```diff
@@ -18,13 +18,34 @@ import Darwin
...
+//===--------------------------------------------------------------------------------------------------------------===//
 // Random Number Generators
-//===------------------------------------------------------------------------------------------===//
+//===--------------------------------------------------------------------------------------------------------------===//
+
+/// A type-erased random number generator.
+///
+/// The `AnyRandomNumberGenerator` type forwards random number generating operations to an underlying random number 
+/// generator, hiding its specific underlying type.
+public struct AnyRandomNumberGenerator: RandomNumberGenerator {
```
rxwei: I don't see why this type is necessary. `RandomNumberGenerator` can be used as an existential already.

eaplatanios: We cannot have a stored property in `Context` without using this struct.

eaplatanios: I addressed the other comments except for this. I don't know if there is another way to define that stored property, but please let me know if you can think of one.

rxwei: Hmm, how so? I think you can add the following to `Context`.

```swift
var randomNumberGenerator: RandomNumberGenerator
```

eaplatanios: I tried that before but it caused an error. Let me check again and I'll report the error here.

eaplatanios: So in the standard library calls to e.g. [`Float.random(in:using:)`](https://developer.apple.com/documentation/swift/float/3020582-random) are made using generic types conforming to `RandomNumberGenerator`. If we try to pass something with type `RandomNumberGenerator` we get errors like this: `error: protocol type 'RandomNumberGenerator' cannot conform to 'RandomNumberGenerator' because only concrete types can conform to protocols`. This is why I introduced the type erased version and maintain the generic function calls.

rxwei: Ah I see. Right, protocol existentials don't conform to protocols. [This post](https://forums.swift.org/t/improving-the-ui-of-generics/22814) talked about ways to make this possible in the future, for example:

```swift
extension any RandomNumericGenerator: RandomNumericGenerator {
    ...
}
```

eaplatanios: Yes I actually found that a bit surprising. Thanks for the reference! I'll take a look at it later today but I imagine that would require more significant changes in the compiler. I can make the new struct internal for now and use it as an implementation details for handling type erasure. Would that be acceptable as a temporary solution?

rxwei: Making the property and the struct internal sounds good!

eaplatanios: Done. Note that even if this is approved and passes tests, it depends on #319 so we should merge that one first.

Sources/TensorFlow/Context.swift
```diff
@@ -83,13 +102,51 @@ public func withContext<R>(_ context: Context, _ body: () throws -> R) rethrows
...
+/// Calls the given closure within a context that has everything identical to the current context
+/// except for the given random number generator.
+///
+/// - Parameters:
+///   - randomNumberGenerator: A random number generator that will be set before the closure gets 
+///     called and restored after the closure returns.
+///   - body: A nullary closure. If the closure has a return value, that value is also used as the
+///     return value of the `withRandomNumberGeneratorForTensorFlow(_:_:)` function.
+/// - Returns: The return value, if any, of the `body` closure.
+public func withRandomNumberGeneratorForTensorFlow<G: RandomNumberGenerator, R>(
```
rxwei: I think the `ForTensorFlow` phrase can be removed here since it is not for TensorFlow ops like the tuple seed.
```suggestion
public func withRandomNumberGenerator<G: RandomNumberGenerator, R>(
```

eaplatanios: The reason I kept it is because we only default to this generator in TensorFlow functions which take generators as arguments so if there is a standard library function that uses a default random number generator (which I'm not aware of but I imagine there probably are), this function will not have any effect there. So, it could cause confusion if not qualified as being specific to TensorFlow.

eaplatanios:
Regarding source compatibility, let's start with #319 which changed a couple of things and I'll fix this after. I'll add TODOs to everything that I add for compatibility reasons, that can be removed later on.

Sources/TensorFlow/Context.swift
rxwei: My initial reasoning behind making the global variable `randomSeedForTensorFlow` was because a global variable is way more public than a type property.

In the `Context` case, since the type is already TensorFlow-specific (we can expect users to know that `Context` is from the `TensorFlow` module), I feel properties and methods don't need to mention TensorFlow. Similarly, `Context.local.randomSeed` is already without `ForTensorFlow`. WDYT?

eaplatanios: I see your point, although this is a global function with no reference to context in the function name or arguments. I can remove the `ForTensorFlow` and make the documentation more explicit about this. Should we also do the same for `withRandomSeedForTensorFlow`?

rxwei: Oh, I misunderstood—I thought it was a method. Keeping `ForTensorFlow` makes sense to me now.

eaplatanios: Sounds good! Thanks for bringing it up.

eaplatanios:
@rxwei this is now backwards compatible. I'm starting the CI tests and it should be ready to merged if approved after.

eaplatanios:
CI tests pass so this should be ready to merge once it gets an approval.

Sources/TensorFlow/Random.swift
```diff
@@ -18,13 +18,34 @@ import Darwin
...
+///
+/// The `AnyRandomNumberGenerator` type forwards random number generating operations to an underlying random number 
+/// generator, hiding its specific underlying type.
+internal struct AnyRandomNumberGenerator: RandomNumberGenerator {
+    @usableFromInline
+    internal var _rng: RandomNumberGenerator
+
+    /// - Parameter rng: A random number generator.
+    @inlinable
+    public init(_ rng: RandomNumberGenerator) {
```
rxwei:
```suggestion
    init(_ rng: RandomNumberGenerator) {
```

Sources/TensorFlow/Random.swift
```diff
@@ -18,13 +18,34 @@ import Darwin
...
+        self._rng = rng
+    }
+
+    @inlinable
+    public mutating func next() -> UInt64 {
```
rxwei:
```suggestion
    mutating func next() -> UInt64 {
```

Sources/TensorFlow/Random.swift
```diff
@@ -18,13 +18,34 @@ import Darwin
...
+/// Generates a new random seed for TensorFlow.
 public func randomSeedForTensorFlow() -> (Int32, Int32) {
     (Int32.random(in: Int32.min..<Int32.max), Int32.random(in: Int32.min..<Int32.max))
 }
 
-//===------------------------------------------------------------------------------------------===//
+//===--------------------------------------------------------------------------------------------------------------===//
 // Random Number Generators
-//===------------------------------------------------------------------------------------------===//
+//===--------------------------------------------------------------------------------------------------------------===//
```
rxwei: 100 columns.

Sources/TensorFlow/Random.swift
```diff
@@ -18,13 +18,34 @@ import Darwin
...
+
+/// A type-erased random number generator.
+///
+/// The `AnyRandomNumberGenerator` type forwards random number generating operations to an underlying random number
```
rxwei: 100 columns.

PR #321, opened 2019-06-30
Update documentation in vanilla Conv layers

jon-tow:
While attempting to build a model with Conv layers, I continually forgot the filterShape argument representation. Hopefully adding the below documentation will be helpful to others.

### DIFF
--- a/Sources/TensorFlow/Layers/Convolutional.swift
+++ b/Sources/TensorFlow/Layers/Convolutional.swift
@@ -68,10 +68,11 @@ public struct Conv1D<Scalar: TensorFlowFloatingPoint>: Layer {
     /// (input width + 2 * padding size - (dilation * (filter width - 1) + 1)) / stride + 1
     ///
     /// and padding size is determined by the padding scheme.
-    /// - Note: Padding size equals zero when using `.valid`.
     ///
     /// - Parameter input: The input to the layer [batch count, input width, input channel count].
     /// - Returns: The output of shape [batch count, output width, output channel count].
+    ///
+    /// - Note: Padding size equals zero when using `.valid`.
     @differentiable
     public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
         let conv = conv2D(
@@ -216,12 +217,13 @@ public struct Conv2D<Scalar: TensorFlowFloatingPoint>: Layer {
     /// / stride width + 1
     ///
     /// and padding sizes are determined by the padding scheme.
-    /// - Note: Padding size equals zero when using `.valid`.
     ///
     /// - Parameter input: The input to the layer of shape
     ///   [batch count, input height, input width, input channel count].
     /// - Returns: The output of shape
     ///   [batch count, output height, output width, output channel count].
+    ///
+    /// - Note: Padding size equals zero when using `.valid`.
     @differentiable
     public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
         return activation(conv2D(
@@ -368,12 +370,13 @@ public struct Conv3D<Scalar: TensorFlowFloatingPoint>: Layer {
     /// / stride width + 1
     ///
     /// and padding sizes are determined by the padding scheme.
-    /// - Note: Padding size equals zero when using `.valid`.
     ///
     /// - Parameter input: The input to the layer of shape
     ///   [batch count, input depth, input height, input width, input channel count].
     /// - Returns: The output of shape
     ///   [batch count, output depth, output height, output width, output channel count].
+    ///
+    /// - Note: Padding size equals zero when using `.valid`.
     @differentiable
     public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
         return activation(conv3D(
--- a/Sources/TensorFlow/Layers/Convolutional.swift
+++ b/Sources/TensorFlow/Layers/Convolutional.swift
@@ -18,9 +18,9 @@
 /// tensor of outputs.
 @frozen
 public struct Conv1D<Scalar: TensorFlowFloatingPoint>: Layer {
-    /// The 3-D convolution kernel `[width, inputChannels, outputChannels]`.
+    /// The 3-D convolution filter.
     public var filter: Tensor<Scalar>
-    /// The bias vector `[outputChannels]`.
+    /// The bias vector.
     public var bias: Tensor<Scalar>
     /// An activation function.
     public typealias Activation = @differentiable (Tensor<Scalar>) -> Tensor<Scalar>
@@ -37,8 +37,9 @@ public struct Conv1D<Scalar: TensorFlowFloatingPoint>: Layer {
     /// dilation and padding.
     ///
     /// - Parameters:
-    ///   - filter: The 3-D convolution kernel `[width, inputChannels, outputChannels]`.
-    ///   - bias: The bias vector `[outputChannels]`.
+    ///   - filter: The 3-D convolution filter of shape
+    ///     `[filter width, input channel count, output channel count]`.
+    ///   - bias: The bias vector of shape `[output channel count]`.
     ///   - activation: The element-wise activation function.
     ///   - stride: The stride of the sliding window for the temporal dimension.
     ///   - padding: The padding algorithm for convolution.
@@ -61,8 +62,8 @@ public struct Conv1D<Scalar: TensorFlowFloatingPoint>: Layer {
 
     /// Returns the output obtained from applying the layer to the given input.
     ///
-    /// - Parameter input: The input to the layer `[batchCount, width, inputChannels]`.
-    /// - Returns: The output `[batchCount, newWidth, outputChannels]`.
+    /// - Parameter input: The input to the layer `[batch count, width, input channel count]`.
+    /// - Returns: The output `[batch count, new width, output channel count]`.
     @differentiable
     public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
         let conv = conv2D(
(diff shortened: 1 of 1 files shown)


Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -61,8 +62,8 @@ public struct Conv1D<Scalar: TensorFlowFloatingPoint>: Layer {
 
     /// Returns the output obtained from applying the layer to the given input.
     ///
-    /// - Parameter input: The input to the layer `[batchCount, width, inputChannels]`.
-    /// - Returns: The output `[batchCount, newWidth, outputChannels]`.
+    /// - Parameter input: The input to the layer `[batch count, width, input channel count]`.
```
eaplatanios: You can call them "input width" and "output width" to be consistent with the channel counts.

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -37,8 +37,9 @@ public struct Conv1D<Scalar: TensorFlowFloatingPoint>: Layer {
     /// dilation and padding.
     ///
     /// - Parameters:
-    ///   - filter: The 3-D convolution kernel `[width, inputChannels, outputChannels]`.
-    ///   - bias: The bias vector `[outputChannels]`.
+    ///   - filter: The 3-D convolution filter of shape
+    ///     `[filter width, input channel count, output channel count]`.
```
rxwei: If this is a textual description, we should remove backquotes.

Sources/TensorFlow/Layers/Convolutional.swift
eaplatanios: Since you're at it, would you mind adding an equation showing how the output width is computed? That is often helpful to people and it would probably be good to do for the 2D and 3D versions also.

jon-tow: Sure thing! Good idea.

Sources/TensorFlow/Layers/Convolutional.swift
jon-tow: Do this for all such filter/bias descriptions?

Sources/TensorFlow/Layers/Convolutional.swift
eaplatanios: Great, thanks! :)

Sources/TensorFlow/Layers/Convolutional.swift
eaplatanios: How about you define `inputChannelCount` etc. as variables in the description of the layer, where you will also have equations in terms of them for computing the output width, and then you keep the backquotes and replace the descriptions with the variable names from earlier on?

jon-tow: So, add properties in those struct definitions for `inputChannelCount` etc.? If so, using such variables to compute outputs wouldn't make much sense given that the `conv` bindings handle it under the hood and don't leak the details up into the Operators api.

eaplatanios: Sorry I was only referring to the documentation string. As in, name these quantities in the doc string and then add the equation for how the output shape is computed, again in the doc string.

jon-tow: @eaplatanios When you get the chance could you please suggest a location in a doc string to add in the quantity names. I've update the source to remove backquotes and added the output computations. The only thing left is to introduce those quantities in the documentation. Thank you!

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -61,8 +62,15 @@ public struct Conv1D<Scalar: TensorFlowFloatingPoint>: Layer {
...
     /// Returns the output obtained from applying the layer to the given input.
     ///
-    /// - Parameter input: The input to the layer `[batchCount, width, inputChannels]`.
-    /// - Returns: The output `[batchCount, newWidth, outputChannels]`.
+    /// - Parameter input: The input to the layer [batch count, input width, input channel count].
+    /// - Returns: The output of shape [batch count, output width, output channel count],
+    ///   where output width is computed as:
+    ///
+    ///   output width =
+    ///   [(input width + 2 * padding size - (dilation * (filter width - 1) + 1)) / stride] + 1
```
```suggestion
    ///   (input width + 2 * padding size - (dilation * (filter width - 1) + 1)) / stride + 1
```

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -61,8 +62,15 @@ public struct Conv1D<Scalar: TensorFlowFloatingPoint>: Layer {
...
+    ///
+    ///   and padding size is determined by the padding scheme. Note that padding size equals zero
+    ///   when using .valid.
```
```suggestion
    ///   when using `.valid`.
```

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -61,8 +62,15 @@ public struct Conv1D<Scalar: TensorFlowFloatingPoint>: Layer {
 
     /// Returns the output obtained from applying the layer to the given input.
     ///
-    /// - Parameter input: The input to the layer `[batchCount, width, inputChannels]`.
-    /// - Returns: The output `[batchCount, newWidth, outputChannels]`.
+    /// - Parameter input: The input to the layer [batch count, input width, input channel count].
+    /// - Returns: The output of shape [batch count, output width, output channel count],
```
rxwei: The return description seems more like a summary of this operation, so I think it should be moved to the summary. Same for other such return descriptions. Returns descriptions are usually a simple noun phrase ending with a period.

jon-tow: Hey @rxwei. Is it okay if I keep breaks in between computations of the output dimensions like those in the comment string below?

/// The output width is computed as:                                                                        
///                                                                          
/// output width =                                   
/// (input width + 2 * padding size - (dilation * (filter width - 1) + 1)) / stride + 1                        
///                              
/// and padding size is determined by the padding scheme. Note that padding size equals zero                   
/// when using `.valid`.

rxwei: Yep, I think it is okay.
I would recommend turning the `Note ...` sentence into a `- Note: Padding size equals ...` documentation item.

Shashi456:
@rxwei the description of inputs for filter and strides make sense, but I don't think we should add the computation equations to documentation.  

I think it becomes a difficult standard to maintain in the long run, plus notations always become an issue if they aren't uniform. 

Finally, neither of the framework we take as benchmarks maintain these either. 

Just a thought, I do like the idea but it had its issues.

jon-tow:
@Shashi456  I like the idea of explicitly showing the output shape computations as I found it very helpful when starting out. I'm all for clarity but I do see your point. 

One thing I noticed was that if and when we implement the data-format arg for layers, the description of shapes will have to change to reflect the possibility of having NCHW vs NHWC...which adds to the maintenance.
 
@rxwei Do you think we should ax the computation docs?

eaplatanios:
I generally feel these formulas are super useful and in this particular case (when it comes to convolutions) confuse a lot of users, so I'm leaning towards adding them. Also, they are not that hard to maintain and do not really depend on the `NCHW` or the `NHWC` format. This would only result in changing the order of the variables in the shapes, but the formulas should stay the same.

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -68,10 +68,11 @@ public struct Conv1D<Scalar: TensorFlowFloatingPoint>: Layer {
     /// (input width + 2 * padding size - (dilation * (filter width - 1) + 1)) / stride + 1
     ///
     /// and padding size is determined by the padding scheme.
-    /// - Note: Padding size equals zero when using `.valid`.
     ///
     /// - Parameter input: The input to the layer [batch count, input width, input channel count].
     /// - Returns: The output of shape [batch count, output width, output channel count].
+    ///
```
rxwei: Remove empty line.

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -216,12 +217,13 @@ public struct Conv2D<Scalar: TensorFlowFloatingPoint>: Layer {
     /// / stride width + 1
     ///
     /// and padding sizes are determined by the padding scheme.
-    /// - Note: Padding size equals zero when using `.valid`.
     ///
     /// - Parameter input: The input to the layer of shape
     ///   [batch count, input height, input width, input channel count].
     /// - Returns: The output of shape
     ///   [batch count, output height, output width, output channel count].
+    ///
```
rxwei: Same here.

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -368,12 +370,13 @@ public struct Conv3D<Scalar: TensorFlowFloatingPoint>: Layer {
     /// / stride width + 1
     ///
     /// and padding sizes are determined by the padding scheme.
-    /// - Note: Padding size equals zero when using `.valid`.
     ///
     /// - Parameter input: The input to the layer of shape
     ///   [batch count, input depth, input height, input width, input channel count].
     /// - Returns: The output of shape
     ///   [batch count, output depth, output height, output width, output channel count].
+    ///
```
rxwei: Same here.

rxwei:
@eaplatanios This PR actually hasn't run CI.

Sources/TensorFlow/Layers/Convolutional.swift
jon-tow: Oops sorry about that. The other occurrences of `- Note:` in the file have an empty line before them.

rxwei: No worries! It would be great to tidy up all of other occurrences as well.

jon-tow: I'm on it!

eaplatanios:
@rxwei sorry I missed that. What's the standard way of handling this? Should we revert this or should I just open a new PR remove the lines you suggested and running tests there? For this PR alone this should be ok as only documentation strings were changed.

eaplatanios:
I could remove the lines as part of #319 to move faster if you'd prefer that.

PR #322, opened 2019-07-01
Make all layers' tangent vectors conform to `VectorProtocol & ElementaryFunctions & PointwiseMultiplicative`.

eaplatanios:
This should be merged after #319 and #320 as I have already merged their changes into this.

You can see the changes introduced in this PR in isolation [here](https://github.com/tensorflow/swift-apis/pull/322/commits/813d58ed8be3deabc782d3b32855b629ef635e94).


rxwei:
I don't think introducing new protocols (`Optimizable` or `ParameterlessLayer`) is the right direction here. Adding dummy protocols is sort of an anti-pattern in Swift. Ultimately we just want certain constraints to be able to trigger derived conformances so that `TangentVector` gets `VectorProtocol` conformance. Have you tried adding an explicit `struct TangentVector` to each layer?

eaplatanios:
Sorry, `Optimizable` is not a new protocol, but rather a type alias grouping multiple constraints together. The issue with `struct TangentVector` is that it needs to be explicitly added to all parameter-less layers which is not very user friendly. Ideally we want users to be able to create parameter-less layers without worrying about `TangentVector` explicitly. Also, just adding that struct without the extra `EmptyParameterSet` struct I added, results in compiler errors. For `EmptyParameterSet` I had to add explicit implementations for `VectorProtocol` because the compiler was unable to derive them otherwise.

The nice thing about this solution is that the only change that needs to be made to support parameter-less layers is that they need to declare conformance to `ParameterlessLayer` instead of `Layer` and everything else works out fine on the compiler side.

eaplatanios:
The `Optimizable` type alias is defined [here](https://github.com/tensorflow/swift-apis/blob/813d58ed8be3deabc782d3b32855b629ef635e94/Sources/TensorFlow/Optimizers/Optimizer.swift#L15).

Sources/TensorFlow/Optimizers/Optimizer.swift
```diff
@@ -12,12 +12,15 @@
...
+public typealias Optimizable =
+    VectorProtocol & ElementaryFunctions & PointwiseMultiplicative & KeyPathIterable
+
 /// A numerical optimizer.
 ///
 /// Optimizers apply an optimization algorithm to update the differentiable models.
 public protocol Optimizer {
     /// The type of the model whose parameters are optimized.
-    associatedtype Model: Differentiable
+    associatedtype Model: Differentiable where Model.TangentVector: Optimizable
```
rxwei: I feel such a requirement makes optimizers no longer general in it forces optimizers to conform to a series of auxiliary protocols, while each optimizer has quite different requirements. I think we should think about the proper Swift Evolution route: propose proper changes to the protocol hierarchy so that a conformance constraint to a certain protocol gives more available operations.

eaplatanios: Yes that's a good point. In this case, I can revert back to the specific requirements of each optimizer and only have layer tangent vectors conform to all of those for now, so that they can be used seamlessly with all optimizers. How does that sound as a temporary solution?

eaplatanios:
I really like the approach you propose and also feel we are not far from that. However, I'm not very familiar with how swift evolution works and how easy it would be to change `Numeric` depending on how often it's used in existing code.

Note that to completely remove the need for `KeyPathIterable` from optimizers we would also need to support `abs(_:)`, which I don't think is in `ElementaryFunctions`.

eaplatanios:
I removed `Optimizable` and so the remaining changes simply make all layers usable with any optimizer we want.

rxwei:
> However, I'm not very familiar with how swift evolution works and how easy it would be to change `Numeric` depending on how often it's used in existing code.

ABI stability for `Numeric` is the biggest issue that's blocking this from happening right now AFAIK.

> Note that to completely remove the need for `KeyPathIterable` from optimizers we would also need to support `abs(_:)`, which I don't think is in `ElementaryFunctions`.

Good point. If `Magnitude` can get removed from `Numeric`, it means that we can derive both `Numeric` and `SignedNumeric`. `abs(_:)` sounds like something that can be added to `SignedNumeric`.

It would be great to ask a quick question about removing `Magnitude` from `Numeric` on the evolution forum. Are you interested in starting a discussion there?

eaplatanios:
I can do that, but I first need to understand a bit more about how `magnitude` is being used. It sounds to me like the semantics of magnitude are equivalent to `abs(_:)`?

Also, just to get a bit of understanding about the evolution process,  how do we go about proposing to remove some functionality? I imagine we need a good reason to do so and in this case I don't necessarily see why it is necessary to remove it. Couldn't we simply make `Magnitude` equal to `Self` for tensors and set the getter to be the `abs(_:)` function? This way we also don't need to add that to `SignedNumeric` and we don't break compatibility with the current `Numeric` as defined in the standard library.

rxwei:
> The nice thing about this solution is that the only change that needs to be made to support parameter-less layers is that they need to declare conformance to `ParameterlessLayer` instead of `Layer` and everything else works out fine on the compiler side.

The issue is that it is a public API that does not guarantee the semantics of conforming types. A type can totally declare a `ParameterlessLayer` and have parameters or its own `TangentVector`, in which case the `ParameterlessLayer` conformance would contradict the fact that such layer has parameters. Any public protocol would change the user model quite significantly, and should be considered thoroughly. In this particular case, instead of providing a default empty `TangentVector` to conforming types, I don't see `ParameterlessLayer` can provide any useful information to a generic algorithm that declares a conformance constraint to this protocol.

I'd recommend against directly dropping in public, user-visible protocols, but instead adding a custom `TangentVector` structure to any such parameterless layer. This won't add any cognitive burden on the user in looking at the protocol hierarchy. After this, a natural next step would be to figure out a default `VectorSpaceScalar` for empty structs. Perhaps we can define an empty struct that conforms to `Numeric` and can act as the default `VectorSpaceScalar` that the compiler derives.

eaplatanios:
Oh I see your point. Yes I also dislike that it does not guarantee the semantics of the conforming type. Let me see if I can work out something simpler with the empty struct.

rxwei:
> I can do that, but I first need to understand a bit more about how `magnitude` is being used. It sounds to me like the semantics of magnitude are equivalent to `abs(_:)`?

That is a good point! Any thoughts, @stephentyrone? It would be great to remove even the `Comparable` constraint from `Magnitude`. The `Comparable` constraint is what's really making `Numeric` non-derivable.

eaplatanios:
Yes I agree. I don't think `Comparable` is necessary there. This actually makes me wonder where that constraint is being used.

eaplatanios:
I just did a quick search through the code and it's only being used in very few places. In principle, we can add the `Comparable` constraint there separately.

rxwei:
Yeah. In the case that the `Comparable` constraint can't be removed, we can technically still derive `Numeric` but it will require deriving `Comparable` for aggregates (comparison followed by reduction), and I'm not sure if it's the right thing.

eaplatanios:
Do you think an evolution pitch would be easier if I first try to remove that constraint and see how many changes this results into in the standard library?

rxwei:
> Do you think an evolution pitch would be easier if I first try to remove that constraint and see how many changes this results into in the standard library?

I suspect there will be some pushback against this on the forum, so it'd be nice to see the reaction from the community first. If there's a serious ABI concern, it's probably just not possible.

eaplatanios:
I created a discussion topic [here](https://forums.swift.org/t/removing-the-comparable-constraint-for-numeric-magnitude/26498). I'm not sure if what I said is sufficient to start this discussion, but please feel free to add any additional comments you may have.

In the meantime, do you think that the changes introduced in this PR are acceptable so we can support optimizers with all layers as long as I remove `ParameterlessLayer` and add `TangentVector` structs to all parameter-less layers? I'll try to push these changes in a little bit.

rxwei:
Nice post! Just a note: `ElementaryFunctions` wasn't designed as part of Swift for TensorFlow. We only made it derivable.

rxwei:
> In the meantime, do you think that the changes introduced in this PR are acceptable so we can support optimizers with all layers as long as I remove `ParameterlessLayer` and add `TangentVector` structs to all parameter-less layers? I'll try to push these changes in a little bit.

Yep!

eaplatanios:
Thanks and good catch! :) I edited the post.

eaplatanios:
So, if I don't define `allDifferentiableVariables` and `move(along:)` as I was doing here:
```swift
public extension ParameterlessLayer {
    var allDifferentiableVariables: EmptyParameterSet {
        get { return EmptyParameterSet() }
        set {}
    }

    mutating func move(along direction: EmptyParameterSet) {}
}
```
I get the following compiler error:
```
Assertion failed: (Storage.hasVal), function getPointer, file /Users/eaplatanios/Development/GitHub/swift-source-temp/llvm/include/llvm/ADT/Optional.h, line 180.
Stack dump:
0.      Program arguments: /Library/Developer/Toolchains/swift-tensorflow-LOCAL-2019-06-30-a.xctoolchain/usr/bin/swift -frontend -c /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Bindings/EagerExecution.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Bindings/RawOpsGenerated.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Bindings/TFTensorOperation.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Context.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/ArrayOps.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/DataTypes.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/DifferentialOperators.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/Execution.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/LazyTensorOperation.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/LazyTensorTFFunctionBuilder.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/LazyTensorTrace.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/PythonConversion.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/Runtime.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/ShapedArray.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/StringTensor.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/Tensor.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/TensorGroup.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/TensorHandle.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/TensorProtocol.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/TensorShape.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/Threading.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Core/Utilities.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Initializers.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layer.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/BackwardsCompatibility.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Convolutional.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Core.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Embedding.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Initialization.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Normalization.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Pooling.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Recurrent.swift -primary-file /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Upsampling.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Loss.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Operators/Basic.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Operators/Comparison.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Operators/Dataset.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Operators/Math.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Operators/NN.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Optimizers/MomentumBased.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Optimizers/Optimizer.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Optimizers/SGD.swift /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Random.swift -emit-module-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Upsampling~partial.swiftmodule -emit-module-doc-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Upsampling~partial.swiftdoc -emit-dependencies-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Upsampling.d -emit-reference-dependencies-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Upsampling.swiftdeps -target x86_64-apple-macosx10.10 -enable-objc-interop -sdk /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk -I /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug -I /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/usr/lib -F /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/Library/Frameworks -enable-testing -g -module-cache-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/ModuleCache -swift-version 4.2 -Onone -D SWIFT_PACKAGE -D DEBUG -color-diagnostics -enable-anonymous-context-mangled-names -parse-as-library -module-name TensorFlow -o /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/TensorFlow.build/Layers/Upsampling.swift.o -index-store-path /Users/eaplatanios/Development/GitHub/swift-apis/.build/x86_64-apple-macosx/debug/index/store -index-system-modules 
1.      Swift version 5.1-dev (LLVM 7d54a7c711, Swift f62bd5687d)
2.      While type-checking 'UpSampling3D' (at /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Upsampling.swift:73:8)
3.      While type-checking 'repeatingElements(_:alongAxis:count:)' (at /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Upsampling.swift:89:13)
4.      While validating 'repeatingElements(_:alongAxis:count:)' (at /Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Upsampling.swift:89:13)
0  swift                    0x00000001128d3955 llvm::sys::PrintStackTrace(llvm::raw_ostream&) + 37
1  swift                    0x00000001128d2998 llvm::sys::RunSignalHandlers() + 248
2  swift                    0x00000001128d3f48 SignalHandler(int) + 264
4  libsystem_platform.dylib 0x0000000000004250 _sigtramp + 2321839888
7  swift                    0x0000000112d5e583 swift::AnyFunctionType::getAutoDiffAssociatedFunctionType(swift::AutoDiffParameterIndices*, unsigned int, unsigned int, swift::AutoDiffAssociatedFunctionKind, llvm::function_ref<llvm::Optional<swift::ProtocolConformanceRef> (swift::CanType, swift::Type, swift::ProtocolDecl*)>, swift::GenericSignature*, bool) (.cold.25) + 35
8  swift                    0x000000010fad941c swift::AnyFunctionType::getAutoDiffAssociatedFunctionType(swift::AutoDiffParameterIndices*, unsigned int, unsigned int, swift::AutoDiffAssociatedFunctionKind, llvm::function_ref<llvm::Optional<swift::ProtocolConformanceRef> (swift::CanType, swift::Type, swift::ProtocolDecl*)>, swift::GenericSignature*, bool) + 3676
9  swift                    0x000000010f6b7d07 swift::ASTVisitor<(anonymous namespace)::AttributeChecker, void, void, void, void, void, void>::visit(swift::DeclAttribute*) + 17655
10 swift                    0x000000010f6b37f1 swift::TypeChecker::checkDeclAttributes(swift::Decl*) + 49
11 swift                    0x000000010f6f4988 swift::TypeChecker::validateDecl(swift::ValueDecl*) + 6008
12 swift                    0x000000010f6fd1c0 (anonymous namespace)::DeclChecker::visitFuncDecl(swift::FuncDecl*) + 48
13 swift                    0x000000010f6efde3 (anonymous namespace)::DeclChecker::visit(swift::Decl*) + 243
14 swift                    0x000000010f6fb5ab (anonymous namespace)::DeclChecker::visitStructDecl(swift::StructDecl*) + 139
15 swift                    0x000000010f6f005d (anonymous namespace)::DeclChecker::visit(swift::Decl*) + 877
16 swift                    0x000000010f6efce6 swift::TypeChecker::typeCheckDecl(swift::Decl*) + 38
17 swift                    0x000000010f790bdc swift::performTypeChecking(swift::SourceFile&, swift::TopLevelContext&, swift::OptionSet<swift::TypeCheckingFlags, unsigned int>, unsigned int, unsigned int, unsigned int, unsigned int, unsigned int) + 940
18 swift                    0x000000010eec9d71 swift::CompilerInstance::parseAndCheckTypesUpTo(swift::CompilerInstance::ImplicitImports const&, swift::SourceFile::ASTStage_t)::$_1::operator()(swift::SourceFile&) const + 81
19 swift                    0x000000010eec574b swift::CompilerInstance::parseAndCheckTypesUpTo(swift::CompilerInstance::ImplicitImports const&, swift::SourceFile::ASTStage_t) + 507
20 swift                    0x000000010eec504a swift::CompilerInstance::performSemaUpTo(swift::SourceFile::ASTStage_t) + 618
21 swift                    0x000000010ecb2e92 performCompile(swift::CompilerInstance&, swift::CompilerInvocation&, llvm::ArrayRef<char const*>, int&, swift::FrontendObserver*, swift::UnifiedStatsReporter*) + 1730
22 swift                    0x000000010ecb17f1 swift::performFrontend(llvm::ArrayRef<char const*>, char const*, void*, swift::FrontendObserver*) + 3025
23 swift                    0x000000010ec59bf9 main + 729
24 libdyld.dylib            0x00007fff757d13d5 start + 1
```
which seems to me like a bug.

An example of how I define a parameter-less layer now, giving me the above error, is this:
```swift
public struct GlobalMaxPool2D<Scalar: TensorFlowFloatingPoint>: Layer {
    public typealias TangentVector = EmptyParameterSet

    public init() {}

    ///
    /// - Returns: The output.
    @differentiable
        return input.max(squeezingAxes: [1, 2])
    }
}
```

eaplatanios:
Side comment about `Magnitude`. As a backup option we can always define it as `Float` or `Double` and then have it be the sum of the magnitudes of all things in the aggregate type. For tensors this could be defined as the Frobenius norm of the tensor. This would give us reasonable semantics.

rxwei:
What if you define both 
```swift
public struct AllDifferentiableVariables: Differentiable {}
```
and
```swift
public struct TangentVector: Differentiable, VectorProtocol {
    public typealias VectorSpaceScalar = Float
}
```
?

eaplatanios:
This results in a lot of errors indicating that no conformances were derived:
```
/Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Convolutional.swift:616:15: error: type 'ZeroPadding3D<Scalar>.TangentVector' does not conform to protocol 'VectorProtocol'
public struct TangentVector: Differentiable, VectorProtocol {
              ^
Swift.VectorProtocol:3:10: note: protocol requires function 'adding' with type '(ZeroPadding3D<Scalar>.TangentVector.VectorSpaceScalar) -> ZeroPadding3D<Scalar>.TangentVector'; do you want to add a stub?
    func adding(_ x: Self.VectorSpaceScalar) -> Self
         ^
Swift.VectorProtocol:5:10: note: protocol requires function 'subtracting' with type '(ZeroPadding3D<Scalar>.TangentVector.VectorSpaceScalar) -> ZeroPadding3D<Scalar>.TangentVector'; do you want to add a stub?
    func subtracting(_ x: Self.VectorSpaceScalar) -> Self
         ^
Swift.VectorProtocol:7:10: note: protocol requires function 'scaled(by:)' with type '(ZeroPadding3D<Scalar>.TangentVector.VectorSpaceScalar) -> ZeroPadding3D<Scalar>.TangentVector'; do you want to add a stub?
    func scaled(by scalar: Self.VectorSpaceScalar) -> Self
         ^
/Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Convolutional.swift:614:15: error: type 'ZeroPadding3D<Scalar>.TangentVector' does not conform to protocol 'ElementaryFunctions'
public struct ZeroPadding3D<Scalar: TensorFlowFloatingPoint>: Layer {
              ^
/Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layers/Convolutional.swift:614:15: error: type 'ZeroPadding3D<Scalar>' does not conform to protocol 'Differentiable'
public struct ZeroPadding3D<Scalar: TensorFlowFloatingPoint>: Layer {
              ^
Swift.Differentiable:2:16: note: candidate has non-matching type 'Self'
    public var allDifferentiableVariables: Self { get set }
               ^
Swift.Differentiable:2:26: note: candidate would match if 'ZeroPadding3D<Scalar>' was the same type as 'ZeroPadding3D<τ_0_0>.TangentVector'
    public mutating func move(along direction: Self.TangentVector)
                         ^
Swift.Differentiable:4:9: note: protocol requires property 'allDifferentiableVariables' with type 'ZeroPadding3D<Scalar>.AllDifferentiableVariables'; do you want to add a stub?
    var allDifferentiableVariables: Self.AllDifferentiableVariables { get set }
        ^
Swift.Differentiable:5:19: note: protocol requires function 'move(along:)' with type '(ZeroPadding3D<Scalar>.TangentVector) -> ()'; do you want to add a stub?
    mutating func move(along direction: Self.TangentVector)
                  ^
```

eaplatanios:
Is there anything similar to the `final` keyword from Java/Scala in Swift? I.e., a way to declare that a type/property cannot be overridden by conforming types?

rxwei:
This probably means derived conformances are currently hard coded to not trigger derive if there are no members. We should fix that.

rxwei:
> Is there anything similar to the `final` keyword from Java/Scala in Swift? I.e., a way to declare that a type/property cannot be overridden by conforming types?

There is `final`, but `final` does not apply to protocols. All protocol requirements are final once implemented.

eaplatanios:
> This probably means derived conformances are currently hard coded to not trigger derive if there are no members. We should fix that.

I looked a bit into this earlier and it seems that [this assertion](https://github.com/apple/swift/blob/dd2067ab0c0e035bd8b421777a4a6030c04e6dd5/lib/Sema/DerivedConformanceDifferentiable.cpp#L256) is failing (at least when the compiler was crashing). Maybe this is also related to what happens when there are no members.

eaplatanios:
Actually never mind, [this line](https://github.com/apple/swift/blob/dd2067ab0c0e035bd8b421777a4a6030c04e6dd5/lib/Sema/DerivedConformanceDifferentiable.cpp#L621) seems to be where the check happens. Or at least one such check.

rxwei:
Yeah, looks like that should be fixed.

Are you also interested in implementing a `@memberwise` attribute for triggering `AdditiveArithmetic`, `VectorProtocol`, `ElementaryFunctions`, and `Differentiable` derived conformances? This is described [here](https://forums.swift.org/t/additivearithmetic-conformance-synthesis-for-structs/26159).

eaplatanios:
Unfortunately I don't have time to look into it this week because I'm rushing with some research projects and will also be traveling Wednesday-Sunday. I may be able to look into it when I get back but I can't know yet.

eaplatanios:
Quick observation: removing the condition in [this line](https://github.com/apple/swift/blob/dd2067ab0c0e035bd8b421777a4a6030c04e6dd5/lib/Sema/DerivedConformanceDifferentiable.cpp#L621) does not fix the problem.

eaplatanios:
> Side comment about `Magnitude`. As a backup option we can always define it as `Float` or `Double` and then have it be the sum of the magnitudes of all things in the aggregate type. For tensors this could be defined as the Frobenius norm of the tensor. This would give us reasonable semantics.

I believe we can just go with this as it's technically not wrong given the loose semantics that `magnitude` already has.

eaplatanios:
@rxwei could we have the automatic derivation for `VectorProtocol` derive `VectorSpaceScalar` as an empty struct itself, for empty structs? I don't why an empty struct cannot conform to `AdditiveArithmetic`.

eaplatanios:
Also, I realized that the `ParameterlessLayer` approach I original implemented for this does guarantee the semantics of the conforming type as it forces `AllDifferentiableVariables` to be equal to `EmptyParameterSet`. No matter what properties the users of the protocol define in conforming structs, these properties will not be differentiating and will thus not be treated as model parameters. So, I don't think semantics is a problem here. What do you think?

rxwei:
> Also, I realized that the `ParameterlessLayer` approach I original implemented for this does guarantee the semantics of the conforming type as it forces `AllDifferentiableVariables` to be equal to `EmptyParameterSet`.

I don't think that's the case. The user can always declare a custom `TangentVector` and go against the "parameterless" promise as specified by such a protocol. Protocols that do not declare new requirements are kind of an anti-pattern in Swift because they do not provide useful information for generic algorithms. There are only extremely few cases where we considered carefully and used these empty protocols, namely `TensorFlowFloatingPoint`, but it was only motivated by the fact that `typealias` doesn't support `where`-clauses yet.

rxwei:
My recommendation is going a different route:
- Phase 1: Make parameterless layers compatible with optimizers (as we discussed upstream).
  - Fix derived conformances for `VectorProtocol` so that it can be derived for empty structs so long as they declare a `VectorSpaceScalar`.
  - Declare an empty `TangentVector` and `AllDifferentiableVariables` in each of the parameterless layers.
- Phase 2: Remove special cases in `TangentVector` conformance derivation.
  - Fix [the bug](https://forums.swift.org/t/synthesized-member-types-and-type-checking/26386) where one cannot extend a compiler-synthesized `TangentVector`.
  - Remove the special case that "if all members conform to `VectorProtocol`/`ElementaryFunctions`, conform `TangentVector` to `VectorProtocol`/`ElementaryFunctions`" from `Differentiable` derived conformances.
  - Extend `TangentVector`s to conform to `VectorProtocol`/`ElementaryFunctions` for each layer.
- Phase 3: Pitch `VectorProtocol` to Swift Evolution. 
  - Carefully examine whether `VectorProtocol` should be derived for empty structs.  If so, consider adding an empty field type to stdlib that can act as the default `VectorSpaceScalar` (I feel it's unlikely, but just something for discussion).
  - Constrain `TangentVector` on a conformance to `VectorProtocol` in addition to `AdditiveArithmetic`.

eaplatanios:
That plan sounds good, although phase 2 seems like it might require a lot
of work. Do you think you’ll be able to tackle phase 1, fixing the derived
conformances?

Regarding the `ParameterlessLayer` protocol approach, note that there is a
`AllDifferentiableVariables == TangentVector` constraint on `Layer` and the
additional constraint `TangentVector == EmptyParameterSet` on
`ParameterlessLayer` so I still don’t think it’s possible to override the
expected behavior. Could you please describe how that would be possible? I
understand the hesitation with respect to introducing a new protocol, but I
also want to understand why it cannot guarantee the desired functionality.

dan-zheng:
(Note: I filed [TF-632](https://bugs.swift.org/browse/TF-632) to track phase 2. I'm interested in working on it but won't start for a while.)

rxwei:
> That plan sounds good, although phase 2 seems like it might require a lot of work. Do you think you’ll be able to tackle phase 1, fixing the derived conformances?

Unfortunately I don't be able to tackle the derived conformances fix this week. Feel free to tackle it when you get a chance. It should be a chance of synthesis condition plus a bunch of tests.

> Could you please describe how that would be possible? I understand the hesitation with respect to introducing a new protocol, but I also want to understand why it cannot guarantee the desired functionality.

Sorry I omitted some details here. `AllDifferentiableVariables` is going to be deprecated and removed from the `Differentiable` protocol. This means that we will not be able to have this sort of mutually recursive constraint when the removal is implemented.

eaplatanios:
I might tackle it if I get a chance in the next few days, or later today.

Regarding your second comment, even if `AllDifferentiableVariables` is
deprecated, `ParameterlessLayer` adds the constraint that `TangentVector ==
EmptyParameterSet` which the user cannot override, so doesn’t this
guarantee the desired behavior?

rxwei:
> Regarding your second comment, even if `AllDifferentiableVariables` is deprecated, `ParameterlessLayer` adds the constraint that `TangentVector == EmptyParameterSet` which the user cannot override, so doesn’t this guarantee the desired behavior?

Yeah, you are right that `TangentVector` is property constrained. The other reservation I had is that adding a new public API `EmptyParameterSet` is pretty suboptimal. Perhaps we can call it `EmptyTangentVector` or something.

eaplatanios:
I could make that change. Also, when we fix the derived conformances for
empty structs we can completely remove that struct and set `TangentVector`
to be an empty struct.

eaplatanios:
Done. This should be ready for review now.

eaplatanios:
Also, is it ok if I make a separate PR to add a `ParameterScalar` associated type to `Layer`? I believe it's safe to assume that all parameters of a layer are of a single type and it makes my life easier with some stuff I'm working on which "collect" all parameters using `KeyPathIterable`. In this case, it's nice to be explicit about the type of the parameters so that "unconsidered" types are not silently ignored (as it happens with the key-path-based optimizers, for example).

I attempted this change and only 2 lines of code change in all of the `TensorFlow` module.

Sources/TensorFlow/Layer.swift
```diff
@@ -41,6 +43,31 @@ public extension Layer {
     }
 }
 
+public struct EmptyTangentVector: Differentiable & VectorProtocol & ElementaryFunctions &
```
dan-zheng: Could you please add a doc comment to newly added types (`EmptyTangentVector` and `ParameterlessLayer`)?

You can use `X, Y` here instead of `X & Y` to save one character per composed protocol.
```suggestion
public struct EmptyTangentVector: Differentiable, VectorProtocol, ElementaryFunctions,
```

rxwei:
> Also, is it ok if I make a separate PR to add a `ParameterScalar` associated type to `Layer`? I believe it's safe to assume that all parameters of a layer are of a single type and it makes my life easier with some stuff I'm working on which "collect" all parameters using `KeyPathIterable`. In this case, it's nice to be explicit about the type of the parameters so that "unconsidered" types are not silently ignored (as it happens with the key-path-based optimizers, for example).

This direction was specifically considered, implemented, and rejected in the `Parameterized`/`ParameterGroup` effort. We've decided not to hard-code any scalar type in the `Layer` protocol because we'd like to keep APIs future-proof (supporting multiple-precision training, for example).

Sources/TensorFlow/Layer.swift
```diff
@@ -41,6 +43,31 @@ public extension Layer {
...
+    public func scaled(by scalar: Float) -> EmptyTangentVector { self }
+    public mutating func scale(by scalar: Float) {}
+}
+
+public protocol ParameterlessLayer: Layer
+where AllDifferentiableVariables == EmptyTangentVector {}
+
+public extension ParameterlessLayer {
+    var allDifferentiableVariables: EmptyTangentVector {
+        get { return EmptyTangentVector() }
```
rxwei:
```suggestion
        get { EmptyTangentVector() }
```

Sources/TensorFlow/Layer.swift
```diff
@@ -41,6 +43,31 @@ public extension Layer {
...
+    public func adding(_ x: Float) -> EmptyTangentVector { self }
+    public mutating func add(_ x: Float) {}
+    public func subtracting(_ x: Float) -> EmptyTangentVector { self }
+    public mutating func subtract(_ x: Float) {}
+    public func scaled(by scalar: Float) -> EmptyTangentVector { self }
+    public mutating func scale(by scalar: Float) {}
+}
+
+public protocol ParameterlessLayer: Layer
+where AllDifferentiableVariables == EmptyTangentVector {}
```
rxwei: Indent by 4

Sources/TensorFlow/Layer.swift
```diff
@@ -56,7 +83,7 @@ public extension Layer {
     @usableFromInline
     internal func _vjpInferring(from input: Input)
         -> (value: Output, pullback: (Output.TangentVector)
-            -> (TangentVector, Input.TangentVector)) {
+            -> (AllDifferentiableVariables, Input.TangentVector)) {
```
rxwei: Could you please revert this change? `TangentVector` is a correct type (though `AllDifferentiableVariables` is constrained to be equal to it) and will be the only correct type after we deprecate `AllDifferentiableVariables`.

rxwei:
The PR title should be corrected to "Make all layers' tangent vectors conform" to reflect what's actually going on.

Sources/TensorFlow/Layer.swift
eaplatanios: I changed this because if I don't I get this error:
```
/Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layer.swift:82:6: error: function result's 'pullback' type does not match 'inferring(from:)'
    @differentiating(inferring(from:))
     ^
/Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layer.swift:85:38: note: 'pullback' does not have expected type '(Self.Output.TangentVector) -> (Self.AllDifferentiableVariables, Self.Input.TangentVector)'
        -> (value: Output, pullback: (Output.TangentVector)
                                     ^~~~~~~~~~~~~~~~~~~~~~
/Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Layer.swift:77:10: note: 'inferring(from:)' defined here
    func inferring(from input: Input) -> Output {
         ^
```

rxwei: Hmm, that looks like a bug. I guess we can deal with it later.

eaplatanios: Yeah I assumed it was a bug, but didn't look into it too much because I was aware that `AllDifferentiableVariables` is going away soon. :)

Sources/TensorFlow/Layer.swift
eaplatanios: Thanks Dan!

eaplatanios:
> The PR title should be corrected to "Make all layers' tangent vectors conform" to reflect what's actually going on.

Good catch, thanks Richard! I made all the requested changes so I'll run tests and merge if they pass.

eaplatanios:
> This direction was specifically considered, implemented, and rejected in the `Parameterized`/`ParameterGroup` effort. We've decided not to hard-code any scalar type in the `Layer` protocol because we'd like to keep APIs future-proof (supporting multiple-precision training, for example).

Is it possible to add a `recursivelyAllWritableKeyPaths(conformingTo:)` function that accepts a protocol that the traversed key path types must conform to?

rxwei:
It is only possible for non-PAT protocols. Protocol metatypes cannot be of PATs, for example:
```swift
test.swift:1:9: error: protocol 'Differentiable' can only be used as a generic constraint because it has Self or associated type requirements
let x = Differentiable.self
        ^
```

eaplatanios:
Thanks!

eaplatanios:
@rxwei @dan-zheng I just bumped into one more requirement for aggregates over tensors that may be good to think about. Say we want to add support for clipping gradient based on the global norm. That requires us to compute the norm of each tensor in the aggregate (`norm[t]`) and then compute:
```
globalNorm = sqrt(sum([norm[t].squared() for t in tensors]))
```
Say we can compute `sqrt` and `squared()` using a conformance to `ElementaryFunctions`. How do we go about the `sum` reduction over the aggregate?

eaplatanios:
For reference, this is the current hack solution I'm using that I don't particularly like since layers don't have a `ParameterScalar` type:
```swift
extension KeyPathIterable {
  public mutating func clipByGlobalNorm<Scalar: TensorFlowFloatingPoint>(clipNorm: Scalar) {
    let clipNorm = Tensor<Scalar>(clipNorm)
    var globalNorm = Tensor<Scalar>(zeros: [])
    for kp in self.recursivelyAllWritableKeyPaths(to: Tensor<Scalar>.self) {
      globalNorm += self[keyPath: kp].squared().sum()
    }
    globalNorm = sqrt(globalNorm)
    for kp in self.recursivelyAllWritableKeyPaths(to: Tensor<Scalar>.self) {
      self[keyPath: kp] *= clipNorm / max(globalNorm, clipNorm)
    }
  }
}
```

eaplatanios:
@rxwei @dan-zheng with respect to the protocols, why do we want to related `VectorProtocol` to `Numeric`? Can't we let `Numeric` be and design completely separate protocols that are defined over tensor structures?

lattner:
I think this PR is raising a bunch of interesting points about tradeoffs in the API design and the direction we're going with compiler synthesis and the core language protocols.  Could one of you bring this to the swift@tensorflow mailing list, and/or a design meeting?

PR #324, opened 2019-07-02
Enhanced the 'Dense' layer to support batch matrix multiplication.

eaplatanios:
This enhances the `Dense` layer to be more aligned with the `matmul` operation. More specifically, it allows it to use the batch variant of `matmul` if the rank of `weight` is equal to `3`. Not sure if we want to merge this here but I use it often in my work and it also aligns `Dense` with the functionality of `matmul`.

### DIFF
--- a/Sources/TensorFlow/Layers/Core.swift
+++ b/Sources/TensorFlow/Layers/Core.swift
@@ -139,10 +139,10 @@ public struct Reshape<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
 /// a weight matrix, `bias` is a bias vector, and `activation` is an element-wise activation
 /// function.
 ///
-/// This layer also supports 3-dimensional weight tensors with 2-dimensional bias matrices. In this 
-/// case the first dimension of both is treated as the batch size that is aligned with the first 
-/// dimension of `input` and the batch variant of the `matmul` operation is used, thus using a 
-/// different weight and bias for each element in input batch.
+/// This layer also supports 3-D weight tensors with 2-D bias matrices. In this case the first
+/// dimension of both is treated as the batch size that is aligned with the first dimension of
+/// `input` and the batch variant of the `matmul(_:_:)` operation is used, thus using a different
+/// weight and bias for each element in input batch.
 @frozen
 public struct Dense<Scalar: TensorFlowFloatingPoint>: Layer {
     /// The weight matrix.
@@ -171,7 +171,7 @@ public struct Dense<Scalar: TensorFlowFloatingPoint>: Layer {
     /// - Returns: The output.
     @differentiable
     public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
-        if weight.rank == 3 {
+        if weight.rankTensor == Tensor<Int>(3) {
             let hidden = matmul(input.expandingShape(at: 1), weight)
             return activation(hidden.squeezingShape(at: 1) + bias)
         }
--- a/Sources/TensorFlow/Layers/Core.swift
+++ b/Sources/TensorFlow/Layers/Core.swift
@@ -151,6 +151,8 @@ public struct Dense<Scalar: TensorFlowFloatingPoint>: Layer {
     public var bias: Tensor<Scalar>
     /// The element-wise activation function.
     @noDerivative public let activation: Activation
+    /// Indicates whether this is a batched dense layer.
+    @noDerivative internal let batched: Bool
     
     /// The element-wise activation function type.
     public typealias Activation = @differentiable (Tensor<Scalar>) -> Tensor<Scalar>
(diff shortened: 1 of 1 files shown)


Sources/TensorFlow/Layers/Core.swift
```diff
@@ -166,6 +171,10 @@ public struct Dense<Scalar: TensorFlowFloatingPoint>: Layer {
     /// - Returns: The output.
     @differentiable
     public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+        if weight.rank == 3 {
```
rxwei: @bgogul Is there a lazy tensor concern here?

Sources/TensorFlow/Layers/Core.swift
```diff
@@ -138,6 +138,11 @@ public struct Reshape<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
 /// `Dense` implements the operation `activation(matmul(input, weight) + bias)`, where `weight` is
 /// a weight matrix, `bias` is a bias vector, and `activation` is an element-wise activation
 /// function.
+///
+/// This layer also supports 3-dimensional weight tensors with 2-dimensional bias matrices. In this
```
rxwei: I think we've been using "n-D" in most API documentation. If so, I suggest changing this. Otherwise never mind.
```suggestion
/// This layer also supports 3-D weight tensors with 2-D bias matrices. In this 
```

Sources/TensorFlow/Layers/Core.swift
```diff
@@ -138,6 +138,11 @@ public struct Reshape<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
 /// `Dense` implements the operation `activation(matmul(input, weight) + bias)`, where `weight` is
 /// a weight matrix, `bias` is a bias vector, and `activation` is an element-wise activation
 /// function.
+///
+/// This layer also supports 3-dimensional weight tensors with 2-dimensional bias matrices. In this 
+/// case the first dimension of both is treated as the batch size that is aligned with the first 
+/// dimension of `input` and the batch variant of the `matmul` operation is used, thus using a
```
rxwei:
```suggestion
/// dimension of `input` and the batch variant of the `matmul(_:_:)` operation is used, thus using a 
```

Sources/TensorFlow/Layers/Core.swift
eaplatanios: If believe there is, but what if I write this as: `weight.rankTensor == Tensor<Int>(3)`?

rxwei: Yeah, that might work a lot better, but I'll wait for @bgogul to confirm.

eaplatanios:
Thanks for the feedback Richard! I addressed all comments and now waiting for confirmation from @bgogul that the current approach is "`LazyTensor`-friendly".

Sources/TensorFlow/Layers/Core.swift
eaplatanios: Actually, given that `if` statements are not currently being lowered to TF ops, this should definitely affect lazy tensor because the condition depends on `weight`. I can move that condition to the initializer and add a boolean stored property named `batched`. Using that in the `if` statement will be safe with lazy tensor. @rxwei what do you think?

rxwei: Hmm, I'd very much prefer keeping the existing semantics if possible.

eaplatanios: This will retain the same semantics. It just moves the rank check to the initializer.

bgogul: Yes, this would cause materialization in lazy tensor, and thereby, break the trace. Even changing the condition to `weight.rankTensor == Tensor<Int>(3)` won't fix it. Adding the `batched` property would fix it though. 

That said, I don't think it is sustainable to have to rewrite code to help lazy tensor in the long run. If the code becomes too awkward and ugly, just do what makes the most sense. The right way to fix this would be to track shapes in the swift side.

eaplatanios: I just pushed the change so you can see what I mean.

eaplatanios: @bgogul Thanks for the confirmation! I already switched to the `batched` property. And yes, I agree with what you say. In this case I believe it's ok to modify the code given how prevalent dense layers are and also given that lowering if-statements to TF ops may take a while.

eaplatanios:
This should be ready for review now. I switched to using the `batched` property.

Sources/TensorFlow/Layers/Core.swift
```diff
@@ -158,6 +165,7 @@ public struct Dense<Scalar: TensorFlowFloatingPoint>: Layer {
         self.weight = weight
```
rxwei: Should we add preconditions here to assert something like `weight.rank <= 3` and `bias.rank <= 2`?

eaplatanios: Done.

PR #327, opened 2019-07-02
Enhanced the 'Tensor.batchGathering(atIndices:)' function.

eaplatanios:
Enhanced the `Tensor.batchGathering(atIndices:)` function so that it has the same functionality as the Python API version. It now supports a variable number of batch dimensions as well as an axis to gather along.

This became possible now due to support for control flow auto-diff. :)

### DIFF
--- a/Sources/TensorFlow/Operators/Basic.swift
+++ b/Sources/TensorFlow/Operators/Basic.swift
@@ -411,8 +411,8 @@ public extension Tensor {
     @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
     func batchGathering<Index: TensorFlowIndex>(
         atIndices indices: Tensor<Index>,
-        alongAxis axis: Int,
-        numBatchDims batchDims: Int
+        alongAxis axis: Int = 1,
+        batchDimCount batchDims: Int = 1
     ) -> Tensor {
         // TODO: precondition(batchDims >= 0 && batchDims < indices.rank,
         //                    "'numBatchDims' must be non-negative and less than 'indices.rank'.")
@@ -440,7 +440,7 @@ public extension Tensor {
             let result = tensor.batchGathering(
                 atIndices: indices,
                 alongAxis: batchDims,
-                numBatchDims: batchDims)
+                batchDimCount: batchDims)
 
             // Move the result dimensions corresponding to self[batchDims ..< axis] to just before
             // the dimensions corresponding to indices[batchDims ...].
--- a/Sources/TensorFlow/Operators/Basic.swift
+++ b/Sources/TensorFlow/Operators/Basic.swift
@@ -390,44 +390,92 @@ public extension Tensor {
         return Raw.gatherV2(params: self, indices: indices, axis: Tensor<Int32>(Int32(axis)))
     }
 
-    /// Returns slices of this tensor at `indices`, while ignoring the first `batchDims` dimensions
-    /// that correspond to batch dimensions. The gather is performed along the first non-batch
-    /// dimension.
+    /// Returns slices of this tensor at `indices` along the `axis` dimension, while ignoring the 
+    /// first `batchDims` dimensions that correspond to batch dimensions. The gather is performed 
+    /// along the first non-batch dimension.
     ///
     /// Performs similar functionality to `gathering`, except that the resulting tensor shape is 
-    /// now:
-    /// ```
-    /// self.shape[..<batchDims] + 
-    ///   indices.shape[batchDims...] + 
-    ///   self.shape[(batchDims + indices.rank + 1)...]
-    /// ```
+    /// now `self.shape[..<axis] + indices.shape[batchDims...] + self.shape[(axis + 1)...]`.
     ///
     /// - Parameters:
     ///   - indices: Contains the indices to gather.
+    ///   - axis: Dimension along which to gather. Negative values wrap around.
     ///   - batchDims: Number of leading batch dimensions to ignore.
     ///
+    /// - Precondition: `axis` must be in the range `[-rank, rank)`, while also being greater than
+    ///   or equal to `batchDims`.
     /// - Precondition: `batchDims` must be less than `indices.rank`.
     ///
     /// - Returns: The gathered tensor.
     @inlinable
     @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
-    func batchGathering<Index: TensorFlowIndex>(atIndices indices: Tensor<Index>) -> Tensor {
-        var batchIndices = indices
-        var accumulated = Tensor<Index>(ones: [])
-        accumulated *= withoutDerivative(at: shapeTensor) { Tensor<Index>($0[1]) }
-        let dValue = withoutDerivative(at: shapeTensor) { $0[0] }
-        let dIndices = Tensor<Index>(
-            rangeFrom: Tensor<Index>(zeros: []),
-            to: Tensor<Index>(dValue),
-            stride: Tensor<Index>(ones: [])
-        ) * accumulated
-        let dShape = Tensor<Int32>(concatenating: [
-            dValue.rankLifted(),
-            Tensor<Int32>([Int32](repeating: 1, count: indices.rank - 1))])
-        batchIndices += dIndices.reshaped(toShape: dShape)
+    func batchGathering<Index: TensorFlowIndex>(
+        atIndices indices: Tensor<Index>,
+        alongAxis axis: Int,
+        numBatchDims batchDims: Int
+    ) -> Tensor {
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -390,44 +390,92 @@ public extension Tensor {
         return Raw.gatherV2(params: self, indices: indices, axis: Tensor<Int32>(Int32(axis)))
     }
 
-    /// Returns slices of this tensor at `indices`, while ignoring the first `batchDims` dimensions
-    /// that correspond to batch dimensions. The gather is performed along the first non-batch
-    /// dimension.
+    /// Returns slices of this tensor at `indices` along the `axis` dimension, while ignoring the 
+    /// first `batchDims` dimensions that correspond to batch dimensions. The gather is performed
```
rxwei: - `batchDims` is undefined.
- "dims" -> "dimensions"

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -390,44 +390,92 @@ public extension Tensor {
...
-            stride: Tensor<Index>(ones: [])
-        ) * accumulated
-        let dShape = Tensor<Int32>(concatenating: [
-            dValue.rankLifted(),
-            Tensor<Int32>([Int32](repeating: 1, count: indices.rank - 1))])
-        batchIndices += dIndices.reshaped(toShape: dShape)
+    func batchGathering<Index: TensorFlowIndex>(
+        atIndices indices: Tensor<Index>,
+        alongAxis axis: Int = 1,
+        batchDimCount batchDims: Int = 1
```
```suggestion
        batchDimensionCount: Int = 1
```

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -390,44 +390,92 @@ public extension Tensor {
...
+            // Adjust axis to be positive.
+            let posAxis = axis < 0 ? axis + rank : axis
+
+            // TODO: precondition(posAxis >= 0 && posAxis < rank, "'axis' is out of range.")
+            // TODO: precondition(batchDims <= posAxis,
+            //                    "'batchDims' must be less than or equal to 'axis'.")
+
+            // Move self[axis] up to self[batchDims].
+            let permutation = Tensor<Int32>(concatenating: [
+                Tensor<Int32>(rangeFrom: 0, to: Int32(batchDims), stride: 1),
```
```suggestion
                Tensor(rangeFrom: 0, to: Int32(batchDims), stride: 1),
```

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -390,44 +390,92 @@ public extension Tensor {
...
+                Tensor<Int32>(Int32(axis)).rankLifted(),
```
```suggestion
                Tensor(Int32(axis)).rankLifted(),
```

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -390,44 +390,92 @@ public extension Tensor {
...
+                Tensor<Int32>(rangeFrom: Int32(batchDims), to: Int32(posAxis), stride: 1),
```
```suggestion
                Tensor(rangeFrom: Int32(batchDims), to: Int32(posAxis), stride: 1),
```

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -390,44 +390,92 @@ public extension Tensor {
...
+                Tensor<Int32>(rangeFrom: Int32(axis) + 1, to: Int32(rank), stride: 1)])
```
```suggestion
                Tensor(rangeFrom: Int32(axis) + 1, to: Int32(rank), stride: 1)])
```

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -390,44 +390,92 @@ public extension Tensor {
...
+            let start = indices.rank + posAxis - batchDims
+            let resultPermutation = Tensor<Int32>(concatenating: [
+                Tensor<Int32>(rangeFrom: 0, to: Int32(batchDims), stride: 1),
+                Tensor<Int32>(rangeFrom: Int32(indices.rank), to: Int32(start), stride: 1),
+                Tensor<Int32>(rangeFrom: Int32(batchDims), to: Int32(indices.rank), stride: 1),
+                Tensor<Int32>(rangeFrom: Int32(start), to: Int32(result.rank), stride: 1)])
+            return result.transposed(withPermutations: resultPermutation)
+        }
+
+        let batchIndices = withoutDerivative(at: { () -> Tensor<Index> in
```
```suggestion
        let batchIndices = withoutDerivative {
```

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -390,44 +390,92 @@ public extension Tensor {
...
+                    Tensor<Int32>([Int32](repeating: 1, count: indices.rank - 1))])
+                batchIndices += dIndices.reshaped(toShape: dShape)
+            }
+            return batchIndices
+        }())
+
         let flatIndices = batchIndices.flattened()
-        let outerShape = withoutDerivative(at: shapeTensor) { $0[2...] }
-        let innerShape = withoutDerivative(at: shapeTensor) { $0[..<2] }.product(squeezingAxes: [0])
+        let outerShape = shapeTensor[Int(batchDims + 1)...]
```
rxwei: `batchDimensionCount` is already `Int`.
```suggestion
        let outerShape = shapeTensor[(batchDimensionCount + 1)...]
```

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -390,44 +390,92 @@ public extension Tensor {
...
+        let innerShape = shapeTensor[..<Int(batchDims + 1)].product(squeezingAxes: [0])
```
rxwei: Same here.

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -390,44 +390,92 @@ public extension Tensor {
...
-    ///   self.shape[(batchDims + indices.rank + 1)...]
-    /// ```
+    /// now `self.shape[..<axis] + indices.shape[batchDims...] + self.shape[(axis + 1)...]`.
     ///
     /// - Parameters:
     ///   - indices: Contains the indices to gather.
+    ///   - axis: Dimension along which to gather. Negative values wrap around.
     ///   - batchDims: Number of leading batch dimensions to ignore.
     ///
+    /// - Precondition: `axis` must be in the range `[-rank, rank)`, while also being greater than
```
rxwei: Use Swift syntax for ranges.
```suggestion
    /// - Precondition: `axis` must be in the range `-rank..<rank`, while also being greater than
```

Sources/TensorFlow/Operators/Basic.swift
eaplatanios: `withoutDerivative(at:)` does not take a closure as argument and so I cannot use the trailing closure syntax.

Sources/TensorFlow/Operators/Basic.swift
eaplatanios: I cannot remove this from any of the calls to this constructor because removing it results in an error about ambiguity at compile time.

eaplatanios:
I made all requested changes except for those that I left comments for above.

Sources/TensorFlow/Operators/Basic.swift
rxwei: It's generic over `T`, so `T` will be inferred to be any trailing closure. Did that not work for you?

Sources/TensorFlow/Operators/Basic.swift
rxwei: Oops. Sorry!

Sources/TensorFlow/Operators/Basic.swift
eaplatanios: I get an error. I believe that's because I don't want to pass the closure, but rather the value it returns and I wrap the whole thing in `withoutDerivative(at:)`. Otherwise, if I make the suggested change I would have to treat `batchIndices` as a function.

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -390,44 +390,97 @@ public extension Tensor {
...
+                Tensor<Int32>(rangeFrom: Int32(indices.rank), to: Int32(start), stride: 1),
+                Tensor<Int32>(
+                    rangeFrom: Int32(batchDimensionCount),
+                    to: Int32(indices.rank),
+                    stride: 1),
+                Tensor<Int32>(rangeFrom: Int32(start), to: Int32(result.rank), stride: 1)])
+            return result.transposed(withPermutations: resultPermutation)
+        }
+
+        let batchIndices = withoutDerivative(at: { () -> Tensor<Index> in
```
```suggestion
        let batchIndices: Tensor<Index> = withoutDerivative(at: {
```

eaplatanios: Done.

PR #328, opened 2019-07-02
Enhanced the random seed generation function.

eaplatanios:
This PR enhances the function that generates default random seeds for TensorFlow. This is needed because the previous approach is weak in multi-process settings, which are quite common when training or evaluating machine learning models. The implementation is inspired from [OpenAI Gym](https://github.com/openai/gym/blob/master/gym/utils/seeding.py).

I also added an implementation on SHA512 in the utilities to avoid having any external library dependencies along with a test for the hash function. The SHA512 implementation is based on that of [CryptoSwift](https://github.com/krzyzanowskim/CryptoSwift).


Sources/TensorFlow/Core/Utilities.swift
```diff
@@ -167,3 +167,176 @@ extension UnsafeMutablePointer where Pointee == CTensorHandle? {
         self.init(unwrapped)
     }
 }
+
+//===------------------------------------------------------------------------------------------===//
+// Hashing
+//===------------------------------------------------------------------------------------------===//
+
+internal extension FixedWidthInteger {
+    @usableFromInline
```
rxwei: No need for `@usableFromInline` for these functions.

Sources/TensorFlow/Core/Utilities.swift
```diff
@@ -167,3 +167,176 @@ extension UnsafeMutablePointer where Pointee == CTensorHandle? {
...
+        self = Self(0)
+        for i in 0..<MemoryLayout<Self>.size {
+            let j = (MemoryLayout<Self>.size - i - 1) * 8
+            self |= count > 0 ? Self(bytes[index.advanced(by: i)]) << j : 0
+        }
+    }
+
+    @usableFromInline
+    func bytes(count byteCount: Int = MemoryLayout<Self>.size) -> [UInt8] {
+        let valuePointer = UnsafeMutablePointer<Self>.allocate(capacity: 1)
```
rxwei: I'd suggest using a local variable here instead of a heap allocation.

```swift
var value = littleEndian
```

Then you can pass pointers to it as `&value`. Also, there's probably a more descriptive name than "value".

Sources/TensorFlow/Core/Utilities.swift
```diff
@@ -167,3 +167,176 @@ extension UnsafeMutablePointer where Pointee == CTensorHandle? {
...
     }
 }
+
+//===------------------------------------------------------------------------------------------===//
+// Hashing
+//===------------------------------------------------------------------------------------------===//
+
+internal extension FixedWidthInteger {
+    @usableFromInline
+    init(fromBytes bytes: ArraySlice<UInt8>, startingAt index: Int) {
```
rxwei: Per Swift API Design Guidelines, the base name of an initializer (aka the type name) should not form a phrase with argument labels. 
```suggestion
    init(bytes: ArraySlice<UInt8>, startIndex: Int) {
```

Sources/TensorFlow/Core/Utilities.swift
```diff
@@ -167,3 +167,176 @@ extension UnsafeMutablePointer where Pointee == CTensorHandle? {
...
+        if bytes.isEmpty { self = 0; return }
+        let count = bytes.count
+        self = Self(0)
```
rxwei: Always prefer initializer delegation in initializers.
```suggestion
        self.init(0)
```

Sources/TensorFlow/Core/Utilities.swift
```diff
@@ -167,3 +167,176 @@ extension UnsafeMutablePointer where Pointee == CTensorHandle? {
...
 }
+
+//===------------------------------------------------------------------------------------------===//
+// Hashing
+//===------------------------------------------------------------------------------------------===//
+
+internal extension FixedWidthInteger {
+    @usableFromInline
+    init(fromBytes bytes: ArraySlice<UInt8>, startingAt index: Int) {
+        if bytes.isEmpty { self = 0; return }
```
rxwei:
```suggestion
        if bytes.isEmpty { self.init(0); return }
```

Sources/TensorFlow/Core/Utilities.swift
```diff
@@ -167,3 +167,176 @@ extension UnsafeMutablePointer where Pointee == CTensorHandle? {
...
+        let count = bytes.count
+        self = Self(0)
+        for i in 0..<MemoryLayout<Self>.size {
+            let j = (MemoryLayout<Self>.size - i - 1) * 8
+            self |= count > 0 ? Self(bytes[index.advanced(by: i)]) << j : 0
+        }
+    }
+
+    @usableFromInline
+    func bytes(count byteCount: Int = MemoryLayout<Self>.size) -> [UInt8] {
```
rxwei: Can this be `fileprivate`?

Sources/TensorFlow/Core/Utilities.swift
```diff
@@ -167,3 +167,176 @@ extension UnsafeMutablePointer where Pointee == CTensorHandle? {
...
+        var bytes = [UInt8](repeating: 0, count: byteCount)
+        for i in 0..<Swift.min(MemoryLayout<Self>.size, byteCount) {
+            bytes[byteCount - 1 - i] = (bytesPointer + i).pointee
+        }
+        valuePointer.deinitialize(count: 1)
+        valuePointer.deallocate()
+        return bytes
+    }
+
+    @usableFromInline
```
rxwei: `fileprivate`?

Sources/TensorFlow/Core/Utilities.swift
```diff
@@ -167,3 +167,176 @@ extension UnsafeMutablePointer where Pointee == CTensorHandle? {
...
+            
+            // Break chunk into sixteen 64-bit words M[j], 0 ≤ j ≤ 15, in big-endian format.
+            // Extend the sixteen 64-bit words into eighty 64-bit words:
+            let M = UnsafeMutablePointer<UInt64>.allocate(capacity: k.count)
+            M.initialize(repeating: 0, count: k.count)
+            defer {
+                M.deinitialize(count: k.count)
+                M.deallocate()
+            }
+            for x in 0..<k.count {
```
rxwei:
```suggestion
            for x in k.indices {
```

Sources/TensorFlow/Core/Utilities.swift
```diff
@@ -167,3 +167,176 @@ extension UnsafeMutablePointer where Pointee == CTensorHandle? {
...
+            0x6a09e667f3bcc908, 0xbb67ae8584caa73b, 0x3c6ef372fe94f82b, 0xa54ff53a5f1d36f1, 
+            0x510e527fade682d1, 0x9b05688c2b3e6c1f, 0x1f83d9abfb41bd6b, 0x5be0cd19137e2179]
+        var index = 0
+        while index < accumulated.count {
+            let chunk = accumulated[index..<(index + blockSize)]
+            index += blockSize
+            
+            // Break chunk into sixteen 64-bit words M[j], 0 ≤ j ≤ 15, in big-endian format.
+            // Extend the sixteen 64-bit words into eighty 64-bit words:
+            let M = UnsafeMutablePointer<UInt64>.allocate(capacity: k.count)
```
rxwei: I'd suggest using an array here: `let M = Array(repeating: 0, count: k.count)`, which is safer and will be promoted to a stack allocation in the optimizer.

Sources/TensorFlow/Core/Utilities.swift
eaplatanios: Sorry I keep forgetting that.

Sources/TensorFlow/Core/Utilities.swift
eaplatanios: It's used in `Random.swift`.

Sources/TensorFlow/Core/Utilities.swift
```diff
@@ -167,3 +167,176 @@ extension UnsafeMutablePointer where Pointee == CTensorHandle? {
...
+            accumulatedHash[4] = (accumulatedHash[4] &+ E)
+            accumulatedHash[5] = (accumulatedHash[5] &+ F)
+            accumulatedHash[6] = (accumulatedHash[6] &+ G)
+            accumulatedHash[7] = (accumulatedHash[7] &+ H)
+        }
+
+        // Step 4: Return the computed hash.
+        var result = [UInt8](repeating: 0, count: digestLength)
+        var position = 0
+        for index in 0..<accumulatedHash.count {
```
rxwei:
```suggestion
        for index in accumulatedHash.indices {
```

Sources/TensorFlow/Core/Utilities.swift
```diff
@@ -167,3 +167,176 @@ extension UnsafeMutablePointer where Pointee == CTensorHandle? {
...
+            repeating: 0,
+            count: msgLength % blockSize < max ?
+                max - 1 - (msgLength % blockSize) :
+                blockSize + max - 1 - (msgLength % blockSize))
+
+        // Step 2: Append the message length as a 64-bit representation of `lengthInBits`.
+        accumulated += lengthBytes
+
+        // Step 3: Process the array bytes.
+        var accumulatedHash: [UInt64] = [
```
rxwei: I'd strongly suggest using a `SIMD8<UInt64>` for this.

eaplatanios:
All the requested changes (except for the `fileprivate` for the `bytes(count:startingAt:)` function are done.

Sources/TensorFlow/Core/Utilities.swift
```diff
@@ -167,3 +167,167 @@ extension UnsafeMutablePointer where Pointee == CTensorHandle? {
...
+                        (M[x - 15] >> 7)
+                    let s1 = M[x - 2].rotate(rightBy: 19) ^ 
+                        M[x - 2].rotate(rightBy: 61) ^ 
+                        (M[x - 2] >> 6)
+                    M[x] = M[x - 16] &+ s0 &+ M[x - 7] &+ s1
+                    break
+                }
+            }
+
+            var A = accumulatedHash[0]
```
rxwei: All these can be vectorized using SIMD so that you won't need indexing.

```swift
// here
var tmp = accumulatedHash // need a better name

// down below
accumulatedHash &+= tmp
```

Sources/TensorFlow/Core/Utilities.swift
```diff
@@ -167,3 +167,167 @@ extension UnsafeMutablePointer where Pointee == CTensorHandle? {
...
+            var B = accumulatedHash[1]
+            var C = accumulatedHash[2]
+            var D = accumulatedHash[3]
+            var E = accumulatedHash[4]
+            var F = accumulatedHash[5]
+            var G = accumulatedHash[6]
+            var H = accumulatedHash[7]
+
+            for j in 0..<k.count {
```
rxwei: I'm not sure if rotation is directly supported by SIMD APIs yet, but it'd be nice to vectorize this as well. (see [proposal](https://github.com/apple/swift-evolution/blob/master/proposals/0251-simd-additions.md)).

Sources/TensorFlow/Core/Utilities.swift
eaplatanios: Done!

Sources/TensorFlow/Core/Utilities.swift
eaplatanios: That would be nice indeed. I believe there must be SIMD implementations of SHA512 out there. I just went for something simple to add support without dependencies but I'm sure we can make it much more efficient.

eaplatanios:
This should be ready for another look now.

Sources/TensorFlow/Core/Utilities.swift
```diff
@@ -167,3 +167,157 @@ extension UnsafeMutablePointer where Pointee == CTensorHandle? {
...
+                    let s1 = M[x - 2].rotate(rightBy: 19) ^ 
+                        M[x - 2].rotate(rightBy: 61) ^ 
+                        (M[x - 2] >> 6)
+                    M[x] = M[x - 16] &+ s0 &+ M[x - 7] &+ s1
+                    break
+                }
+            }
+
+            var hashCopy = accumulatedHash
+            for j in 0..<k.count {
```
rxwei:
```suggestion
            for j in 0..<k.indices {
```

Sources/TensorFlow/Core/Utilities.swift
```diff
@@ -167,3 +167,157 @@ extension UnsafeMutablePointer where Pointee == CTensorHandle? {
...
+                let maj = (hashCopy[0] & hashCopy[1]) ^ 
+                    (hashCopy[0] & hashCopy[2]) ^
+                    (hashCopy[1] & hashCopy[2])
+                let t2 = s0 &+ maj
+                let s1 = hashCopy[4].rotate(rightBy: 14) ^
+                    hashCopy[4].rotate(rightBy: 18) ^
+                    hashCopy[4].rotate(rightBy: 41)
+                let ch = (hashCopy[4] & hashCopy[5]) ^ ((~hashCopy[4]) & hashCopy[6])
+                let t1 = hashCopy[7] &+ s1 &+ ch &+ k[j] &+ UInt64(M[j])
+                hashCopy[7] = hashCopy[6]
```
rxwei: @stephentyrone Could you please advise?

Sources/TensorFlow/Core/Utilities.swift
rxwei: @stephentyrone Could you please advise us?

Sources/TensorFlow/Core/Utilities.swift
```diff
@@ -167,3 +167,157 @@ extension UnsafeMutablePointer where Pointee == CTensorHandle? {
...
+                hashCopy[3] = hashCopy[2]
+                hashCopy[2] = hashCopy[1]
+                hashCopy[1] = hashCopy[0]
+                hashCopy[0] = t1 &+ t2
+            }
+            accumulatedHash &+= hashCopy
+        }
+
+        // Step 4: Return the computed hash.
+        var result = [UInt8](repeating: 0, count: digestLength)
```
rxwei: Should we use `SIMD64<UInt8>` instead?

rxwei: That way, the following loop can also be vectorized.

eaplatanios: I use `[UInt8]` because we also define it over `[UInt8]` and we only afterwards break it down to a pair of `Int32`s. Do you think using `SIMD` would be useful here?

rxwei: If it has a fixed size, it'd be really great to use SIMD.

Sources/TensorFlow/Core/Utilities.swift
eaplatanios: I found [this](https://www.intel.com/content/dam/www/public/us/en/documents/white-papers/fast-sha512-implementations-ia-processors-paper.pdf) which seems relevant. :)

Sources/TensorFlow/Core/Utilities.swift
```diff
@@ -167,3 +167,157 @@ extension UnsafeMutablePointer where Pointee == CTensorHandle? {
...
+            let bytesPointer = UnsafeMutablePointer<UInt8>(OpaquePointer(pointer))
+            var bytes = [UInt8](repeating: 0, count: byteCount)
+            for i in 0..<Swift.min(MemoryLayout<Self>.size, byteCount) {
+                bytes[byteCount - 1 - i] = (bytesPointer + i).pointee
+            }
+            return bytes
+        }
+    }
+
+    fileprivate func rotate(rightBy count: Self) -> Self {
```
rxwei: If a method's phrase has the syntactic structure `(VP V Adv (PP P NP))`, it is more common to put the adverb in the base name than in the argument label with the prepositional phrase.

```suggestion
    fileprivate func rotateRight(by count: Self) -> Self {
```

eaplatanios: I'm actually working on removing the need for rotations now to better utilize SIMD. :)

Sources/TensorFlow/Core/Utilities.swift
```diff
@@ -167,3 +167,153 @@ extension UnsafeMutablePointer where Pointee == CTensorHandle? {
...
+                hashCopy[2] = hashCopy[1]
+                hashCopy[1] = hashCopy[0]
+                hashCopy[0] = t1 &+ t2
+            }
+            accumulatedHash &+= hashCopy
+        }
+
+        // Step 4: Return the computed hash.
+        var result = SIMD64<UInt8>()
+        // var result = [UInt8](repeating: 0, count: digestLength)
```
rxwei: Could this be removed?

eaplatanios: Yes sorry it's still a WIP because I'm looking into whether I can vectorize some more parts. If I can't I'll just go ahead and merge this a bit later.

By the way, should we make the `sha512()` extension public? It's a nice implementation of SHA512 that's missing from the standard library anyway and can come in handy. I think it's fine either way so it's your call.

eaplatanios: @rxwei what do you think about making `sha512()` public? This is ready to merge otherwise as I don't see a simple way to vectorize the rest.

I also have an `sha1()` implementation for Swift Retro that I've been building and I could open a separate PR adding a `Hashing.swift` utilities file with both implementations.

rxwei: Let's make it internal for now. Would be great to bring this up on the Swift forum.

Sources/TensorFlow/Core/Utilities.swift
```diff
@@ -167,3 +167,153 @@ extension UnsafeMutablePointer where Pointee == CTensorHandle? {
...
+        for i in 0..<MemoryLayout<Self>.size {
+            let j = (MemoryLayout<Self>.size - i - 1) * 8
+            self |= count > 0 ? Self(bytes[index.advanced(by: i)]) << j : 0
+        }
+    }
+
+    func bytes(count byteCount: Int = MemoryLayout<Self>.size) -> [UInt8] {
+        var littleEndianValue = littleEndian
+        return withUnsafePointer(to: &littleEndianValue) { pointer -> [UInt8] in
+            let bytesPointer = UnsafeMutablePointer<UInt8>(OpaquePointer(pointer))
```
rxwei: `OpaquePointer` is for interoperability with opaque C types. This is not the best way of turning some value into bytes. Instead, it would be highly preferable to use `UnsafeBufferPointer` (which is a `Collection`) and `UnsafeBufferPointer.withMemoryRebound(to:_:)`. Then, you can initialize an array directly without doing pointer arithmetics.

eaplatanios: Thanks for the suggestion! I'll fix that soon, after I open the PR for conditional differentiable conformance.

Shashi456:
@eaplatanios, If you need any help with all these tasks you are undertaking, I'd love to help you out. I really like the enhancements we've had in the past week. So if there are trivial/non-trivial tasks you need some help with, do tag me.

eaplatanios:
@Shashi456 Thanks a lot! Sorry for the late response but I was traveling. I don't have something planned right now, but I will definitely tag you if anything comes up. :)

Sources/TensorFlow/Core/Utilities.swift
eaplatanios: Addressed in #337.

PR #330, opened 2019-07-03
_vjpConv2DBackpropInput incorrectly passing shape to conv2DBackpropFi…

sjaz24:
…lter filterSizes parameter instead of filter shape.

### DIFF
--- a/Sources/TensorFlow/Operators/NN.swift
+++ b/Sources/TensorFlow/Operators/NN.swift
@@ -185,7 +185,7 @@ func _vjpConv2DBackpropInput<Scalar: TensorFlowFloatingPoint>(
     let value = conv2DBackpropInput(x, shape: shape, filter: filter,
                                     strides: strides, padding: padding, dilations: dilations)
     return (value, { v in
-        (conv2DBackpropFilter(x, input: v, filterSizes: shape, strides: strides,
+        (conv2DBackpropFilter(x, input: v, filterSizes: filter.shapeTensor, strides: strides,
                               padding: padding, dilations: dilations),
          conv2D(v, filter: filter, strides: strides, padding: padding, dilations: dilations))
     })


sjaz24:
Not sure why it is saying "Need a CLA for one or more commit authors". I've gone through the individual CLA process twice now.

Shashi456:
@sjaz24 don't mind that, it'll resolve itself in a while.

Sources/TensorFlow/Operators/NN.swift
```diff
@@ -185,7 +185,7 @@ func _vjpConv2DBackpropInput<Scalar: TensorFlowFloatingPoint>(
     let value = conv2DBackpropInput(x, shape: shape, filter: filter,
                                     strides: strides, padding: padding, dilations: dilations)
     return (value, { v in
```
```suggestion
     return (value, { [filterShape = filter.shapeTensor] v in
```

Sources/TensorFlow/Operators/NN.swift
```diff
@@ -185,7 +185,7 @@ func _vjpConv2DBackpropInput<Scalar: TensorFlowFloatingPoint>(
...
-        (conv2DBackpropFilter(x, input: v, filterSizes: shape, strides: strides,
+        (conv2DBackpropFilter(x, input: v, filterSizes: filter.shapeTensor, strides: strides,
```
rxwei: Using `filter.shapeTensor` here is less than ideal because the `filter` value would be captured in the backward pass, while we only need its `shapeTensor`. Instead, you can explicitly capture `filter.shapeTensor` as a separate value (`filterShape`, in my suggestion above), and use `filterShape` here directly.
```suggestion
        (conv2DBackpropFilter(x, input: v, filterSizes: filterShape, strides: strides,
```

rxwei:
@sjaz24 It is possible that commit https://github.com/tensorflow/swift-apis/pull/330/commits/23176cc1820c81fe49c9e90fa504da46590e8421 isn't associated with the email address which you signed the CLA with. Could you please verify that?

Sources/TensorFlow/Operators/NN.swift
sjaz24: Yes, that makes total sense.

sjaz24:
Yes that's probably it. I just took my gmail when I did the CLA and only asked for my github username but my github is using a different email.

rxwei:
Could you try resetting the author of both commits?

rxwei:
Have you tried `git config --global user.email ...` before resetting commits? You could also squash all commits into one so that you can reset the email address more easily.

sjaz24:
I've created a new branch with everything i think set up correctly. Sorry had to verify my other email and add it to the CLA. Can I reject this PR and submit the other one.

rxwei:
Feel free to start a separate PR. Sorry for the hassle!

sjaz24:
No hassle. I appreciate the opportunity to contribute. I've opened the other PR and it seem to be good this time. Sorry about the confusion.

PR #331, opened 2019-07-03
_vjpConv2DBackpropInput using shape instead of using filter size for …

sjaz24:
…filterSizes argument.

### DIFF
--- a/Sources/TensorFlow/Operators/NN.swift
+++ b/Sources/TensorFlow/Operators/NN.swift
@@ -184,8 +184,8 @@ func _vjpConv2DBackpropInput<Scalar: TensorFlowFloatingPoint>(
 ) -> (Tensor<Scalar>, (Tensor<Scalar>) -> (Tensor<Scalar>, Tensor<Scalar>)) {
     let value = conv2DBackpropInput(x, shape: shape, filter: filter,
                                     strides: strides, padding: padding, dilations: dilations)
-    return (value, { [filterShape = filter.shapeTensor] v in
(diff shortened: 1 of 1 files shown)


Sources/TensorFlow/Operators/NN.swift
```diff
@@ -184,8 +184,8 @@ func _vjpConv2DBackpropInput<Scalar: TensorFlowFloatingPoint>(
 ) -> (Tensor<Scalar>, (Tensor<Scalar>) -> (Tensor<Scalar>, Tensor<Scalar>)) {
     let value = conv2DBackpropInput(x, shape: shape, filter: filter,
                                     strides: strides, padding: padding, dilations: dilations)
-    return (value, { v in
-        (conv2DBackpropFilter(x, input: v, filterSizes: shape, strides: strides,
+    return (value, { [filterShape = filter.shapeTensor] v in
```
rxwei: Actually this pullback is already capturing `filter` (on line 190), so I think your original implementation is better! I reverted it for you.

sjaz24: Thanks!

Shashi456:
@sjaz24, would this solve 221?

sjaz24:
As I mentioned in my comment on #288, I'm not having issues with the original code as it was before your change. Your test passes for me. The issue I had was with the calculation of the gradients during backprop which this PR fixed. I'm wondering if your issue was with your original test. It looks like you might have had Conv2D instead of TransposeConv2D. Have you tried running your updated test against the original code?

Issue #332, opened 2019-07-07
Another issue with _vjpConv2DBackpropInput


sjaz24:
After fixing issue #329, I've encountered another issue during backpropagation that results in the following type of error:

```
Fatal error: Incompatible shapes: [1,32,32,64] vs. [4,4,32,64]: file /Users/stephenjohnson/Projects/Conv2dTransposeTest/.build/checkouts/swift-apis/Sources/TensorFlow/Bindings/EagerExecution.swift, line 299
Illegal instruction: 4
```

It took me a while to track down, but I believe it is because conv2DBackpropInput has differentiable order of wrt: (x, filter)

```
/// TensorFlow builtin conv2d gradient helper for the input.
@differentiable(wrt: (x, filter), vjp: _vjpConv2DBackpropInput)
@usableFromInline
func conv2DBackpropInput<Scalar: TensorFlowFloatingPoint>(
```
but _vjpConv2DBackpropInput is returning a tuple where conv2DBackpropFilter call is the first tuple item and the conv2D call is the second tuple item. 

```
return (value, { v in
    (conv2DBackpropFilter(x, input: v, filterSizes: shape, strides: strides,
                          padding: padding, dilations: dilations),
     conv2D(v, filter: filter, strides: strides, padding: padding, dilations: dilations))
```

When I switch them around, everything appears to work fine. I'm going to submit a PR.

rxwei:
Great catch!

sjaz24:
Thank you!

PR #333, opened 2019-07-07
Another fix for vjp conv2 d backprop input

sjaz24:
This is a fix for issue #332


saeta:
@sjaz24 thank you very much for this! Quick question: would it be possible to add a test here (e.g. based off of your code that triggered the issue) to ensure we don't regress our behavior in the future? Thanks! -Brennan

Shashi456:
@saeta I'm adding a rest in #288

saeta:
Thanks @Shashi456 !

PR #336, opened 2019-07-08
[Tensor] Add precondition in 'Tensor.init(arrayLiteral:)' to reject 'init()' calls.

rxwei:
A long-standing issue in Swift is that labeled variadic parameters can accept 0 arguments when a label does not exist. Here is a related discussion on Swift Evolution: [Should labeled variadic parameters accept 0 arguments?](https://forums.swift.org/t/should-labeled-variadic-parameters-accept-0-arguments/12500)

When there is no `Tensor.init()`, an `init()` expression will resolve to a call to `init(arrayLiteral:)` with no arguments, which is semantically invalid for `Tensor`, `ShapedArray`, and `ShapedArraySlice`. Declaring an `init()` with an `@available(*, unavailable)` does not work becuase the type checker would still resolve `init()` calls to `init(arrayLiteral:)`. Before Swift supports finer control over labeled variadic parameters, the only thing we can do is to fail early at run-time.

XCTest does not yet provide test utilities for crashers. We can add such a test utility along with tests for such run-time errors later.

Resolves [TF-644](https://bugs.swift.org/browse/TF-644).

### DIFF
--- a/Sources/TensorFlow/Core/ShapedArray.swift
+++ b/Sources/TensorFlow/Core/ShapedArray.swift
@@ -703,6 +703,7 @@ extension ShapedArray: ExpressibleByArrayLiteral where Scalar: TensorFlowScalar 
     public typealias ArrayLiteralElement = _TensorElementLiteral<Scalar>
     @inlinable
     public init(arrayLiteral elements: _TensorElementLiteral<Scalar>...) {
+        precondition(!elements.isEmpty, "Cannot create a 'ShapedArray' with no elements")
         self = Tensor<Scalar>(_tensorElementLiterals: elements).array
     }
 }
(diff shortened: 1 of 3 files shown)


Sources/TensorFlow/Core/ShapedArray.swift
```diff
@@ -703,6 +703,7 @@ extension ShapedArray: ExpressibleByArrayLiteral where Scalar: TensorFlowScalar
     public typealias ArrayLiteralElement = _TensorElementLiteral<Scalar>
     @inlinable
     public init(arrayLiteral elements: _TensorElementLiteral<Scalar>...) {
+        precondition(!elements.isEmpty, "Cannot create a 'ShapedArray' with no elements")
```
eaplatanios: Should we end the error messages with "." (i.e., periods) so they form full sentences?

rxwei: Good suggestion.  Done.

PR #339, opened 2019-07-09
Added support for the SHA1 hashing algorithm.

eaplatanios:
@rxwei I use this in my Gym Retro interface and so I wanted to add it here along with SHA512. I plan to start a discussion in the Swift forums about making these algorithms available in the standard library and I thought it'd be nice to have a working implementation.

Regarding SIMD, I think we should use `ContiguousArray<UInt8>` for representing the hashes (or just `[UInt8]`) because not all can be represented using SIMD. For example, in `SHA1` the hash consists of only 5 bytes and there is no `SIMD5`. What do you think?


rxwei:
I think we can add a 3-bit padding in a `SIMD8` and still use SIMD because it's faster.

eaplatanios:
Sounds good. I added a 3-byte zero padding in the accumulator and a 12-byte zero padding in the result to facilitate for this.

PR #340, opened 2019-07-09
Standardized the use of random seeds.

eaplatanios:
I made `Initializers.swift` also use the same interface for random seeds that we use for layer initialization. This allows for a consistent notion of random seeds and how they work across the library. I also moved constructors that use stateful Swift random number generators in `BackwardsCompatibility.swift`. I also introduced a `TensorFlowRandomSeed` typealias to allow for easier refactoring if we decide to change things in the future.

### DIFF
--- a/Sources/TensorFlow/BackwardsCompatibility.swift
+++ b/Sources/TensorFlow/BackwardsCompatibility.swift
@@ -27,8 +27,8 @@ public extension Tensor where Scalar == Int32 {
     init<G: RandomNumberGenerator>(
         randomUniform shape: TensorShape,
         generator: inout G,
-        lowerBound: Scalar = Scalar.self.min,
-        upperBound: Scalar = Scalar.self.max
+        lowerBound: Scalar = Scalar.min,
+        upperBound: Scalar = Scalar.max
     ) {
         let dist = UniformIntegerDistribution<Scalar>(
             lowerBound: lowerBound,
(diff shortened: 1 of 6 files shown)


Sources/TensorFlow/BackwardsCompatibility.swift
```diff
@@ -14,6 +14,137 @@
...
+    /// - Parameters:
+    ///   - shape: The dimensions of the tensor.
+    ///   - generator: Random number generator to use.
+    ///   - lowerBound: The lower bound of the distribution.
+    ///   - upperBound: The upper bound of the distribution.
+    @available(*, deprecated, message: "This API will be removed after Swift for TensorFlow 0.4.")
+    init<G: RandomNumberGenerator>(
+        randomUniform shape: TensorShape,
+        generator: inout G,
+        lowerBound: Scalar = Scalar.self.min,
```
rxwei: No need to use metatype here.
```suggestion
        lowerBound: Scalar = Scalar.min,
```

Sources/TensorFlow/BackwardsCompatibility.swift
```diff
@@ -14,6 +14,136 @@
...
+    /// uniform distribution, using the default random number generator.
+    ///
+    /// - Parameters:
+    ///   - shape: The dimensions of the tensor.
+    ///   - lowerBound: The lower bound of the distribution.
+    ///   - upperBound: The upper bound of the distribution.
+    @available(*, deprecated, message: "This API will be removed after Swift for TensorFlow 0.4.")
+    init(
+        randomUniform shape: TensorShape,
+        lowerBound: Scalar = Scalar.self.min,
```
rxwei: Same here.

PR #341, opened 2019-07-10
Fix sigmoid grad

t-ae:
I tried the code below on colab.

```swift
import TensorFlow

@differentiable
public func mysigmoid<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
    return 1 / (1 + exp(-x))
}

let dsqrt: (Tensor<Float>)->Tensor<Float> = gradient(of: sqrt)
let dsigmoid: (Tensor<Float>)->Tensor<Float> = gradient(of: sigmoid)
let dmysigmioid: (Tensor<Float>)->Tensor<Float> = gradient(of: mysigmoid)

// Check sigmoid and mysigmoid are same
let ten: Tensor<Float> = [-1, -0.5, 0, 0.5, 1]
print(sigmoid(ten))
print(mysigmoid(ten))

// The gradient of sqrt where x=0.5 (= 0.7071...)
print(dsqrt(Tensor<Float>(0.5))) // 0.36787945

// The gradient of mysigmoid where x=0.9 (=0.2055...)
print(dmysigmoid(Tensor<Float>(0.9))) // 0.2055003

// Expects 0.2055...
print(dsigmoid(Tensor<Float>(0.9))) // 0.09000002

// This returns 0.2055..
print(dsigmoid(sigmoid(Tensor<Float>(0.9)))) // 0.20550032
```

It looks `dsigmoid` returns wrong value.

`sigmoid` and its gradient is defined here.
https://github.com/tensorflow/swift-apis/blob/323b5640c48822340f2f5410d6bff51972fc9f45/Sources/TensorFlow/Operators/Math.swift#L962-L975

And `Raw.sigmoidGrad` is here.
https://raw.githubusercontent.com/tensorflow/swift-apis/323b5640c48822340f2f5410d6bff51972fc9f45/Sources/TensorFlow/Bindings/RawOpsGenerated.swift

```swift
/// Computes the gradient of the sigmoid of `x` wrt its input.
///
/// Specifically, `grad = dy * y * (1 - y)`, where `y = sigmoid(x)`, and
/// `dy` is the corresponding input gradient.
@inlinable @inline(__always)
public static func sigmoidGrad<T: FloatingPoint & TensorFlowScalar>(
    _ y: Tensor<T>,
    dy: Tensor<T>
)
```

`_vjpSigmoid` passes `x` but what `sigmoidGrad` requires is `y`.


t-ae:
By the way, I added test code but couldn't test. (but tried equivalent code on colab.)
When I run `swift build`, it fails with the message below.

```
Assertion failed: (isa<GenericTypeParamDecl>(D) && "unexpected decl kind"), function isDeclXRef, file /Users/danielzheng/swift-tf/swift/lib/Serialization/Serialization.cpp, line 2043.
```

(It's odd message, since I'm not danielzheng, of course)

I want someone to test my code before merge.

t-ae:
@googlebot I signed it!

eaplatanios:
Thanks for catching this @t-ae!

eaplatanios:
@t-ae some tests fail with this error:
```
Test Case 'LayerTests.testSequential' started at 2019-07-10 03:42:15.515
/swift-apis/Tests/TensorFlowTests/LayerTests.swift:53: error: LayerTests.testSequential : XCTAssertEqual failed: ("0.24223351") is not equal to ("0.25301588") +/- ("0.0001") -
/swift-apis/Tests/TensorFlowTests/LayerTests.swift:53: error: LayerTests.testSequential : XCTAssertEqual failed: ("0.20923203") is not equal to ("0.21743035") +/- ("0.0001") -
/swift-apis/Tests/TensorFlowTests/LayerTests.swift:53: error: LayerTests.testSequential : XCTAssertEqual failed: ("0.32405508") is not equal to ("0.32044548") +/- ("0.0001") -
/swift-apis/Tests/TensorFlowTests/LayerTests.swift:53: error: LayerTests.testSequential : XCTAssertEqual failed: ("0.28849685") is not equal to ("0.2807928") +/- ("0.0001") -
```
Could you please update them appropriately?

t-ae:
@eaplatanios 
That test uses `sigmoid` and its gradient so result is changed.

It's you who added this test. How did you get these expected values?
https://github.com/tensorflow/swift-apis/blame/5b4300e9746204cbde006f8f7af59964d0e88a9c/Tests/TensorFlowTests/LayerTests.swift#L28-L57

I suspect these values were got from this code itself.
It's easy to simply update values to match results. But it's not valid for testing IMO.

t-ae:
As I wrote above, I can't build in my environment currently. I'll try that at first.
I'm glad if someone update test while I'm trying to build.

t-ae:
OK, I managed to build.

I'm not sure what is the main purpose of this test.
(Building `Sequential` model? its differentiability?)

Anyway, it looks the result values are not what count.
So I changed the test to simply check if loss is descending.

eaplatanios:
Yes a lot of the tests are not super well organized/justified right now. Usually we try to make them match some existing working implementation (e.g., using Python). This looks good to me now. Thanks for updating it!

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -43,16 +43,19 @@ final class LayerTests: XCTestCase {
             let optimizer = SGD(for: model)
             let x = Tensor<Float>([[0, 0], [0, 1], [1, 0], [1, 1]])
             let y = Tensor<Float>([0, 1, 1, 0])
+
```
rxwei: Remove redundant empty lines.

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -43,16 +43,19 @@ final class LayerTests: XCTestCase {
             let optimizer = SGD(for: model)
             let x = Tensor<Float>([[0, 0], [0, 1], [1, 0], [1, 1]])
             let y = Tensor<Float>([0, 1, 1, 0])
+            
+            let initialLoss = meanSquaredError(predicted: model(x).squeezingShape(at: 1), expected: y)
+
```
rxwei: Same here.

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -43,16 +43,19 @@ final class LayerTests: XCTestCase {
...
-                    let ŷ = model(x).squeezingShape(at: 1)
-                    return (y - ŷ).squared().sum()
+                    meanSquaredError(predicted: model(x).squeezingShape(at: 1), expected: y)
                 }
                 optimizer.update(&model, along: 𝛁model)
             }
-            assertEqual(
-                model.inferring(from: [[0, 0], [0, 1], [1, 0], [1, 1]]),
-                [0.25301588, 0.21743035, 0.32044548, 0.2807928], accuracy: 0.0001)
+
```
rxwei: Same here.

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -43,16 +43,19 @@ final class LayerTests: XCTestCase {
...
+                    meanSquaredError(predicted: model(x).squeezingShape(at: 1), expected: y)
                 }
                 optimizer.update(&model, along: 𝛁model)
             }
-            assertEqual(
-                model.inferring(from: [[0, 0], [0, 1], [1, 0], [1, 1]]),
-                [0.25301588, 0.21743035, 0.32044548, 0.2807928], accuracy: 0.0001)
+            
+            let updatedLoss = meanSquaredError(predicted: model(x).squeezingShape(at: 1), expected: y)
+
```
rxwei: Same here.

Tests/TensorFlowTests/LayerTests.swift
t-ae: Removed.

Tests/TensorFlowTests/TensorAutoDiffTests.swift
```diff
@@ -411,6 +416,7 @@ final class TensorAutoDiffTests: XCTestCase {
         ("testConcatenationPlusPlus", testConcatenationPlusPlus),
         ("testConcatenated", testConcatenated),
         ("testTransposed", testTransposed),
+        ("testSigmoid", testSigmoid),
```
rxwei: The definition of `testSigmoid()` is placed between `testRelu()` and `testSoftmax()`. This line should be moved below `("testRelu", testRelu)` for consistency.

t-ae: Moved `testSigmoid` below `testRelu` line.

eaplatanios: Sorry to be nitpicky about this but it'd be better to move the `testSigmoid()` definition above the `testRelu()` definition instead so it matches the order in which they are defined in `Math.swift`.

t-ae: It makes sense.
Moved `testRelu` definition.
(In first commit, I was puzzled which should be appear first `testRelu` or `testSigmoid`. And somehow left them inconsistent order😅)

eaplatanios: Thanks @t-ae.

eaplatanios: Thanks @t-ae. :)

Issue #343, opened 2019-07-10
Remove 'PointwiseMultiplicative' and 'ElementaryFunctions' requirements from 'Layer'


rxwei:
#322 broke some existing code, and these requirements are pretty heavyweight. We should drop these requirements for now.

Shashi456:
Can I take this up? This would mean I'll have to remove layer :PointwiseMultiplicative' and 'ElementaryFunctions, from every layer it was added to ?

eaplatanios:
@rxwei could you possibly add a test with some example code that breaks? It would be great to being in tests for breaking code so we avoid issues in the future and can better think of an appropriate design.

rxwei:
Let's hold off on this change for now. I've experienced breakage in custom defined parameterless layers, but it was simply because they were not using `ParameterlessLayer`. Thanks @Shashi456 for taking a stab!

PR #344, opened 2019-07-10
Remove Layer requirement of PointwiseMultiplicative and ElementaryFunctions

Shashi456:
Fix: #343 

I was wondering if i should add a FIXIT by mentioning we reverted the change.

### DIFF
--- a/Sources/TensorFlow/Layer.swift
+++ b/Sources/TensorFlow/Layer.swift
@@ -20,8 +20,7 @@
 /// `Layer` instances define a differentiable `applied(to:)` method for mapping inputs to
 /// outputs.
 public protocol Layer: Differentiable, KeyPathIterable
-    where TangentVector: VectorProtocol & ElementaryFunctions &
-                         PointwiseMultiplicative & KeyPathIterable,
+    where TangentVector: VectorProtocol & KeyPathIterable,
           AllDifferentiableVariables == TangentVector {
     /// The input type of the layer.
     associatedtype Input: Differentiable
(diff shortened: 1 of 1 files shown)


rxwei:
I don't think it would be this trivial. Did you verify that it compiles?

Shashi456:
Yeah, forgot to do it it on the latest build.  just pulling it and compiling right now.

Sources/TensorFlow/Layers/Core.swift
```diff
@@ -28,7 +28,7 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
 /// Dropout consists in randomly setting a fraction of input units to `0` at each update during
 /// training time, which helps prevent overfitting.
 @frozen
-public struct Dropout<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
+public struct Dropout<Scalar: TensorFlowFloatingPoint>: Layer {
```
rxwei: These should _not_ be changed.

Sources/TensorFlow/Layer.swift
```diff
@@ -44,8 +43,7 @@ public extension Layer {
 }
 
 /// An empty struct representing empty `TangentVector`s for parameterless layers.
-public struct EmptyTangentVector: Differentiable, VectorProtocol, ElementaryFunctions,
-                                  PointwiseMultiplicative, KeyPathIterable {
+public struct EmptyTangentVector: Differentiable, VectorProtocol, KeyPathIterable {
```
rxwei: Conformances should not be dropped from this type. Only the `Layer` protocol.

Shashi456:
@rxwei yeah I will revert them, I pushed them to this branch by mistake. I was just getting some SIL errors and changed the wrong conformance.

Shashi456:
@rxwei You were right it's not as trivial. One evident error is ` error: type 'Function<Input, Output>.AllDifferentiableVariables' does not conform to protocol 'VectorProtocol'`

Shashi456:
@rxwei this commit passes on the latest build without any errors.

eaplatanios:
This should make layers not directly usable with the momentum-based optimizers such as Adam, AdaGrad, RMSProp, etc. I would be surprised if it doesn't. Maybe we can manually add them to layers that can conform to them so we keep them compatible with the optimizers.

eaplatanios:
It also breaks SGD. I just tested on my machine.

Shashi456:
@eaplatanios I was wondering that it would break optimizers. 

@rxwei closed #343. I'll close this too.

PR #346, opened 2019-07-10
Split the `Layer` protocol into `Layer` and `Module` where `Module` does not require `Input: Differentiable`.

eaplatanios:
`Layer` currently requires `Input: Differentiable`. This results in boilerplate code for layers with non-differentiable inputs (e.g., embedding layers that take integer tensors as input). However, the constraint is required for chaining together layers (e.g., as done in the `Sequential` layer). This PR includes the following changes:

- Introduced a new `Module` protocol that is equivalent to the current `Layer` with the `Input: Differentiable` constraint removed and `callAsFunction` annotated with `@differentiable(wrt: self)` as opposed to `@differentiable(wrt: (self, input))`.
- Conformed `Layer` to `Module` with the additional `@differentiable(wrt: (self, input))` attribute on `callAsFunction`.
- Switched `Embedding` from conforming to `Layer` to conforming to `Module` and removed the boilerplate `EmbeddingInput` struct.
- Adapted `Sequential` so that it can work with both `Module`s and `Layer`s, with the constraint that `Module`s can only appear in the beginning of a sequential layer chain.

### DIFF
--- a/Sources/TensorFlow/Bindings/EagerExecution.swift
+++ b/Sources/TensorFlow/Bindings/EagerExecution.swift
@@ -68,7 +68,6 @@ internal struct TFE_Op: TFTensorOperation {
         let count = input._tensorHandleCount
         var buffer = UnsafeMutableBufferPointer<CTensorHandle>.allocate(capacity: Int(count))
         defer { buffer.deallocate() }
-        let pointer = UnsafeMutablePointer<OpaquePointer?>(buffer.baseAddress)
         input._unpackTensorHandles(into: buffer.baseAddress)
         for i in 0..<Int(count) {
             TFE_OpAddInput(op, buffer[i], status)
--- a/Sources/TensorFlow/Layer.swift
+++ b/Sources/TensorFlow/Layer.swift
@@ -48,7 +48,7 @@ public protocol Layer: XXX where Input: Differentiable {
 public extension Layer {
     @differentiable
     func call(_  input: Input) -> Output {
-        return callAsFunction(input)
+        callAsFunction(input)
     }
 }
 
@@ -69,7 +69,10 @@ public struct EmptyTangentVector: Differentiable, VectorProtocol, ElementaryFunc
 /// A parameterless neural network layer.
 ///
 /// The `TangentVector` of parameterless layers is always `EmptyTangentVector`.
-public protocol ParameterlessLayer: Layer where AllDifferentiableVariables == EmptyTangentVector {}
+public protocol ParameterlessLayer: Layer where AllDifferentiableVariables == EmptyTangentVector {
+    @differentiable
+    func callAsFunction(_ input: Input) -> Output
+}
 
 public extension ParameterlessLayer {
     var allDifferentiableVariables: EmptyTangentVector {
--- a/Sources/TensorFlow/Layer.swift
+++ b/Sources/TensorFlow/Layer.swift
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-public protocol XXX: Differentiable, KeyPathIterable
+public protocol Module: Differentiable, KeyPathIterable
     where TangentVector: VectorProtocol & ElementaryFunctions &
                          PointwiseMultiplicative & KeyPathIterable,
           AllDifferentiableVariables == TangentVector {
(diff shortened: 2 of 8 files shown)


rxwei:
> 1. Rename `XXX` to something better :) I don't like `Module` because modules are usually used to refer to what we call `Layer`, which would cause confusion. Do you have any other ideas? I thought of `InputLayer` but that is too restrictive a description semantically.

`InputLayer` would make models look weird. `Model` would make the `Embedding` layer look weird. Maybe there's something in between?

rxwei:
It seems to be a bug. Do you wanna fix it? :)
You can start by pulling one of those parameterless layers out to a single file, compile it with `-Xllvm -debug-only=differentiation`, and look at the activity analysis result.

Shashi456:
Can we simply name them LayerV1, LayerV2?

rxwei:
> Can we simply name them LayerV1, LayerV2?

We cannot. These are not user-friendly names but merely a reflection of implementation details that users do not need to know.

Shashi456:
Sorry for bouncing name ideas, but `FeatureLayer` ? Initial layers since most often embedding are often feature representations of the input. And even in sequential, the name would seem neat.

eaplatanios:
@rxwei this is ready for review. All tests pass now. I went with `Module` after all, as you initially suggested. I couldn't come up with something better and I think this is not a good reason to delay this further. I haven't been able to debug the invalid `result does not depend on differentiation arguments` warnings but have verified that all tests pass and thus the warnings have no effect on the derivatives. I believe activity analysis may be being triggered twice, once for `Module` and once for `Layer`, due to the overridden function declaration. The warning is probably then triggered by the `Module` function analysis, but the `Layer` function analysis still goes through and results in gradients being propagated correctly.

In either case, since tests pass and this removes an annoying restriction I believe we can go ahead review and merge this.

eaplatanios:
Has the CI toolchain been updated? Because I believe the errors are due to an older toolchain.

rxwei:
Going with Module for now sounds good. We should bring this up on the mailing list as soon as possible and talk about this on Friday's design review.

eaplatanios:
@rxwei I'll being it up later today. Is the toolchain updated in the CI? Because I think the test failures are due to an older toolchain.

rxwei:
IIRC nightly toolchains are tracking the tensorflow-0.4 branch right now. @bgogul

eaplatanios:
I see. That would mean they do not include the fix from apple/swift#25974 which is what causing the CI failures I believe.

eaplatanios:
@rxwei I ran tests about an hour ago. Unless the toolchain has been updated in the meantime, the tests won't pass.

rxwei:
A new toolchain was produced at ~10:45ish.

eaplatanios:
Coincidentally, [that's exactly when I ran them](https://source.cloud.google.com/results/invocations/a66d2554-cbfa-49bf-9dc5-b0b164a4ee1a/targets/s4tf%2Fapis%2Fubuntu%2Fpresubmit/log) so it may succeed now. :)

eaplatanios:
This passes locally so I'm not sure what's wrong here. Are you sure the toolchain is the latest one from the `tensorflow` branch? If it is from the `tensorflow-0.4` branch are you sure that the conditional differentiable PR has been merged there?

JacopoMangiavacchi:
I see we already have an initializer for using a pretrained embeddings table but it is not clear for me how we can froze also the embedding table itself. Keras use to have a **trainable** parameter in the constructor.  Don't you think we should introduce that parameter too to this initializer?  If value is true we should do not differentiate the weights table too isn't it?

eaplatanios:
@JacopoMangiavacchi there is separate ongoing work on using property wrappers to allow freezing trainable parameters more generally for layers. For that reason, for now we follow the same convention we've been using for layers all along.

rxwei:
@eaplatanios Could you update the PR description to accurately describe the changes?

eaplatanios:
@rxwei done.

rxwei:
Sorry, I meant PR description (not just the title). If you could list out the various changes, that'd be great. We want to include that in the commit log.

eaplatanios:
Sorry my bad. I just updated the description.

PR #348, opened 2019-07-10
Implement isAlmostEqual method for floating-point tensors.

dominikgrewe:


Sources/TensorFlow/Operators/Comparison.swift
```diff
@@ -227,3 +227,16 @@ public extension StringTensor {
         return Raw.equal(self, other)
     }
 }
+
+public extension Tensor where Scalar: FloatingPoint {
+    /// Returns true if tensors are of equal shape and all pairs of elements are approximately
```
rxwei:
```suggestion
    /// Returns `true` if tensors are of equal shape and all pairs of scalars are approximately
```

Sources/TensorFlow/Operators/Comparison.swift
```diff
@@ -227,3 +227,16 @@ public extension StringTensor {
...
+    /// equal.
+    @inlinable
+    func isAlmostEqual(
+        to other: Tensor,
+        tolerance: Scalar = Scalar.ulpOfOne.squareRoot()
+    ) -> Bool {
+        return self.shape == other.shape &&
+          zip(self.scalars, other.scalars).allSatisfy{abs($0.0 - $0.1) < tolerance}
```
rxwei:
```suggestion
            zip(self.scalars, other.scalars).allSatisfy { abs($0.0 - $0.1) < tolerance }
```

Tests/TensorFlowTests/OperatorTests/ComparisonTests.swift
```diff
@@ -28,8 +28,22 @@ final class ComparisonOperatorTests: XCTestCase {
         XCTAssertTrue(x < y)
     }
 
+    func testIsAlmostEqual() {
```
dan-zheng: Could you please add some tests involving approximate self equality of nan/infinity?
`Tensor<Float>([.nan, .infinity])`

Sources/TensorFlow/Operators/Comparison.swift
```diff
@@ -227,3 +227,16 @@ public extension StringTensor {
...
+public extension Tensor where Scalar: FloatingPoint {
+    /// Returns `true` if tensors are of equal shape and all pairs of scalars are approximately
+    /// equal.
+    @inlinable
+    func isAlmostEqual(
+        to other: Tensor,
+        tolerance: Scalar = Scalar.ulpOfOne.squareRoot()
+    ) -> Bool {
+        return self.shape == other.shape &&
+            zip(self.scalars, other.scalars).allSatisfy { abs($0.0 - $0.1) < tolerance }
```
dan-zheng: Note: this implementation is a bit simpler than the [floating-point approximate equality pitch](https://forums.swift.org/t/approximate-equality-for-floating-point/22420) implementation, which performs two kinds of scaling (one to handle infinite values, one in the general case).

If the pitch becomes an accepted proposal, we may want to adopt its implementation.

eaplatanios:
Can we change the test helpers that test for approximate equality to use this function? Also, isn't `isApproximatelyEqual` a better name for this?

Sources/TensorFlow/Operators/Comparison.swift
```diff
@@ -227,3 +227,16 @@ public extension StringTensor {
...
+
+public extension Tensor where Scalar: FloatingPoint {
+    /// Returns `true` if tensors are of equal shape and all pairs of scalars are approximately
+    /// equal.
+    @inlinable
+    func isAlmostEqual(
+        to other: Tensor,
+        tolerance: Scalar = Scalar.ulpOfOne.squareRoot()
+    ) -> Bool {
+        return self.shape == other.shape &&
```
eaplatanios: You can use `Tensor.elementsApproximatelyEqual(_:tolerance:)` directly to make this faster. Also, maybe we should also add the `to` argument label to all these other functions that check for element-wise equality.

rxwei:
> Can we change the test helpers that test for approximate equality to use this function? Also, isn't `isApproximatelyEqual` a better name for this?

This is to follow [SE-0259](https://github.com/apple/swift-evolution/blob/master/proposals/0259-approximately-equal.md). If that gets changed, we should change ours to follow.

Tests/TensorFlowTests/OperatorTests/ComparisonTests.swift
```diff
@@ -28,8 +28,22 @@ final class ComparisonOperatorTests: XCTestCase {
...
+        let x = Tensor<Float>([0.1, 0.2, 0.3, 0.4])
+        let y = Tensor<Float>([0.0999, 0.20001, 0.2998, 0.4])
+        XCTAssertTrue(x.isAlmostEqual(to: y, tolerance: 0.01))
+    }
+
+    func testIsNotAlmostEqual() {
```
eaplatanios: It may be better to put this also in `testIsAlmostEqual()` since they are both testing the same function (i.e., `isAlmostEqual`).

Sources/TensorFlow/Operators/Comparison.swift
dominikgrewe: Do you want me to adopt the same scaling? Should we add this to `elementsApproximatelyEqual` and then call that here as suggested by @eaplatanios ?

Sources/TensorFlow/Operators/Comparison.swift
rxwei: > Also, maybe we should also add the `to` argument label to all these other functions that check for element-wise equality.

This `elementsApproximatelyEqual(_:tolerance:)` method name is a sentence rather than a verb phrase where "equal" is a transitive verb. Same for `Sequence.elementsEqual(_:)`.

We should change "approximately" to "almost" to for consistency with `isAlmostEqual(to:)`,

Sources/TensorFlow/Operators/Comparison.swift
dan-zheng: For simplicity, not adopting for now SGTM.
We can re-evaluate if [SE-0259](https://github.com/apple/swift-evolution/blob/master/proposals/0259-approximately-equal.md) is accepted.

Sources/TensorFlow/Operators/Comparison.swift
dominikgrewe: I've renamed `elementsApproximatelyEqual` to `elementsAlmostEqual` and this is now called from `isAlmostEqual`. For consistency I also changed the condition on `Scalar` to now conform to `TensorFlowFloatingPoint` so that it's convertible to `Double`. Ptal.

Sources/TensorFlow/Operators/Comparison.swift
```diff
@@ -227,3 +227,16 @@ public extension StringTensor {
...
+
+public extension Tensor where Scalar: TensorFlowFloatingPoint {
+    /// Returns `true` if tensors are of equal shape and all pairs of scalars are approximately
+    /// equal.
+    @inlinable
+    func isAlmostEqual(
+        to other: Tensor,
+        tolerance: Scalar = Scalar.ulpOfOne.squareRoot()
+    ) -> Bool {
+        return self.shape == other.shape &&
```
eaplatanios: I'm wondering if we want to have this check for mismatching shapes here. @rxwei wouldn't a precondition be better? I believe it would be misleading for this to return `false` when the shapes are mismatching as it should be expected that it operates over tensors with matching shapes.

Another issue is that of broadcasting. If we want to support broadcasting for this function we should remove that check altogether.

rxwei: +1

PR #350, opened 2019-07-10
fix _vjpRsqrt and add test

mikowals:
This fixes `_vjpRsqrt` again and adds a test.  The previous fix had `pow(x, 3 / 2)` which was doing division of integers and rounding 3 divided by 2 down to 1.

Rather than just make those floating point I changed the formula to match the [documentation of `Raw.rsqrtGrad` ](https://github.com/tensorflow/swift-apis/blob/master/Sources/TensorFlow/Bindings/RawOpsGenerated.swift).

### DIFF
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -855,7 +855,7 @@ internal func _vjpRsqrt<T: TensorFlowFloatingPoint>(
     _ x: Tensor<T>
 ) -> (Tensor<T>, (Tensor<T>) -> Tensor<T>) {
     let value = rsqrt(x)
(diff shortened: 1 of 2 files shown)


eaplatanios:
Why not use `Raw.rsqrtGrad` directly?

mikowals:
I didn't use `Raw.rsqrtGrad` only because the original implementation did not.

eaplatanios:
I'm not aware why that was not used, but since you're fixing this, could you please switch it to use `Raw.rsqrtGrad`?

mikowals:
I  updated the code to use `Raw.rsqrtGrad()`.

eaplatanios:
Great, thanks! :)

PR #351, opened 2019-07-11
Added support for negative 'axis' argument values to 'Tensor.unstacked(alongAxis:)'.

eaplatanios:
Without this fix the gradients of `Tensor.init(stacking:alongAxis:)` are wrong when `axis` is negative. It also makes it consistent with other functions that take in axis arguments.

### DIFF
--- a/Sources/TensorFlow/Operators/Basic.swift
+++ b/Sources/TensorFlow/Operators/Basic.swift
@@ -34,9 +34,10 @@ public extension TensorFlowScalar {
 }
 
 public extension Tensor {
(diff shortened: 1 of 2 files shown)


eaplatanios:
@dan-zheng I added a test as you requested, but it currently causes a compiler crash that is irrelevant to this PR. I filed [TF-653](https://bugs.swift.org/browse/TF-653) to track it.

eaplatanios:
@rxwei I commented out the failing test and added a TODO related to TF-653. Could we merge this in the meantime since it corrects some unexpected behavior for `Tensor.unstacked(alongAxis:)`?

rxwei:
SGTM

PR #353, opened 2019-07-11
Added support for 'Tensor.cumulativeSum(alongAxis:exclusive:reverse:)'.

eaplatanios:


Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1895,11 +1969,22 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
...
-	      let unsqueezed = v.expandingShape(at: axes.scalars.map { Int($0) })
+        return (value, { [shape = shapeTensor] seed in
+	      let unsqueezed = seed.expandingShape(at: axes.scalars.map { Int($0) })
 	      return unsqueezed.broadcasted(toShape: shape) / Tensor(count)
 	    })
     }
+
+    @inlinable
+    func _vjpCumulativeSum(
+        alongAxis axis: Tensor<Int32>,
```
rxwei: Shouldn’t this be “axes” instead, since it’s a tensor?

eaplatanios: Cumulative sum can only be performed over a single axis. Since we cannot enforce that in the tensor type, I hope the name make it clear. I can also add a precondition although that would cause lazy tensor materialization and so I avoided it.

rxwei: Why not take an Int instead?

eaplatanios: To allow it to be used without lazy tensor materialization in cases where that axis is the result of some other tensor operation (e.g., `Tensor.rank` or `Tensor.shape[i]`). I have also added an overload taking `Int` directly.

rxwei: Oh ok!

eaplatanios: Tests pass so if you approve this, I can go ahead and merge it.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1889,6 +1889,80 @@ public extension Tensor where Scalar: Numeric {
...
+    /// By setting the `exclusive` argument to `true`, an exclusive cumulative sum is performed
+    /// instead:
+    /// ```
+    /// Tensor<Float>([a, b, c]).cumulativeSum(exclusive: true) = Tensor<Float>([0, a, a + b])
+    /// ```
+    /// By setting the `reverse` argument to `true`, the cumulative sum is performed in the
+    /// opposite direction:
+    /// ```
+    /// Tensor<Float>([a, b, c]).cumulativeSum(reverse: true) = 
+    ///   Tensor<Float>([a + b + c, a + b, a])
```
```suggestion
    ///     Tensor<Float>([a + b + c, a + b, a])
```

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1889,6 +1889,80 @@ public extension Tensor where Scalar: Numeric {
...
+    /// ```
+    /// By setting the `exclusive` argument to `true`, an exclusive cumulative sum is performed
+    /// instead:
+    /// ```
+    /// Tensor<Float>([a, b, c]).cumulativeSum(exclusive: true) = Tensor<Float>([0, a, a + b])
+    /// ```
+    /// By setting the `reverse` argument to `true`, the cumulative sum is performed in the
+    /// opposite direction:
+    /// ```
+    /// Tensor<Float>([a, b, c]).cumulativeSum(reverse: true) =
```
```suggestion
    /// Tensor<Float>([a, b, c]).cumulativeSum(reverse: true) == 
```

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1901,8 +1975,8 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
     @inlinable
     func _vjpSum(squeezingAxes axes: Tensor<Int32>) -> (Tensor, (Tensor) -> Tensor) {
         let value = sum(squeezingAxes: axes)
-        return (value, { [shape = shapeTensor] v in
-	      let unsqueezed = v.expandingShape(at: axes.scalars.map { Int($0) })
+        return (value, { [shape = shapeTensor] seed in
+	      let unsqueezed = seed.expandingShape(at: axes.scalars.map { Int($0) })
```
rxwei: The indentation here seems wrong.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1901,8 +1975,8 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
     @inlinable
     func _vjpSum(squeezingAxes axes: Tensor<Int32>) -> (Tensor, (Tensor) -> Tensor) {
         let value = sum(squeezingAxes: axes)
-        return (value, { [shape = shapeTensor] v in
-	      let unsqueezed = v.expandingShape(at: axes.scalars.map { Int($0) })
+        return (value, { [shape = shapeTensor] seed in
```
rxwei: I'd suggest using "v" instead of "seed". We've been avoiding the word seed in both the AD implementation and the library.

eaplatanios: I addressed the comments, but I'm curious, why don't we use a more descriptive name?

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1902,8 +1976,8 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
     func _vjpSum(squeezingAxes axes: Tensor<Int32>) -> (Tensor, (Tensor) -> Tensor) {
         let value = sum(squeezingAxes: axes)
         return (value, { [shape = shapeTensor] v in
-	      let unsqueezed = v.expandingShape(at: axes.scalars.map { Int($0) })
-	      return unsqueezed.broadcasted(toShape: shape)
+	        let unsqueezed = v.expandingShape(at: axes.scalars.map { Int($0) })
```
rxwei: Indentation is still incorrect here. It should indent by 4 from `return ...`.

Sources/TensorFlow/Operators/Math.swift
rxwei: You could call it "tangent" or "direction", but "seed" is only confusing because only the really traditional automatic differentiation community calls it "seed".

Sources/TensorFlow/Operators/Math.swift
eaplatanios: Sorry there was a tab there. Good catch.

Sources/TensorFlow/Operators/Math.swift
eaplatanios: I see...yeah a more descriptive name would be better but the best I can think of at the moment is `backpropagatedGradient` which is too long. Let's leave it for now I guess. :)

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1976,8 +1976,8 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
     func _vjpSum(squeezingAxes axes: Tensor<Int32>) -> (Tensor, (Tensor) -> Tensor) {
         let value = sum(squeezingAxes: axes)
         return (value, { [shape = shapeTensor] v in
-	        let unsqueezed = v.expandingShape(at: axes.scalars.map { Int($0) })
-	        return unsqueezed.broadcasted(toShape: shape)
+            let unsqueezed = v.expandingShape(at: axes.scalars.map { Int($0) })
+            return unsqueezed.broadcasted(toShape: shape)
 	    })
```
rxwei: The `}` should be un-indented too.

PR #354, opened 2019-07-11
Added support for computing tensor moments.

eaplatanios:
This also removes the need for a custom VJP for batch normalization.


rxwei:
Awesome

eaplatanios:
Done in #355 :)

PR #355, opened 2019-07-11
Added a test comparing the old batch-norm VJP with the automatically derived one.

eaplatanios:
As requested in #354 by @dan-zheng.

They two implementations give the same result.

### DIFF
--- a/Tests/TensorFlowTests/TensorAutoDiffTests.swift
+++ b/Tests/TensorFlowTests/TensorAutoDiffTests.swift
@@ -25,58 +25,6 @@ func vjpFoo(_ x: Tensor<Float>) -> (Tensor<Float>, (Tensor<Float>) -> Tensor<Flo
     return (foo(x), { v in v })
 }
 
-@differentiable(wrt: (x, offset, scale), vjp: _vjpOldBatchNormalized)
-func oldBatchNormalized(
-    _ x: Tensor<Float>,
-    alongAxis axis: Int,
-    offset: Tensor<Float> = Tensor(0),
-    scale: Tensor<Float> = Tensor(1),
-    epsilon: Float = 0.001
-) -> Tensor<Float> {
-    let mean = x.mean(alongAxes: axis)
-    let squaredDiff: Tensor = Raw.squaredDifference(x, mean)
-    let variance = squaredDiff.mean(alongAxes: axis)
-    let inv = rsqrt(variance + epsilon) * scale
-    return x * inv + offset - mean * inv
-}
-
-func _vjpOldBatchNormalized(
-    _ x: Tensor<Float>,
-    alongAxis axis: Int,
-    offset: Tensor<Float>,
-    scale: Tensor<Float>,
-    epsilon: Float
-) -> (Tensor<Float>, (Tensor<Float>) -> (Tensor<Float>, Tensor<Float>, Tensor<Float>)) {
-    let value = oldBatchNormalized(
-        x,
-        alongAxis: axis,
-        offset: offset,
-        scale: scale,
-        epsilon: epsilon)
-    return (value, { v in
-        let mean = x.mean(alongAxes: axis)
-        let squaredDiff: Tensor = Raw.squaredDifference(x, mean)
-        let variance = squaredDiff.mean(alongAxes: axis)
-
-        let diff = x - mean
-        let inv = rsqrt(variance + epsilon)
-        let norm = diff * inv
-
-        let dNorm = v * scale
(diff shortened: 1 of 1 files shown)


Tests/TensorFlowTests/TensorAutoDiffTests.swift
```diff
@@ -392,7 +444,20 @@ final class TensorAutoDiffTests: XCTestCase {
         let expected: Tensor<Float> = inputTensor
         XCTAssertEqual(expected, pb(inputTensor))
     }
-    
+
+    func testBatchNormalized() {
+        let x = Tensor<Float>([
```
dan-zheng: Could you please use a value of `x` for which `gradient(at: x) { $0.batchNormalized(alongAxis: 1).sum() }` contains less zeros?

Try this one:
```swift
let x = Tensor<Float>([
    [  -1.0474433,  -0.11914538,  -0.08634827,   0.15446888,   -1.0572497],
    [   1.5165012,    0.3753972,  -0.30856386,   -0.3100725,   -1.9584457],
    [ 0.006384419,    1.4424847,   0.91568077,   0.66328526,   -1.0794537],
    [    1.056803,   0.14263044,   -1.8308276,    0.4189805,    0.6933893],
    [  0.30175626,  -0.16121633,   -0.4191958,  -0.53092813, -0.029484272]])

print(gradient(at: x) { $0.batchNormalized(alongAxis: 1).sum() })
// [[-1.1920929e-07, -1.1920929e-07, -1.1920929e-07, -1.1920929e-07, -1.1920929e-07],
//  [-1.7881393e-07, -1.7881393e-07, -1.7881393e-07, -1.7881393e-07, -1.7881393e-07],
//  [-2.3841858e-07, -2.3841858e-07, -2.3841858e-07, -2.3841858e-07, -2.3841858e-07],
//  [ 5.9604645e-08,  5.9604645e-08,  5.9604645e-08,  5.9604645e-08,  5.9604645e-08],
//  [ 2.3841858e-07,  2.3841858e-07,  2.3841858e-07,  2.3841858e-07,  2.3841858e-07]]
```

eaplatanios: Ok it turns out the numbers are different and I believe the previous implementation was wrong. The numbers differ by quite a lot sometimes actually.

I rederived it and instead of this:
```swift
        let dNorm = v * scale
        let dVariance = -(dNorm * diff).sum(alongAxes: axis) / 2 * TensorFlow.pow(inv, -3)
        // Note: `dMean` is split into two lines to avoid the "compiler is unable to 
        // type-check this expression in reasonable time" error.
        var dMean = (-dNorm * inv).sum(alongAxes: axis)
        dMean = dMean + dVariance * (-diff * 2).mean(alongAxes: axis)
        let dim = Tensor<Float>(x.shapeTensor[axis])
        let tmp = (dNorm * inv) + (dVariance * 2 * dMean / dim)
        let dSelf = tmp + (dMean / dim)
```
which was in the old implementation, I get this:
```swift
        let dNormInv = v * scale * inv
        let tmp1 = dNormInv.sum(alongAxes: axis)
        let tmp2 = value * (dNormInv * value).sum(alongAxes: axis)
        let dSelf = dNormInv - (tmp1 + tmp2) / Tensor<Float>(x.shapeTensor[axis])
```
Using this also results in the same numbers as the new one which is automatically derived. I can add a test for this with correct values, but be aware that this might break existing code that was differentiating through batch norm layers because the gradient was computed incorrectly.

dan-zheng: Adding a test with correct gradients values SGTM! Verifying gradients against Python TF would be great.

eaplatanios: Unfortunately I don't have time to check what happens with Python TF now, but I'm fairly confident the current implementation is correct given that I get the same result with derivation by hand and with the automated gradient derivation in Swift. I may look into Python TF tomorrow or during the weekend if I get a break.

eaplatanios: @dan-zheng I just added a reference value computed using the python API and the test passes. I also included the Python code in a comment so it can be reproduced if needed.

dan-zheng: Thanks so much! This is a great precedent for verifying gradients by providing Python code.

eaplatanios: No problem :)

PR #356, opened 2019-07-12
Standardized the loss functions.

eaplatanios:
This PR makes the following changes:

1. Removed reductions (i.e., `mean()` or `sum()`) from the loss functions. The are two main reasons for that: (i) they were being used inconsistently (sometimes sum was used and sometimes mean) and users might run into issues with their learning rates expecting mean or sum (it's better for them to be explicitly aware of what reduction they're using to reduce the chance of bugs), and (ii) they are irreversible but sometimes the user may want the element-wise loss (e.g., `softmaxCrossEntropy` is also the probability density function of the Categorical distribution and we might want to use it as such -- in this case the reduction does not make sense).
2. Added `@differentiable` to some instances of `Tensor.max(...)` and `Tensor.min(...)` where it was missing.
3. Removed the `assertElementsEqual` helper from the loss tests file and switched to using the standard helper we have for all our test suites.

Note that I kept the `mean()` reduction only for `meanAbsoluteError`, `meanSquaredError`, `meanSquaredLogarithmicError`, and `meanAbsolutePercentageError` where it is explicit from the name.

After this PR, the `squaredHingeLoss` function is rendered kind of useless as it is just defined as `hingeLoss(...).squared()` so we may want to remove it. Also, for a lot of these functions, I'm not sure if `Loss.swift` is the right place; maybe we should move all of them to `Math.swift` since they are pretty general math computations, some of which can be used for many other things than losses (e.g., the example mentioned above about `softmaxCrossEntropy`).

This will require a minimal set of changes to existing code. Namely, wherever a loss function is used, `.mean()` or `.sum()` will need to be added, depending on which loss function that is. This will allow behavior to stay the same after the update.

@rxwei @dan-zheng what do you guys think about this?

### DIFF
--- a/Sources/TensorFlow/Loss.swift
+++ b/Sources/TensorFlow/Loss.swift
@@ -17,6 +17,7 @@
 /// - Parameters:
 ///   - predicted: Predicted outputs from a neural network.
 ///   - expected: Expected values, i.e. targets, that correspond to the correct output.
+///   - reduction: Reduction to apply on the computed element-wise loss values.
 @differentiable(wrt: predicted)
 public func l1Loss<Scalar: TensorFlowFloatingPoint>(
     predicted: Tensor<Scalar>,
@@ -31,6 +32,7 @@ public func l1Loss<Scalar: TensorFlowFloatingPoint>(
 /// - Parameters:
 ///   - predicted: Predicted outputs from a neural network.
 ///   - expected: Expected values, i.e. targets, that correspond to the correct output.
+///   - reduction: Reduction to apply on the computed element-wise loss values.
 @differentiable(wrt: predicted)
 public func l2Loss<Scalar: TensorFlowFloatingPoint>(
     predicted: Tensor<Scalar>,
@@ -102,6 +104,7 @@ public func meanAbsolutePercentageError<Scalar: TensorFlowFloatingPoint>(
 /// - Parameters:
 ///   - predicted: Predicted outputs from a neural network.
 ///   - expected: Expected values, i.e. targets, that correspond to the correct output.
+///   - reduction: Reduction to apply on the computed element-wise loss values.
 @differentiable(wrt: predicted)
 public func hingeLoss<Scalar: TensorFlowFloatingPoint>(
     predicted: Tensor<Scalar>,
@@ -116,6 +119,7 @@ public func hingeLoss<Scalar: TensorFlowFloatingPoint>(
 /// - Parameters:
 ///   - predicted: Predicted outputs from a neural network.
 ///   - expected: Expected values, i.e. targets, that correspond to the correct output.
+///   - reduction: Reduction to apply on the computed element-wise loss values.
 @differentiable(wrt: predicted)
 public func squaredHingeLoss<Scalar: TensorFlowFloatingPoint>(
     predicted: Tensor<Scalar>,
@@ -130,6 +134,7 @@ public func squaredHingeLoss<Scalar: TensorFlowFloatingPoint>(
 /// - Parameters:
 ///   - predicted: Predicted outputs from a neural network.
 ///   - expected: Expected values, i.e. targets, that correspond to the correct output.
+///   - reduction: Reduction to apply on the computed element-wise loss values.
 @differentiable(wrt: predicted)
 public func categoricalHingeLoss<Scalar: TensorFlowFloatingPoint>(
     predicted: Tensor<Scalar>,
@@ -141,11 +146,13 @@ public func categoricalHingeLoss<Scalar: TensorFlowFloatingPoint>(
     return reduction(max(Tensor(0), negative - positive + Tensor(1)))
 }
 
-/// Returns the logarithm of the hyperbolic cosine of the error between predictions and expectations.
+/// Returns the logarithm of the hyperbolic cosine of the error between predictions and
+/// expectations.
 ///
 /// - Parameters:
 ///   - predicted: Predicted outputs from a neural network.
 ///   - expected: Expected values, i.e. targets, that correspond to the correct output.
+///   - reduction: Reduction to apply on the computed element-wise loss values.
 @differentiable(wrt: predicted)
 public func logCoshLoss<Scalar: TensorFlowFloatingPoint>(
     predicted: Tensor<Scalar>,
@@ -161,6 +168,7 @@ public func logCoshLoss<Scalar: TensorFlowFloatingPoint>(
 /// - Parameters:
 ///   - predicted: Predicted outputs from a neural network.
 ///   - expected: Expected values, i.e. targets, that correspond to the correct output.
+///   - reduction: Reduction to apply on the computed element-wise loss values.
 @differentiable(wrt: predicted)
 public func poissonLoss<Scalar: TensorFlowFloatingPoint>(
     predicted: Tensor<Scalar>,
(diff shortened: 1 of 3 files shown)


Shashi456:
I believe we should keep them in loss file itself. 

I think we should add reduction functions, instead of removing them entirely with `none` as default. More often than not, I've seen loss function usage inspired from the formula and they always either have sum or mean. I know standardization is a good move forward, but it would be better to have a reduction option.

Finally was there reason to call other l1 and l2 loss functions from other loss functions? Just trying to understand this, I thought since it adds to additional function calling, is that overhead minimal?

eaplatanios:
> I believe we should keep them in loss file itself.

I'm fine with that. It's just a thought that I believed made sense to consider.

> I think we should add reduction functions, instead of removing them entirely with none as default. More often than not, I've seen loss function usage inspired from the formula and they always either have sum or mean. I know standardization is a good move forward, but it would be better to have a reduction option.

The problem with reductions is that they are irreversible operations and they do not let these functions be used for any other purpose (like the categorical distribution example mentioned in the original post). Also, in reproducing existing results in literature it's important to be explicitly aware of which reduction is being used rather than relying on a default one as that will result in significantly different learning rates.

> Finally was there reason to call other l1 and l2 loss functions from other loss functions? Just trying to understand this, I thought since it adds to additional function calling, is that overhead minimal?

Reduce code duplication and improve maintainability I guess. There should be no overhead at all as the compiler should inline those.

rxwei:
I'm +1 on the direction. How do you compare this with introducing a `reducer:` parameter in each loss function with a default argument? The new approach that you implemented would be breaking all existing models, so we need to discuss these tradeoffs. Would you be interested in starting a discussion on the mailing list?

eaplatanios:
Adding a reducer/reduction parameter with a default set to whatever the reduction was before sounds good to me in order to also avoid code breakages. I don't have time to start a discussion on the mailing list now, but I could it to this PR so we maintain source compatibility until we have a lengthier discussion. How does that sound?

Shashi456:
@eaplatanios, source compatibility sounds good. Should I start the discussion meanwhile? We could add community suggested modifications in a subsequent PR.

eaplatanios:
Thanks @Shashi456! Feel free to start the discussion. If we arrive at a consensus early I can also integrate them as part of this PR. Let's see what people think. I'll add support for reduction functions later today.

Shashi456:
Started a discussion [here](https://groups.google.com/a/tensorflow.org/forum/#!topic/swift/b5ZdHjyHHBg).

If you want me to add anything specific there, I will do so.

eaplatanios:
@rxwei I added support for reduction functions, but I currently get the following crash when compiling:
```
Assertion failed: (NextValue < Values.size()), function claimNext, file /Users/eaplatanios/Development/GitHub/swift-source-temp/swift/lib/IRGen/Explosion.h, line 129.
Stack dump:
1.      Swift version 5.1-dev (LLVM 7d54a7c711, Swift c1dfcde1a4)
2.      While emitting IR SIL function "@$s10TensorFlow6l1Loss9predicted8expected9reductionAA0A0VyxGAH_A3HXFtAA0aB13FloatingPointRzlFfA1_".
 for expression at [/Users/eaplatanios/Development/GitHub/swift-apis/Sources/TensorFlow/Loss.swift:24:69 - line:24:80] RangeText="{ $0.sum() "
0  swift                    0x000000010771df95 llvm::sys::PrintStackTrace(llvm::raw_ostream&) + 37
1  swift                    0x000000010771cfd8 llvm::sys::RunSignalHandlers() + 248
2  swift                    0x000000010771e588 SignalHandler(int) + 264
4  swift                    0x0000000109b2bda8 cmark_strbuf__initbuf + 132691
7  swift                    0x00000001077f5433 swift::irgen::ScalarPairTypeInfo<(anonymous namespace)::FuncTypeInfo, swift::irgen::ReferenceTypeInfo>::initialize(swift::irgen::IRGenFunction&, swift::irgen::Explosion&, swift::irgen::Address, bool) const (.cold.3) + 35
8  swift                    0x0000000103bc525d swift::irgen::ScalarPairTypeInfo<(anonymous namespace)::FuncTypeInfo, swift::irgen::ReferenceTypeInfo>::initialize(swift::irgen::IRGenFunction&, swift::irgen::Explosion&, swift::irgen::Address, bool) const + 461
9  swift                    0x0000000103b8be33 swift::irgen::RecordTypeInfo<(anonymous namespace)::DiffFuncTypeInfo, swift::irgen::LoadableTypeInfo, (anonymous namespace)::DiffFuncFieldInfo, true, true>::initialize(swift::irgen::IRGenFunction&, swift::irgen::Explosion&, swift::irgen::Address, bool) const + 147
10 swift                    0x0000000103c6ab2e swift::SILInstructionVisitor<(anonymous namespace)::IRGenSILFunction, void>::visit(swift::SILInstruction*) + 54414
11 swift                    0x0000000103c5a493 swift::irgen::IRGenModule::emitSILFunction(swift::SILFunction*) + 8051
12 swift                    0x0000000103b79bfe swift::irgen::IRGenerator::emitGlobalTopLevel() + 734
13 swift                    0x0000000103c355c2 performIRGeneration(swift::IRGenOptions&, swift::ModuleDecl*, std::__1::unique_ptr<swift::SILModule, std::__1::default_delete<swift::SILModule> >, llvm::StringRef, swift::PrimarySpecificPaths const&, llvm::LLVMContext&, swift::SourceFile*, llvm::GlobalVariable**) + 1282
14 swift                    0x0000000103c35ab2 swift::performIRGeneration(swift::IRGenOptions&, swift::SourceFile&, std::__1::unique_ptr<swift::SILModule, std::__1::default_delete<swift::SILModule> >, llvm::StringRef, swift::PrimarySpecificPaths const&, llvm::LLVMContext&, llvm::GlobalVariable**) + 82
15 swift                    0x0000000103ae620f performCompile(swift::CompilerInstance&, swift::CompilerInvocation&, llvm::ArrayRef<char const*>, int&, swift::FrontendObserver*, swift::UnifiedStatsReporter*) + 13423
16 swift                    0x0000000103ae1dc1 swift::performFrontend(llvm::ArrayRef<char const*>, char const*, void*, swift::FrontendObserver*) + 3025
17 swift                    0x0000000103a8a1d9 main + 729
18 libdyld.dylib            0x00007fff757d13d5 start + 1
```
I could probably avoid the error by making reductions `enum` or `protocol` based but I avoided that because: (1) the current solution is more flexible, and (2) this looks like a compiler bug that should not exist. :)

eaplatanios:
@rxwei all tests pass locally with the latest toolchain. It looks like the memory leak PR may have also fixed the above crash (or the merge from master). This should be ready for review now.

Shashi456:
@eaplatanios, you might have to add documentation wrt the reduction parameter.

eaplatanios:
Good point @Shashi456! I'll add that later today.

eaplatanios:
@rxwei this has now been updated with doc strings for the reduction arguments and is ready for review.

Sources/TensorFlow/Loss.swift
```diff
@@ -17,23 +17,42 @@
 /// - Parameters:
 ///   - predicted: Predicted outputs from a neural network.
 ///   - expected: Expected values, i.e. targets, that correspond to the correct output.
+///   - reduction: Reduction to apply on the computed element-wise loss values.
```
Shashi456: @eaplatanios Just a suggestion, If there aren't too many reductions available, could we mention what they are ? or do swift reductions necessarily have to be given in the ` $0."function"` format?

eaplatanios: You can pass any closure you want as a reduction. In principle this is more of a "postprocessing" function but I believe calling it "reduction" is fine as that is the intended use and it's mainly added to maintain source compatibility with existing code.

PR #363, opened 2019-07-14
Added a couple missing 'Tensor.moments' functions.

eaplatanios:


Sources/TensorFlow/Operators/Math.swift
```diff
@@ -2317,12 +2317,24 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
     /// dimensions are removed.
```
rxwei: Add a precondition which states that `axes.rank` must be `0`.

eaplatanios: It has to be rank `1`, but I didn't add it because it would cause lazy tensor materialization. The same is true for all reduction ops and we don't use any preconditions there.

rxwei: Oh sorry I meant rank `1`.  Precondition should be added in comments at least.

eaplatanios: Good point. I added it in the documentation strings.

Issue #365, opened 2019-07-15
Make more math functions differentiable


rxwei:
These functions should be made `@differentiable`.

```swift
/// Returns the power of the scalar to the tensor, broadcasting the scalar.
@inlinable
// @differentiable
public func pow<T: TensorFlowFloatingPoint>(_ lhs: T, _ rhs: Tensor<T>) -> Tensor<T> {
    pow(Tensor(lhs), rhs)
}

/// Returns the power of the tensor to the scalar, broadcasting the scalar.
@inlinable
// @differentiable
public func pow<T: TensorFlowFloatingPoint>(_ lhs: Tensor<T>, _ rhs: T) -> Tensor<T> {
    pow(lhs, Tensor(rhs))
}

/// Returns the power of the tensor to the scalar, broadcasting the scalar.
@inlinable
// @differentiable
public func pow<T: TensorFlowFloatingPoint>(_ x: Tensor<T>, _ n: Int) -> Tensor<T> {
    pow(x, Tensor(T(n)))
}

/// Returns the element-wise `n`th root of the tensor.
@inlinable
// @differentiable
public func root<T: TensorFlowFloatingPoint>(_ x: Tensor<T>, _ n: Int) -> Tensor<T> {
    sign(x) * pow(abs(x), Tensor(T(1) / T(n)))
}
```

Shashi456:
@rxwei Can this be closed wrt #366 ?

PR #366, opened 2019-07-15
Add @differentiable support to pow(_:_:) and root(_:_:)

jon-tow:
Addresses #365.


jon-tow:
@rxwei Sure thing! I'll open a PR for it in just a few.

PR #367, opened 2019-07-15
Remove check for equal shapes in `isAlmostEqual`

dominikgrewe:


Sources/TensorFlow/Operators/Comparison.swift
```diff
@@ -229,14 +229,12 @@ public extension StringTensor {
...
-    /// equal.
+    /// Returns `true` if all pairs of scalars are approximately equal.
     @inlinable
     func isAlmostEqual(
         to other: Tensor,
         tolerance: Scalar = Scalar.ulpOfOne.squareRoot()
     ) -> Bool {
-        return self.shape == other.shape &&
-            self.elementsAlmostEqual(other, tolerance: tolerance).all()
+        return self.elementsAlmostEqual(other, tolerance: tolerance).all()
```
eaplatanios:
```suggestion
        elementsAlmostEqual(other, tolerance: tolerance).all()
```

Sources/TensorFlow/Operators/Comparison.swift
```diff
@@ -229,14 +229,12 @@ public extension StringTensor {
 }
 
 public extension Tensor where Scalar: TensorFlowFloatingPoint {
-    /// Returns `true` if tensors are of equal shape and all pairs of scalars are approximately
-    /// equal.
+    /// Returns `true` if all pairs of scalars are approximately equal.
```
eaplatanios:
```suggestion
    /// Returns `true` if all elements of this tensor are approximately equal to those of `other`.
```

Sources/TensorFlow/Operators/Comparison.swift
```diff
@@ -229,14 +229,12 @@ public extension StringTensor {
 }
 
 public extension Tensor where Scalar: TensorFlowFloatingPoint {
-    /// Returns `true` if tensors are of equal shape and all pairs of scalars are approximately
-    /// equal.
+    /// Returns `true` if all elements of this tensor are approximately equal to those of `other`.
```
rxwei: Semantically, does `isAlmostEqual(to:tolerance:)` support arguments with incompatible shapes? If only compatible shapes are accepted, saying "all elements ... are approximately equal" sounds like a scalar comparison on flattened arrays. It'd be good to say something about shapes here as well as in `elementsAlmostEqual(_:tolerance:)`.

dominikgrewe: The underlying TF op requires the shapes to be the same, see https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/cwise_ops_common.h#L172

I added that condition to the comment.

Is this something we want to catch at the Swift level? It seems generally we don't catch shape errors at the Swift level, right?

saeta: @bgogul has been working on tracking Tensor shapes in the Swift Tensor object as part of his work on Lazy Tensor. I'll let him comment further. :-)

eaplatanios: @saeta could we also discuss this during Friday's meeting?

Sources/TensorFlow/Operators/Comparison.swift
```diff
@@ -210,6 +210,7 @@ public extension Tensor where Scalar: Equatable {
 public extension Tensor where Scalar: TensorFlowFloatingPoint & Equatable {
     /// Returns a tensor of Boolean values indicating whether the elements of `self` are
     /// approximately equal to those of `other`.
+    /// `self` and `other` must be of the same shape.
```
rxwei:
```suggestion
    /// - Precondition: `self` and `other` must be of the same shape.
```
Add precondition comments as a `- Precondition:` doc item so that they get better rendered.

Sources/TensorFlow/Operators/Comparison.swift
```diff
@@ -229,14 +230,13 @@ public extension StringTensor {
 }
 
 public extension Tensor where Scalar: TensorFlowFloatingPoint {
-    /// Returns `true` if tensors are of equal shape and all pairs of scalars are approximately
-    /// equal.
+    /// Returns `true` if all elements of `self` are approximately equal to those of `other`.
+    /// `self` and `other` must be of the same shape.
```
rxwei:
```suggestion
    /// - Precondition: `self` and `other` must be of the same shape.
```

Sources/TensorFlow/Operators/Comparison.swift
bgogul: Here is a work in progress prototype of shape inference: https://github.com/tensorflow/swift-apis/pull/388/files

Note that this requires some changes from TF. So, you won't be able to try it right away. However, this gives the gist of what will be available on the Swift side. Essentially, we will be able to examine shapes on Swift side without actually executing the op (with a few exceptions, where we may need to materialize.)

Happy to fill in more during the open design meeting this Friday.

dominikgrewe:
Can this be merged?

rxwei:
Yes. Thanks for the ping!

PR #370, opened 2019-07-17
Refactor DeviceScopes out into a separate structure.

bgogul:


Sources/TensorFlow/Core/Runtime.swift
```diff
@@ -1281,6 +1264,32 @@ class _ThreadLocalState {
...
+    }
+
+    @usableFromInline
+    func popDevice() {
+        internalConsistencyCheck(deviceStack.popLast() != nil)
+    }
+}
+
 @usableFromInline
 @_cdecl("_swift_tfc_OpSetDeviceFromScope")
```
rxwei: I think this can be deleted now.

Sources/TensorFlow/Core/Runtime.swift
```diff
@@ -1281,6 +1264,32 @@ class _ThreadLocalState {
...
 
+/// Stack of devices that models nested calls to withDevice/withDefaultDevice. Devices are
+/// represented by their names in TensorFlow notation. See documentation for
+/// `withDevice(named:perform:)` to learn about device names.
+///
+/// All TensorFlow operations will be put on the topmost device on the stack. When the stack is
+/// empty or the topmost device is `nil`, that allows TensorFlow to place operations on any device
+/// that it sees fit.
+@usableFromInline
+class DeviceScopes {
```
pschuh: Why class? is there a reason for multiple threadlocalstate to share the same devicestack?

bgogul: Doesn't have to be. Changed `class` -> `struct` and made the related changes.

Sources/TensorFlow/Core/Runtime.swift
bgogul: This is still used in EagerExecution.swift. However, those use cases can be rewritten. Will send out a separate PR for that.

bgogul: It looks like you were only referring to `_cdecl`. Will send out a PR cleaning this up and a few other uses.

rxwei: Yeah it'd be nice to have EagerExecution.swift use C API directly.

PR #371, opened 2019-07-17
Cleanup: remove `_cdecl` and `_silgen` annotations.

bgogul:


rxwei:
Are these functions used anywhere? I think it's not just annotation - the functions could be removed too.

bgogul:
> Are these functions used anywhere? I think it's not just annotation - the functions could be removed too.

Actually, they are being used in a bunch of places: `Runtime.swift`, `ArrayOps.swift`, and `EagerExecution.swift`.

PR #372, opened 2019-07-17
Make liveness tracking of `LazyTensorOperations` a thread local state.

bgogul:


Sources/TensorFlow/Core/LazyTensorContext.swift
```diff
@@ -0,0 +1,73 @@
+/// A class to keep track of runtime information about `LazyTensorOperation`
+/// instances created by the program. This will be manaaged as a thread local
```
pschuh: manaaged ?

rxwei:
You could put all thread-local state into `Context` as private/internal properties, and deprecate `_ThreadLocalState`.

Sources/TensorFlow/Core/LazyTensorContext.swift
bgogul: ah. fixed now. :)

bgogul:
> You could put all thread-local stuff into `Context` as private/internal properties, and deprecate `_ThreadLocalState`.

Good idea. Let me do that in a follow up PR.

PR #374, opened 2019-07-18
Add '.swift-format'.

rxwei:
Add a `swift-format` configuration.

```json
{
    "version": 1,
    "lineLength": 100,
    "indentation": {
        "spaces": 4
    },
    "maximumBlankLines": 1,
    "respectsExistingLineBreaks": true,
    "blankLineBetweenMembers": {
        "ignoreSingleLineProperties": true
    },
    "lineBreakBeforeControlFlowKeywords": false,
    "lineBreakBeforeEachArgument": false
}
```

Note: `swift-format` built using swift:master currently has weird behavior on functions with tensorflow-branch-only attributes such as `@differentiable`. Use with caution. A potential solution is to build `swift-format` with a Swift for TensorFlow toolchain, but we currently need to update our branch to support building `swift-format`.

### DIFF
--- a/.swift-format
+++ b/.swift-format
@@ -1,0 +1,14 @@
+{
+    "version": 1,
+    "lineLength": 100,
+    "indentation": {
+        "spaces": 4
+    },
+    "maximumBlankLines": 1,
+    "respectsExistingLineBreaks": true,
+    "blankLineBetweenMembers": {
+        "ignoreSingleLineProperties": true
+    },
+    "lineBreakBeforeControlFlowKeywords": false,
+    "lineBreakBeforeEachArgument": false
+}


eaplatanios:
@rxwei I have tried to use swift-format before but it produces weird behavior as you mention. Shouldn't we wait until we can actually use it before merging this so it's not confusing/misleading.

rxwei:
> @rxwei I have tried to use swift-format before but it produces weird behavior as you mention. Shouldn't we wait until we can actually use it before merging this so it's not confusing/misleading.

Which swift-format did you use? The new swift-format is entirely different from the one in apple/swift repo.

eaplatanios:
I had tried the one from the google pitch for a standardized format. The
branch was called “format” I believe. Is that different than the new one?

rxwei:
Got it. Yes, this is the new one. We are working on resolving the bug.

PR #375, opened 2019-07-18
Use multiple `TFE_OpAddInput` calls instead of one `TFE_OpAddIputList`

bgogul:
A bug in `TFE_OpAddInputList` implementation is causing `Raw.ZipDataset` test to fail. This PR replaces a single call to `TFE_OpAddInputList` with a  sequence of calls to `TFE_OpAddInput`, which has the same effect as `TFE_OpAddInput` implementation works for `Raw.ZipDataset`.

This PR also fixes the inconsistency between `EagerExecution.swift.gyb` and `EagerExecution.swift`.

### DIFF
--- a/Sources/TensorFlow/Bindings/EagerExecution.swift
+++ b/Sources/TensorFlow/Bindings/EagerExecution.swift
@@ -70,8 +70,8 @@ internal struct TFE_Op: TFTensorOperation {
         defer { buffer.deallocate() }
         let pointer = UnsafeMutablePointer<OpaquePointer?>(buffer.baseAddress)
         input._unpackTensorHandles(into: buffer.baseAddress)
-        for i in 0..<count {
-            TFE_OpAddInput(op, buffer[Int(i)], status)
+        for i in 0..<Int(count) {
+            TFE_OpAddInput(op, buffer[i], status)
             checkOk(status)
         }
     }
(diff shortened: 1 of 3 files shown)


Sources/TensorFlow/Bindings/EagerExecution.swift
```diff
@@ -70,8 +70,10 @@ internal struct TFE_Op: TFTensorOperation {
         defer { buffer.deallocate() }
         let pointer = UnsafeMutablePointer<OpaquePointer?>(buffer.baseAddress)
         input._unpackTensorHandles(into: buffer.baseAddress)
-        TFE_OpAddInputList(op, pointer, count, status)
-        // TODO: checkOk(status)
+        for i in 0..<count {
```
rxwei:
```suggestion
        for i in 0..<Int(count) {
```

Sources/TensorFlow/Bindings/EagerExecution.swift
```diff
@@ -70,8 +70,10 @@ internal struct TFE_Op: TFTensorOperation {
...
+            TFE_OpAddInput(op, buffer[Int(i)], status)
```
rxwei:
```suggestion
            TFE_OpAddInput(op, buffer[i], status)
```

Sources/TensorFlow/Bindings/EagerExecution.swift
rxwei: This pushes integer conversion outside the loop.

Sources/TensorFlow/Bindings/EagerExecution.swift.gyb
```diff
@@ -70,8 +70,10 @@ internal struct TFE_Op: TFTensorOperation {
         defer { buffer.deallocate() }
         let pointer = UnsafeMutablePointer<OpaquePointer?>(buffer.baseAddress)
         input._unpackTensorHandles(into: buffer.baseAddress)
-        TFE_OpAddInputList(op, pointer, count, status)
-        // TODO: checkOk(status)
+        for i in 0..<count {
```
rxwei: Same.

PR #377, opened 2019-07-20
Experimental support for installing tensorflow directly into a blank swift toolchain.

pschuh:


tools/install.sh
```diff
@@ -0,0 +1,51 @@
...
+install_header "$TENSORFLOW_DIRECTORY/tensorflow/c/eager/c_api.h" c_api_eager.h
+cp tools/module.modulemap "$USR_DIRECTORY/lib/swift/linux/x86_64/modulemaps/CTensorFlow/"
+
+$USR_DIRECTORY/bin/swift build -Xswiftc -module-link-name -Xswiftc TensorFlow
+
+BIN_DIR='.build/x86_64-unknown-linux/debug'
+
+cp $BIN_DIR/libTensorFlow.so $USR_DIRECTORY/lib/swift/linux/
+cp $BIN_DIR/TensorFlow.swiftdoc $USR_DIRECTORY/lib/swift/linux/x86_64/
+cp $BIN_DIR/TensorFlow.swiftmodule $USR_DIRECTORY/lib/swift/linux/x86_64/
```
bgogul: Do we need the `.swfitinterface` files as well?

tools/install.sh
```diff
@@ -0,0 +1,51 @@
...
+  echo "Install header: " $1 $2
+  fix_tf_header $1 "$USR_DIRECTORY/lib/swift/linux/x86_64/modulemaps/CTensorFlow/$2" ""
+}
+
+mkdir -p $USR_DIRECTORY/lib/swift/linux
+copy_file $TENSORFLOW_BIN_DIRECTORY libtensorflow.so $USR_DIRECTORY/lib/swift/linux
+copy_file $TENSORFLOW_BIN_DIRECTORY libtensorflow_framework.so $USR_DIRECTORY/lib/swift/linux
+
+mkdir -p $USR_DIRECTORY/lib/swift/linux/x86_64/modulemaps/CTensorFlow
+install_header "$TENSORFLOW_DIRECTORY/tensorflow/c/c_api.h" c_api.h
```
bgogul: Is the plan to eventually have these headers as part of the tensorflow archive?

pschuh: One plan to investigate is to use the headers directly from the tensorflow installation. That would require using them as implementationOnly from inside swift-apis so that they do not need to be exported alongside swfit-apis.

tools/install.sh
pschuh: That is just for module stability. see: https://forums.swift.org/t/plan-for-module-stability/14551

PR #380, opened 2019-07-22
Dense layer tests

jkrukowski:
Some tests for #77


Shashi456:
Thanks @jkrukowski.

eaplatanios:
Thanks @jkrukowski! Given that it's a single layer I think it's fine to keep it in one test.

PR #381, opened 2019-07-22
Adds a mechanism to extract traces explicitly using LazyTensor.

bgogul:
This PR uses `LazyTensor` to perform extract traces explicitly. This mechanism will be used to replace the existing tracing mechanism in `Runtime.swift`.

### DIFF
--- a/Sources/TensorFlow/Core/LazyTensorTrace.swift
+++ b/Sources/TensorFlow/Core/LazyTensorTrace.swift
@@ -89,6 +89,48 @@ class LazyTensorTraceBuilder {
         return materializationTraceInfo([lazyOperation])
     }
 
+    /// Trace the given function and return the trace.
+    static func trace<In: TensorGroup, Out: TensorGroup>(
+        _ fn: (In) -> Out
+    ) -> LazyTensorTrace {
+        assert(_RuntimeConfig.useLazyTensor, "Lazy tensor is not enabled for tracing.")
+
+        // Set up inputs for running `fn`
+        let inputs = In._typeList.map { Self.makePlaceholder(with: $0) }
+        let inputHandles = inputs.map { LazyTensorHandle(_lazy: $0, index: 0) }
+        let input = In(_handles: inputHandles)
+
+        // Run the function.
+        let output: TensorArrayProtocol = fn(input)
+
+        // Set up the closure that determines if a `LazyTensorOperation` should be an output.
+        let outputLazyOperations = output._tensorHandles.map { (handle: _AnyTensorHandle) -> LazyTensorOperation in 
+            let lazyOp = lazyTensorOperation(handle)
+            assert(lazyOp != nil, "Found a non-lazy tensor in output when tracing.")
+            return lazyOp!
+        }
+        let outputIds = Set<ObjectIdentifier>(outputLazyOperations.map {
+                ObjectIdentifier($0)
+            })
+        let isOutput: (LazyTensorOperation) -> Bool = { outputIds.contains(ObjectIdentifier($0)) }
+
+        // Create the builder and get the trace.
+        let builder = LazyTensorTraceBuilder()
+        builder.neverPromoteConstants = true
+        builder.isOutput = isOutput
+        /// Set up the inputs for the builder as we need to have specific order.
+        for inputOp in inputs {
+            let id = ObjectIdentifier(inputOp)
+            builder.updateOperationAndCache(id, inputOp)
+        }
+        builder.inputs = inputs
+        for lazyOp in outputLazyOperations { _ = builder.collectLazyOperation(lazyOp) }
+        return LazyTensorTrace(
+            inputs: builder.inputs,
+            operations: builder.operations,
+            outputs: builder.outputs)
+    }
+
     // inputs will be "placeholder" nodes.
     private var inputs: [LazyTensorOperation] = []
     private var inputValues: [TFETensorHandle] = []
@@ -119,6 +161,17 @@ class LazyTensorTraceBuilder {
             "value": LazyTensorOperation.Attribute.constTensor(handle)]
         updateOperationAndCache(ObjectIdentifier(handle), result)
         return LazyTensorHandle(_lazy: result, index: 0)
+    }
+
+    /// Extract the LazyTensorOperation (if any) for this handle.
+    private static func lazyTensorOperation(_ handle: _AnyTensorHandle) -> LazyTensorOperation? {
+        guard let lazyTensorHandle = handle as? LazyTensorHandle else {
+            return nil
+        }
+        guard case let .symbolic(lazyOp, _, _)  = lazyTensorHandle.handle else {
+            return nil
+        }
+        return lazyOp
     }
 
     private static func makePlaceholder(with dtype: TensorDataType) -> LazyTensorOperation {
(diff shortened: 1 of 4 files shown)


Sources/TensorFlow/Core/LazyTensorTrace.swift
```diff
@@ -89,13 +89,55 @@ class LazyTensorTraceBuilder {
         return materializationTraceInfo([lazyOperation])
     }
 
+    /// Trace the given function and return the trace.
+    static func trace<In: TensorGroup, Out: TensorGroup>(_ fn: (In) -> Out) -> LazyTensorTrace {
+        precondition(_RuntimeConfig.useLazyTensor, "Lazy tensor is not enabled for tracing.")
+
+        // Set up inputs for running `fn`
```
rxwei:
```suggestion
        // Set up inputs for running `fn`.
```

Sources/TensorFlow/Core/LazyTensorTrace.swift
```diff
@@ -89,13 +89,55 @@ class LazyTensorTraceBuilder {
...
+        let output: TensorArrayProtocol = fn(input)
+
+        // Set up the closure that determines if a `LazyTensorOperation` should be an output.
+        let outputLazyOperations = output._tensorHandles.map {
+            (handle: _AnyTensorHandle) -> LazyTensorOperation in
+            let lazyOp = lazyTensorOperation(handle)
+            precondition(lazyOp != nil, "Found a non-lazy tensor in output when tracing.")
+            return lazyOp!
+        }
+        let outputIds = Set<ObjectIdentifier>(outputLazyOperations.map { ObjectIdentifier($0) })
```
rxwei: `Id` -> `ID` for all occurrences.

Sources/TensorFlow/Core/LazyTensorTrace.swift
```diff
@@ -89,13 +89,55 @@ class LazyTensorTraceBuilder {
...
+        let isOutput: (LazyTensorOperation) -> Bool = { outputIds.contains(ObjectIdentifier($0)) }
+
+        // Create the builder and get the trace.
+        let builder = LazyTensorTraceBuilder()
+        builder.neverPromoteConstants = true
+        builder.isOutput = isOutput
+        /// Set up the inputs for the builder as we need to have them in a specific order.
```
rxwei: Use 2 slashes for non-documentation comments.
```suggestion
        // Set up the inputs for the builder as we need to have them in a specific order.
```

Sources/TensorFlow/Core/LazyTensorTrace.swift
```diff
@@ -117,14 +159,28 @@ class LazyTensorTraceBuilder {
...
+        guard let lazyTensorHandle = handle as? LazyTensorHandle else {
+            return nil
+        }
+        guard case let .symbolic(lazyOp, _, _)  = lazyTensorHandle.handle else {
+            return nil
+        }
+        return lazyOp
+    }
+
+    private static func makePlaceholder(with dtype: TensorDataType) -> LazyTensorOperation {
```
rxwei: * Factory methods should not form a phrases with the first argument label.
* `with` as an argument label is often regarded as not being meaningful in Swift API design, so it should be avoided.

```suggestion
    private static func makePlaceholder(dataType: TensorDataType) -> LazyTensorOperation {
```

Sources/TensorFlow/Core/LazyTensorTrace.swift
```diff
@@ -117,14 +159,28 @@ class LazyTensorTraceBuilder {
...
         let placeholder = LazyTensorOperation("Placeholder", 1)
+        let dtypeAttr = LazyTensorOperation.Attribute.tensorDataTypeValue(dtype)
         placeholder.attributes = ["dtype": dtypeAttr]
+        return placeholder
+    }
+
+    private func makePlaceholderTensor(with handle: TFETensorHandle) -> LazyTensorHandle {
```
rxwei:
```suggestion
    private func makePlaceholderTensor(handle: TFETensorHandle) -> LazyTensorHandle {
```

bgogul:
@rxwei Thanks for the review. I addressed your comments. PTAL.

Sources/TensorFlow/Core/LazyTensorTrace.swift
```diff
@@ -117,14 +158,28 @@ class LazyTensorTraceBuilder {
         return LazyTensorHandle(_lazy: result, index: 0)
     }
 
-    private func makePlaceholderTensor(
-        with handle: TFETensorHandle
-    ) -> LazyTensorHandle {
-        let cTensorHandle = handle._cTensorHandle
-        let dtype = TensorDataType(TFE_TensorHandleDataType(cTensorHandle))
-        let dtypeAttr = LazyTensorOperation.Attribute.tensorDataTypeValue(dtype)
+    /// Extract the LazyTensorOperation (if any) for this handle.
```
rxwei:
```suggestion
    /// Returns the `LazyTensorOperation`, if any, for this handle.
```
#160

Sources/TensorFlow/Core/LazyTensorTrace.swift
```diff
@@ -138,9 +193,9 @@ class LazyTensorTraceBuilder {
...
             return LazyTensorHandle(_lazy: lazyOp, index: 0)
         }
-        return asConst
+        return asConst || neverPromoteConstants
             ? makeConstTensor(with: handle)
-            : makePlaceholderTensor(with: handle)
+            : makePlaceholderTensor(handle: handle)
     }
 
     /// Return the original tensor or a concrete tensor that is promoted to a
```
rxwei:
```suggestion
    /// Returns the original tensor or a concrete tensor that is promoted to a
```

Sources/TensorFlow/Core/LazyTensorTrace.swift
```diff
@@ -89,13 +89,54 @@ class LazyTensorTraceBuilder {
...
+        let output: TensorArrayProtocol = fn(input)
+
+        // Set up the closure that determines if a `LazyTensorOperation` should be an output.
+        let outputLazyOperations = output._tensorHandles.map {
+            (handle: _AnyTensorHandle) -> LazyTensorOperation in
+            let lazyOp = lazyTensorOperation(handle)
+            precondition(lazyOp != nil, "Found a non-lazy tensor in output when tracing.")
+            return lazyOp!
+        }
+        let outputIDs = Set<ObjectIdentifier>(outputLazyOperations.map { ObjectIdentifier($0) })
```
rxwei: Use `.lazy` to avoid creating a temporary array.
```suggestion
        let outputIDs = Set<ObjectIdentifier>(outputLazyOperations.lazy.map { ObjectIdentifier($0) })
```

Sources/TensorFlow/Core/LazyTensorTrace.swift
```diff
@@ -117,14 +159,28 @@ class LazyTensorTraceBuilder {
...
-    private func makePlaceholderTensor(
-        with handle: TFETensorHandle
-    ) -> LazyTensorHandle {
-        let cTensorHandle = handle._cTensorHandle
-        let dtype = TensorDataType(TFE_TensorHandleDataType(cTensorHandle))
-        let dtypeAttr = LazyTensorOperation.Attribute.tensorDataTypeValue(dtype)
+    /// Returns the `LazyTensorOperation`, if any, for this handle.
+    private static func lazyTensorOperation(_ handle: _AnyTensorHandle) -> LazyTensorOperation? {
+        guard let lazyTensorHandle = handle as? LazyTensorHandle else {
+            return nil
```
rxwei: These two `guard` statements can be folded into one.

```swift
guard let lazyTensorHandle = handle as? LazyTensorHandle,
      case let .symbolic(lazyOp, _, _)  = lazyTensorHandle.handle else {
    return nil
}
```

rxwei: Or even better:

```swift
guard case let .symbolic(lazyOp, _, _)?  = (handle as? LazyTensorHandle)?.handle else {
    return nil
}
```

PR #382, opened 2019-07-22
Cleanup: replace stack of `operationsTracker` in `LazyTensorContext` into a single tracker.

bgogul:
We don't need a stack of `operationsTracker` at this point.

### DIFF
--- a/Sources/TensorFlow/Core/LazyTensorContext.swift
+++ b/Sources/TensorFlow/Core/LazyTensorContext.swift
@@ -61,13 +61,13 @@ class LazyTensorOperationsTracker {
 }
 
 struct LazyTensorContext {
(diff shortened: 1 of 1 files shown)


Sources/TensorFlow/Core/LazyTensorContext.swift
```diff
@@ -61,13 +61,13 @@ class LazyTensorOperationsTracker {
 }
 
 struct LazyTensorContext {
-    var scopes = [LazyTensorOperationsTracker()]
+    private var _operationsTracker = LazyTensorOperationsTracker()
```
rxwei: If the access level is less than public, it doesn't need to be underscored.

Issue #364, opened 2019-07-14
Bad Memory Leak


eaplatanios:
@rxwei @dan-zheng I get insane memory usage when I try to run some simple reinforcement learning experiments. It looks like there is a memory bug somewhere related to partial applies in auto-diff (purely a guess based on what I saw in the profiler report). Attached is a screenshot which I hope you find helpful in figuring this out:

<img width="1321" alt="Screen Shot 2019-07-14 at 7 25 12 PM" src="https://user-images.githubusercontent.com/1294940/61190643-66ed8900-a66d-11e9-934b-e5babcf6e04c.png">

Note that this is pretty important as it makes it impossible for me to run any experiments currently.

rxwei:
I’m on it.

eaplatanios:
Thanks Richard! I'm also currently compiling the latest apple/swift to see if anything has changed.

eaplatanios:
@rxwei I can confirm the leak is still there even when I use a toolchain built from the latest commit on apple/swift.

eaplatanios:
@rxwei let me know if I can help with this in any way. Although I’m not familiar with memory management during the autodiff pass.

rxwei:
I'm working on a fix [here](https://github.com/rxwei/swift/tree/ownership).

eaplatanios:
Do you have an ETA on when this might be fixed? I may take a look tomorrow but I don't know if it'll be productive given that I'm not familiar with this aspect of the compiler.

rxwei:
I'll try my best to get it land before EOD tomorrow.

eaplatanios:
That's great! Thanks a lot Richard! In that case I may take a shot at one of the other auto-diff bugs in the meantime. I assume we're not ready yet for the `inout` support changes we discussed previously, right?

rxwei:
This week we are aiming to land two things that would allow you to start the `inout` work.

eaplatanios:
That's great news! Thanks Richard!

eaplatanios:
@rxwei so I did some tests after the leak fix PR landed and I noticed that a portion of the leak still exists for me even if I remove all control flow. It's significantly reduced, but still unusable for anything more than toy experiments. I'll try to get an instruments trace uploaded and sent to you if you think that helps.

rxwei:
A trace would be very useful. Note that custom initializer VJPs have leaks ([TF-667](https://bugs.swift.org/browse/TF-667)) and there is a workaround.

[...]

PR #384, opened 2019-07-24
use .moments() in LayerNorm and BatchNorm layers

mikowals:
Mean and variance in the layers are now calculated using `Tensor.moments()`.  I also added tests for both BatchNorm and LayerNorm layers.  

The tests turned up a flaw in how the shape of the `scale` and `offset` which were always of shape `[featureCount]` irrespective of the input shape or axis for normalisation.  That shape leads to incorrect broadcasting when the axis being normalized along is not the last axis.

I have fixed this by always reshaping `scale` and `offset` before they are used.  This seems hacky in that I get the shapes from the calculated `mean` and `variance`.  Without the input shape being known at initialization time though I couldn't see a better way to do this.  

I think the axis argument is probably there to be consistent with Keras but most of the Swift api layers assume inputs and activations are NHWC.  So requiring NHWC, eliminating the axis argument, and the setting the correct shapes in `init()` would be another option.

### DIFF
--- a/Sources/TensorFlow/Layers/Normalization.swift
+++ b/Sources/TensorFlow/Layers/Normalization.swift
@@ -89,7 +89,7 @@ public struct BatchNorm<Scalar: TensorFlowFloatingPoint>: Layer {
     ///
     /// - Parameter input: The input to the layer.
     /// - Returns: The output.
-    @differentiable
+    @differentiable(vjp: _vjpApplied)
     public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
         switch Context.local.learningPhase {
         case .training: return applyingTraining(to: input)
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -70,48 +70,33 @@ public struct BatchNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
+        let offsetShape = runningMean.value.shape
+        let inv = rsqrt(runningVariance.value + epsilon) * scale.reshaped(to: scaleShape)
+        return (input - runningMean.value) * inv + offset.reshaped(to: offsetShape)
     }
 
     /// Returns the output obtained from applying the layer to the given input.
     ///
     /// - Parameter input: The input to the layer.
     /// - Returns: The output.
-    @differentiable(vjp: _vjpApplied(to:))
```
rxwei: Control flow differentiation is known to have memory leaks (#364).  Though it's going to be fixed some time this week, I'd suggest making this change in a later PR instead.

mikowals: I added back _vjpApplied() as suggested.

rxwei: Could you please revert the change so that the original indentation won't get modified?

mikowals: I have reverted to the original for these lines now.

BradLarson:
@mikowals - The inference changes introduced here were triggering shape mismatches within the MiniGo sample project and its tests on tensorflow/swift-models. I've submitted a PR to roll those back here: https://github.com/tensorflow/swift-apis/pull/426 , but if you have a better way to approach this, I'm all ears.

PR #385, opened 2019-07-24
Rewrite tracing utilities in terms of LazyTensor. 

bgogul:
This PR reimplements `_tffunc` and `_graph` in `Runtime.swift` using the explicit tracing mechanism in LazyTensor.

### DIFF
--- a/Sources/TensorFlow/Core/LazyTensorTFFunctionBuilder.swift
+++ b/Sources/TensorFlow/Core/LazyTensorTFFunctionBuilder.swift
@@ -269,7 +269,7 @@ class TFFunction {
         checkOk(status)
     }
 
-    func execute(_ inputs: [TFETensorHandle], useXLA: Bool = false) -> [TFETensorHandle] {
+    func execute(_ inputs: [TFETensorHandle], usingXLA: Bool = false) -> [TFETensorHandle] {
         let status: CTFStatus = TF_NewStatus()
         defer { TF_DeleteStatus(status) }
 
@@ -286,7 +286,7 @@ class TFFunction {
             checkOk(status)
         }
 
-        if useXLA {
+        if usingXLA {
             debugLog("Enabling XLA compilation")
             TFE_OpSetAttrBool(eagerOp, "_XlaCompile", 1)
         }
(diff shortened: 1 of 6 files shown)


Sources/TensorFlow/Core/LazyTensorTFFunctionBuilder.swift
```diff
@@ -268,7 +269,7 @@ class TFFunction {
         checkOk(status)
     }
 
-    func execute(_ inputs: [TFETensorHandle]) -> [TFETensorHandle] {
+    func execute(_ inputs: [TFETensorHandle], useXLA: Bool = false) -> [TFETensorHandle] {
```
rxwei:
```suggestion
    func execute(_ inputs: [TFETensorHandle], usingXLA: Bool = false) -> [TFETensorHandle] {
```

Sources/TensorFlow/Core/Runtime.swift
```diff
@@ -934,65 +934,33 @@ public func _tffunc<State: _TensorArrayProtocolEnhanced, Data: TensorGroup>(
     }
 }
 
+internal func _trace<In: TensorGroup, Out: TensorGroup>(_ fn: (In) -> Out) -> TFFunction {
+    let useLazyTensor = _RuntimeConfig.useLazyTensor
```
rxwei: This seems not thread-safe.

bgogul: Ah.. you are right. Hold the review on this.

bgogul: This should be resolved now.

Sources/TensorFlow/Core/Runtime.swift
```diff
@@ -934,65 +934,33 @@ public func _tffunc<State: _TensorArrayProtocolEnhanced, Data: TensorGroup>(
...
-        let inputTensors = inputStateTensorHandles.map {
-            TFETensorHandle(_owning: $0)
-        }
-        debugLog("Executing trace graph function.")
-        let returnValues = traceContext.execute(traceeInputs: inputTensors, useXLA: useXLA)
-
-        debugLog("Creating output model instance.")
-        return Out(_owning: returnValues)
+    let tffunc = _trace(fn)
+    return {(input: In) -> Out in
```
rxwei:
```suggestion
    return { input in
```

PR #387, opened 2019-07-24
Make useLazyTensor a thread-local state.

bgogul:


Sources/TensorFlow/Core/Runtime.swift
```diff
@@ -1201,6 +1201,17 @@ class _ThreadLocalState {
 
     var lazyTensorContext = LazyTensorContext()
 
+    static var useLazyTensor: Bool {
+        get {
+            _ThreadLocalState.local.lazyTensorEnabled ?? _RuntimeConfig.useLazyTensor
+        }
+        set(newValue) {
```
rxwei:
```suggestion
        set {
```

Sources/TensorFlow/Core/Runtime.swift
```diff
@@ -1201,6 +1201,17 @@ class _ThreadLocalState {
...
+            _ThreadLocalState.local.lazyTensorEnabled = newValue
+        }
+    }
+
+    private var lazyTensorEnabled: Bool? = nil
```
rxwei: Why is this optional?

bgogul: This may not always be set. If this is not set, we should use the global `_RuntimeConfig.useLazyTensor` property to determine if lazy tensor is enabled.  I added a comment to explain this.

rxwei: Hmm, isn't using the global `_RuntimeConfig.useLazyTensor` still not thread-safe? Would it make sense to make this flag the only one?

Sources/TensorFlow/Core/Runtime.swift
```diff
@@ -1201,6 +1201,20 @@ class _ThreadLocalState {
...
+        get {
+            _ThreadLocalState.local.lazyTensorEnabled ?? _RuntimeConfig.useLazyTensor
+        }
+        set {
+            _ThreadLocalState.local.lazyTensorEnabled = newValue
+        }
+    }
+
+    /// When true, use lazy evaluation. If this is not set, we should use the
+    /// value of _RuntimeConfig.useLazyTensor to determine if lazy evaluation
```
rxwei:
```suggestion
    /// value of `_RuntimeConfig.useLazyTensor` to determine if lazy evaluation
```

bgogul:
> Hmm, isn't using the global _RuntimeConfig.useLazyTensor still not thread-safe? Would it make sense to make this flag the only one?

`_RuntimeConfig.useLazyTensor` should only be written to once at the start of the program and read afterwards.

bgogul:
> > Hmm, isn't using the global _RuntimeConfig.useLazyTensor still not thread-safe? Would it make sense to make this flag the only one?
> 
> `_RuntimeConfig.useLazyTensor` should only be written to once at the start of the program and read afterwards.

Also, `_RuntimeConfig.useLazyTensor` is a temporary flag until we make lazy tensor the default behavior in code.

PR #388, opened 2019-07-25
Shape inference of simple operations in LazyTensor.

bgogul:
This PR incorporates shape tracking in lazy tensor by using the shape inference functions associated with the TF operations.

### DIFF
--- a/Sources/TensorFlow/Core/LazyTensorContext.swift
+++ b/Sources/TensorFlow/Core/LazyTensorContext.swift
@@ -62,22 +62,21 @@ class LazyTensorOperationsTracker {
 
 struct LazyTensorContext {
     private var operationsTracker = LazyTensorOperationsTracker()
-    private var _shapeTrackingEnabled = true
+    private var isShapeTrackingEnabled = true
 
-    static private var threadLocalContext: LazyTensorContext {
+    static private var local: LazyTensorContext {
         _read { yield _ThreadLocalState.local.lazyTensorContext }
         _modify { yield &_ThreadLocalState.local.lazyTensorContext }
     }
 
     static var operationsTracker: LazyTensorOperationsTracker {
-        return threadLocalContext.operationsTracker
+        return local.operationsTracker
     }
 
-    /// A flag that determines whether we should track shapes.
-    /// We will need to disable shape tracking within certain contexts.
-    /// e.g., we won't be able to compute shapes when tracing.
-    static var shapeTrackingEnabled: Bool {
-        get { threadLocalContext._shapeTrackingEnabled }
-        set { threadLocalContext._shapeTrackingEnabled = newValue }
+    /// A flag that determines whether we should track shapes. We will need to disable shape
+    /// tracking within certain contexts. e.g., we won't be able to compute shapes when tracing.
+    static var isShapeTrackingEnabled: Bool {
+        get { local.isShapeTrackingEnabled }
+        set { local.isShapeTrackingEnabled = newValue }
     }
 }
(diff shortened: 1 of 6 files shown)


Sources/TensorFlow/Core/Tensor.swift
```diff
@@ -51,10 +51,15 @@ extension Tensor: AnyTensor {
 
 public extension Tensor {
     /// The number of dimensions of the `Tensor`.
-    @inlinable
+    // @inlinable
```
eaplatanios: Why remove the `@inlinable`?

bgogul: Just a temporary hack as `LazyTensorHandle` is not public. We need to make it `@usableFromInline`.

bgogul:
Folks, this PR is ready for review.

Sources/TensorFlow/Core/LazyTensorOperation.swift
```diff
@@ -67,12 +67,29 @@ class LazyTensorHandle: _AnyTensorHandle {
     }
 
     /// The number of dimensions of the underlying `Tensor`.
-    @inlinable
-    var rank: Int { _tfeTensorHandle.rank }
+    @usableFromInline
+    var rank: Int {
+        @_semantics("autodiff.nonvarying")
+        get { return shape.rank }
```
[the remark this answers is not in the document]
bgogul: Unfortunately, that is not possible with the TF API. In any case, in both the eager and lazy paths, we will have the shapes available anyway?

bgogul:
Gentle ping. Can someone take a look at this PR, please?

Sources/TensorFlow/Core/LazyTensorContext.swift
```diff
@@ -62,12 +62,22 @@ class LazyTensorOperationsTracker {
 
 struct LazyTensorContext {
     private var operationsTracker = LazyTensorOperationsTracker()
+    private var _shapeTrackingEnabled = true
```
rxwei:
```suggestion
    private var isShapeTrackingEnabled = true
```

No need for underscore in a less-than-public type.

Sources/TensorFlow/Core/LazyTensorContext.swift
```diff
@@ -62,12 +62,22 @@ class LazyTensorOperationsTracker {
...
 
     static private var threadLocalContext: LazyTensorContext {
```
rxwei: Consider renaming to just `local`. It's a convention used in a lot of other Swift APIs.

Sources/TensorFlow/Core/LazyTensorContext.swift
```diff
@@ -62,12 +62,22 @@ class LazyTensorOperationsTracker {
...
-        _ThreadLocalState.local.lazyTensorContext
+        _read { yield _ThreadLocalState.local.lazyTensorContext }
+        _modify { yield &_ThreadLocalState.local.lazyTensorContext }
     }
 
     static var operationsTracker: LazyTensorOperationsTracker {
         return threadLocalContext.operationsTracker
     }
+
+    /// A flag that determines whether we should track shapes.
```
rxwei: Fit to 100 columns? This seems to be using an 80-column limit.

Sources/TensorFlow/Core/LazyTensorContext.swift
```diff
@@ -62,12 +62,22 @@ class LazyTensorOperationsTracker {
...
+    /// We will need to disable shape tracking within certain contexts.
+    /// e.g., we won't be able to compute shapes when tracing.
+    static var shapeTrackingEnabled: Bool {
```
rxwei:
```suggestion
    static var isShapeTrackingEnabled: Bool {
```

PR #393, opened 2019-07-26
[Runtime] Merge '_ThreadLocalState' into 'Context'.

rxwei:
`Context` is effectively a user-oriented thread-local state, same as the internal-facing `_ThreadLocalState`. We don't need two thread-local states. This PR moves properties of `_ThreadLocalState` into `Context` as `internal` and `private` properties and removes `_ThreadLocalState`.

### DIFF
--- a/Sources/TensorFlow/Context.swift
+++ b/Sources/TensorFlow/Context.swift
@@ -44,28 +44,6 @@ public enum LearningPhase {
 ///   Context.local.learningPhase = .inference
 ///   ```
 public struct Context {
-    /// The learning phase.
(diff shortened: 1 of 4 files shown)


pschuh:
Please just merge _ThreadLocalState and ContextManager.

bgogul:
> Please just merge _ThreadLocalState and ContextManager.

+1 on this. `_ThreadLocalState` is more similar with `ContextManager`. Also, ThreadLocalState might be a better name for the merged struct.

rxwei:
Yep, I realized that I overlooked. Will update this.

PR #397, opened 2019-07-28
Fix in vjpConv2DBackpropFilter

Shashi456:
The differentiation is wrt `(x, input)`:
```
/// TensorFlow builtin conv2d gradient helper for the filter.
@differentiable(wrt: (x, input), vjp: _vjpConv2DBackpropFilter)
@usableFromInline
```
While the order of the arguments returned in the vjp is input, x.
Another fix is in the input function the shape needed is the input shape and the one being used is filter shape.


Shashi456:
@rxwei, Actually this is the case for depthwise and conv3d as well, Should i reflect the change there as well?

dan-zheng:
> @rxwei, Actually this is the case for depthwise and conv3d as well, Should i reflect the change there as well?

Yes, please change those as well.

Please also add differentiation tests for the updated operators in `Tests/TensorFlowTests/TensorAutoDiffTests.swift`. 🙂 

You can check correctness against Python TF as is done [here](https://github.com/tensorflow/swift-apis/blob/728a2d6c2146d12d8a16428f6c9a3bce24492bb7/Tests/TensorFlowTests/TensorAutoDiffTests.swift#L404).

Shashi456:
can this be tested and merged? Conv2d Differentiation test was added in #443, I'll do the same for conv3d and add a differentiation test today in another PR.

rxwei:
Please do a `git pull` so that the test that was added will run as part of CI.

Shashi456:
This can be merged.

PR #399, opened 2019-07-29
Test swift-models.

pschuh:
Basic testing of building swift-models in addition to swift-apis

### DIFF
--- a/Sources/TensorFlow/Loss.swift
+++ b/Sources/TensorFlow/Loss.swift
@@ -52,7 +52,7 @@ public func meanAbsoluteError<Scalar: TensorFlowFloatingPoint>(
     predicted: Tensor<Scalar>,
     expected: Tensor<Scalar>
 ) -> Tensor<Scalar> {
-    l1Loss(predicted: predicted, expected: expected, reduction: { $0.mean() })
+    l1Loss(predicted: predicted, expected: expected, reduction: _mean)
 }
 
 /// Returns the mean squared error between predictions and expectations.
@@ -65,7 +65,7 @@ public func meanSquaredError<Scalar: TensorFlowFloatingPoint>(
     predicted: Tensor<Scalar>,
     expected: Tensor<Scalar>
 ) -> Tensor<Scalar> {
-    l2Loss(predicted: predicted, expected: expected, reduction: { $0.mean() })
+    l2Loss(predicted: predicted, expected: expected, reduction: _mean)
 }
 
 /// Returns the mean squared logarithmic error between predictions and expectations.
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/Loss.swift
```diff
@@ -194,6 +194,14 @@ public func kullbackLeiblerDivergence<Scalar: TensorFlowFloatingPoint>(
     reduction(expected * log(expected / predicted))
 }
 
+/// Workaround for cross-module default parameter @differentiable functions.
+@differentiable
+public func _justTakesTheMean<Scalar: TensorFlowFloatingPoint>(
```
dan-zheng: Could you please rename this to just `mean`? If that's ambiguous, then `_mean`?
```suggestion
public func mean<Scalar: TensorFlowFloatingPoint>(
```

Sources/TensorFlow/Loss.swift
```diff
@@ -52,7 +52,7 @@ public func meanAbsoluteError<Scalar: TensorFlowFloatingPoint>(
     predicted: Tensor<Scalar>,
     expected: Tensor<Scalar>
 ) -> Tensor<Scalar> {
-    l1Loss(predicted: predicted, expected: expected, reduction: { $0.mean() })
```
eaplatanios: I believe there is also a couple of loss functions that use `sum()`. Shouldn't we modify those too?

dan-zheng: The PR tests integrating TensorFlow directly into a toolchain:
* Delete `TensorFlow` module and artifacts from an existing toolchain.
* Use toolchain to build `swift-apis`.
* Copy built `TensorFlow` artifacts back into the toolchain.

`_mean` is a workaround for linker errors encountered during CI for `swift-models`:
```
<compiler-generated>:0: error: undefined reference to 'AD__$s10TensorFlow19softmaxCrossEntropy6logits6labels9reductionAA0A0VyxGAH_AGys5Int32VGA2HXFtAA0aB13FloatingPointRzlFfA1_A2HcfU___jvp_src_0_wrt_0'
<compiler-generated>:0: error: undefined reference to 'AD__$s10TensorFlow19softmaxCrossEntropy6logits6labels9reductionAA0A0VyxGAH_AGys5Int32VGA2HXFtAA0aB13FloatingPointRzlFfA1_A2HcfU___vjp_src_0_wrt_0'
<compiler-generated>:0: error: undefined reference to 'AD__$s10TensorFlow19softmaxCrossEntropy6logits6labels9reductionAA0A0VyxGAH_AGys5Int32VGA2HXFtAA0aB13FloatingPointRzlFfA1_A2HcfU___pullback_src_0_wrt_0'
```

Adding `_mean` and renaming usages was sufficient to fix these errors, so `_sum` is not necessary at the moment. We're debugging these linker errors so we can remove `_mean` ASAP.

eaplatanios: Great, thanks Dan!

rxwei: Is there a bug filed for the linker error?

dan-zheng: Filed [TF-690](https://bugs.swift.org/browse/TF-690].

PR #401, opened 2019-07-30
Move the computation of shape and rank to `TFETensorHandle`.

bgogul:
This is a necessary step to incorporate shape inference into `LazyTensorHandle`.

### DIFF
--- a/Sources/TensorFlow/Core/LazyTensorOperation.swift
+++ b/Sources/TensorFlow/Core/LazyTensorOperation.swift
@@ -65,6 +65,14 @@ class LazyTensorHandle: _AnyTensorHandle {
             LazyTensorContext.operationsTracker.decrementRefCount(op, isLive: isLive)
         }
     }
+
+    /// The number of dimensions of the underlying `Tensor`.
+    @inlinable
+    var rank: Int { _tfeTensorHandle.rank }
+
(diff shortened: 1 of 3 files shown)


bgogul:
Thanks for the review, @dan-zheng

eaplatanios:
@bgogul That's great! Are we currently using a version of TF that supports the C API for shapes or has that not been merged yet?

bgogul:
> @bgogul That's great! Are we currently using a version of TF that supports the C API for shapes or has that not been merged yet?

We have the shape inference API now. I just merged it in this morning. Unfortunately, swift-apis does not build with the swift compiler at head. :( @dan-zheng is looking into it.

eaplatanios:
That's great! Thanks for the update @bgogul! :) It would be nice to get some benchmarking suite up at some point to see how large an impact changes like this one combined with lazy tensor make.

bgogul:
> That's great! Thanks for the update @bgogul! :) It would be nice to get some benchmarking suite up at some point to see how large an impact changes like this one combined with lazy tensor make.

Certainly! We have been having performance benchmarking on our radar for a while (https://bugs.swift.org/browse/TF-663). If you have any suggestions, please do share on the ticket.

PR #403, opened 2019-07-31
Disable failing BasicOperatortests.testVJPConcatentation.

bgogul:
Related to https://bugs.swift.org/browse/TF-698

### DIFF
--- a/Tests/TensorFlowTests/OperatorTests/BasicTests.swift
+++ b/Tests/TensorFlowTests/OperatorTests/BasicTests.swift
@@ -366,6 +366,8 @@ final class BasicOperatorTests: XCTestCase {
             ShapedArray(shape: [2, 6], scalars: [0, 1, 2, 6, 7, 8, 3, 4, 5, 9, 10, 11]))
     }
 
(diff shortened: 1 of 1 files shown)


eaplatanios:
@bgogul @dan-zheng I don't think it's a good idea to remove this test. It's quite simple and should be working. If the new toolchain results in incompatible shapes for some reason, we should probably find out why that happens. This kind of operation is common across many models and so existing code may break if this test fails.

dan-zheng:
> @bgogul @dan-zheng I don't think it's a good idea to remove this test. It's quite simple and should be working. If the new toolchain results in incompatible shapes for some reason, we should probably find out why that happens. This kind of operation is common across many models and so existing code may break if this test fails.

I'm investigating the failure now. I haven't narrowed down the cause yet; once I do, I'll revert the `apple/swift` PR that introduced the breakage if an immediate fix isn't obvious.

eaplatanios:
That's great. Thanks Dan!

dan-zheng:
Fixed derivatives of `Tensor.concatenated(with:)` and `Tensor.++` in https://github.com/apple/swift/pull/26448.

PR #405, opened 2019-07-31
Add an API to get the underlying `LazyTensorOperation` of a `LazyTensorHandle`.

bgogul:


Tests/TensorFlowTests/LazyTensorHandleTests.swift
```diff
@@ -52,6 +52,21 @@ final class LazyTensorHandleTests: XCTestCase {
         XCTAssertEqual(liveSymTensor.description, "%0.0*")
     }
 
+    func testLazyTensorOperationProperty() {
+        let zero = Tensor<Float>(0.0)
+        let zeroTFEHandle = zero.handle.handle._tfeTensorHandle
+        let concTensor = LazyTensorHandle(zeroTFEHandle)
+        XCTAssertTrue(concTensor.lazyTensorOperation == nil)
```
dan-zheng:
```suggestion
        XCTAssertNil(concTensor.lazyTensorOperation)
```

Tests/TensorFlowTests/LazyTensorHandleTests.swift
```diff
@@ -52,6 +52,21 @@ final class LazyTensorHandleTests: XCTestCase {
...
+
+        let op = LazyTensorOperation(
+            _id: "0", name: "IdentityN", outputCount: 3)
+        let symTensor = LazyTensorHandle(_lazy: op, index: 0)
+        let lazyTensorOperation = symTensor.lazyTensorOperation
+        XCTAssertFalse(lazyTensorOperation == nil)
```
dan-zheng:
```suggestion
        XCTAssertNotNil(lazyTensorOperation)
```

Tests/TensorFlowTests/LazyTensorHandleTests.swift
```diff
@@ -52,6 +52,21 @@ final class LazyTensorHandleTests: XCTestCase {
...
+        // Checks that returned value is the same as the one that we passed in.
+        XCTAssertTrue(lazyTensorOperation! === op)
```
rxwei: [`===(_:_:)`](https://developer.apple.com/documentation/swift/1538988) is defined as a function that takes optionals. Force unwrapping should be avoided.
```suggestion
        XCTAssertTrue(lazyTensorOperation === op)
```

bgogul: ah I see. Thanks!

Tests/TensorFlowTests/LazyTensorHandleTests.swift
bgogul: Thanks for the suggestion! Fixed it!

PR #406, opened 2019-07-31
Materialize inputs en masse when executing a 0-output lazy operation.

bgogul:
This PR does not introduce any new functionality. 
Related to https://bugs.swift.org/browse/TF-604

### DIFF
--- a/Sources/TensorFlow/Core/LazyTensorOperation.swift
+++ b/Sources/TensorFlow/Core/LazyTensorOperation.swift
@@ -384,11 +384,9 @@ extension LazyTensorOperation: TFTensorOperation {
                     unmaterializedInputs.append(lazyOperation)
                 }
             case .list(let values):
-                for value in values {
-                    if let lazyOperation = value.lazyTensorOperation {
-                        unmaterializedInputs.append(lazyOperation)
-                    }
-                }
+                unmaterializedInputs.append(
+                    contentsOf: values.lazy.compactMap { $0.lazyTensorOperation }
+                )
             }
         }
         // Materialize the inputs now.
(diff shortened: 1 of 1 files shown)


Sources/TensorFlow/Core/LazyTensorOperation.swift
```diff
@@ -374,9 +374,28 @@ extension LazyTensorOperation: TFTensorOperation {
...
+        var unmaterializedInputs = Array<LazyTensorOperation>()
+        unmaterializedInputs.reserveCapacity(inputs.count)
+        for input in inputs {
+            switch input {
+            case .single(let v):
+                if let lazyOperation = v.lazyTensorOperation {
+                    unmaterializedInputs.append(lazyOperation)
+                }
+            case .list(let values):
+                for value in values {
```
dan-zheng: Consider using [`Array.compactMap(_:)`](https://developer.apple.com/documentation/swift/array/2957701-compactmap):
```swift
                unmaterializedInputs.append(
                    contentsOf: values.compactMap { $0.lazyTensorOperation }
                )
```

bgogul: Thanks for the suggestion, @dan-zheng However, isn't the explicit loop slightly more efficient as we don't need to create the temporary array?

dan-zheng: > Thanks for the suggestion, @dan-zheng. However, isn't the explicit loop slightly more efficient as we don't need to create the temporary array?

Yes, oops. You can use a [`LazySequence`](https://developer.apple.com/documentation/swift/lazysequence), which will not materialize a temporary array:
```diff
                 unmaterializedInputs.append(
-                    contentsOf: values.compactMap { $0.lazyTensorOperation }
+                    contentsOf: values.lazy.compactMap { $0.lazyTensorOperation }
                 )
```

bgogul: Gotcha, thanks!

Issue #408, opened 2019-08-01
Port tensor control flow AD tests


dan-zheng:
[`test/TensorFlowRuntime/tensor_autodiff_control_flow.swift`](https://github.com/apple/swift/blob/tensorflow/test/TensorFlowRuntime/tensor_autodiff_control_flow.swift) from `apple/swift` needs to be ported to [`Tests/TensorFlowTests/TensorAutoDiffTests.swift`](https://github.com/tensorflow/swift-apis/blob/master/Tests/TensorFlowTests/TensorAutoDiffTests.swift).

This enables removing the deprecated `TensorFlowUnittest` module from `apple/swift` - see https://github.com/apple/swift/pull/26447.

rxwei:
I don’t think this needs to block #26447. These tensor tests are merely using additive arithmetics. Please just make sure the same arithmetics are tested using ‘Tracked’ in the AD suite, and then delete the test file in order to merge #26447.

Shashi456:
@dan-zheng I think this issue can be closed.

dan-zheng:
> @dan-zheng I think this issue can be closed.

Thanks! Closing.

PR #413, opened 2019-08-02
[Tests] Port tensor control flow AD tests

jon-tow:
#408


Tests/TensorFlowTests/TensorAutoDiffTests.swift
```diff
@@ -41,6 +41,116 @@ final class TensorAutoDiffTests: XCTestCase {
...
+                y.0 = w.1
+                y.1 = w.0
+                z.0.0 = z.0.0 - y.0
+                z.0.1 = z.0.1 + y.0
+            } else {
+                z = ((y.0 - x, y.1 + x), x)
+            }
+            return y.0 + y.1 - z.0.0 + z.0.1
+        }
+        XCTAssertTrue(([8], [2]) == valueWithGradient(at: Tensor(4), in: condNestedTupleVar))
```
rxwei: Use XCTAssertEqual instead?

rxwei:
Was it tested locally?

jon-tow:
@rxwei Yup. It surely worked on the previous nightly toolchain:
`(Swift version 5.1-dev (LLVM 200186e28b, Swift b2720a3460)`
I now get the error after updating to the latest nightly. 

I'm not sure why `testConcatenationPlusPlus()` is failing because the return expression in the `gradient(at:)` call works fine when checked in the REPL with the same tensors. I'll try to see what `gradient(at:)` is doing with the arguments when I get the chance.

rxwei:
It should be a compiler bug that’s introduced recently. The fix is https://github.com/apple/swift/pull/26448 and we need to wait for that to get in.

rxwei:
The nightly toolchain used for CI is not up to date because of toolchain build failures. The team is looking into this.

jon-tow:
@dan-zheng Would you like me to port over the new tests in `swift/test/AutoDiff/control_flow.swift`?

dan-zheng:
> @dan-zheng Would you like me to port over the new tests in `swift/test/AutoDiff/control_flow.swift`?

The goal is to move all `Tensor`-specific AD tests to `tensorflow/swift-apis`. The `Tensor` control flow AD tests are the last of these.

`AutoDiff` tests from `apple/swift` (including `swift/test/AutoDiff/control_flow.swift`) are intentionally `Tensor`-independent and should not be ported to `tensorflow/swift-apis`.

marcrasi:
@jon-tow It looks like tests on this are failing because it's based off an older commit of swift-apis. If you merge master into this, we can probably rerun tests and merge this. Could you try that when you get a chance?

jon-tow:
Hey @marcrasi. I merged master into the branch here but I'm getting compiler errors on conditionals that compare `Tensor<Float>`s with scalar literals. E.g. `if x > 0` where `x` is a `Tensor<Float>`. This worked just fine a few months ago and any form of casting seems to not resolve the issue. 
Any suggestions would be appreciated. Thanks!

marcrasi:
It looks like the comparisons with scalars were deprecated and then removed (https://github.com/tensorflow/swift-apis/pull/516). The suggestion in the deprecation warning is to use `(lhs .> rhs).all()`. Does that fix things?

jon-tow:
It sure does! Thanks for pointing that out, invoking `all()` resolved the issue. :)

marcrasi:
Oh no, a very recent new commit made a backwards-incompatible change so CI is still failing. I think if you merge from master again we should be able to get CI passing.

jon-tow:
Sure thing! Not a problem. Let me know if you need me to address anything.

PR #426, opened 2019-08-08
Reverting inference behavior for BatchNorm

BradLarson:
This reverts inference behavior for BatchNorm to what it was before PR #384. The reshaping introduced there was triggering shape mismatches in the MiniGo tests and preventing the demo from running due to runtime errors.


dan-zheng:
I wonder if it's possible to check in a "MiniGo shape mismatch reproducer" test to ensure the bug in https://github.com/tensorflow/swift-apis/pull/384 won't happen again?

(I don't mean to block progress at all)

BradLarson:
That's a good idea. We need any kind of inference tests for BatchNorm and LayerNorm, because I believe both are only tested under training conditions. That would also help verify numerical correctness, like has been done on the training side.

eaplatanios:
I don't really understand why these reshapes were there. I'm preparing another PR that removes them from a couple other places where they also shouldn't be.

PR #427, opened 2019-08-08
Disable shape inference (temporarily) to move tensor commit-id.

bgogul:
We removed a parameter from `TFE_Shapes`. This is just a temporary change to smoothly transition the swift-apis repo without CI breakages.

### DIFF
--- a/Sources/TensorFlow/Core/LazyTensorShapeInference.swift
+++ b/Sources/TensorFlow/Core/LazyTensorShapeInference.swift
@@ -45,59 +45,61 @@ extension LazyTensorOperation {
 
     func updateOutputShapes() {
         outputShapes = Array<TensorShape?>(repeating: nil, count: outputCount)
-        // let status = TF_NewStatus()
-        // defer { TF_DeleteStatus(status) }
+        /*
+        let status = TF_NewStatus()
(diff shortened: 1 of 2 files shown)


bgogul:
> Use "/*" style?

There are some embedded '/*' comments in the code.

bgogul:
> > Use "/*" style?
> 
> There are some embedded '/*' comments in the code.

Did not realize swift allowed nested '/*'. Thanks, @pschuh

Sources/TensorFlow/Core/LazyTensorShapeInference.swift
```diff
@@ -44,6 +44,8 @@ extension LazyTensorOperation {
     }
 
     func updateOutputShapes() {
+        outputShapes = Array<TensorShape?>(repeating: nil, count: outputCount)
```
rxwei: Redundant generic argument.
```swift
outputShapes = Array(repeating: nil, count: outputCount)
```

bgogul: @rxwei Sorry, I just noticed this comment. I will be reverting this change shortly anyway.

PR #428, opened 2019-08-09
Normalization layers fix (fixes #384 and #426).

eaplatanios:
This PR fixes the shape related issues in the normalization layers (i.e., fixes #384 and #426). The role of `LayerNorm` is unclear to me currently as I believe it's not really the same as the layer norm in the Python API. In either case, this PR fixes shape issues in a principled way. We can deal with what `LayerNorm` is supposed to be later.

cc @rxwei @BradLarson @mikowals

### DIFF
--- a/Sources/TensorFlow/Layers/Normalization.swift
+++ b/Sources/TensorFlow/Layers/Normalization.swift
@@ -79,8 +79,6 @@ public struct BatchNorm<Scalar: TensorFlowFloatingPoint>: Layer {
             broadcastShape[positiveAxis] = input.shape[positiveAxis]
             offset = offset.reshaped(to: broadcastShape)
             scale = scale.reshaped(to: broadcastShape)
-            runningMean.value = runningMean.value.reshaped(to: broadcastShape)
-            runningVariance.value = runningVariance.value.reshaped(to: broadcastShape)
         }
         switch Context.local.learningPhase {
         case .training:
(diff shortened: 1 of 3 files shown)


eaplatanios:
@BradLarson could you please verify if your use cases works fine with these changes?

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -71,16 +71,26 @@ public struct BatchNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
+        let positiveAxis = (input.rank + axis) % input.rank
+        var offset = self.offset
+        var scale = self.scale
+        if positiveAxis != input.rank - 1 {
+            var broadcastShape = TensorShape([Int](repeating: 1, count: input.rank))
+            broadcastShape[positiveAxis] = input.shape[positiveAxis]
+            offset = offset.reshaped(to: broadcastShape)
+            scale = scale.reshaped(to: broadcastShape)
+            runningMean.value = runningMean.value.reshaped(to: broadcastShape)
+            runningVariance.value = runningVariance.value.reshaped(to: broadcastShape)
```
mikowals: Are these reshapes of `runningMean.value` and `runningVariance.value` necessary?  I think either have rank 0 or the same rank as input, so they already always broadcast correctly.

eaplatanios: Good point. I removed these reshapes.

eaplatanios:
Great, thanks @BradLarson!

mikowals:
@BradLarson I have an test that is part of #429.  I will make it part of a separate pull request later today.

BradLarson:
@mikowals - Excellent, thanks. Better test coverage overall is definitely something we're interested in.

PR #429, opened 2019-08-09
fix BatchNorm inference scale and offset shapes, add test

mikowals:
Fix the scale and offset reshapes done in #384 and partially reverted in #426.  

I also added a test for inference.  

I will do another pull request today to make similar changes in LayerNorm.


mikowals:
Closing this.  The fix is already done in #428.

PR #430, opened 2019-08-09
Added support for the 'selu' activation function.

eaplatanios:
Added support for the `selu` activation function, along with tests for `selu` and `relu`, and also made a few other minor changes.

### DIFF
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -1205,7 +1205,7 @@ func _vjpLeakyRelu<T: TensorFlowFloatingPoint>(
     })
 }
 
-/// Returns `scale * alpha * (exp(x) - 1)` if `x < 0`, and  `scale * x` otherwise.
+/// Returns `scale * alpha * (exp(x) - 1)` if `x < 0`, and `scale * x` otherwise.
 ///
 /// - Note: This is designed to be used together with the variance scaling layer initializers.
 ///   Please refer to [Self-Normalizing Neural Networks](https://arxiv.org/abs/1706.02515) for more
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -1602,8 +1602,10 @@ public extension Tensor where Scalar: Numeric & Comparable {
     /// - Parameter axes: The dimensions to reduce.
     /// - Precondition: Each value in `axes` must be in the range `-rank..<rank`.
     @inlinable
+    @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
     func min(alongAxes axes: [Int]) -> Tensor {
-        let axes = axes.map(Int32.init)
+        // TODO(TF-433): Remove workaround for differentiating `map`.
+        let axes = {axes.map(Int32.init)}()
         return min(alongAxes: Tensor<Int32>(axes))
     }
 
@@ -1612,6 +1614,7 @@ public extension Tensor where Scalar: Numeric & Comparable {
     /// - Parameter axes: The dimensions to reduce.
     /// - Precondition: Each value in `axes` must be in the range `-rank..<rank`.
     @inlinable
+    @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
     func min(alongAxes axes: Int...) -> Tensor {
         min(alongAxes: axes)
     }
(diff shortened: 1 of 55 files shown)


Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1185,24 +1200,30 @@ func _vjpLeakyRelu<T: TensorFlowFloatingPoint>(
...
 ) -> (Tensor<T>, (Tensor<T>) -> Tensor<T>) {
-    return (leakyRelu(x, alpha: alpha), { v in
+    (leakyRelu(x, alpha: alpha), { v in
         Raw.leakyReluGrad(gradients: v, features: x, alpha: alpha)
     })
 }
 
-/// Returns a tensor by applying the ReLU activation function to the specified tensor element-wise.
-/// Specifically, computes `max(0, x)`.
+/// Returns `scale * alpha * (exp(x) - 1)` if `x < 0`, and  `scale * x` otherwise.
```
```suggestion
/// Returns `scale * alpha * (exp(x) - 1)` if `x < 0`, and `scale * x` otherwise.
```

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1185,24 +1200,30 @@ func _vjpLeakyRelu<T: TensorFlowFloatingPoint>(
...
 ) -> (Tensor<T>, (Tensor<T>) -> Tensor<T>) {
-    return (leakyRelu(x, alpha: alpha), { v in
+    (leakyRelu(x, alpha: alpha), { v in
         Raw.leakyReluGrad(gradients: v, features: x, alpha: alpha)
     })
 }
 
-/// Returns a tensor by applying the ReLU activation function to the specified tensor element-wise.
-/// Specifically, computes `max(0, x)`.
+/// Returns `scale * alpha * (exp(x) - 1)` if `x < 0`, and `scale * x` otherwise.
```
Shashi456: Just a small suggestion in accordance with the rest of the activation function docs.

```suggestion
/// Returns  a tensor by applying the SeLU activation function to the specified tensor element-wise.
/// Specifically, computes `scale * alpha * (exp(x) - 1)` if `x < 0`, and `scale * x` otherwise.
```

eaplatanios: That's a good observation. I generally tend towards less verbosity in documentation strings (while not sacrificing clarity) and so would suggest changing the other activation function doc strings to be more concise instead. @Shashi456 @rxwei what do you think?

rxwei: How about:

> Returns a tensor by applying the SeLU activation function, namely `scale * alpha * (exp(x) - 1)` if `x < 0`, and `scale * x` otherwise.

rxwei: I'm positive about mentioning "the SeLU activation function" somewhere, which is a term that people can google to find the exact definition. Googling just the function name "selu" will not give an exact definition.

rxwei:
Why does it involve changes to trace tests?

Sources/TensorFlow/Operators/Math.swift
Shashi456: The adherence to this format was introduced in #160.

rxwei:
Could you separate changes to other ops to a different PR?

eaplatanios:
I updated the add function to use `Raw.addV2` instead of `Raw.add` to match the Python API behavior. The `AddV2` op was introduced recently. This change does not affect any behavior other than renaming the op in tracing tests.

eaplatanios:
> Could you separate changes to other ops to a different PR?

Sounds good. I will do that.

Sources/TensorFlow/Operators/Math.swift
eaplatanios: Sounds good. I'll make the change.

eaplatanios:
Done.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1168,6 +1168,21 @@ public func gelu<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
...
+@differentiable(vjp: _vjpRelu(_:))
+public func relu<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
+    Raw.relu(features: x)
+}
+
+@inlinable
+func _vjpRelu<T: TensorFlowFloatingPoint>(
+    _ x: Tensor<T>
+) -> (Tensor<T>, (Tensor<T>) -> Tensor<T>) {
+    (relu(x), { v in Tensor(x .> 0) * v })
```
Shashi456: I'm not sure about the performance issues, but @eaplatanios there is a `reluGrad` available as well, What do you think about using that instead?

eaplatanios: Good point! I switched to using that.

PR #431, opened 2019-08-09
Added initial support for learning rate schedules. 

eaplatanios:
This is an initial proposal for supporting learning rate schedules. If this is accepted, I plan to propose a way to integrate in optimizers so that users do not have to explicitly call `optimizer.learningRate = learningRate(forStep: ...)`, which is what is currently required.

Specifically, this PR adds support for the following learning rate schedules:
- `FixedLearningRate`
- `LinearlyDecayedLearningRate`
- `ExponentiallyDecayedLearningRate`
- `RSqrtDecayedLearningRate`
- `CosineDecayedLearningRate`
- `CycleLinear10xDecayedLearningRate`
- `LinearlyWarmedUpLearningRate`
- `ExponentiallyWarmedUpLearningRate`


rxwei:
Great work! The team is working hard on differentiation infra right now. I’ll try to do a thorough review next week if not this weekend.

eaplatanios:
Sounds good Richard! I'll keep updating this PR in the meantime.

rxwei:
Quick note: 4-space indentation :)

rxwei:
Oh, you already updated it. Nvm!

Shashi456:
@eaplatanios Consider adding the Triangular learning rate policy as well. [Paper](https://arxiv.org/pdf/1506.01186.pdf), Leslie smith 2015. Or i could add it, once this is merged.

saeta:
We could chat about this at the design review today, but we already have a lot on the docket and might not get to it.

eaplatanios:
@Shashi456 yes sure. This is by no means exhaustive. It's mostly a proposal for an API that I want to follow up with integration in optimizers. I just added a few implementations to show that the API is sufficient to support them.

@saeta sounds good! :)

eaplatanios:
@saeta could you give an example of how the `learningRate` argument would not be required? Even in the case of integrating with optimizers, I was just thinking of optimizers taking a learning rate schedule as when initialized and using that internally, but that would still require a `learningRate` argument in `callAsFunction`.

eaplatanios:
@jph00 I looked into harebrain and one concern I have is that all hyper parameters are string-indexed and float-valued. Please correct me if I'm wrong but I believe this is not very type safe. E.g., it lets you define schedule for parameters that your optimizer may not be using at all.

Sources/TensorFlow/Optimizers/LearningRates.swift
```diff
@@ -0,0 +1,348 @@
...
+/// ```
+/// schedule1(step: step, learningRate: schedule2(step: step, learningRate: learningRate))
+/// ```
+public struct ComposedLearningRateSchedule<
+    Schedule1: LearningRateSchedule,
+    Schedule2: LearningRateSchedule
+>: LearningRateSchedule where Schedule1.Scalar == Schedule2.Scalar {
+    public typealias Scalar = Schedule1.Scalar
+
+    public let schedule1: Schedule1
```
sjaz24: IMHO and perhaps a nitpick, but to me using 1 and 2 on the names is a bit misleading as schedule2 is performed before schedule1 and these numbers could be confused to denote the ordering of the schedules.

Sources/TensorFlow/Optimizers/LearningRates.swift
```diff
@@ -0,0 +1,348 @@
...
+    @inlinable
+    public init(slope: Scalar, lowerBound: Scalar = Scalar(0), startStep: UInt64 = 0) {
+        self.slope = slope
+        self.lowerBound = lowerBound
+        self.startStep = startStep
+    }
+
+    @inlinable
+    public func callAsFunction(step: UInt64, learningRate: Scalar) -> Scalar {
+        if step < startStep { return learningRate }
```
sjaz24: I don't know what the particular coding style of the developers of TensorFlow for Swift have, but my personal preference and so I'm not really suggesting a change, but just pointing out, that a guard condition for things like this would be my preference. 

```
guard step >= startStep else {
    return learningRate
}
```

saeta:
Hey @eaplatanios ! Great question. I was thinking something along the lines of:

```swift
public struct ExponentialLearningRateDecay<
    Scalar: FloatingPoint & ElementaryFunctions,
>: LearningRateSchedule {
    public let underlying: LearningRateSchedule
    public let decayRate: Scalar
    public let decayStepCount: UInt64
    public let staircase: Bool
    public let lowerBound: Scalar
    public let startStep: UInt64

     /// Creates a new exponential learning rate decay schedule.
    ///
    /// - Parameters:
    ///   - decayRate: Decay rate.
    ///   - decayStepCount: Decay step count.
    ///   - staircase: If `true`, the decay will occur at discrete intervals.
    ///   - lowerBound: Minimum decayed learning rate value as a fraction of the original learning
    ///     rate value.
    ///   - startStep: Step after which to start decaying the learning rate.
    @inlinable
    public init(
        _ underlying: LearningRateSchedule,
        decayRate: Scalar,
        decayStepCount: UInt64,
        staircase: Bool = false,
        lowerBound: Scalar = Scalar(0),
        startStep: UInt64 = 0
    ) {
        self.underlying = underlying
        self.decayRate = decayRate
        self.decayStepCount = decayStepCount
        self.staircase = staircase
        self.lowerBound = lowerBound
        self.startStep = startStep
    }

     @inlinable
    public func callAsFunction(step: UInt64) -> Scalar {
        let learningRate = underlying(step)
        if step < startStep { return learningRate }
        let step = step - startStep
        let power = Scalar(step) / Scalar(decayStepCount)
        let decay = Scalar.pow(decayRate, staircase ? power.rounded(.down) : power)
        return learningRate * ((1 - lowerBound) * decay + lowerBound)
    }
}
```

Does that make sense? (Note: I suspect that this isn't quite right, because of the associated type on `LearningRateSchedule`. This requires an extra type parameter, and some extra constraints if I had to hazard a guess.)

If you wanted to get cute, I suspect you could have the constant learning rate type conform to `ExpressibleByFloatLiteral` so that you could write something like:

```swift
ExponentialLearningRateDecay(0.01, ...)
```

WDYT?

eaplatanios:
@saeta yeah I totally see your point. I'll make some changes in this PR and see what people think. Brennan thanks a lot for the suggestion! :)

eaplatanios:
@saeta I updated the PR per your suggestions. :)

Sources/TensorFlow/Optimizers/LearningRates.swift
eaplatanios: Yes, that's a good question. I'm not sure what's the most common convention. Maybe @rxwei can comment on the style conventions for this library.

Sources/TensorFlow/Optimizers/LearningRates.swift
eaplatanios: That's a good point. The composed schedule is not there anymore so this confusion is now avoided. :)

saeta:
@eaplatanios you're totally welcome; that's why we have design reviews! :-) Thank _you_ for putting in the work to actually making this go! +1000!

eaplatanios:
No problem @saeta! The design meetings have been super useful. I'm also happy to hear feedback on this design and refine until everyone is happy. :) I'll update the optimizers in a separate PR after we eventually settle on the learning rates design.

rahulbhalley:
@eaplatanios I think it's important to add the references to papers which introduced the learning rates you have made the implementation of. 🙂

jph00:
Many thanks for the comment @eaplatanios . And thanks for encouraging the design review @saeta .

Maybe we should step back for a moment. How do we go about doing a design review for a piece of deep learning functionality? I think we have to start with: what deep learning problem is the design attempting to help solve? If we don't start there, then it's a bit like reviewing a tornado warning app in the absence of a tornado! ;)


In this case, what we want to support is annealing of training hyper-parameters. So, what are some of the things that need to be annealed during training? Here's a small but somewhat representative subset:

- Optimizer params
  - Learning rate
  - Momentum
  - Epsilon
  - Weight decay
- Model params
  - Dropout 
  - Variational distribution params
- Other training regularization params
  - AR
  - TAR
  - Augmentation params
- Data params
  - Image size
  - Batch size
- ... and much more

It's not actually possible to specify all these up-front, because research and practitioners continuously add more and more, and they are very specific to the details of model, optimizer, data, etc. In addition, a hyperparameter meaning can change subtly (e.g. 'eps' could be pre or post sqrt) or significantly (e.g. 'eps' might refer to an epsilon in a totally different area) from problem to problem.

Furthermore, practitioners often want to combine, add, and remove combinations of optimizers, architectures, regularization methods, etc as they build and test their model. As they do so, they don't want to be having the totally change the annealing methods they're using. Instead, generally they'd rather that unused annealing is quietly ignored (perhaps with an optional warning). And they need to be able to, at any time, decide to add or remove annealing from any of their params. The params may also be calculated based on the value of other params, in some other part of the overall system.

These requirements are why, over time, deep learning libs have generally moved towards more flexible ways to store and access hyperparams, and away from type-restricted mechanisms which contained assumptions about what hyperparams might be set and when/how they might be set. For instance, keras struggled for a long time with the implications of an early design choice that made it very hard to update params at any time other than the end of an epoch. So it's not really about "type safety", but about meeting user requirements.

So, for instance, in harebrain you'll see that there's a totally generalized optimizer (the first in any library, as far as we know) and a totally general system for getting and setting any param - including ones we haven't thought of yet. We showed examples of how to use this system to implement a wide variety of training methods - for instance during the course I co-taught with @lattner a new optimizer called [LAMB](https://arxiv.org/abs/1904.00962) came out from Google Brain, and thanks to this general platform we were able to have an implementation within one day and just 12 lines of code [here](https://github.com/fastai/course-v3/blob/master/nbs/swift/09_optimizer.ipynb). And any of the params in that new optimizer were, with no additional work, immediately accessible to any annealing method, because that's how the system is designed.

This issue is critical not just for researchers, but for both beginner and experienced practitioners. For instance, the current best practice for training the most popular types of models is [1cycle training](https://sgugger.github.io/the-1cycle-policy.html). fastai uses by default a slightly improved (unpublished) version that uses cosine annealing, which is [shown here](https://docs.fast.ai/callbacks.one_cycle.html). 1cycle training requires annealing not just LR, but also momentum. (More recently it's becoming clear that the next generation of best practices are going to involve annealing a far wider range of hyperparams.)

Is the current approach in harebrain the optimal API for Swift? It's very likely that it isn't! However I think if we are looking to improve the API, we should *start* with the functionality that practitioners require, and *then* refactor it and analyze redesigns based on user requirements. i.e. in this case it would be great to see PRs to harebrain that show how to improve the dev UX, without regressing needed functionality.

HTH! :)

eaplatanios:
@jph00 Thanks a lot for the detailed response! I completely agree with you regarding making general APIs that allow for the functionality you describe and thanks a lot for pointing these use cases out. I also believe that we don't need to sacrifice type safety to achieve that generality. For now, I will adapt this PR to support the functionality you describe. In fact, it almost already does (other than naming which is specific to learning rates).

My main concern with string-valued keys that are common for all optimizers is that this design is super prone to errors that will be hard for many users to debug. E.g., say someone adds a momentum decay schedule while using a gradient descent optimizer with no momentum terms. It's very easy to think that momentum is being used just because of adding a schedule for it, irrespective of whether the current optimizer is making use of momentum. I believe there is a nice way to achieve what you describe in a type-safe manner and so I'll go ahead and adapt this PR. I generally tend to be biased towards not letting users use APIs in ways that are unsupported at all as that eventually saves them a lot of time. At the same time, I agree we should not be sacrificing generality to achieve that.

Note here that we're also talking about a very limited family of optimizers that are currently popular in deep learning (e.g., simple gradient-descent-based optimizers) that is by no means general. For example, most optimizers in deep learning libraries do not allow for line search or things like conjugate gradients methods which are super useful in many cases (e.g., probabilistic generative models which are also gaining popularity again). If we were to write a general-purpose numerical optimization package we could do a much better job at being general in a type safe manner (e.g., [this](https://github.com/eaplatanios/makina/tree/master/optimization/src/main/java/makina/optimization) is a Java optimization package I wrote a few years back that is much more general than what existing deep learning libraries provide and that is also type safe).

To that end, I would like to propose that we separate the high-level library components from the `TensorFlow` module. For example, `Layer`s and `Optimizer`s could be moved to a separate module (which could be called `DeepLearning`) that can still reside in this repository. This will allow practitioners to import just core `TensorFlow` and implement their own high-level APIs on top of it, that better serve their purposes. What do you think? We could also have a `DeepLearningExperimental` module if we want to play around with ideas without providing any API stability guarantees.

eaplatanios:
By the way, I just want to say that it's great to be finally getting this discussion going so thanks @saeta for bringing it up in the design review and thanks @jph00 and others for being actively involved in the discussion.

eaplatanios:
@jph00 @saeta @rxwei should I go ahead and make changes to generalize this over arbitrary hyper parameters or should we defer for a future design review discussion?

rxwei:
How about bringing this up in Friday's meeting?

eaplatanios:
Sounds good @rxwei! In either case, I just updated this PR so that it implements a general notion of `ScheduledParameter`s that can be used for learning rates or other hyperparameters of optimizers, etc. I will leave modifying optimizers for a separate PR, if this one gets accepted.

jph00:
It appears that this still only handles optimizer params and can't be used in any of the other categories I listed. It would be much easier if this PR was to SwiftAI instead of swift-apis, since we're trying to keep high-level APIs separated, and SwiftAI already has rich integration tests and a fleshed-out API covering much of the functionality needed for training models.

No decision has been made yet on whether SwiftAI will remain in a separate repo or will be integrated into swift-apis, but in either case having higher-level APIs implemented in incompatible ways in two places will be very confusing!

I'm working on defining a clearer boundary between SwiftAI and swift-apis at the moment. The decision on where those boundaries go may involve having some things that are currently in SwiftAI moved into swift-apis, although my understanding is that the preference is generally to keep high-level APIs out of the swift-apis repo and instead just have a minimal set of protocols defined there.

If possible, I'd love to see this re-submitted as a PR to improve the schedulers in SwiftAI.

eaplatanios:
@jph00 thanks for the feedback! The current design can actually work across all the use cases you mentioned if we introduced a global way of keeping track of the training step (e.g., through the context that we use for layers). The reason I haven’t made an effort to integrate this with SwiftAI is because I’m not familiar with that library and don’t have the bandwidth currently to look into it in depth. In general though I feel we should allow for multiple high-level APIs as there are many designs that could work, and we should probably separate them from the `TensorFlow` module. Another reason I made this PR here is that there is already a lot of high-level APIs in this repo and I’m not sure where we draw the line and decide to move things over.

jph00:
Yes there's a lot of benefits to keeping high level APIs separate. There current high level APIs in swift-apis are somewhat of a legacy thing (they were there before the SwiftAI effort). I'd be happy to help get you up to speed with SwiftAI if you get the bandwidth to contribute in the future - one important benefit there is that schedulers need to be integrated with all the functionality for layerwise training, such as discriminative learning rates. That functionality is currently only in SwiftAI, and not in swift-api (although perhaps this is a protocol that should be included in the core of swift-api, so that all high level APIs can interoperate).

Also, swift-api doesn't currently have any generic optimizer - only a handful of specific optimizers that can't be combined or decoupled. A scheduler really needs to work with a generic optimizer framework to support scheduling the various pieces in novograd, radam, lookahead, etc. For instance, here's something that combines radam and lookahead: https://forums.fast.ai/t/meet-radam-imo-the-new-state-of-the-art-ai-optimizer/52656/26 . SwiftAI has a generic optimizer which supports mixing and matching in this way, and also results in far less code (and more understandable code!) I'll show examples tomorrow.

eaplatanios:
Thanks @jph00! I'll attend your presentation today and may follow up afterwards if needed. I'm quite busy until mid-September or so but I'll try to look into SwiftAI before then. :)

eaplatanios:
@jph00 thanks for the presentation on Friday! I've also been relying a lot on callbacks and believe are a great way to increase API versatility. To that end, I was wondering if it would be worth exploring a "function wrapper" feature for Swift, similar to Swift property wrappers. Currently, whenever we want to provide callbacks we need to pass in extra arguments to functions or use nested classes (e.g., similar to environment wrappers in OpenAI gym). This is especially true if we want to support callbacks that modify function arguments (e.g., observation normalization in a gym environment). In this case, it's quite ugly having to force users who implement environments to apply the callbacks themselves. An alternative is to have a default function that wraps another function with a callback (e.g., `Environment.step` and `Environment._step`) and have the users implement the latter one. However, that is also not elegant. Something like function wrappers (which would act similar to python decorators) would allow us to improve this API significantly. @rxwei do you think it might be a good idea to pursue such a design? If my example is confusing, I can elaborate a bit with a detailed use case.

eaplatanios:
Is there any interest in pushing forward with this PR?

pschuh:
I think in the spirit of Jeremy's comments, it would be best if this PR could go in as a part of splitting out the apis like so:

Tensorflow { // a module that just re-exports:
  Tensorflow_base
  Tensorflow_optimizer
}
This way we can make sure that things are layering properly. Ideally, the current Optimizer code could be moved into that other module as well. These shuffles will be easier when the modules are no longer built from the apple/swift repository. Once that move is finished, we should be able to commit this as the first CL in the optimizer target.

jph00:
Sounds like a useful direction Parker. A few important (IMO) things to consider:

I'd suggest trying to avoid blessing *any* high-level APIs as being part of s4tf, or within the s4tf repo. As soon as you do so, you will find that nearly all innovation in high-level APIs will disappear, since everyone will naturally use the official APIs. This is what's happened with keras, for instance (which, when it was chosen for tf, was indeed the best around, but there hasn't been much competition after that to help keep it fresh).

If you do decide to bless a high-level API in this way, make very sure it's the absolute best possible design, and compare plenty of designs in practice to confirm this. In particular, make sure that it's written in a way to every possible variant can be plugged in without modifying the lib (including variant that haven't been invented yet!)

Which leads me to the next issue: the most important thing isn't the high-level APIs, but the mid-level APIs. These are the things that let people build the high level stuff by taking advantage of carefully designed decoupled components. For instance, optimizers should be built on top of a generic optimizer class, which refactors stat tracking and step development, so the every optimizer can be built with those existing components. A generic training loop should be provided, so that everything changing the training loop can simply use callbacks to do so. A generic data processing API should be used by the generic training loop, so that anything reading or modifying data can do so in a consistent way.

Without the mid-tier APIs. you don't really end up with a proper high-level API, but something more like tf.contrib - a bunch of random stuff which doesn't really work together.

(And of course I'd be remiss not to mention that these mid-tier APIs are already provided by SwiftAI. So, for instance, a complete set of schedulers is already available in SwiftAI, which (as is required by modern neural net design) schedule not just LR, but *any* hyperparam, and handle discriminative LR, which is necessary for transfer learning, and is built entirely using callbacks, so doesn't change the training loop and requires very little code.)

texasmichelle:
Reminder to @pschuh to follow-up

pschuh:
Sorry for the delay on this, but we want to be more thorough about new apis going through swift-models first.

PR #433, opened 2019-08-09
Switched from using 'Raw.add' to using 'Raw.addV2'.

eaplatanios:
This is done so the add operator's behavior matches that of the Python API.


eaplatanios:
@dan-zheng yeah me too. I couldn't really find documentation and didn't look at the kernel implementations.

eaplatanios:
@dan-zheng I found out. `AddV2` is marked `is_commutative` and `is_aggregate`, which I assume can be used for optimizations by grappler maybe.

Issue #432, opened 2019-08-09
Rename /tools to /Tools for consistency


rxwei:

Shashi456:
@rxwei Should I leave this for a new contributor :P?

eaplatanios:
Done in #434.

PR #434, opened 2019-08-09
Rename `tools` to `Tools`

jon-tow:
#432


eaplatanios:
I believe this does not currently remove the `tools` directory. Could you also do something like `git rm tools` and the commit and push?

jon-tow:
@eaplatanios Thanks for the pointer.

eaplatanios:
Great, thanks! :)

PR #436, opened 2019-08-09
Documentation update for gradient computation

Shashi456:
Sorry for being meticulous about it, but the docs mention that `print(t.gradient(y,x))` directly would give you the value, whereas you would get `Tensor("AddN_45:0", shape=(5, 5), dtype=float32)`. This change allows for the values to be printed.


eaplatanios:
That is because you're using graph mode. With eager mode or TF 2.0 you don't need the eval call.

Shashi456:
@eaplatanios Alright thanks, I know it was graph mode, but for universal compatibility I thought this could work. I'll close the pull request.

PR #437, opened 2019-08-09
Adding LAMB optimizer

Shashi456:
Adding the LAMB optimizer, I just wanted some help to check in terms of correctness. 
image

A more readable version :
image

I adopted this from a [TensorFlow](https://github.com/cybertronai/pytorch-lamb/blob/master/pytorch_lamb/lamb.py) and [pyTorch](https://github.com/btahir/tensorflow-LAMB/blob/master/tensorflow_LAMB.py) version. 

My major doubt is r1 is to be updated wrt the parameters of the model while r2 wrt the adam based gradients. Am i doing that right? 

I'll update the PR with the test later.

### DIFF
--- a/Sources/TensorFlow/Optimizers/MomentumBased.swift
+++ b/Sources/TensorFlow/Optimizers/MomentumBased.swift
@@ -14,8 +14,8 @@
 
 /// RMSProp optimizer.
 ///
-/// It is recommended to leave the parameters of this optimizer at their default values (except for 
-/// the learning rate, which can be freely tuned). This optimizer is usually a good choice for 
+/// It is recommended to leave the parameters of this optimizer at their default values (except for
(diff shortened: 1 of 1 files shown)


eaplatanios:
@Shashi456 not sure what the status of this PR is, but before I go through it could you please make it valid Swift code? It looks like you've copied some Python code over but not converted it.

Shashi456:
@eapltanios it did compile locally in when I did it then. I'm mostly going to close this PR and rework this.

eaplatanios:
This PR definitely contains code that is not valid Swift code and looks like Python. Maybe something was not pushed. I have a working version of the LAMB optimizer so I could open a PR in the next few days adding it if that's ok with you.

Shashi456:
@eaplatanios, please do add it. I'll go ahead and close the PR. Will you add both lamb and Lars?

PR #439, opened 2019-08-10
Preserve order of outputs when generating explicit traces

bgogul:


Tests/TensorFlowTests/LazyTensorExplicitTraceTests.swift
```diff
@@ -157,6 +187,8 @@ final class LazyTensorExplicitTraceTests: XCTestCase {
         ("testClosureCapturesOfTensors", testClosureCapturesOfTensors),
         ("testClosureCapturesOfNonTensors", testClosureCapturesOfNonTensors),
         ("testNestedTracing", testNestedTracing),
-        ("testCallableTrace", testCallableTrace)
+        ("testCallableTrace", testCallableTrace),
+        ("testTraceWithOutputSameAsInput", testTraceWithOutputSameAsInput),
+        ("testRetainsidenticaloutputs", testRetainsIdenticalOutputs)
```
eaplatanios:
```suggestion
        ("testRetainsIdenticalOutputs", testRetainsIdenticalOutputs)
```

Sources/TensorFlow/Core/LazyTensorTrace.swift
```diff
@@ -243,7 +243,7 @@ class LazyTensorTraceBuilder {
         newLazyOp.inputs = lazyOp.inputs.map { maybePromotedInput($0) }
         updateOperationAndCache(id, newLazyOp)
 
-        if isOutput(lazyOp) {
+        if generateOutputs &&  LazyTensorHandle.isLive(lazyOp) {
```
rxwei:
```suggestion
        if generateOutputs && LazyTensorHandle.isLive(lazyOp) {
```

PR #440, opened 2019-08-10
Fix LayerNorm test and modify scale and offset broadcast to match Python

mikowals:
The original test was passing:

- because it tested a square 5 x 5 matrix hiding broadcast error, and
- checking vs Python showed values from testBatchNorm should not be used for testLayerNorm.

Also fixed typo where `positiveAxis` was not being used in the call to moments.


Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -171,11 +171,11 @@ public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
     @differentiable
     public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
         let positiveAxis = (input.rank + axis) % input.rank
-        var broadcastShape = input.shape
-        broadcastShape[positiveAxis] = 1
-        let offset = self.offset.reshaped(to: broadcastShape)
-        let scale = self.scale.reshaped(to: broadcastShape)
-        let moments = input.moments(alongAxes: axis)
+        var broadcastShape = Array(repeating: 1, count: input.rank)
```
rxwei:
```suggestion
        var broadcastShape = TensorShape(Array(repeating: 1, count: input.rank))
```

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -171,11 +171,11 @@ public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
+        broadcastShape[positiveAxis] = input.shape[positiveAxis]
+        let offset = self.offset.reshaped(to: TensorShape(broadcastShape))
```
rxwei:
```suggestion
        let offset = self.offset.reshaped(to: broadcastShape)
```

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -171,11 +171,11 @@ public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
+        let scale = self.scale.reshaped(to: TensorShape(broadcastShape))
```
rxwei:
```suggestion
        let scale = self.scale.reshaped(to: broadcastShape)
```

mikowals:
I made the suggested changes.

BradLarson:
We're still encountering problems due to the changes in BatchNorm and LayerNorm. The MiniGo tests in tensorflow/swift-models crash due to shape errors (just run "swift test" on tensorflow/swift-models), and the Transformer example crashes at runtime.

The tests here don't check for those cases, so I might expand the test cases to include those specific shapes. Also, the tests here aren't run in inference, because testBatchNormInference has the following inside it:

```
```

Changing that to .inference causes the tests to fail. I'll try to work on this over the next few days, unless you have an easy fix for these issues (again, checking against the MiniGo tests and Transformer model at runtime are great ways to verify).

mikowals:
Hi @BradLarson, the `Context.local.learningPhase = .training` at [line 526](https://github.com/tensorflow/swift-apis/blob/master/Tests/TensorFlowTests/LayerTests.swift#L526) is there to run a single training step in the next line.  That training step is there to test whether the moving averages tracked by `BatchNorm` are updated during training. Changing line 526 to `.inference` would be expected to make the test fail with the second call to `bnLayer.inferring` getting the same values as the first call before the training step.

The original problem turned up by the MiniGo tests was that the `runningMean.value` and `runningVariance.value` in `BatchNorm` were initialised with a shape of `[1]` so running inference before training would try to reshape a `[featureCount]` tensor to `[1]` and fail.  Inference before training is tested in [line 524](https://github.com/tensorflow/swift-apis/blob/master/Tests/TensorFlowTests/LayerTests.swift#L524).

Remaining problems being turned up by the MiniGo tests might be related to using an `axis` value in `BatchNorm` or `LayerNorm` not covered in `LayerTests.swift` or in using a input shape that is not 2D which is also currently not tested.  

Could you open an issue with the error message from the MiniGo test failure?

BradLarson:
@mikowals - Sorry about the misread on the tests, sounds good.

The shape mismatches appear to have been fixed somewhere around August 16, so this is no longer an issue. To be on the safe side, I added a couple of tests here: https://github.com/tensorflow/swift-apis/pull/462 that I verified fail in the cases where the two models hit runtime shape errors. Hopefully, this should guard against any regressions.

PR #441, opened 2019-08-10
[Layers] Update `DepthwiseConv2D(filterShape:...)` `bias` initialization

jon-tow:
**Issue**: `DepthwiseConv2D(filterShape:strides:padding:activation:filterInitializer:...)`
 produces a shape mismatch error in the conv-bias sum of `callAsFunction()` when initialized under a 
`filterShape` with channel-multiplier greater than `1`.
 More specifically, this is caused by initializing the `bias` vector with dimension equal to said channel-multiplier. 

**Fix** :This PR adds the proper `bias` vector initialization by computing the dimensionality as:
`bias vector dimension = output channel count = (input channel count * channel multiplier)`

### DIFF
--- a/Sources/TensorFlow/Layers/Convolutional.swift
+++ b/Sources/TensorFlow/Layers/Convolutional.swift
@@ -70,8 +70,8 @@ public struct Conv1D<Scalar: TensorFlowFloatingPoint>: Layer {
     ///
     /// and padding size is determined by the padding scheme.
     ///
-    /// - Parameter input: The input to the layer [batch count, input width, input channel count].
-    /// - Returns: The output of shape [batch count, output width, output channel count].
+    /// - Parameter input: The input to the layer [batch size, input width, input channel count].
+    /// - Returns: The output of shape [batch size, output width, output channel count].
     ///
     /// - Note: Padding size equals zero when using `.valid`.
     @differentiable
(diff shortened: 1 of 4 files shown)


rxwei:
Thanks for the patch. Could you please add a test?

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -495,8 +495,10 @@ public struct DepthwiseConv2D<Scalar: TensorFlowFloatingPoint>: Layer {
 
     /// Returns the output obtained from applying the layer to the given input.
     ///
-    /// - Parameter input: The input to the layer.
-    /// - Returns: The output.
+    /// - Parameter input: The input to the layer of shape,
+    ///   [batch count, input height, input width, input channel count]
```
```suggestion
    ///   [batch size, input height, input width, input channel count]
```

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -495,8 +495,10 @@ public struct DepthwiseConv2D<Scalar: TensorFlowFloatingPoint>: Layer {
...
+    /// - Returns: The output of shape,
+    ///   [batch count, output height, output width, input channel count * channel multiplier]
```
```suggestion
    ///   [batch size, output height, output width, input channel count * channel multiplier]
```

Issue #422, opened 2019-08-08
softmaxCrossEntropy can't be used directly as a loss function


sgugger:
Because it has a third argument (reduction) with a default value, ` softmaxCrossEntropy` can no longer be used as a loss function in the `Learner` API of the harebrain notebooks. A workaround is to define a new function that has only two arguments and pass it:
```
@differentiable(wrt: logits)
func crossEntropy(logits: Tensor<Float>, labels: Tensor<Int32>) {
    return softmaxCrossEntropy(logits: logits, labels: labels
}
```
but it's a bit of a waste of code...

Shashi456:
@sgugger so this decision was taken in the #356 PR, in an attempt to standardize loss functions. The reduction was added to maintain source compatibility.

sgugger:
I think it's a good idea to have the reduction parameter in the loss function in theory. However, it's very annoying that swift doesn't consider that a function with three arguments, the third one having a default value, can be interpreted as a function of two arguments.

rxwei:
@Shashi456 Thanks for chiming in to provide the context. The core issue here is not about whether we should have a reduction argument, but Swift's lack of support for function coercion with default parameters. Since reduction is a default parameter in all loss functions, ideally we should be able to coerce them to a function type that excludes the defaulted parameter.
```swift
func foo(x: T, y: T = z) -> T { ... }
foo as (T) -> T // Nope, Swift does not support it.
```

Right now there is a workaround: anywhere you passed `softmaxCrossEntropy`, you pass `{ softmaxCrossEntropy(logins: $0, labels: $1) }` instead. This is less heavyweight than declaring a new function. cc @sgugger 

To fix the breakage in the short term, we can consider making the `reduction` parameter not be a default parameter, but declare an overload that does not take a `reduction` but instead passes `mean` in, and do this for all loss functions. @eaplatanios what do you think?

To keep discussions going, we should start a discussion on the [Swift Evolution forum](https://forums.swift.org/c/evolution) about coercing functions with default parameters.

Shashi456:
@rxwei, would there be any way to do that generically for all the losses ?

rxwei:
> @rxwei, would there be any way to do that generically for all the losses ?
> Like we made a function layer for differentiable functions.

There is not, unfortunately.

eaplatanios:
[...]

PR #442, opened 2019-08-11
Add overloaded loss functions

Shashi456:
Fix: #422 
Added overloaded functions for losses which have a reduction parameter. 
cc: @saeta, @eaplatanios, @rxwei.

Build passes locally.

One final question, @rxwei you mentioned that the reduction parameter in the original loss functions should no longer be made default, Should I go ahead and make those changes as well?

### DIFF
--- a/Sources/TensorFlow/BackwardsCompatibility.swift
+++ b/Sources/TensorFlow/BackwardsCompatibility.swift
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//===------------------------------------------------------------------------------------------===//
+// Layers
+//===------------------------------------------------------------------------------------------===//
 // TODO: Remove this file after 0.4.
 
 public extension Tensor where Scalar == Int32 {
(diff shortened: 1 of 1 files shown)


Shashi456:
@rxwei done.

Sources/TensorFlow/BackwardsCompatibility.swift
```diff
@@ -615,3 +618,138 @@ public extension Dense {
...
+///   - predicted: Predicted outputs from a neural network.
+///   - expected: Expected values, i.e. targets, that correspond to the correct output.
+@differentiable(wrt: predicted)
+public func l2Loss<Scalar: TensorFlowFloatingPoint>(
+    predicted: Tensor<Scalar>,
+    expected: Tensor<Scalar>
+) -> Tensor<Scalar> {
+    l2Loss(predicted: predicted, expected: expected, reduction: { $0.sum() })
+}
+
```
rxwei: Where is `meanAbsoluteError`?

rxwei:
Oh actually my bad. I didn't notice that they have no reduction parameter.

Sources/TensorFlow/BackwardsCompatibility.swift
Shashi456: @rxwei, these functions are already two parameter functions and don't need to be overloaded.

rxwei:
> One final question, @rxwei you mentioned that the reduction parameter in the original loss functions should no longer be made default, Should I go ahead and make those changes as well?

There's no need to do that at the moment because there's no overload resolution ambiguity in the type checker.

Shashi456:
Kokoro surprisingly doesn't seem to be getting triggered.

Shashi456:
@rxwei this can be merged.

Issue #402, opened 2019-07-30
Add derivative tests for layers


dan-zheng:
Layers currently lack derivative tests.

* Tensor operation derivatives are tested in `Tests/TensorFlowTests/TensorAutoDiffTests.swift`.
* Layers are tested in `Tests/TensorFlowTests/LayerTests.swift`. (#77)

Consider adding layer derivative tests to `Tests/TensorFlowTests/LayerTests.swift`, alongside layer "forward-pass" tests.

Shashi456:
@dan-zheng might be the right time to make a LayerTests directory and have forwardpass and differentiation tests as files?

jkrukowski:
@Shashi456 @dan-zheng just wanted to ask you, what's the preferred way of implementing derivative tests for layers. Here is one of the first attempts https://github.com/tensorflow/swift-apis/pull/443 I'm thinking to use `pullback` function for that, something like
```swift
// ...
let seed = Tensor<Float>([[1, 1]])
let backprop = pullback(at: input, layer) { $1($0) }
let (_, 𝛁layer) = backprop(seed)
// ...
```

Shashi456:
The preferred way is at it is done for batch normalization in `tensorAutoDiffTests.swift`. You can check that [here](https://github.com/tensorflow/swift-apis/blob/bc047be5c55d2125f4115d80dd76811163c9eb00/Tests/TensorFlowTests/TensorAutoDiffTests.swift#L409).

jon-tow:
**Layer Gradient Tests**

- [ ] Sequential
- [ ] Conv1D
- [ ] Conv2D
- [ ] Conv3D
- [ ] DepthConv2D
- [ ] SeparableConv2D
- [ ] ZeroPadding1D
- [ ] ZeroPadding2D
- [ ] ZeroPadding3D
- [x] MaxPool1D 
- [x] MaxPool2D
- [x] MaxPool3D
- [ ] AvgPool1D
- [ ] AvgPool2D
- [ ] AvgPool3D
- [ ] GlobalAvgPool1D
- [ ] GlobalAvgPool2D
- [ ] GlobalAvgPool3D
- [ ] GlobalMaxPool1D
- [ ] GlobalMaxPool2D
- [ ] GlobalMaxPool3D
- [ ] UpSampling1D
- [ ] UpSampling2D
- [ ] UpSampling3D
- [ ] Reshape
- [ ] Flatten
- [ ] Embedding
- [x] Dense
[...]

PR #444, opened 2019-08-13
Add `MaxPool` derivative tests

jon-tow:
#402


eaplatanios:
Could you please add comments with python code that generates the reference values? You can take a look at the batch normalization tests for an example.

jon-tow:
@eaplatanios There's no python code. It's from paper.

eaplatanios:
I guess that's true about max pool so we could leave it without.However, using a python reference script can generally help us avoid mistakes that can result even from typos. I'm fine either way.

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -230,6 +254,18 @@ final class LayerTests: XCTestCase {
         XCTAssertEqual(output, expected)
     }
 
+    func testMaxPool3DGradient(){
+        let layer = MaxPool3D<Float>(poolSize: (2, 2, 2), strides: (1, 1, 1), padding: .valid)
+        let input = Tensor(shape: [1, 2, 2, 2, 1], scalars: (0..<8).map(Float.init))
+        let computedGradient = gradient(at: input, layer) { $1($0).sum() }
+        let expectedGradient = Tensor<Float>([[
```
Shashi456:
```suggestion
        let expectedGradient =  Tensor<Float>(shape: [1, 2, 2, 2, 1),
                                              scalars: [0, 0, 0, 0, 0, 0, 0, 1])
 ```

Shashi456: Just a suggestion for better readability.

jon-tow: I was going to do that but it seemed easier to visualize which indices carry gradient information from above.

jon-tow:
@eaplatanios I'll add the scripts for consistency.

jon-tow:
@rxwei Sorry about that. I was just about to merge branches.

PR #445, opened 2019-08-14
Remove references to TFE_TensorHandleIsConcrete

jaingaurav:


Sources/TensorFlow/Core/Tensor.swift
```diff
@@ -158,7 +158,6 @@ public extension Tensor {
     @inlinable
     var array: ShapedArray<Scalar> {
         debugLog("Returning a host copy of array.")
-        internalConsistencyCheck(handle.isConcrete)
 	    return handle.makeHostCopy()
```
eaplatanios:
```suggestion
      return handle.makeHostCopy()
```

jaingaurav: Thanks, it should be fixed now. I believe there was a mixture of some tabs in the file which caused the indentation issues.

eaplatanios: Thanks @jaingaurav! Yes it looked like some tabs were being used.

Sources/TensorFlow/Core/Tensor.swift
```diff
@@ -224,8 +223,8 @@ public extension Tensor {
             provided.
             """)
         self = scalars.withUnsafeBufferPointer { bufferPointer in
-	        Tensor(shape: shape, scalars: bufferPointer)
-	    }
+          Tensor(shape: shape, scalars: bufferPointer)
```
eaplatanios:
```suggestion
            Tensor(shape: shape, scalars: bufferPointer)
```

bgogul:
Thanks, @jaingaurav. It looks like there are a bunch of places where the function is still being used: 

https://source.cloud.google.com/results/invocations/22c80a3d-3b56-4666-a31d-935c77b552fb/targets/s4tf%2Fapis%2Fubuntu%2Fpresubmit/log

If you want to test this, You will need to get the latest swift toolchain from https://github.com/tensorflow/swift/blob/master/Installation.md

If it makes it easier for you, I can go ahead and fix this. Can you allow me (bgogul) to push changes to your branch?

PR #446, opened 2019-08-14
Add support for `tensorFunctionPointer` attribute in LazyTensor.

bgogul:


Sources/TensorFlow/Core/LazyTensorTFFunctionBuilder.swift
```diff
@@ -163,6 +163,12 @@ class TFGraph {
                     }
                 }
             }
+        case .tensorFunctionPointer(let value):
+            value.name.utf8CString.withUnsafeBufferPointer { buffer in
```
rxwei: You can use C APIs on Swift strings directly. The type checker has special cases for implicit conversion.

```suggestion
            TF_SetAttrFuncName(description, name, value.name, value.name.count)
```

PR #448, opened 2019-08-15
 Fused the LSTM cell matrix multiplication operations. 

eaplatanios:


eaplatanios:
Done. Note that if we resolve SR-10697/TF-507 we should be able to get improved performance.

PR #450, opened 2019-08-16
Shape check in `==` and `!=`

t-ae:
Resolved TODO in `==` and `!=`.

> // TODO: This is not correct due to broadcasting.

IMO, it doesn't make sense `[1, 2] == [[1, 2], [1, 2]]` is `true`, so I simply disabled broadcasting.

There was some test codes don't check equality strictly. Also fixed them.

### DIFF
--- a/Tests/TensorFlowTests/LayerTests.swift
+++ b/Tests/TensorFlowTests/LayerTests.swift
@@ -282,7 +282,7 @@ final class LayerTests: XCTestCase {
         let layer = GlobalMaxPool1D<Float>()
         let input = Tensor(shape: [1, 10, 1], scalars: (0..<10).map(Float.init))
         let output = layer.inferring(from: input)
-        let expected = Tensor<Float>([9])
+        let expected = Tensor<Float>([[9]])
         XCTAssertEqual(output, expected)
     }
 
@@ -290,7 +290,7 @@ final class LayerTests: XCTestCase {
         let layer = GlobalMaxPool2D<Float>()
         let input = Tensor(shape: [1, 2, 10, 1], scalars: (0..<20).map(Float.init))
         let output = layer.inferring(from: input)
-        let expected = Tensor<Float>([19])
+        let expected = Tensor<Float>([[19]])
         XCTAssertEqual(output, expected)
     }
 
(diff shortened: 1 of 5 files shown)


Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -282,23 +282,23 @@ final class LayerTests: XCTestCase {
         let layer = GlobalMaxPool1D<Float>()
         let input = Tensor(shape: [1, 10, 1], scalars: (0..<10).map(Float.init))
         let output = layer.inferring(from: input)
-        let expected = Tensor<Float>([9])
+        let expected = Tensor<Float>([[9]])
```
Shashi456: These are unrelated changes wrt to this PR, Are they absolutely needed? Because the tests passed before the change as well. But if it's still necessary I'd suggest opening a separate PR since they are layer test fixes.

Tests/TensorFlowTests/TensorAutoDiffTests.swift
```diff
@@ -46,7 +46,7 @@ final class TensorAutoDiffTests: XCTestCase {
         func negate<T : TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
             return (1 - x).sum()
         }
-        XCTAssertEqual(gradient(at: Tensor([0.1, 0.2, 0.3]), in: negate), Tensor(-1))
+        XCTAssertEqual(gradient(at: Tensor([0.1, 0.2, 0.3]), in: negate), Tensor([-1, -1, -1]))
```
Shashi456: Afaik, broadcasting is at works here.

Tests/TensorFlowTests/LayerTests.swift
t-ae: `GlobalMaxPool1D` outputs tensor which has shape [BatchSize, Channels].
In here input shape is `[1, 10, 1]` and output shape is `[1, 1]`.
I disabled broadcasting in this PR so this test doesn't pass without change.

Tests/TensorFlowTests/TensorAutoDiffTests.swift
t-ae: The TODO message originally written in `==` is:

> // TODO: This is not correct due to broadcasting.

It make sense to disable broadcastring. So I disabled it in this PR.

t-ae: My first PR comment didn't described the change enough.
I updated it.

Shashi456: In this specific case, since its a 1-D broadcast, I thought it was okay. The case you mentioned in the original comment though makes sense.

Issue #452, opened 2019-08-16
Re-implement 'TensorBuffer.create(create:withInitializer:)' using new Array initializer


rxwei:
[`Array.init(unsafeUninitializedCapacity:initializingWith:)`](https://developer.apple.com/documentation/swift/array/3200717-init) was introduced in [SE-0245](https://github.com/apple/swift-evolution/blob/master/proposals/0245-array-uninitialized-initializer.md). This solves a problem in the core tensor API where we allocate an empty array in [`TensorBuffer.init(create:withInitializer:)`](https://github.com/tensorflow/swift-apis/blob/3629ddc49b706df89f89fa3e92e495e6e3f42332/Sources/TensorFlow/Core/ShapedArray.swift#L85) and fill it in using TensorFlow C API (passed in as the `body` argument). Before this new `Array` initializer was available, we had to create a dummy pointer and use that to unsafely "initialize" an `Array` with unsafe data before we can call `body` on its buffer. Now we can replace the use of this dummy pointer with a call to `Array.init(unsafeUninitializedCapacity:initializingWith:)` and with a trailing closure that calls `body`.

PR #454, opened 2019-08-16
Re-implement `TensorBuffer.create(create:withInitializer:)` using new…

jon-tow:
#452


Sources/TensorFlow/Core/ShapedArray.swift
```diff
@@ -86,15 +86,10 @@ extension TensorBuffer {
...
-        /// Since `Scalar` may be any generic type, it is not possible to construct
-        /// an instance of `Scalar` directly for use with the
-        /// `Array(repeating:count:)` initializer. The workaround here is to
-        /// allocate a dummy `Scalar` pointer of size 1 and to use the pointee value
-        /// as the `repeatedValue` of the initializer.
-        let dummyPointer = UnsafeMutablePointer<Scalar>.allocate(capacity: 1)
-        var array = Array(repeating: dummyPointer.move(), count: count)
-        array.withUnsafeMutableBufferPointer { body($0) }
-        dummyPointer.deallocate()
+        let array = Array<Scalar>(unsafeUninitializedCapacity: 1) { buffer, initializedCount in
```
rxwei: The capacity here is incorrect. The buffer under `dummyPointer` was allocated with capacity `1` because we were to load its value (an uninitialized `Scalar`) and repeat it `count` times to initialize the array. Now that we are to allocate the array directly, we should allocate the array with a capacity equal to the number of elements we want to add to the array, aka. `count`.

```suggestion
        let array = Array<Scalar>(unsafeUninitializedCapacity: count) { buffer, initializedCount in
```

jon-tow: Ah sorry about that @rxwei. I was confused about that and should've asked in the issue. Changing now.

rxwei:
It looks like your branch happens to be on an old commit that's failing to build. Could you do a pull?

jon-tow:
@rxwei Done!

PR #174, opened 2019-06-04
Adding conv transpose 1d & 3d

Shashi456:
I'll add the tests. but till then could someone check the logic once and tell me if i'm wrong or right?

PR #458, opened 2019-08-20
Add Separable Conv1D layer

Shashi456:
Test and build pass.
Just a question, do we hope to eventually support NCWH formats? A lot of code would need to be modified in that case and I'm realizing that writing these layers. 
Could i open an issue for that or would you want to support all layers first before supporting multiple formats?

### DIFF
--- a/Tests/TensorFlowTests/LayerTests.swift
+++ b/Tests/TensorFlowTests/LayerTests.swift
@@ -222,11 +222,11 @@ final class LayerTests: XCTestCase {
                                            padding: .same)
         let input = Tensor(shape: [2, 2, 2], scalars: (0..<8).map(Float.init))
         let output = layer.inferring(from: input)
-        let expected = Tensor<Float>(shape: [2, 2, 1],
-                                     scalars: [17, 45, 73, 101])
+        let expected = Tensor<Float>(shape: [2, 2, 1], scalars: [17, 45, 73, 101])
         XCTAssertEqual(output, expected)
     }
 
+        
     func testSeparableConv2D() {
         let depthwiseFilter =  Tensor(shape: [2, 2, 2, 2], scalars: (0..<16).map(Float.init))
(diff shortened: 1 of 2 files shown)


jon-tow:
@Shashi456 A slight nitpick; could you please move the `Separable` layers below the `DepthwiseConv`s since the two are closely related and `ZeroPadding`s stands between them. 
It'd then also match the testing order.

Shashi456:
@jon-tow alright. But would a layer reorg be appropriate in this PR?

Shashi456:
@rxwei done. With this and transposed conv1d and 3d, we are done with basically supporting all convolutional layers :). Extensive testing and more functionality remains.

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -210,6 +210,22 @@ final class LayerTests: XCTestCase {
         XCTAssertEqual(output, expected)
     }
 
+    func testSeparableConv1D() {
+        let depthwiseFilter =  Tensor(shape: [2, 2, 2], scalars: (0..<8).map(Float.init))
```
rxwei:
```suggestion
        let depthwiseFilter = Tensor(shape: [2, 2, 2], scalars: (0..<8).map(Float.init))
```

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -210,6 +210,22 @@ final class LayerTests: XCTestCase {
...
+        let pointwiseFilter =  Tensor(shape: [1, 4, 1], scalars: (0..<4).map(Float.init))
```
rxwei:
```suggestion
        let pointwiseFilter = Tensor(shape: [1, 4, 1], scalars: (0..<4).map(Float.init))
```

Shashi456:
Can someone trigger a build on this ? And sorry for the persistence but #288 and #174 can be merged.

PR #459, opened 2019-08-20
Minor tweak: materialize inputs on demand in lazy tensor

bgogul:
Instead of iterating over all the lazy tensor operations and converting symbolic inputs that are materialized to concrete inputs, this PR triggers the conversion on demand only when it is needed during trace generation.


Sources/TensorFlow/Core/LazyTensorOperation.swift
```diff
@@ -848,7 +848,7 @@ extension LazyTensorOperation {
 
     /// Converts symbolic tensor inputs to concrete inputs if the
```
rxwei: Fit the comment to size? Looks like they are 80-column formatted.

PR #461, opened 2019-08-21
Lazy Tensor: Do not trigger materialization during shape inference

bgogul:
This PR avoids triggering materialization during the shape inference process. This tweak avoid reduces the overhead due to shape inference.

### DIFF
--- a/Sources/TensorFlow/Core/LazyTensorShapeInference.swift
+++ b/Sources/TensorFlow/Core/LazyTensorShapeInference.swift
@@ -47,15 +47,30 @@ extension LazyTensorOperation {
         let status = TF_NewStatus()
         defer { TF_DeleteStatus(status) }
 
-        let inputShapes: [TensorShape] = inputs.lazy.flatMap { (input: Input) -> [TensorShape] in
+        /// Returns shape only if it has already been computed.
+        func shapeIfAvailable(_ handle: LazyTensorHandle) -> TensorShape? {
+            switch handle.handle {
+            case .symbolic(let op, let index, _):
+                if let shape = op.outputShapes[index] { return shape }
+                return nil
+            case .concrete(let tfeHandle, _): return tfeHandle.shape
+            }
+        }
+
+        LazyTensorHandle._materializationCallback("inputshape_start")
+        let inputShapes: [TensorShape?] = inputs.lazy.flatMap { (input: Input) -> [TensorShape?] in
             switch input {
-            case .single(let handle): return [handle.shape]
-            case .list(let values): return values.lazy.map { $0.shape }
+            case .single(let handle): return [shapeIfAvailable(handle)]
+            case .list(let values): return values.lazy.map { shapeIfAvailable($0) }
             }
         }
         let inputShapeList = TF_NewShapeAndTypeList(/*num_shapes*/ Int32(inputShapes.count))
         defer { TF_DeleteShapeAndTypeList(inputShapeList) }
         for (i, shape) in inputShapes.enumerated() {
+            guard let shape = shape else {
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/Core/LazyTensorShapeInference.swift
```diff
@@ -47,15 +47,30 @@ extension LazyTensorOperation {
         let status = TF_NewStatus()
         defer { TF_DeleteStatus(status) }
 
-        let inputShapes: [TensorShape] = inputs.lazy.flatMap { (input: Input) -> [TensorShape] in
+        /// Returns shape only if it has already been computed.
+        func shapeIfAvailable(_ handle: LazyTensorHandle) -> TensorShape? {
+            switch handle.handle {
+            case .symbolic(let op, let index, _):
+                if let shape = op.outputShapes[index] { return shape }
```
rxwei:
```suggestion
                return op.outputShapes[index]
```

Sources/TensorFlow/Core/LazyTensorShapeInference.swift
```diff
@@ -47,15 +47,30 @@ extension LazyTensorOperation {
         let status = TF_NewStatus()
         defer { TF_DeleteStatus(status) }
 
-        let inputShapes: [TensorShape] = inputs.lazy.flatMap { (input: Input) -> [TensorShape] in
+        /// Returns shape only if it has already been computed.
+        func shapeIfAvailable(_ handle: LazyTensorHandle) -> TensorShape? {
```
rxwei:
```suggestion
        func shape(for handle: LazyTensorHandle) -> TensorShape? {
```
1. The `IfAvailable` semantics is already implied by the optional return type.
2. Then handle needs an argument label to form a phrase with the method base name.

Sources/TensorFlow/Core/LazyTensorShapeInference.swift
```diff
@@ -47,15 +47,30 @@ extension LazyTensorOperation {
...
+            switch handle.handle {
+            case .symbolic(let op, let index, _):
+                if let shape = op.outputShapes[index] { return shape }
+                return nil
+            case .concrete(let tfeHandle, _): return tfeHandle.shape
+            }
+        }
+
+        LazyTensorHandle._materializationCallback("inputshape_start")
```
rxwei:
```suggestion
        LazyTensorHandle.materializationCallback("inputshape_start")
```
The underscore can be dropped since it's internal.

Sources/TensorFlow/Core/LazyTensorShapeInference.swift
```diff
@@ -47,15 +47,30 @@ extension LazyTensorOperation {
...
+        let inputShapes: [TensorShape?] = inputs.lazy.flatMap { (input: Input) -> [TensorShape?] in
```
rxwei: Redundant type information.
```suggestion
        let inputShapes: [TensorShape?] = inputs.lazy.flatMap { input in
```

Sources/TensorFlow/Core/LazyTensorShapeInference.swift
bgogul: Oops. this should not have been in the PR at all.

Sources/TensorFlow/Core/LazyTensorShapeInference.swift
bgogul: The type checker does not like this change. :(

Sources/TensorFlow/Core/LazyTensorShapeInference.swift
bgogul: Thanks! Done.

PR #466, opened 2019-08-22
Added back a test for 'Tensor.init(stacking:alongAxis:)'.

eaplatanios:
@rxwei I believe this has also now been fixed in apple/swift.


Tests/TensorFlowTests/TensorAutoDiffTests.swift
```diff
@@ -153,18 +153,17 @@ final class TensorAutoDiffTests: XCTestCase {
...
-    //     XCTAssertEqual(a1, grads.0)
-    //     XCTAssertEqual(b1, grads.1)
-    // }
+    func testTensorInitStacking() {
+        let a1 = Tensor<Float>([1, 2, 3, 4, 5])
+        let b1 = Tensor<Float>([6, 7, 8, 9, 10])
+        let a2 = Tensor<Float>([1, 1, 1, 1, 1])
+        let b2 = Tensor<Float>([1, 1, 1, 1, 1])
+        let grads = gradient(at: a2, b2) { a, b in
+            Tensor<Float>(stacking: [a1 * a, b1 * b], alongAxis: -1).sum()
```
rxwei: Are you still getting a warning?

eaplatanios: I haven't been able to compile the toolchain from master since the merge that happened today so I couldn't check. I get failures when compiling LLVM but I also recently updated to Catalina so I'm not sure if that's the issue yet.

eaplatanios: This seems to be passing the CI server so should we go ahead and merge it?

rxwei: yes

PR #468, opened 2019-08-22
Update 'Array.DifferentiableView' uses to 'Array.TangentVector'.

rxwei:
This is necessary for https://github.com/apple/swift/pull/26771.


shabalind:
Hi @rxwei, we liked the direction, but it seems like PR has been stale for quite some time. Please reopen if you feel like the issue is still relevant! Thanks!

PR #467, opened 2019-08-22
Add missing tests for XCTestManifests

t-ae:
There's some tests not included in XCTestManifests.swift.
They are not executed on Ubuntu. So CI can miss test failures.

I think it's better to use `swift test --enable-test-discovery` and quit maintaining XCTestManifests manually. What do you think?

PR #470, opened 2019-08-22
Add GlobalMaxPool gradient tests

Shashi456:


Shashi456:
@rxwei the latest builds won't work. There's a test error as I mentioned in #467 .
```
/swift-apis/Tests/TensorFlowTests/TensorAutoDiffTests.swift:164: error: TensorAutoDiffTests.testTensorInitStacking : XCTAssertEqual failed: ("[1.0, 2.0, 3.0, 4.0, 5.0]") is not equal to ("0.0") -
/swift-apis/Tests/TensorFlowTests/TensorAutoDiffTests.swift:165: error: TensorAutoDiffTests.testTensorInitStacking : XCTAssertEqual failed: ("[ 6.0,  7.0,  8.0,  9.0, 10.0]") is not equal to ("0.0") -
```

Same build error in #471 as well.

eaplatanios:
@rxwei this test passed when we merged it last night. Did the toolchain get updated since then?

Shashi456:
@eaplatanios so when you built it, they weren't getting tested. `TensorAutoDiffTests` were introduced in #469

rxwei:
Toolchain builds are blocked on [TF-770](https://bugs.swift.org/browse/TF-770).

eaplatanios:
@Shashi456 good catch. In that case we need to wait for [TF-770](https://bugs.swift.org/browse/TF-770) to be resolved as this test should only pass with a newer toolchain.

@rxwei oh that's the same issue I was having on my mac. Good to know it's reproducible.

Shashi456:
Merged into master, this shouldn't fail anymore.

Shashi456:
This can be merged as well.

PR #473, opened 2019-08-22
Adding a 'linspace' Tensor initializer and tests

BradLarson:
This adds a `Tensor.init(linearSpaceFrom:to:count:)` initializer that has both scalar and `Tensor` parameter variants.

It also adds tests, based on the values generated by Numpy's `linspace()` operation.


Sources/TensorFlow/Initializers.swift
```diff
@@ -327,6 +327,37 @@ public extension Tensor where Scalar: Numeric {
...
+    /// including an end value, spaced evenly to generate the number of values specified.
+    ///
+    /// - Parameters:
+    ///   - start: The starting value to use for the sequence. If the sequence contains any values,
+    ///     the first one is `start`.
+    ///   - end: An end value to limit the sequence. `end` is the last element of the resulting
+    ///     sequence.
+    ///   - count: The number of values in the resulting sequence. `count` must be positive.
+    @inlinable
+    init(linSpaceFrom start: Scalar, to end: Scalar, count: Int) {
```
eaplatanios: Since we're using full names in the API so far, it may be better to rename the argument label to:
```suggestion
    init(linearSpaceFrom start: Scalar, to end: Scalar, count: Int) {
```

BradLarson: That makes sense to me. Wasn't sure if we wanted the `linspace` for familiarity when comparing to Numpy and Python TF, but I'm all for descriptive names.

Sources/TensorFlow/Initializers.swift
```diff
@@ -327,6 +327,37 @@ public extension Tensor where Scalar: Numeric {
...
+
+    /// Creates a 1-D tensor representing a sequence from a starting value, up to and
+    /// including an end value, spaced evenly to generate the number of values specified.
+    ///
+    /// - Parameters:
+    ///   - start: The starting value to use for the sequence. If the sequence contains any values,
+    ///     the first one is `start`.
+    ///   - end: An end value to limit the sequence. `end` is the last element of the resulting
+    ///     sequence.
+    ///   - count: The number of values in the resulting sequence. `count` must be positive.
```
rxwei: Add a `- Precondition: ` to assert that `start`, `end`, and `count` must be scalars?

Tests/TensorFlowTests/InitializerTests.swift
```diff
@@ -37,6 +37,32 @@ final class InitializerTests: XCTestCase {
...
+        assertEqual(
+            Tensor<Float>(
+                [
+                    2, 1.78888889, 1.57777778, 1.36666667, 1.15555556, 0.94444444, 0.73333333,
+                    0.52222222, 0.31111111, 0.1
+                ]), linearSpaceTensor2, accuracy: 0.001)
+
+        let linearSpaceTensor3 = Tensor<Float>(linearSpaceFrom: -5.0, to: 8.5, count: 20)
+        assertEqual(
+            Tensor<Float>(
```
rxwei: This call to `init(_:)` may not be necessary because `Tensor` is expressible by array literal.

Sources/TensorFlow/Initializers.swift
BradLarson: Done.

Tests/TensorFlowTests/InitializerTests.swift
BradLarson: Good call, that simplifies things. Fixed that.

eaplatanios:
The failures were due to a different issue so I'm restarting tests now.

rxwei:
Could you pull from master?

PR #475, opened 2019-08-23
Fix a compiler warning in lazy tensor.

bgogul:
With the latest toolchain, we get the following warning: 
```
warning: cannot match several associated values at once, implicitly tupling the associated values and trying to match that instead
        case .concrete(_): return nil
```

This PR fixes the warning.

### DIFF
--- a/Sources/TensorFlow/Core/LazyTensorOperation.swift
+++ b/Sources/TensorFlow/Core/LazyTensorOperation.swift
@@ -95,7 +95,7 @@ class LazyTensorHandle: _AnyTensorHandle {
     var lazyTensorOperation: LazyTensorOperation? {
         switch handle {
         case .symbolic(let op, _, _): return op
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/Core/LazyTensorOperation.swift
```diff
@@ -95,7 +95,7 @@ class LazyTensorHandle: _AnyTensorHandle {
     var lazyTensorOperation: LazyTensorOperation? {
         switch handle {
         case .symbolic(let op, _, _): return op
-        case .concrete(_): return nil
+        case .concrete(_, _): return nil
```
rxwei: You can also just write
```suggestion
        case .concrete: return nil
```

bgogul: Even better! Done.

PR #476, opened 2019-08-23
Lazy tensor: automatically promote constants to inputs based on history

bgogul:
This PR implements the logic to promote constants to inputs (https://bugs.swift.org/browse/TF-704). Suppose that we have the following traces:

```
lazyTrace_sum2(%0: float) -> (float) {
   %1 = Const[dtype: float, value: 2.0]() 
   %2 = Add[T: float](%0, %1)
   return %2
}
```

```
lazyTrace_sum2(%0: float) -> (float) {
   %1 = Const[dtype: float, value: 3.0]() 
   %2 = Add[T: float](%0, %1)
   return %2
}
```
The above two functions are different and will need to be compiled individually. Suppose that we promote these constants to inputs as follows:
```
lazyTrace_sum3(%0: float, %1: float) -> (float) {
   %2 = Add[T: float](%0, %1)
   return %2
}
```
Only the inputs change and the function remains constant, which eliminates the additional compilation overhead.


Tests/TensorFlowTests/LazyTensorTraceCacheTests.swift
```diff
@@ -0,0 +1,131 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import XCTest
+
+@testable import TensorFlow
+import CTensorFlow
+
+final class LazyTensorTraceCacheTests : LazyTensorTestCase {
```
eaplatanios:
```suggestion
final class LazyTensorTraceCacheTests: LazyTensorTestCase {
```

Tests/TensorFlowTests/LazyTensorTraceCacheTests.swift
```diff
@@ -0,0 +1,131 @@
...
+            lazyTrace_3() -> (%2) {
+              %0 = Const[dtype: float, value: [[1.0, 2.0, 3.0]]]()
+              %1 = Const[dtype: float, value: 5.0]()
+              %2 = Mul[T: float](%0, %1)
+            }
+            """)
+        XCTAssertEqual(y.scalars, [5.0, 10.0, 15.0])
+    }
+
+    func testDontPromoteEqualConstants() {
```
eaplatanios:
```suggestion
    func testDoNotPromoteEqualConstants() {
```

Tests/TensorFlowTests/LazyTensorTraceCacheTests.swift
```diff
@@ -0,0 +1,131 @@
...
+
+    private func lazyTrace<T: TensorFlowScalar>(
+        _ input: Tensor<T>
+    ) -> LazyTensorTrace {
+        let lazyOperation = lazyTensorOperation(input)!
+        return LazyTensorTraceBuilder.materializationTraceInfo(lazyOperation).trace
+    }
+
+    static var allTests = [
+        ("testAutoConstPromotion", testAutoConstPromotion),
```
eaplatanios:
```suggestion
        ("testConstantPromotion", testConstantPromotion),
```

Tests/TensorFlowTests/LazyTensorTraceCacheTests.swift
```diff
@@ -0,0 +1,131 @@
...
+        ("testDontPromoteEqualConstants", testDontPromoteEqualConstants)
```
eaplatanios:
```suggestion
        ("testDoNotPromoteEqualConstants", testDoNotPromoteEqualConstants)
```

bgogul:
> > @rxwei this is not super urgent, but can you take a look at this when you get a chance?
> 
> Sorry for the delay!

No worries. I addressed your comments and left some clarification questions.

PR #477, opened 2019-08-23
Remove Comparable conformance of Tensor

### LINKED ISSUE #451: Problem in comparison operators
https://github.com/tensorflow/swift-apis/blob/3629ddc49b706df89f89fa3e92e495e6e3f42332/Sources/TensorFlow/Operators/Comparison.swift#L104-L132

Documentation comments says they compare two tensors "lexicographically".
But it seems the code doing is not lexicographical comparison.

I tried to fix it, but soon got stuck.
In the first place, is lexicographical order defined on tensors?

t-ae:
Resolve #451. 
1. Remove `Comparable` conformance of `Tensor`.
2. Update documentation comments of comparison operators.


rxwei:
The comparison operators were only added to satisfy `Tensor`'s conformance to `Comparable`. So we should remove these operators as well.

rxwei:
I'd suggest marking those operators deprecated first and with a message that suggests what to use instead.

t-ae:
Added deprecated messages.
Should I move them into BackwardsCompatibility.swift?

Shashi456:
@t-ae yeah that might be a good idea.

rxwei:
Yes. Thanks!

Issue #479, opened 2019-08-23
Update 'min(_:_:)' and 'max(_:_:)' gradients to match Python TensorFlow.


rxwei:
IIRC, `tf.minimum` and `tf.maximum`'s gradients are not rounded.  We should update our implementation to match that.

rxwei:
Note: Fixing this is a high priority.

dan-zheng:
Here are the Python TensorFlow gradients for `minimum` and `maximum`, which we want to mimic:
```python
import tensorflow as tf
tf.enable_eager_execution()

x = tf.constant([4, 5, 3], dtype=tf.float32)
y = tf.constant([4, 2, 6], dtype=tf.float32)

with tf.GradientTape(persistent=True) as t:
  t.watch([x, y])
  result = tf.reduce_sum(tf.math.minimum(x, y))
print(t.gradient(result, [x, y]))
# tf.Tensor([1. 0. 1.], shape=(3,), dtype=float32)
# tf.Tensor([0. 1. 0.], shape=(3,), dtype=float32)

with tf.GradientTape(persistent=True) as t:
  t.watch([x, y])
  result = tf.reduce_sum(tf.math.minimum(y, x))
print(t.gradient(result, [x, y]))
# tf.Tensor([0. 0. 1.], shape=(3,), dtype=float32)
# tf.Tensor([1. 1. 0.], shape=(3,), dtype=float32)
```

Here is our current behavior, which is incorrect. I believe our logic was originally adapted from Tangent, which performs tie breaking (derivative is sometimes 0.5 before scaling).

```swift
import TensorFlow

let x = Tensor([4, 5, 3])
let y = Tensor([4, 2, 6])
print(gradient(at: x, y) { x, y in min(x, y).sum().scalarized() })
// ([0.5, 0.0, 1.0], [0.5, 1.0, 0.0])
print(gradient(at: x, y) { x, y in min(y, x).sum().scalarized() })
// ([0.5, 0.0, 1.0], [0.5, 1.0, 0.0])
```

eaplatanios:
Thanks for clarifying @dan-zheng! In order to have all the relevant information here, the Python API gradient implementation can be found [here](https://github.com/tensorflow/tensorflow/blob/752556174a1dcf68693784a8c78f07ba818e230f/tensorflow/python/ops/math_grad.py#L1417).

PR #480, opened 2019-08-24
Update `min(_:_:)` and `max(_:_:)` gradients to match Python TensorFlow

jon-tow:
#479

### DIFF
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -1399,9 +1399,9 @@ internal func _vjpMinMaxHelper<T: TensorFlowFloatingPoint>(
     originalValue: Tensor<T>,
     seed: Tensor<T>
 ) -> (Tensor<T>, Tensor<T>) {
-    let denominator = 1 + Tensor<T>(x .== y)
-    let lhsGrad = seed * Tensor<T>(x .== originalValue) / denominator
-    let rhsGrad = seed * Tensor<T>(y .== originalValue) / denominator
+    let tieBreakerMask = Tensor<T>(x .!= y)
+    let lhsGrad = seed * Tensor<T>(x .== originalValue)
+    let rhsGrad = seed * Tensor<T>(y .== originalValue) * tieBreakerMask
     let (lhsShape, rhsShape) = (x.shapeTensor, y.shapeTensor)
     let (lhsAxes, rhsAxes) = Raw.broadcastGradientArgs(s0: lhsShape, s1: rhsShape)
     return (lhsGrad.sum(squeezingAxes: lhsAxes).reshaped(toShape: lhsShape),
--- a/Tests/TensorFlowTests/TensorAutoDiffTests.swift
+++ b/Tests/TensorFlowTests/TensorAutoDiffTests.swift
@@ -153,6 +153,42 @@ final class TensorAutoDiffTests: XCTestCase {
         XCTAssertEqual(varianceGradAlongAxes(input), expected)
     }
 
+    func testMin() {
+        func f(a: Tensor<Float>, b: Tensor<Float>) -> Tensor<Float> { min(a, b).sum() }
+        var a = Tensor<Float>([[3.0, -2.0], [0.3, 10.0]])
+        var b = Tensor<Float>([[9.0, -3.0], [0.1, 10.0]])
+        var computedGradient = gradient(at: a, b, in: f)
+        var expectedGradient: (Tensor<Float>, Tensor<Float>) = (
+            [[1.0, 1.0], [1.0, 1.0]], [[0.0, 0.0], [1.0, 0.0]])
+        XCTAssertEqual(computedGradient.0, expectedGradient.0)
+        XCTAssertEqual(computedGradient.1, expectedGradient.1)
+
+        a = Tensor<Float>([[3.0, -2.0], [0.3, 10.0]])
+        b = Tensor<Float>([9.0, -3.0])
+        computedGradient = gradient(at: a, b, in: f)
+        expectedGradient = ([[1.0, 1.0], [1.0, 0.0]], [0.1, 1.0])
+        XCTAssertEqual(computedGradient.0, expectedGradient.0)
+        XCTAssertEqual(computedGradient.1, expectedGradient.1)
+    }
+
+    func testMax() {
+        func f(a: Tensor<Float>, b: Tensor<Float>) -> Tensor<Float> { max(a, b).sum() }
+        var a = Tensor<Float>([[3.0, -2.0], [0.3, 10.0]])
+        var b = Tensor<Float>([[9.0, -3.0], [0.5, 10.0]])
+        var computedGradient = gradient(at: a, b, in: f)
+        var expectedGradient: (Tensor<Float>, Tensor<Float>) = (
+            [[0.0, 1.0], [0.0, 1.0]], [[1.0, 0.0], [1.0, 0.0]])
+        XCTAssertEqual(computedGradient.0, expectedGradient.0)
+        XCTAssertEqual(computedGradient.1, expectedGradient.1)
+
+        a = Tensor<Float>([[3.0, -2.0], [0.3, 10.0]])
+        b = Tensor<Float>([9.0, -3.0])
+        computedGradient = gradient(at: a, b, in: f)
+        expectedGradient = ([[0.0, 1.0], [0.0, 1.0]], [2.0, 0.0])
+        XCTAssertEqual(computedGradient.0, expectedGradient.0)
+        XCTAssertEqual(computedGradient.1, expectedGradient.1)
+    }
+
     /*TODO:(https://bugs.swift.org/browse/TF-771): Disabling this case as assertions fail.
     func testTensorInitStacking() {
         let a1 = Tensor<Float>([1, 2, 3, 4, 5])
@@ -449,6 +485,8 @@ final class TensorAutoDiffTests: XCTestCase {
         ("testSum", testSum),
         ("testMean", testMean),
         ("testVariance", testVariance),
+        ("testMin", testMax),
+        ("testMax", testMax),
         // TODO(https://bugs.swift.org/browse/TF-771): Disabling the failing test.
         // ("testTensorInitStacking", testTensorInitStacking),
         ("testExpandingShape", testExpandingShape),


Tests/TensorFlowTests/TensorAutoDiffTests.swift
```diff
@@ -153,6 +153,42 @@ final class TensorAutoDiffTests: XCTestCase {
         XCTAssertEqual(varianceGradAlongAxes(input), expected)
     }
 
+    func testMin() {
```
rxwei: Please add the corresponding Python TensorFlow (2.0) reproducer in a block comment for both min and max tests.

dan-zheng: Here's a start:
```python
import tensorflow as tf

tf.enable_eager_execution()

def test_op(x, y, op):
  with tf.GradientTape(persistent=True) as t:
    t.watch(x)
    t.watch(y)
    result = tf.reduce_sum(op(x, y))
  
  print(t.gradient(result, x))
  print(t.gradient(result, y))
  
  with tf.GradientTape(persistent=True) as t:
    t.watch(x)
    t.watch(y)
    result = tf.reduce_sum(op(y, x))
  
  print(t.gradient(result, x))
  print(t.gradient(result, y))

x = tf.constant([4, 5, 3], dtype=tf.float32)
y = tf.constant([4, 2, 6], dtype=tf.float32)
test_op(x, y, tf.math.maximum)
test_op(x, y, tf.math.minimum)

x = tf.constant([1, 5, 3], dtype=tf.float32)
y = tf.constant([4], dtype=tf.float32)
test_op(x, y, tf.math.maximum)
test_op(x, y, tf.math.minimum)

x = tf.constant([1, 5, 3], dtype=tf.float32)
y = tf.constant(2, dtype=tf.float32)
test_op(x, y, tf.math.maximum)
test_op(x, y, tf.math.minimum)
```

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1399,9 +1399,9 @@ internal func _vjpMinMaxHelper<T: TensorFlowFloatingPoint>(
     originalValue: Tensor<T>,
     seed: Tensor<T>
 ) -> (Tensor<T>, Tensor<T>) {
-    let denominator = 1 + Tensor<T>(x .== y)
-    let lhsGrad = seed * Tensor<T>(x .== originalValue) / denominator
-    let rhsGrad = seed * Tensor<T>(y .== originalValue) / denominator
+    let tieBreakerMask = Tensor<T>(x .!= y)
```
dan-zheng: Are you sure this implementation is correct?

Here's my result, after porting directly from [`MaximumMinimumGradHelper` in `tensorflow/core/ops/math_grad.cc`](https://github.com/tensorflow/tensorflow/blob/dac733509fbe366804656a71f15ba0712bb5e69d/tensorflow/core/ops/math_grad.cc#L614):
```diff
diff --git a/Sources/TensorFlow/Operators/Math.swift b/Sources/TensorFlow/Operators/Math.swift
index 81d890f..c8688af 100644
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -1344,7 +1344,7 @@ internal func _vjpMax<T: TensorFlowFloatingPoint>(
     _ y: Tensor<T>
 ) -> (Tensor<T>, (Tensor<T>) -> (Tensor<T>, Tensor<T>)) {
     let value = max(x, y)
-    return (value, { v in _vjpMinMaxHelper(x, y, originalValue: value, seed: v) })
+    return (value, { v in _vjpMinMaxHelper(x, y, originalValue: value, seed: v, comparisonOperation: .<=) })
 }

 /// Returns the element-wise maximum of the scalar and the tensor, broadcasting the scalar.
@@ -1375,7 +1375,7 @@ internal func _vjpMin<T: TensorFlowFloatingPoint>(
     _ y: Tensor<T>
 ) -> (Tensor<T>, (Tensor<T>) -> (Tensor<T>, Tensor<T>)) {
     let value = min(x, y)
-    return (value, { v in _vjpMinMaxHelper(x, y, originalValue: value, seed: v) })
+    return (value, { v in _vjpMinMaxHelper(x, y, originalValue: value, seed: v, comparisonOperation: .>=) })
 }

 /// Returns the element-wise minimum of the scalar and the tensor, broadcasting the scalar.
@@ -1397,11 +1397,12 @@ internal func _vjpMinMaxHelper<T: TensorFlowFloatingPoint>(
     _ x: Tensor<T>,
     _ y: Tensor<T>,
     originalValue: Tensor<T>,
-    seed: Tensor<T>
+    seed: Tensor<T>,
+    comparisonOperation: (Tensor<T>, Tensor<T>) -> Tensor<Bool>
 ) -> (Tensor<T>, Tensor<T>) {
-    let denominator = 1 + Tensor<T>(x .== y)
-    let lhsGrad = seed * Tensor<T>(x .== originalValue) / denominator
-    let rhsGrad = seed * Tensor<T>(y .== originalValue) / denominator
+    let mask: Tensor<T> = Raw.cast(comparisonOperation(x, y))
+    let lhsGrad = seed * mask
+    let rhsGrad = seed - lhsGrad
     let (lhsShape, rhsShape) = (x.shapeTensor, y.shapeTensor)
     let (lhsAxes, rhsAxes) = Raw.broadcastGradientArgs(s0: lhsShape, s1: rhsShape)
     return (lhsGrad.sum(squeezingAxes: lhsAxes).reshaped(toShape: lhsShape),
```

Tests/TensorFlowTests/TensorAutoDiffTests.swift
jon-tow: I did the following in colab:
```
with tf.GradientTape() as t:
  t.watch([a, b])
  y = tf.math.reduce_sum(tf.minimum(a, b))
grads = t.gradient(y, [a, b])
```
which reproduced the discrepancies in the `swift-api` version which averages the ties.

Sources/TensorFlow/Operators/Math.swift
jon-tow: I couldn't get your solution working under the tests I've written. I'll look into it more.
My idea was just to multiply out the ties from the `rhsGrad` with the mask which doesn't seem efficient enough now that you point it out.

jon-tow: @dan-zheng Ah I got it :smile:! I had misread a piece. Thank you!

jon-tow:
@dan-zheng @rxwei This should be ready.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1344,7 +1344,8 @@ internal func _vjpMax<T: TensorFlowFloatingPoint>(
     _ y: Tensor<T>
 ) -> (Tensor<T>, (Tensor<T>) -> (Tensor<T>, Tensor<T>)) {
     let value = max(x, y)
-    return (value, { v in _vjpMinMaxHelper(x, y, originalValue: value, seed: v) })
+    return (value,
```
dan-zheng: Formatting nit - how about:
```swift
    return (value, { v in
      _vjpMinMaxHelper(x, y, originalValue: value, seed: v, comparisonOperation: .>=)
    })
```

Similar formatting below too, please.

jon-tow: Sure thing. Done!

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1397,11 +1401,12 @@ internal func _vjpMinMaxHelper<T: TensorFlowFloatingPoint>(
...
     _ y: Tensor<T>,
     originalValue: Tensor<T>,
-    seed: Tensor<T>
+    seed: Tensor<T>,
+    comparisonOperation: (Tensor<T>, Tensor<T>) -> Tensor<Bool>
 ) -> (Tensor<T>, Tensor<T>) {
-    let denominator = 1 + Tensor<T>(x .== y)
-    let lhsGrad = seed * Tensor<T>(x .== originalValue) / denominator
-    let rhsGrad = seed * Tensor<T>(y .== originalValue) / denominator
+    let mask: Tensor<T> = Raw.cast(comparisonOperation(x, y))
```
rxwei: Tensor has a casting initializer.
```suggestion
    let mask: Tensor<T>(comparisonOperation(x, y))
```

dan-zheng:
There appears to be one failure:
```
Test Case 'LossTests.testSigmoidCrossEntropyGrad' started at 2019-08-24 07:41:17.269
/swift-apis/Tests/TensorFlowTests/LossTests.swift:219: error: LossTests.testSigmoidCrossEntropyGrad : XCTAssertEqual failed: ("0.125") is not equal to ("0.0625") +/- ("1e-06") -
```

It's likely because `sigmoidCrossEntropy` internally calls `max`:
```swift
@differentiable(wrt: logits)
    logits: Tensor<Scalar>,
    labels: Tensor<Scalar>,
    reduction: @differentiable (Tensor<Scalar>) -> Tensor<Scalar> = _mean
) -> Tensor<Scalar> {
    let maxLogitsWithZero = max(logits, Tensor(0)) // `max` called here
    let result = log(1 + exp(-abs(logits)))
    return reduction(maxLogitsWithZero - logits * labels + result)
}
```

Could you please fix? Updating the expected value in the test should be good - validating the expected value against [`tf.losses.sigmoid_cross_entropy`](https://www.tensorflow.org/api_docs/python/tf/losses/sigmoid_cross_entropy) would be great (I'm not sure how [`expectedGradientsBeforeMean`](https://github.com/tensorflow/swift-apis/blob/2c3d3f677689ff3b5f08b6932c4ec515f02d8a4e/Tests/TensorFlowTests/LossTests.swift#L209) was computed).

jon-tow:
`tf.losses.sigmoid_cross_entropy` is computed via [`tf.python.ops.nn_impl.sigmoid_cross_entropy_with_logits`.](https://github.com/tensorflow/tensorflow/blob/00fad90125b18b80fe054de1055770cfb8fe4ba3/tensorflow/python/ops/nn_impl.py#L109) which is a bit different than the swift-api version in that it uses custom `max` and `abs` in order to compute gradients at `0`. I'm still trying to find my way around this but will update as soon as possible.

Sources/TensorFlow/Loss.swift
```diff
@@ -287,8 +287,6 @@ public func sigmoidCrossEntropy<Scalar: TensorFlowFloatingPoint>(
 ) -> Tensor<Scalar> {
     // This numerically stable implementation is based on the TensorFlow Python API.
     let maxLogitsWithZero = max(logits, Tensor(0))
-    // Note: `result` is split into two lines to avoid the "compiler is unable to type-check this
-    // expression in reasonable time" error.
-    let result = log(1 + exp(-abs(logits)))
-    return reduction(maxLogitsWithZero - logits * labels + result)
+    let negAbsLogits = max(logits, -logits) // Custom `abs` to compute gradients at `0`.
```
dan-zheng: Could you please clarify why `max(logits, -logits)` works while `abs(logits)` does not? Is the derivative of `abs` inconsistent with Python TensorFlow?

The old code seems to have been derived directly from the [doc comment for `tf.losses.sigmoid_cross_entropy_with_logits`](https://github.com/tensorflow/tensorflow/blob/00fad90125b18b80fe054de1055770cfb8fe4ba3/tensorflow/python/ops/nn_impl.py#L136):
```
  Hence, to ensure stability and avoid overflow, the implementation uses this
  equivalent formulation
      max(x, 0) - x * z + log(1 + exp(-abs(x)))
```

jon-tow: Well both work. It's just a matter of how we approach the evaluation of gradients on `0` entries of the `logits`. Recall that `abs` and `max` are both non-differentiable at that point. Now, in our case there happens to be a discrepancy in the gradient responses of `abs` and `max(x, -x)` for reasons I've yet to see. It could just be how `max` breaks ties. Note that this response is also happening in Python TF (and PyTorch if that matters). 
I chose to use `max(logits, -logits)` because it matches the Python TF implementation of `tf.losses.sigmoidCrossEntropy`. Using `abs` works just as well but the gradients through any `0` entries of `logits` are going to be different. If you'd like, I'd gladly change over to using `abs` as it seems to keep more `0` entries at `0` which could, ever so slightly, help loss convergence.

On another note, I ran a few gradient tests on `abs` and did not find inconsistencies between the Python TF and swift-api versions. All is well on that end.

eaplatanios: I believe I know what the reason for this is. In this PR, both min and max gradients check for `<=` and `>=` when computing gradients, which means that at 0 they set the gradient to `1`. `abs` uses `sign` which sets it to  0. From a subgradient optimization point of view, I believe `0` will behave better and I would thus rather use `abs`. We should probably change the gradients of `min` and `max` to do the same though, which can be easily done by changing `.<=` and `.>=` to `.>` and `.<` when calling `_vjpMinMaxHelper`. What do you think?

jon-tow: @eaplatanios That's exactly it! I also prefer the behavior of `abs` but wasn't sure about breaking parity with the TF implementation. 
I'm +1 on taking out the lower and upper bounds in the grad helper. I'll make the above changes and test things out. Thanks for shedding some light on this. :).

eaplatanios: That's great, thanks! :) Regarding breaking parity with the TF implementation, I believe that consistency in the behavior of `abs(x)` and `max(x, -x)` is more important and would thus also lean towards ignoring parity with the Python API in this case. @dan-zheng @rxwei what do you guys think?

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1397,11 +1401,12 @@ internal func _vjpMinMaxHelper<T: TensorFlowFloatingPoint>(
...
-    seed: Tensor<T>
+    seed: Tensor<T>,
+    comparisonOperation: (Tensor<T>, Tensor<T>) -> Tensor<Bool>
 ) -> (Tensor<T>, Tensor<T>) {
-    let denominator = 1 + Tensor<T>(x .== y)
-    let lhsGrad = seed * Tensor<T>(x .== originalValue) / denominator
-    let rhsGrad = seed * Tensor<T>(y .== originalValue) / denominator
+    let mask = Tensor<T>(comparisonOperation(x, y))
+    let lhsGrad = seed * mask
+    let rhsGrad = seed - lhsGrad
```
eaplatanios: It may be better to use `seed * (1 - mask)` here from a numerical precision point of view.

Sources/TensorFlow/Loss.swift
dan-zheng: > Regarding breaking parity with the TF implementation, I believe that consistency in the behavior of abs(x) and max(x, -x) is more important and would thus also lean towards ignoring parity with the Python API in this case.

Unfortunately, I believe maintaining parity with Python TensorFlow gradient calculations is important. We want to the ability to check that gradient calculations exactly match Python TensorFlow.

Sources/TensorFlow/Operators/Math.swift
dan-zheng: Could we please try to match TensorFlow gradient implementations exactly (see [here](https://github.com/tensorflow/swift-apis/pull/480#discussion_r317415394))?

I believe that involves using `let rhsGrad = seed - lhsGrad`: Python TensorFlow reference implementation [here](https://github.com/tensorflow/tensorflow/blob/dac733509fbe366804656a71f15ba0712bb5e69d/tensorflow/core/ops/math_grad.cc#L614).

eaplatanios: @dan-zheng I believe the code you're referring to is not that of the Python API gradients, but rather the C++ API gradients. The Python API gradients are defined [here](https://github.com/tensorflow/tensorflow/blob/752556174a1dcf68693784a8c78f07ba818e230f/tensorflow/python/ops/math_grad.py#L213). I initially implemented the gradients of the min and max to match that implementation, before this PR. @rxwei which implementation were you referring to in #479?

In general, it seems that there are multiple versions of these gradients in the various TF APIs. All are correct given that min/max have a non-differentiable point. All we care about is for a subgradient and all these implementations provide valid subgradients and should all pretty much result in the same behavior as it's highly unlikely a value would equal the non-differentiable point (unless intentionally done so).

I understand that we're using the Python API to generally verify gradients so we should probably stick to what's done there so differences don't accumulate and make it. hard to verify other gradients. Could you please confirm that [this](https://github.com/tensorflow/tensorflow/blob/752556174a1dcf68693784a8c78f07ba818e230f/tensorflow/python/ops/math_grad.py#L213) is the Python API implementation? If so, @rxwei could you please point to the issue related to #479?

Sources/TensorFlow/Loss.swift
eaplatanios: Let's just go with the Python version since all implementations are technically correct. The inconsistency pointed out should not really matter in practice. However, please check my other response for some clarification as to which one is the correct Python implementation. :)

Sources/TensorFlow/Operators/Math.swift
dan-zheng: I left [a comment](https://github.com/tensorflow/swift-apis/issues/479#issuecomment-524669755) on #479 describing the exact gradient calculation mismatch for `tf.minimum`/`tf.maximum`.

> @dan-zheng I believe the code you're referring to is not that of the Python API gradients, but rather the C++ API gradients. The Python API gradients are defined [here](https://github.com/tensorflow/tensorflow/blob/752556174a1dcf68693784a8c78f07ba818e230f/tensorflow/python/ops/math_grad.py#L213).

Thanks for pointing out the distinction between Python and C++ gradient implementations! Do you happen to know when each implementation is used?

(It seems likely to me that gradients registered in Python via `@ops.RegisterGradient` are used to compute gradients in Python - that makes me wonder when the C++ gradient implementations are used.)

---

You linked to `_MinOrMaxGrad`, but I think we actually want to match [`_MaximumMinimumGrad`](https://github.com/tensorflow/tensorflow/blob/752556174a1dcf68693784a8c78f07ba818e230f/tensorflow/python/ops/math_grad.py#L1417) (which seems similar but perhaps subtly different from the current Swift implementation in this PR). The bottom line is the gradient calculation mismatch should be fixed 🙂

jon-tow: For what it's worth, I could not produce gradients equivalent to the Python TF 2.0-Beta implementation when using the previous implementation.

eaplatanios: @dan-zheng you're right sorry. In that case, we should switch to `seed * (1 - mask)` to match that behavior. Thanks for helping clarify this! Regarding the Python vs C++ gradients, I had asked that same question a couple years back and I think the answer was that C++ gradients are only used by the C++ API for TensorFlow (i.e., not the Python API or the runtime). The Python API gradients are all registered in Python and the only way they can have C++ implementations is through custom ops and corresponding kernels. @alextp may be able to verify this.

Tests/TensorFlowTests/TensorAutoDiffTests.swift
```diff
@@ -153,6 +153,68 @@ final class TensorAutoDiffTests: XCTestCase {
...
+        // The expected gradient values were computed using the following TensorFlow 2.0 Beta1
+        // Python code with respective `a` and `b` tensors :
+        // ```
+        // with tf.GradientTape() as t:
+        //     t.watch([a, b])
+        //     y = tf.math.reduce_sum(tf.minimum(a, b))
+        // print(t.gradient(y, [a, b]))
+        // ```
+        func f(a: Tensor<Float>, b: Tensor<Float>) -> Tensor<Float> { min(a, b).sum() }
+        var a = Tensor<Float>([4, 5, 3])
```
dan-zheng: Could you please use `let` variables instead of `var`?
You can use `do { ... }` blocks to avoid variable redeclaration issues.
```swift
do {
    let a = Tensor<Float>([4, 5, 3])
    ...
}

do {
    let a = Tensor<Float>([[3.0, -2.0], [0.3, 10.0]])
    ...
}
```

dan-zheng: Done in https://github.com/tensorflow/swift-apis/pull/480/commits/7eba493331c07f204d487e4869af137516d496b0.

jon-tow:
No problem @dan-zheng. Thanks for the guidance!

PR #484, opened 2019-08-24
Remove redundant testing of optimizers

Shashi456:
We already test all of the optimizers in `SequentialTests.swift` in the same fashion i.e XOR gate classifier.
Having this file is mostly redundant testing, and is incomplete as well since it only has a test for Adagrad.


eaplatanios:
@Shashi456 can you also remove [this line](https://github.com/tensorflow/swift-apis/blob/2c3d3f677689ff3b5f08b6932c4ec515f02d8a4e/Tests/TensorFlowTests/XCTestManifests.swift#L36) please?

Shashi456:
@eaplatanios done

Shashi456:
Can someone add a test to this?

PR #485, opened 2019-08-26
Fix unused variable warning.

dan-zheng:
Note that the call to `BatchNorm.callAsFunction` cannot be removed because it updates running mean and variance.


Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -919,12 +919,13 @@ final class LayerTests: XCTestCase {
         let x = Tensor<Float>(rangeFrom: 0, to: 20, stride: 1).reshaped(to: [4,5])
         let epsilon = Tensor<Float>(0.001)
         let bnLayer = BatchNorm<Float>(featureCount: 5, axis: 1, epsilon: epsilon)
-        // Test inferrence before any training is only changed by epsilon value.
+        // Test inference before any training.
         assertEqual(bnLayer.inferring(from: x), x / TensorFlow.sqrt(1 + epsilon), accuracy: 1e-5)
-        // Test inferrence after single training step.
+        // Perform one training step, updating running mean and variance.
```
eaplatanios:
```suggestion
        // Perform one training step, updating the running mean and variance.
```

PR #492, opened 2019-08-28
Add ZeroPadding*D gradient tests

t-ae:
Related issue: #402
Add gradient tests for `ZeroPadding1D`, `ZeroPadding2D`, `ZeroPadding3D`.

### DIFF
--- a/Tests/TensorFlowTests/LayerTests.swift
+++ b/Tests/TensorFlowTests/LayerTests.swift
@@ -255,9 +255,9 @@ final class LayerTests: XCTestCase {
     func testZeroPadding1DGradient() {
         let input = Tensor<Float>([0.0, 1.0, 2.0])
         let layer = ZeroPadding1D<Float>(padding: 2)
-        let computedGradient = gradient(at: input) { layer($0).sum() }
+        let computedGradient = gradient(at: input, layer) { $1($0).sum() }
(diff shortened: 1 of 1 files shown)


Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -251,6 +251,14 @@ final class LayerTests: XCTestCase {
         let expected = Tensor<Float>([0.0, 0.0, 0.0, 1.0, 2.0, 0.0, 0.0])
         XCTAssertEqual(output, expected)
     }
+    
+    func testZeroPadding1DGradient() {
+        let input = Tensor<Float>([0.0, 1.0, 2.0])
+        let layer = ZeroPadding1D<Float>(padding: 2)
+        let computedGradient = gradient(at: input) { layer($0).sum() }
```
Shashi456: You'll need to add the TensorFlow python equivalent.

Also just a question, you don't need use `computedGradient.0` in the test because you take the gradient only at the input? while the other tests do so because the compute the gradient in the `gradient(at: input, layer` fashion.

t-ae: Updated fashion.
I'll add equivalent code after get answer for #494.

t-ae:
Move to #498

PR #493, opened 2019-08-28
Add gradient test for Embedding

t-ae:
Related issue: #402
Add gradient test for `Embedding`.

### DIFF
--- a/Tests/TensorFlowTests/LayerTests.swift
+++ b/Tests/TensorFlowTests/LayerTests.swift
@@ -758,27 +758,37 @@ final class LayerTests: XCTestCase {
     }
     
     func testEmbeddingGradient() {
-        let layer = Embedding<Float>(vocabularySize: 4, embeddingSize: 5)
+        let embeddings = Tensor<Float>([
+            [0.0, 0.2, 0.1],
+            [0.1, 0.7, 0.5],
+            [0.2, 0.4, 0.6],
+            [0.3, 0.2, 0.3]])
+        let layer = Embedding<Float>(embeddings: embeddings)
         let input = Tensor<Int32>(shape: [2, 3], scalars: [0, 1, 2, 1, 2, 2])
         let grad = gradient(at: layer) { $0(input).sum() }
         // The expected value of the gradient was computed using the following Python code:
         // ```
-        // indices = tf.convert_to_tensor([[0, 1, 2], [1, 2, 2]], dtype=tf.int32)
-        // embedding = tf.random.normal([4, 5])
+        // indices = tf.constant([0, 1, 2, 1, 2, 2], dtype=tf.int32)
+        // embeddings = tf.constant([
+        //      [0.0, 0.2, 0.1],
+        //      [0.1, 0.7, 0.5],
+        //      [0.2, 0.4, 0.6],
+        //      [0.3, 0.2, 0.3]])
+        // layer = tf.keras.layers.Embedding(4, 3, weights=[embeddings])
         // with tf.GradientTape() as t:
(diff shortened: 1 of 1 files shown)


Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -756,6 +756,19 @@ final class LayerTests: XCTestCase {
         let expected = Tensor<Float>([[[0.4, 0.3], [0.2, 0.1]], [[0.2, 0.1],[0.2, 0.1]]])
         XCTAssertEqual(output, expected)
     }
+    
+    func testEmbeddingGradient() {
+        let layer = Embedding<Float>(vocabularySize: 4, embeddingSize: 5)
```
Shashi456: Add a Tensorflow Python equivalent, afaik our functionalities do match.

t-ae: Added equivalent code.
Keras has `Embedding` layer but it takes one-hot vectors as input.
So I used `tf.gather` instead.

Shashi456: Should we also update our implementation to support one hot o.O?

t-ae: @jon-tow 
Thank you for clarification. Taking one-hot vectors was my misunderstanding.

I found that `gather` and `Embedding` give different gradients even though these outputs are same.
`swift-apis`'s `Embedding` is `gather`-based, so it doesn't match to `tf.keras.layer.Embedding`.
I'll look into it.

```python
def gather():
    indices = tf.convert_to_tensor([0, 1, 2, 1, 2, 2], dtype=tf.int32)
    embedding = tf.convert_to_tensor([
        [0.4, 0.2, 0.1, 0.0, 0.1],
        [0.9, 0.7, 0.5, 0.2, 0.1],
        [0.2, 0.4, 0.6, 0.3, 0.2],
        [0.1, 0.3, 0.8, 0.3, 0.3]])
    with tf.GradientTape() as t:
        t.watch(embedding)
        output = tf.gather(embedding, indices)
        y = tf.reduce_sum(output)
    grad_slice = t.gradient(y, embedding) # IndexedSlice
    grad = tf.zeros_like(embedding).numpy()
    for index in grad_slice.indices:
        grad[index] += grad_slice.values[index].numpy()
    print(output)
    print(grad)

def embedding():
    indices = tf.convert_to_tensor([0, 1, 2, 1, 2, 2], dtype=tf.int32)
    embeddings = tf.constant([
        [0.4, 0.2, 0.1, 0.0, 0.1],
        [0.9, 0.7, 0.5, 0.2, 0.1],
        [0.2, 0.4, 0.6, 0.3, 0.2],
        [0.1, 0.3, 0.8, 0.3, 0.3]])
    layer = tf.keras.layers.Embedding(4, 5, weights=[embeddings])

    with tf.GradientTape() as t:
        t.watch(layer.weights)
        output = layer(indices)
        y = tf.reduce_mean(output)
    grad_slice = t.gradient(y, layer.weights)[0] # IndexedSlice
    grad = tf.zeros_like(embeddings).numpy()
    for index in grad_slice.indices:
        grad[index] += grad_slice.values[index].numpy()
    print(output)
    print(grad)

gather()
embedding()
```

t-ae: @jon-tow 
Thank you for your clarification.
I updated python code to use `tf.keras.layer.Embedding`.

rxwei:
How about applying #499 going forward?

t-ae:
Updated.

PR #496, opened 2019-08-28
Use docker container with swift dependencies installed.

bgogul:
This should speed up the kokoro CI a bit.


bgogul:
Doesn't make much of a difference in runtime. Closing this PR to keep things simple.

PR #497, opened 2019-08-29
Remove deprecated APIs for 0.5

### LINKED ISSUE #495: Remove deprecated APIs for 0.5.
The Swift for TensorFlow team is preparing a 0.5 release of Swift for TensorFlow. When we transitioned from 0.3 to 0.4, we deprecated some APIs moved them to [BackwardsCompatibility.swift](https://github.com/tensorflow/swift-apis/blob/master/Sources/TensorFlow/BackwardsCompatibility.swift), e.g. `Conv2D.init(filterShape:strides:padding:dilations:activation:seed:)`.

**Task**: Remove APIs in [BackwardsCompatibility.swift](https://github.com/tensorflow/swift-apis/blob/master/Sources/TensorFlow/BackwardsCompatibility.swift) that are marked to be removed after 0.4.

jon-tow:
Resolves #495.


rxwei:
Looks like we need to update swift-models.

Shashi456:
@rxwei but this error shouldn't really pop up here 
```
/swift-models/Catch/main.swift:49:28: error: unable to infer complex closure return type; add explicit type to disambiguate
    var model = Sequential {
                           ^
```

jon-tow:
@Shashi456 It's a sequential through the deprecated the `Dense(..:generator:)` initializer. Not too familiar with the file and currently don't understand the context of it. I'll look into it.

rxwei:
That initializer is obsolete, so we should switch to using `withRandomSeedForTensorFlow(...) {}`.

Shashi456:
@rxwei is there a way to have global seeds? because the locality of `withRandomSeedForTensorflow` makes `model` inaccessible to other functions in its extension.

rxwei:
You don't actually need a seed. Just drop those random seed arguments since they have a default argument.

Shashi456:
@rxwei alright, putting a PR now.

eaplatanios:
> @rxwei is there a way to have global seeds? because the locality of `withRandomSeedForTensorflow` makes `model` inaccessible to other functions in its extension.

In general though if you want to do that one way is to return `model` from within the `withRandomSeedForTensorflow` closure.

Issue #499, opened 2019-08-29
Guidelines for Python code examples in derivative tests.


rxwei:
A good practice has been exercised over the past few months in adding layer tests: we add a Python TensorFlow reproducer. 

What has really helped in assuring our layers' correctness is adding a Python reproducer per layer test like ([this](https://github.com/tensorflow/swift-apis/blob/9616543321a279ebb550fbfdd37fb10d63842f25/Tests/TensorFlowTests/LayerTests.swift#L310)). Thank you @eaplatanios, @Shashi456, @jon-tow and @t-ae for doing this! As we make more progress towards finishing #402, we need to make sure all the reproducers we add are high-quality (@dan-zheng also mentioned this to me earlier). I propose that we follow the following guidelines in future Python code examples:

1. Use TensorFlow 2.0 and tf.keras.
2. Include **all code** necessary for reproducing the gradients. This includes any `import` statements.
3. Leave no blank lines.
4. Call `print()` on the gradient, so that reproducing is as easy as copy-paste.
5. Zero base indentation. Currently I'm seeing:
    ```swift
    // ```
    //   maxpool2D = tf.keras.layers.MaxPool2D(strides=(2, 2))
    //   with tf.GradientTape() as t:
    ```
    ... which should be changed to:
    ```swift
    // ```
    // import tensorflow as tf
    // maxpool2D = tf.keras.layers.MaxPool2D(strides=(2, 2))
    // with tf.GradientTape() as t:
    ```
6. Add a sentence above the code block:
    ```
    // The expected value of the gradient was computed using the following Python code:
    ```

Shashi456:
What about adding the variable used in the code as well? We don't exactly add x/input to the reproducers.

rxwei:
Yes, good point. Updated.

PR #500, opened 2019-08-29
Standardization of layer gradient examples

Shashi456:
#499


Shashi456:
Except for convolution and batchnorm, all other layers are done using tf.keras.

We approach the initialization of convolution layers differently, while the functionality remains at par.

dan-zheng:
This is huge, thanks @Shashi456! 🙂

PR #502, opened 2019-09-01
Add depthwiseConv2D gradient test and some vjp fixes

Shashi456:
Built and tested locally and ran swift-format

### DIFF
--- a/Sources/TensorFlow/Operators/NN.swift
+++ b/Sources/TensorFlow/Operators/NN.swift
@@ -329,7 +329,7 @@ public func depthwiseConv2D<Scalar: TensorFlowFloatingPoint>(
     filter: Tensor<Scalar>,
     strides: (Int, Int, Int, Int),
     padding: Padding
-    ) -> Tensor<Scalar> {
+) -> Tensor<Scalar> {
     return Raw.depthwiseConv2dNative(
         input,
         filter: filter,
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/Operators/NN.swift
```diff
@@ -343,18 +343,14 @@ func _vjpDepthwiseConv2D<Scalar: TensorFlowFloatingPoint>(
     filter: Tensor<Scalar>,
     strides: (Int, Int, Int, Int),
     padding: Padding
-) -> (Tensor<Scalar>, (Tensor<Scalar>) -> (Tensor<Scalar>, Tensor<Scalar>)) {
+    ) -> (Tensor<Scalar>, (Tensor<Scalar>) -> (Tensor<Scalar>, Tensor<Scalar>)) {
```
Shashi456: This is a swift-format change @rxwei, Should i keep this or revert it? Because if it's to be kept, we should probably standardize this for the other functions as well.

rxwei: This does not seem like what swift-format would do. How did you get the swift-format? Could you build and run this [swift-format](https://github.com/apple/swift-format)?

Shashi456: so when I run the command it's usually in the `swift-format.exe <file>` way.
I'll run the other swift-format if it's buildable on WSL.

Shashi456:
@rxwei i'll take a look at swift-format later, could you put a build on this? The separable convolution gradients depend on the depthwise function.

Shashi456:
Looks like the build didn't get triggered.

Shashi456:
@rxwei sorry for pinging you again, could you put a force build this didn't build the first time.

Shashi456:
@rxwei done.

rxwei:
> I did run swift-format on this file, this is just plain weird xD.

It's always good to proof-read the diff before or after sending a PR.

Issue #518, opened 2019-09-29
RNN gradients are wrong


eaplatanios:
[This line](https://github.com/tensorflow/swift-apis/blob/27ea161a69ed207c26028286ce751888c6f9a12d/Sources/TensorFlow/Layers/Recurrent.swift#L323) should be `𝛁cell += new𝛁cell` instead. There is also an issue with the initial state initialization to a zeros tensor with batch size 1 that does not broadcast. I don't have time to open a PR right now because I'm traveling but will try to open one once I get a chance.

rxwei:
Nice catch!

Shashi456:
Should I take care of this @eaplatanios

eaplatanios:
@Shashi456 thanks! Feel free to fix the `+=` bug. The initial state issue is a less trivial API design question that I haven't thought about much yet. There are a couple of possible solutions that I used in swift-rl, so I'll try to look into it sometime soon.

sgugger:
I think this has been solved by the two PRs merged. Feel free to reopen if there is still a problem.

Shashi456:
@sgugger the tests against these changes were remaining so I thought we'd keep this PR open. #555, #554 track the tests. But since we have separate issues for them maybe it's okay.

sgugger:
Like you said there are already two issues tracking this. I wanted to do a bit of clean-up, but if anyone feels strongly about this issue remaining open, they can click the button :)

PR #519, opened 2019-10-01
Fix RNN gradient accumulation.

Shashi456:
partial fix for #518.

### DIFF
--- a/Sources/TensorFlow/Layers/Recurrent.swift
+++ b/Sources/TensorFlow/Layers/Recurrent.swift
@@ -48,7 +48,7 @@ extension RNNCellOutput: EuclideanDifferentiable
 
 /// A recurrent neural network cell.
 public protocol RNNCell: Layer
-    where Input == RNNCellInput<TimeStepInput, State>, 
+    where Input == RNNCellInput<TimeStepInput, State>,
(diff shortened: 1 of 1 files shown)


rxwei:
Could you please add a unit test?

Shashi456:
@rxwei how would a unit test for gradient accumulation look like?

rxwei:
Just add a unit test for `RNN`.

eaplatanios:
You can check whether the gradient of the update matrix is non-zero for
time steps before the last one I believe. The bug should make these
gradients zero and this should be fixed now.

rxwei:
@Shashi456 Did you have a chance to update this PR?

Shashi456:
@rxwei sorry for the delay, I'll write the rnn test in a day or 2.

saeta:
Hi @Shashi456 ! Just following up here; do you think you can add a test or two? Thanks so much for this PR! -Brennan

Shashi456:
@saeta, when we say unit test we mean an rnn gradient test? Sorry for the confusion

marcrasi:
Hi, we'll merge this now to unblock #522. We're still interested in having a test, so I'll create a github issue for that and we can discuss details about the test there.

PR #520, opened 2019-10-01
Deprecate Tensor.droppingOut

apaszke-g:
It's public, but it doesn't check the learning phase, which might lead to unexpected behavior!

### DIFF
--- a/Sources/TensorFlow/Layers/Core.swift
+++ b/Sources/TensorFlow/Layers/Core.swift
@@ -12,14 +12,26 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-public extension Tensor where Scalar: TensorFlowFloatingPoint {
+fileprivate extension Tensor where Scalar: TensorFlowFloatingPoint {
     /// Computes dropout given a probability.
     @differentiable(wrt: self where Scalar: Differentiable)
-    func droppingOut(probability: Double) -> Tensor {
+    fileprivate func _droppingOut(probability: Double) -> Tensor {
         let noise = Tensor(randomUniform: shape)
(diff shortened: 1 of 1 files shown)


Sources/TensorFlow/Layers/Core.swift
```diff
@@ -12,17 +12,29 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-public extension Tensor where Scalar: TensorFlowFloatingPoint {
+fileprivate extension Tensor where Scalar: TensorFlowFloatingPoint {
     /// Computes dropout given a probability.
     @differentiable(wrt: self where Scalar: Differentiable)
-    func droppingOut(probability: Double) -> Tensor {
+    fileprivate func _droppingOut(probability: Double) -> Tensor {
```
rxwei: Since the `extension` has been marked as `fileprivate`, access modifiers on individual method declarations can be dropped.
```suggestion
    func _droppingOut(probability: Double) -> Tensor {
```

Sources/TensorFlow/Layers/Core.swift
```diff
@@ -12,17 +12,29 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-public extension Tensor where Scalar: TensorFlowFloatingPoint {
+fileprivate extension Tensor where Scalar: TensorFlowFloatingPoint {
     /// Computes dropout given a probability.
```
rxwei:
```suggestion
    /// Computes dropout given a probability.
    // TODO: Remove the underscore after `droppingOut(probability:)` has been deprecated.
```

PR #521, opened 2019-10-02
Added back the 'Tensor.init(stacking:alongAxis:)' test.

eaplatanios:


eaplatanios:
I believe this failure may be related to [TF-681](https://bugs.swift.org/projects/TF/issues/TF-681?filter=allopenissues) actually.

PR #522, opened 2019-10-02
Fixed a couple RNN bugs.

eaplatanios:
@rxwei @Shashi456 this tackles #518 and also includes the fix of #519. I haven't added a test yet, but will try to add one tomorrow. The fixed issues are:

- Backpropagated gradients for RNN cells were computed wrongly, resulting in being unable to train RNNs (especially in cases where only the last time step output is being used -- e.g., sequence classification)
- The default zero state for RNNs always had a batch size of 1 that could not be broadcast and caused failures (e.g., when concatenating the input and hidden state in LSTM cells). Now `zeroState` is a function that takes an example `input` as argument so that it can account for the batch size. In principle we can also switch it to just take the batch size as input, but in either case, until we can discuss the design of RNNs more extensively this solution suffices and fixes the current failures.


Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -299,10 +308,10 @@ public struct RNN<Cell: RNNCell>: Layer {
           (Array<Cell.TimeStepOutput>.TangentVector)
               -> (TangentVector, Array<Cell.TimeStepInput>.TangentVector)) {
         let timeStepCount = inputs.count
-        var currentHiddenState = cell.zeroState
-        var timeStepOutputs: [Cell.TimeStepOutput] = []
+        var currentHiddenState = cell.zeroState(for: inputs[0])
+        var timeStepOutputs = [Cell.TimeStepOutput]()
         timeStepOutputs.reserveCapacity(timeStepCount)
-        var backpropagators: [Cell.Backpropagator] = []
+        var backpropagators = [Cell.Backpropagator]()
```
rxwei: We actually prefer literal initialization for arrays in this code base. Could you revert it?

eaplatanios:
I reverted the refactoring changes as suggested.

Sources/TensorFlow/Layers/Recurrent.swift
eaplatanios: Sure, I wasn't aware. :)

eaplatanios:
There is only a couple variable renames I kept to make all names consistent across the `callAsFunction` and its corresponding VJP.

rxwei:
Let's wait on #519 and also add a unit test.

eaplatanios:
I haven't really found time to add a test for this PR yet, but I believe it should be merged anyway because it fixes a bug in the RNN gradients that will almost definitely cause undesired consequences for users, and add a TODO for a test. What do you think?

marcrasi:
That sounds good. I will merge and create an issue for adding a test.

Issue #517, opened 2019-09-29
Add preconditions


dan-zheng:
Preconditions (for shape-checking, valid value ranges, etc) would be valuable for better source locations in error messages and to avoid uninformative TensorFlow runtime errors.

Example: `matmul(_:_:)` currently crashes with an uninformative stack trace.
```swift
import TensorFlow
let x = Tensor([1, 1])
print(matmul(x, x)) // invalid
```
```
$ swift matmul.swift
Fatal error: In[0] is not a matrix. Instead it has shape [2]: file /usr/local/src/swift-build/tensorflow-swift-apis/Sources/TensorFlow/Bindings/EagerExecution.swift, line 300 # unuseful source location
Stack dump: ...
```

Shashi456:
Could we start with making a list of what methods need a precondition ?

dan-zheng:
> Could we start with making a list of what methods need a precondition ?

I'd recommend looking through files in `Sources/TensorFlow/Operators` and `Sources/TensorFlow/Layers` and look for "precondition" comments on properties/parameters that aren't actually checked.

Examples:
* `func conv2D`, [``/// - Precondition: `input` must have rank `4`.``](https://github.com/tensorflow/swift-apis/blob/27ea161a69ed207c26028286ce751888c6f9a12d/Sources/TensorFlow/Operators/NN.swift#L83)
  * Missing a precondition at top of function.
* `Conv1D.filter`, [`/// The 3-D convolution filter.`](https://github.com/tensorflow/swift-apis/blob/27ea161a69ed207c26028286ce751888c6f9a12d/Sources/TensorFlow/Layers/Convolutional.swift#L21)
  * Missing precondition in initializer.
* Other shape-related functions: `matmul`, `reshape`, etc
* Probability-related APIs (`Tensor.droppingOut`, `Dropout` layer) should have `0..<1` preconditions for probability values.

Shashi456:
@dan-zheng I'll try to put up a list in the next few hours

ocampor:
@dan-zheng I can create a runtime performance analysis on some of the functions with preconditions. Do you have any preferred methodology to assess the impact? Any recommendation to make the analysis more reliable?

Shashi456:
A list of preconditions :
- [ ] 
- [ ] 
- [ ] 
- [ ]

ocampor:
Today I’m going to spend some time looking the missing preconditions.

rickwierenga:
[...]

PR #523, opened 2019-10-02
Add assertions for conv2D

ocampor:
This is an example of a precondition for `conv2D` as described in #517 

[Here](https://gist.github.com/ocampor/adf9769fa4e6e5d43ea1146623758350) is a Gist with some quick code to measure the impact. I used assert instead of precondition as it is supposed not to affect performance in shipping code [here](https://developer.apple.com/documentation/swift/1541112-assert)

If this is satisfactory, I can start adding the rest of the preconditions mentioned in #517


Sources/TensorFlow/Operators/NN.swift
```diff
@@ -90,6 +90,8 @@ public func conv2D<Scalar: TensorFlowFloatingPoint>(
     padding: Padding = .valid,
     dilations: (Int, Int, Int, Int) = (1, 1, 1, 1)
 ) -> Tensor<Scalar> {
+    assert(input.shape.rank == 4, "input must have rank 4")
```
Shashi456: Is there a reason to use an `assert` instead of `precondition` ?
Also would it make sense to put the filter precondition in the initialization? Since each precondition adds some overhead, we might be reducing some by checking for filter at init and input at call.
WDYT?

rxwei: Please use `precondition` instead of `assert`.

ocampor: @rxwei perfect. I changed the assertions for preconditions. 
@Shashi456 maybe I am missing something, but as the function receives both of the parameters, you need to check the precondition every time or there is a risk of failing without error.

Sources/TensorFlow/Operators/NN.swift
```diff
@@ -90,6 +90,8 @@ public func conv2D<Scalar: TensorFlowFloatingPoint>(
     padding: Padding = .valid,
     dilations: (Int, Int, Int, Int) = (1, 1, 1, 1)
 ) -> Tensor<Scalar> {
+    precondition(input.shape.rank == 4, "input must have rank 4")
```
rxwei: Precondition and assertion messages should be a full sentence ([example](https://github.com/tensorflow/swift-apis/blob/6309c55e3acd71caa550c64b9d35c6c43fcaafbb/Sources/TensorFlow/Core/ShapedArray.swift#L607)) ending with a period.
```suggestion
    precondition(input.shape.rank == 4, "The input must have rank 4.")
```

Sources/TensorFlow/Operators/NN.swift
```diff
@@ -90,6 +90,8 @@ public func conv2D<Scalar: TensorFlowFloatingPoint>(
...
+    precondition(filter.shape.rank == 4, "filter must have rank 4")
```
```suggestion
    precondition(filter.shape.rank == 4, "The filter must have rank 4.")
```

ocampor:
@rxwei I included your comments and the rest of the preconditions for the file `NN.swift`. Should I continue adding the rest of the preconditions in the list, or would it be better to merge this one first, and then create another PR for the rest?

dan-zheng:
> @rxwei I included your comments and the rest of the preconditions for the file `NN.swift`. Should I continue adding the rest of the preconditions in the list, or would it be better to merge this one first, and then create another PR for the rest?

Sorry for the late response. The team is starting a weekly routine for reviewing `tensorflow/swift-apis` PRs, so we'll be more responsive!

Incrementally adding related preconditions sounds good to me. Using consistent style in all precondition messages would be great.

PR #525, opened 2019-10-02
Change BatchNorm momentum and epsilon to scalars.

dan-zheng:
Change `BatchNorm.momentum` and `BatchNorm.epsilon` to `Scalar` instead of `Tensor<Scalar>`. Semantically, these properties are scalars.

It is not clear why `BatchNorm` originally defined these properties as `Tensor<Scalar>` since the beginning: https://github.com/tensorflow/swift-apis/commit/6ca38130ab7cc60fd76072c3dc41795a41d770b1

### DIFF
--- a/Sources/TensorFlow/Initializers.swift
+++ b/Sources/TensorFlow/Initializers.swift
@@ -446,7 +446,7 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
         standardDeviation: Tensor<Scalar> = Tensor<Scalar>(1),
         seed: TensorFlowSeed = TensorFlow.Context.local.randomSeed
     ) {
-        let sample: Tensor<Scalar> = Raw.statelessTruncatedNormal(
+        let sample: Tensor<Scalar> = _Raw.statelessTruncatedNormal(
             shape: Tensor<Int32>((0..<shape.rank).map { Int32(shape[$0]) }),
             seed: Tensor<Int32>([seed.graph, seed.op]))
         self = standardDeviation * sample + mean
--- a/Sources/TensorFlow/Layers/Normalization.swift
+++ b/Sources/TensorFlow/Layers/Normalization.swift
@@ -25,13 +25,13 @@ public struct BatchNorm<Scalar: TensorFlowFloatingPoint>: Layer {
     /// The feature dimension.
     @noDerivative public let axis: Int
     /// The momentum for the running mean and running variance.
-    @noDerivative public let momentum: Tensor<Scalar>
+    @noDerivative public let momentum: Scalar
     /// The offset value, also known as beta.
     public var offset: Tensor<Scalar>
     /// The scale value, also known as gamma.
     public var scale: Tensor<Scalar>
     /// The variance epsilon value.
-    @noDerivative public let epsilon: Tensor<Scalar>
+    @noDerivative public let epsilon: Scalar
     /// The running mean.
     @noDerivative public let runningMean: Parameter<Scalar>
     /// The running variance.
@@ -49,10 +49,10 @@ public struct BatchNorm<Scalar: TensorFlowFloatingPoint>: Layer {
     ///   - runningVariance: The running variance.
     public init(
         axis: Int,
-        momentum: Tensor<Scalar>,
+        momentum: Scalar,
         offset: Tensor<Scalar>,
         scale: Tensor<Scalar>,
-        epsilon: Tensor<Scalar>,
+        epsilon: Scalar,
         runningMean: Tensor<Scalar>,
         runningVariance: Tensor<Scalar>
     ) {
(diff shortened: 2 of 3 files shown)


dan-zheng:
I will trigger a full Kokoro build to check if other repositories ([tensorflow/swift-models](https://github.com/tensorflow/swift-models), [fastai/swiftai](https://github.com/fastai/swiftai)) are broken.

dan-zheng:
> `Scalar` is a generic parameter. It will turn `BatchNorm` into an address-only type, which will be lowered as a SIL address type. In the meantime, the `BatchNorm.TangentVector` is loadable and will lowered to a SIL object type. This will break the current assumptions in the differentiation transform.

Can you recommend a crash reproducer? I tested a few but they didn't crash.

rxwei:
> It is not clear why `BatchNorm` originally defined these properties as `Tensor<Scalar>` since the beginning: [6ca3813](https://github.com/tensorflow/swift-apis/commit/6ca38130ab7cc60fd76072c3dc41795a41d770b1)

It was originally `Tensor<Scalar>` exactly because of the type mismatch issue. This will be fixed by [TF-625](https://bugs.swift.org/browse/TF-625).

rxwei:
I don't yet have a reproducer, but I'll look at swift-models and fast.ai notebooks test results. I'd be wary of turning this on just because no easy reproducer is found. It is technically pretty clear that lowering isn't able to handle such cases.

dan-zheng:
CI failed because tensorflow/swift-models requires changes:
```
/swift-models/MiniGo/Models/GoModel.swift:57:21: error: cannot invoke initializer for type 'BatchNorm<_>' with an argument list of type '(featureCount: Int, momentum: Tensor<Float>, epsilon: Tensor<Float>)'
        self.norm = BatchNorm(
                    ^
/swift-models/MiniGo/Models/GoModel.swift:57:21: note: expected an argument list of type '(featureCount: Int, axis: Int, momentum: Scalar, epsilon: Scalar)'
        self.norm = BatchNorm(
                    ^
```

Not pursuing further for now due to low-priority. https://github.com/apple/swift/pull/27298 will fix [TF-625](https://bugs.swift.org/browse/TF-625).

rxwei:
One possible reproducer at the SIL level is the following:
```swift
%f = function_ref @BatchNorm.callAsFunction(_:)
%f' = differentiable_function %f : $(*BatchNorm<Scalar>, Tensor<Scalar>) -> Tensor<Scalar>
%f_vjp = differentiable_function_extract [vjp] %f'
```

... after canonicalization:
```swift
%f = function_ref @BatchNorm.callAsFunction(_:)
%f' = differentiable_function %f : $(@in_guaranteed BatchNorm<Scalar>, @guaranteed Tensor<Scalar>) -> @owned Tensor<Scalar> with { %f_jvp : $(@in BatchNorm<Scalar>, @guaranteed Tensor<Scalar>) -> (@owned Tensor<Scalar>, (@guaranteed BatchNorm<Scalar>.TangentVector, @guaranteed Tensor<Scalar>) -> Tensor<Scalar>), %f_vjp : $(@in_guaranteed BatchNorm<Scalar>, @guaranteed Tensor<Scalar>) -> (@owned Tensor<Scalar>, (@guaranteed Tensor<Scalar>) -> (@owned BatchNorm<Scalar>.TangentVector, @owned Tensor<Scalar>) }
%f_vjp = differentiable_function_extract [vjp] %f'
```

According to the `differentiable_function_extract` type calculation rule, `%f_vjp` will have `$(@in_guaranteed BatchNorm<Scalar>, @guaranteed Tensor<Scalar>) -> (@owned Tensor<Scalar>, (@guaranteed Tensor<Scalar>) -> (@out BatchNorm<Scalar>.TangentVector, @owned Tensor<Scalar>)`, which doesn't match what's in the canonicalized `differentiable_function` instruction.

It's possible that the current transform happens to handle this case despite formal type mismatches. For example, the transform sometimes checks the type of the actual SIL values instead of formal types, which is very suboptimal.

If pullback-based tests are not failing, it's possible that a differential-based test will trigger this.

saeta:
Notes: 
 - Change LayerNorm as well?
 - Be sure to include this in the v0.6 release notes as API breaking.

dan-zheng:
`LayerNorm.epsilon` has also been updated.

---

Note: this will be an API breaking change in Swift for TensorFlow 0.6.

Deprecating the old `BatchNorm` and `LayerNorm` initializers is tricky because it causes ambiguity problems:

```
/Users/danielzheng/swift-apis/Tests/TensorFlowTests/LayerTests.swift:1171:23: error: ambiguous use of 'init(featureCount:axis:momentum:epsilon:)'
        let bnLayer = BatchNorm<Float>(featureCount: 5, axis: 0)
                      ^
TensorFlow.BatchNorm:12:12: note: found this candidate
    public init(featureCount: Int, axis: Int = -1, momentum: Scalar = 0.99, epsilon: Scalar = 0.001)
           ^
TensorFlow.BatchNorm:14:12: note: found this candidate
    public init(featureCount: Int, axis: Int = -1, momentum: TensorFlow.Tensor<Scalar> = Tensor(0.99), epsilon: TensorFlow.Tensor<Scalar> = Tensor(0.001))
           ^
```

PR #526, opened 2019-10-04
Refactor transposed(withPermutations:)

ocampor:
This is a proposal to refactor `transposed(withPermutations:)` to `transposed(permutation:)` according to the description in #524.

I ran all the tests and everything is working as expected. I didn't move `transposed(withPermutations:)` to the backward compatibility file because it depends on some internal methods. What do you suggest @rxwei ?

PR #527, opened 2019-10-05
Introduce renaming of method transpose(withPermutations:)

### LINKED ISSUE #524: Rename 'transposed(withPermutations:)' to 'transposed(permutation:)'.
`transposed(withPermutations:)` is not a perfect name because
1. "With" is often considered as a needless word, especially in an argument label where a noun phrase follows. We've been dropping "with" in cases like this for APIs added earlier this year, but we have not revisited `transposed(withPermutations:)` which was introduced since the beginning.
2. The argument is one permutation. So a plural doesn't make sense. Thanks @apaszke-g for pointing this out!

We should deprecate the current API and consider renaming it to `transposed(permutation:)`. A PR is welcome, but it should not be merged until it's discussed in a design review.

ocampor:
Following the recommendation of @rxwei in https://github.com/tensorflow/swift-apis/pull/526#pullrequestreview-297762689. This is a PR that introduces the renaming of method transposed(withPermutations:) to transposed(permutation:).

Resolves #524 .

### DIFF
--- a/Sources/TensorFlow/Operators/Basic.swift
+++ b/Sources/TensorFlow/Operators/Basic.swift
@@ -280,7 +280,7 @@ public extension Tensor {
     @inlinable
     @differentiable(
         wrt: self,
-        vjp: _vjpTransposed(withPermutations:) where Scalar: TensorFlowFloatingPoint)
+        vjp: _vjpTransposed(permutation:) where Scalar: TensorFlowFloatingPoint)
     func transposed(permutation: Tensor<Int32>) -> Tensor {
         Raw.transpose(self, perm: permutation)
     }
@@ -288,9 +288,7 @@ public extension Tensor {
     /// Returns a transposed tensor, with dimensions permuted in the specified order.
     @available(*, deprecated, renamed: "transposed(permutation:)")
     @inlinable
-    @differentiable(
-        wrt: self,
-        vjp: _vjpTransposed(withPermutations:) where Scalar: TensorFlowFloatingPoint)
+    @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
     func transposed(withPermutations permutations: Tensor<Int32>) -> Tensor {
         transposed(permutation: permutations)
     }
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -278,11 +278,21 @@ infix operator ++: AdditionPrecedence
...
+    func transposed(permutation: Tensor<Int32>) -> Tensor {
+        Raw.transpose(self, perm: permutation)
+    }
+
+    /// Returns a transposed tensor, with dimensions permuted in the specified order.
+    @available(*, deprecated, renamed: "transposed(permutation:)")
+    @inlinable
     @differentiable(
         wrt: self,
         vjp: _vjpTransposed(withPermutations:) where Scalar: TensorFlowFloatingPoint)
```
rxwei: Remove the `vjp: _vjpTransposed(withPermutations:)`  from the `@differentiable` attribute on the old method, since it can be differentiated through. Then the `@differentiable` attribute can be on a single line:

```swift
@differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
```

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -278,11 +278,21 @@ infix operator ++: AdditionPrecedence
 public extension Tensor {
     /// Returns a transposed tensor, with dimensions permuted in the specified order.
     @inlinable
+    @differentiable(
+        wrt: self,
+        vjp: _vjpTransposed(withPermutations:) where Scalar: TensorFlowFloatingPoint)
```
rxwei: Also, rename method `_vjpTransposed(withPermutations:)` to Also, rename `_vjpTransposed(permutations:)`.

```suggestion
        vjp: _vjpTransposed(permutations:) where Scalar: TensorFlowFloatingPoint)
```

Sources/TensorFlow/Operators/Basic.swift
ocampor: Very good.

Sources/TensorFlow/Operators/Basic.swift
ocampor: Would it be also a good idea to use the new `transposed(permutation:)` in methods that use the old one in the same file, to avoid warnings when compiling?

For example,  in `_vjpGathering` 

```swift
let transposedValues = values.transposed(withPermutations: permutations)
```

or overloading methods.

```swift
func transposed(withPermutations permutations: [Int]) -> Tensor {
        let permutations = permutations.map(Int32.init)
        return transposed(withPermutations: Tensor<Int32>(permutations))
}
```

rxwei: Yes.

ocampor: Thanks, all right. It is done.

rxwei:
@ocampor Could you bring this up in one of the design reviews?

ocampor:
@rxwei very happy to share it in one design review. I am just afraid that it is a little bit shallow, engineering-wise.

rxwei:
You could bring this (the suggested naming) up as a quick question near the end of a design review, and see if there's any concerns in the community :)

Shashi456:
discussed in the design review today, this PR is good to go.

PR #529, opened 2019-10-09
Added initial support for a TensorFlow checkpoint file reader.

eaplatanios:
This PR adds initial support for a TensorFlow checkpoint file reader. I have not added tests yet because: (i) I don't know what's a good way to include resource files for swift tests, and (ii) I don't have a readily available small checkpoint. This is preliminary work while I'm trying to design a more standardized way of saving/loading checkpoints. Merging this now would unblock me and I can make subsequent PRs once I have a more concrete solution for the checkpointing issue.

P.S. I'm currently working on some ideas using `Codable` and `CodingKeys` to add compatibility with existing Python TF checkpoints.


rxwei:
`init(owning:)` is considered an implementation detail, and it should be hidden whenever possible. Instead, have you considered writing the the checkpoint reader within the library?

eaplatanios:
I can do that. I haven’t tested things yet so I wasn’t sure if you’d like
it to be merged. I will try to open a PR tomorrow. :)

eaplatanios:
> `init(owning:)` is considered an implementation detail, and it should be hidden whenever possible. Instead, have you considered writing the the checkpoint reader within the library?

I changed this PR to add the checkpoint reader instead and updated the PR message.

eaplatanios:
> How do you load pretrained models o.O ?

Currently working in getting a hopefully nice solution. Will open another PR when I have one. :)

eaplatanios:
@rxwei on a separate note, at some point we should update the `checkOk` helper to throw an exception instead of a fatal error and mark all relevant functions that use it as `throws`.

eaplatanios:
@rxwei @dan-zheng gentle ping for a review.

Sources/TensorFlow/Core/Serialization.swift
```diff
@@ -0,0 +1,85 @@
...
+    @inlinable
+    public func dataType(ofTensorNamed name: String) -> TensorDataType {
+        TensorDataType(TF_CheckpointReaderGetVariableDataType(handle, name))
+    }
+
+    /// Loads and returns the value of the tensor with the provided name stored in the checkpoint.
+    @inlinable
+    public func load<Scalar: _TensorFlowDataTypeCompatible>(
+        tensorNamed name: String
+    ) -> ShapedArray<Scalar> {
```
dan-zheng: I'm curious if there's some reason why `load` returns `ShapedArray<Scalar>` rather than `Tensor<Scalar>`?

eaplatanios: This is just because it is stored in host memory anyway and would require an internal conversion to Tensor. In this case, I thought it’d be nice to let the user decide to convert if they want to. What do you think?

eaplatanios: @dan-zheng do you think we can get this merged soon? I believe it is useful irrespective of how we design high-level serialization APIs as it allows us to read checkpoint files written by the TF Python API.

dan-zheng: > This is just because it is stored in host memory anyway and would require an internal conversion to Tensor. In this case, I thought it’d be nice to let the user decide to convert if they want to. What do you think?

I see, thanks for explaining! I'm curious if you use `ShapedArray` much in your projects, in addition to `Tensor`?

eaplatanios:
Thanks @dan-zheng ! I'll post examples and tests once I settle on a design. I have been mostly experimenting with loading BERT checkpoints for now and thinking about how the loading mechanism can become more modular. By the way, I have been unable to build the S4TF toolchain on my Mac lately. The build script hangs while building SwiftPM. This is the command it runs and the stacktrace I get when I kill the process:
```
/Users/eaplatanios/Development/GitHub/swift-source-temp/build/buildbot_osx/swiftpm-macosx-x86_64/x86_64-apple-macosx/debug/swiftc --driver-mode=swift -L /Users/eaplatanios/Development/GitHub/swift-source-temp/build/buildbot_osx/swiftpm-macosx-x86_64/x86_64-apple-macosx/lib/swift/pm/4_2 -lPackageDescription -swift-version 4.2 -I /Users/eaplatanios/Development/GitHub/swift-source-temp/build/buildbot_osx/swiftpm-macosx-x86_64/x86_64-apple-macosx/lib/swift/pm/4_2 -target x86_64-apple-macosx10.10 -sdk /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk -package-description-version 4.2 /Users/eaplatanios/Development/GitHub/swift-source-temp/swiftpm/Package.swift -fileno 8

Traceback (most recent call last):
  File "/Users/eaplatanios/Development/GitHub/swift-source-temp/swiftpm/Utilities/bootstrap", line 1565, in <module>
    main()
  File "/Users/eaplatanios/Development/GitHub/swift-source-temp/swiftpm/Utilities/bootstrap", line 1334, in main
    result = subprocess.call(cmd, cwd=g_project_root)
  File "/usr/local/Cellar/python@2/2.7.16_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 172, in call
    return Popen(*popenargs, **kwargs).wait()
  File "/usr/local/Cellar/python@2/2.7.16_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 1099, in wait
    pid, sts = _eintr_retry_call(os.waitpid, self.pid, 0)
  File "/usr/local/Cellar/python@2/2.7.16_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 125, in _eintr_retry_call
    return func(*args)
```
I even let it run overnight and there was still a swift process fully consuming a single CPU core running all night. I'm using Catalina 10.15.1 Beta (19B77a) and Xcode 11.1 (11A1027). Do you have any idea what may be wrong?

Sources/TensorFlow/Core/Serialization.swift
eaplatanios: Not really. The APIs for `ShapedArray` are very limited currently and I also mostly use GPUs for compute so I use `Tensor`s almost always. I just like to keep `ShapedArray` in APIs that could be used in both ways, so that it's up to the user to convert to a `Tensor`.

eaplatanios:
@dan-zheng if I kill the background swift process itself, I get the following error:
```
/Users/eaplatanios/Development/GitHub/swift-source-temp/swiftpm: error: manifest parse error(s):
/Users/eaplatanios/Development/GitHub/swift-source-temp/build/buildbot_osx/swift-macosx-x86_64/lib/swift/macosx/Swift.swiftmodule/x86_64.swiftinterface:1252:16: error: referencing struct 'DifferentiableView' on 'Array' requires that 'Element' conform to '_Differentiable'
    if index < base.count {
               ^
/Users/eaplatanios/Development/GitHub/swift-source-temp/build/buildbot_osx/swift-macosx-x86_64/lib/swift/macosx/Swift.swiftmodule/x86_64.swiftinterface:1209:1: note: where 'Element' = 'Element'
extension Array where Element : Swift._Differentiable {
^
/Users/eaplatanios/Development/GitHub/swift-source-temp/build/buildbot_osx/swift-macosx-x86_64/lib/swift/macosx/Swift.swiftmodule/x86_64.swiftinterface:1253:14: error: referencing struct 'DifferentiableView' on 'Array' requires that 'Element' conform to '_Differentiable'
      return base[index]
             ^
/Users/eaplatanios/Development/GitHub/swift-source-temp/build/buildbot_osx/swift-macosx-x86_64/lib/swift/macosx/Swift.swiftmodule/x86_64.swiftinterface:1209:1: note: where 'Element' = 'Element'
extension Array where Element : Swift._Differentiable {
^
--- bootstrap: error: build failed with exit status 1
```
Maybe this is not the best way to discuss this, but I'm wondering if there is a quick fix. It's weird that there is an error related to array differentiability when compiling SwiftPM.

eaplatanios:
The swift interface file referenced does not include at all the `& Differentiable` constraint shown [here](https://github.com/apple/swift/blob/7498390f620ee3b9e754eed6ddba58f5f61bf714/stdlib/public/core/Array.swift#L2003).

Sources/TensorFlow/Core/Serialization.swift
```diff
@@ -0,0 +1,85 @@
...
+        self.checkpointPath = checkpointPath
+    }
+
+    deinit {
+        TF_DeleteCheckpointReader(handle)
+    }
+
+    /// Returns `true` if the checkpoint contains a tensor with the provided name.
+    @inlinable
+    public func contains(tensorNamed name: String) -> Bool {
```
rxwei: `containsTensor(named:)`

Sources/TensorFlow/Core/Serialization.swift
```diff
@@ -0,0 +1,85 @@
...
+        TF_CheckpointReaderHasTensor(handle, name) > 0
+    }
+
+    /// Returns the shape of the tensor with the provided name stored in the checkpoint.
+    @inlinable
+    public func shape(ofTensorNamed name: String) -> TensorShape {
```
rxwei: `shapeOfTensor(named:)`

Sources/TensorFlow/Core/Serialization.swift
```diff
@@ -0,0 +1,85 @@
...
+        defer { dimensions.deallocate() }
+        TF_CheckpointReaderGetVariableShape(handle, name, dimensions, rank, status)
+        checkOk(status)
+        let dimensionsBufferPointer = UnsafeBufferPointer(start: dimensions, count: Int(rank))
+        return TensorShape([Int64](dimensionsBufferPointer).map(Int.init))
+    }
+
+    /// Returns the data type of the tensor with the provided name stored in the checkpoint.
+    @inlinable
+    public func dataType(ofTensorNamed name: String) -> TensorDataType {
```
rxwei: `TensorDataType` is an internal type that was never designed to be exposed to the user. It should have been named to `_TensorFlowDataType` according to the original design. A bigger problem is, in Swift for TensorFlow APIs, that we have never exposed the notion of "data type" to the user. The notion has always been "scalar type".

Sources/TensorFlow/Core/Serialization.swift
```diff
@@ -0,0 +1,85 @@
...
+        TensorDataType(TF_CheckpointReaderGetVariableDataType(handle, name))
+    }
+
+    /// Loads and returns the value of the tensor with the provided name stored in the checkpoint.
+    @inlinable
+    public func load<Scalar: _TensorFlowDataTypeCompatible>(
```
rxwei: `loadTensor(named:)`

Sources/TensorFlow/Core/Serialization.swift
```diff
@@ -0,0 +1,85 @@
...
+
+/// A TensorFlow checkpoint file reader.
+public class TensorFlowCheckpointReader {
+    @usableFromInline internal let status: OpaquePointer
+    @usableFromInline internal let handle: OpaquePointer
+
+    /// URL of the checkpoint file.
+    public let checkpointPath: URL
+
+    /// Number of tensors stored in the checkpoint.
```
rxwei: Always add a definite article in doc comments that describe an entity.

Sources/TensorFlow/Core/Serialization.swift
```diff
@@ -0,0 +1,85 @@
...
+    public var tensorCount: Int { Int(TF_CheckpointReaderSize(handle)) }
+
+    /// Names of the tensors stored in the checkpoint.
```
rxwei: Always add a definite article in doc comments that describe an entity.

Sources/TensorFlow/Core/Serialization.swift
```diff
@@ -0,0 +1,85 @@
...
+
+import CTensorFlow
+import Foundation
+
+/// A TensorFlow checkpoint file reader.
+public class TensorFlowCheckpointReader {
+    @usableFromInline internal let status: OpaquePointer
+    @usableFromInline internal let handle: OpaquePointer
+
+    /// URL of the checkpoint file.
```
rxwei: Always add a definite article in doc comments that describe an entity.

Sources/TensorFlow/Core/Serialization.swift
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /// Returns `true` if the checkpoint contains a tensor with the provided name.
+    @inlinable
+    public func contains(tensorNamed name: String) -> Bool {
+        TF_CheckpointReaderHasTensor(handle, name) > 0
+    }
+
+    /// Returns the shape of the tensor with the provided name stored in the checkpoint.
+    @inlinable
```
rxwei: I'd suggest not doing `@inlinable` and `@usableFromInline` preemptively.

dan-zheng:
> By the way, I have been unable to build the S4TF toolchain on my Mac lately. The build script hangs while building SwiftPM. This is the command it runs and the stacktrace I get when I kill the process:

Thanks for sharing, and sorry for not catching the error! I just replicated the same hang while building SwiftPM, filed [TF-934](https://bugs.swift.org/browse/TF-934). I think the macOS toolchain issue started after we merged `swift-DEVELOPMENT-SNAPSHOT-2019-10-08-a` into `tensorflow` (https://github.com/apple/swift/pull/27751).

The team is looking into another merge from `swift-DEVELOPMENT-SNAPSHOT-2019-10-13-a`, perhaps that will naturally resolve the issue. If not, we will investigate further.

eaplatanios:
Great, thanks @dan-zheng! I'll also post there if I find a way to work around this issue.

Sources/TensorFlow/Core/Serialization.swift
eaplatanios: Thanks Richard! I'll address all your comments in a separate PR since this was merged already.

Sources/TensorFlow/Core/Serialization.swift
eaplatanios: In principle, we could make this function internal. The issue though is that there is not way for a user to inspect the data types of tensors stored in a checkpoint. I could make it return `String` with the name of the data type instead. Do you think that would be better?

rxwei: How about returning a metatype, `Any.Type`?

eaplatanios: Sounds good. Will open a PR in the next few minutes.

PR #530, opened 2019-10-14
Explore learning phase as a layer parameter.

saeta:
Instead of tracking learning phase as an implicit (thread-local) parameter to
certain layers, this PR proposes managing the learning phase as an explicit
parameter on layers.

This PR explores this design, and shows that it can be done with a minimum of
fuss. By building on top of the KeyPathIterable infrastructure, we can ensure
there's a minimum amount of boilerplate, but retain full functionality.
Although this has not been benchmarked in a large model, it is expected to not
have significant performance implications.

Advantages of this approach are:
 1. The phase can be easily changed with straight-forward, imperative code.
 2. Different parts of the model can have different states. This can be useful
    when performing certain kinds of research or other experimentation.
 3. No spooky-action-at-a-distance with thread-local and/or implicit
    parameters.

If this direction is a good one, we should consider making the following
changes:

 1. Make Module extend KeyPathIterable
 2. Build some sort of smooth bridge between this new API and the existing API.
    (Potentially by searching for all key paths to both LearningPhase.self and
    Optional<LearningPhase>.self.)
 3. Add more tests.

### DIFF
--- a/Sources/TensorFlow/Layer.swift
+++ b/Sources/TensorFlow/Layer.swift
@@ -119,6 +119,34 @@ public extension Layer {
     }
 }
 
+
+
+public extension Layer where Self: KeyPathIterable {
+    var allLearningPhases: LearningPhase? {
+        get {
+            let allPhases = self.recursivelyAllKeyPaths(to: LearningPhase.self)
+            if allPhases.isEmpty {
+                return nil
+            }
+            let first = self[keyPath: allPhases[0]]
+            for kp in allPhases {
+                if self[keyPath: kp] != first {
+                    return nil
+                }
+            }
+            return first
+        }
+        set {
+            guard let phase = newValue else {
+                fatalError("Cannot set allLearningPhases to nil!")
+            }
+            for kp in self.recursivelyAllWritableKeyPaths(to: LearningPhase.self) {
+                self[keyPath: kp] = phase
+            }
+        }
+    }
+}
+
 public extension Differentiable {
     /// Returns the output computed by applying a sequence of layers to the previous layer's output,
     /// except that the first layer's input is `self`.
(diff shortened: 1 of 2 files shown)


dan-zheng:
For reference, PyTorch uses imperative learning phase APIs [`module.train`](https://pytorch.org/docs/stable/nn.html#torch.nn.Module.train) and [`module.eval`](https://pytorch.org/docs/stable/nn.html#torch.nn.Module.eval), which recurse on submodules. These APIs are no-ops for most modules, but are overwritten by dropout and batchnorm.

I think this design could work in Swift, but I probably wouldn't pursue it unless there's a strong reason to change.

Shashi456:
Just a simple question, does this somehow pave ways for freezability of layers?

dan-zheng:
> Just a simple question, does this somehow pave ways for freezability of layers?

"Freezability" is actually orthogonal. "Learning phase" is relevant for layers dropout and batchnorm which have different behavior during training versus inference.

Shashi456:
oh alright, @dan-zheng thanks for the swift response.

eaplatanios:
I have mixed feelings about this. On one hand, I think the thread-local approach can sometimes have unexpected results, although I'm not sure how common that is in practice. On the other hand, I find the `allLearningPhases` name a bit confusing as, in my understanding, it really is a single optional learning phase that is only non-`nil` if it is the same for all nested modules. Intuitively, I feel a learning phase should not have different values for different parts of the model but could have different values for different consumers of a model (e.g., training on one thread and evaluating on another). For this reason, I would lean a bit more towards the current approach which also allows for the second use-case while avoiding a model copy. @saeta do you have any use case examples where a different phase is used for different parts of the model? This actually makes me think that layers sometimes may need to declare other types of context in which they operate and that should be "settable" externally, but I think that's worth a completely separate discussion. :)

Also, regarding the syntax a simple fix would be to wrap the current approach in helper functions similar to PyTorch, as @dan-zheng pointed out.

dan-zheng:
After discussion: this direction needs more exploration. @saeta: could you please reinvestigate when you have time? Closing for now.

PR #531, opened 2019-10-20
Addressed comments from #529.

eaplatanios:
This PR addresses @rxwei's comments from #529.


Sources/TensorFlow/Core/Serialization.swift
```diff
@@ -67,16 +64,30 @@ public class TensorFlowCheckpointReader {
         return TensorShape([Int64](dimensionsBufferPointer).map(Int.init))
     }
 
-    /// Returns the data type of the tensor with the provided name stored in the checkpoint.
-    @inlinable
-    public func dataType(ofTensorNamed name: String) -> TensorDataType {
-        TensorDataType(TF_CheckpointReaderGetVariableDataType(handle, name))
+    /// Returns the type of the tensor with the provided name stored in the checkpoint.
+    public func typeOfTensor(named name: String) -> Any.Type {
```
rxwei:
```suggestion
    public func scalarTypeOfTensor(named name: String) -> Any.Type {
```

Sources/TensorFlow/Core/Serialization.swift
```diff
@@ -20,13 +20,13 @@ public class TensorFlowCheckpointReader {
     @usableFromInline internal let status: OpaquePointer
```
rxwei: `@usableFromInline` can be all removed.

eaplatanios: Done.

Sources/TensorFlow/Core/Serialization.swift
eaplatanios: Done.

rxwei: Oh actually you can make it more specific: return `TensorFlowScalar.self`.

eaplatanios: That does not work because of the `String` type.

PR #532, opened 2019-10-21
Added support for a constant and a truncated Normal initializer.

eaplatanios:


eaplatanios:
Thanks Dan! And yes, it wasn't actually brought up during a design review. Unfortunately this semester the design review meetings clash with another meeting and I haven't been able to make it so far. :/

Sources/TensorFlow/Layers/Initialization.swift
```diff
@@ -19,6 +19,14 @@ public func zeros<Scalar: TensorFlowFloatingPoint>() -> ParameterInitializer<Sca
     { Tensor(zeros: $0) }
 }
 
+/// Returns a function that creates a tensor by initializing it to the provided value. Note that
+/// broadcasting of the provided value is supported.
+public func constantInitializer<Scalar: TensorFlowFloatingPoint>(
+    value: Tensor<Scalar>
+) -> ParameterInitializer<Scalar> {
+    { value.broadcasted(to: $0) }
```
dan-zheng: On second glance, I think broadcasting for a "constant initializer" may be controversial. I think broadcasting can lead to difficult to debug issues, like accidentally extending unexpected dimensions (batch vs spatial): http://nlp.seas.harvard.edu/NamedTensor#trap-2-broadcasting-by-alignment

I think the following initializer variant is less controversial:
* Broadcasting a scalar value to a given shape.

I wonder if there are good ways to specify principled broadcasting for non-scalar constant values?

eaplatanios: I agree and that's one of the concerns I had, but constant initializers are often used when loading from checkpoints. We could potentially have two constant initializers: (i) broadcasting a scalar to the shape, and (ii) taking in a tensor, verifying its shape matches the requested shape using a precondition, and returning that same tensor.

dan-zheng:
> Thanks Dan! And yes, it wasn't actually brought up during a design review. Unfortunately this semester the design review meetings clash with another meeting and I haven't been able to make it so far. :/

Thanks for sharing - you must be busy, take care!

---

Context: recently (since `Sequential` was added), @rxwei mentioned that we've adopted an informal policy of discussing API additions/changes during a design review before merging them. At the same time, we definitely care about moving fast and unblocking progress. I'll bring up whether we should make the policy explicit in the next design review.

Sources/TensorFlow/Layers/Initialization.swift
dan-zheng: > (ii) taking in a tensor, verifying its shape matches the requested shape using a precondition, and returning that same tensor.

Yes - I can't think of a better way to do constant tensor initialization without broadcasting. The precondition is a bit weird but it should be relatively easy to debug and triggers right when a layer is defined.

dan-zheng: I wonder if the constant initializer is blocking your work? Discussing this during the next design review would be great - if you're not free, I'm happy to mention it and explain constant initialization tradeoffs.

eaplatanios: @dan-zheng I addressed your concern by implementing the changes suggested above. This is not blocking me at all at it can also be implemented outside of swift-apis. :) I'm only opening this PR as I was planning to upstream some work I've been doing on getting BERT running in swift that may be useful to others. It would be greatly appreciated if you could bring this up in the next design review. :) I'll try to also open PRs for a multi-head attention layer and a transformer layer before then, in case you want to also bring that up. :)

eaplatanios:
> Context: recently (since `Sequential` was added), @rxwei mentioned that we've adopted an informal policy of discussing API additions/changes during a design review before merging them. At the same time, we definitely care about moving fast and unblocking progress. I'll bring up whether we should make the policy explicit in the next design review.

I think that's definitely a good idea. :)

Sources/TensorFlow/Layers/Initialization.swift
dan-zheng: Great, thank you! I added all of these to the design meeting notes (http://bit.ly/s4tf-design).

PR #533, opened 2019-10-24
Add huber loss support

jrabary:
This PR add huber loss support. Used for object detection for instance.


jrabary:
@googlebot I fixed it.

PR #540, opened 2019-11-03
[Layering]: Create Tensor library & move Random.

saeta:
This is the first of a series of changes to implement the [Layering Swift APIs
Proposal](https://docs.google.com/document/d/1HO_sMhZJHxlDqw4Pjz4qva2s5yzgwxel-82yoK8O6L4/edit#).
This first change creates the new Tensor library and moves the random number
utilities over.


eaplatanios:
@saeta @dan-zheng This PR was merged before making associated changes in apple/swift and so the toolchain won't build anymore. The relevant CMake configurations need to be changed in apple/swift for this change to be compatible.

dan-zheng:
> @saeta @dan-zheng This PR was merged before making associated changes in apple/swift and so the toolchain won't build anymore. The relevant CMake configurations need to be changed in apple/swift for this change to be compatible.

Thanks for reporting! Could you please share the exact error you encountered?

I'll try to reproduce soon today. [`stdlib/public/TensorFlow/CMakeLists.txt`](https://github.com/apple/swift/blob/b60de112837a192af09d234bc46471ac316f334a/stdlib/public/TensorFlow/CMakeLists.txt#L39) uses a glob to find Swift sources and the results are cached, so wiping `swift-macosx-x86_64/CMakeCache.txt` in the build subdirectory may be necessary.

saeta:
@eaplatanios Sorry about that! I got the following error; is this what you got as well?

```
<unknown>:0: error: filename "Utilities.swift" used twice: '/usr/local/google/home/saeta/src/s4tf6/tensorflow-swift-apis/Sources/Tensor/Utilities.swift' and '/usr/local/google/home/saeta/src/s4tf6/tensorflow-swift-apis/Sources/TensorFlow/Core/Utilities.swift'
<unknown>:0: note: filenames are used to distinguish private declarations with the same name
```

dan-zheng:
> @eaplatanios Sorry about that! I got the following error; is this what you got as well?
> 
> ```
> <unknown>:0: error: filename "Utilities.swift" used twice: '/usr/local/google/home/saeta/src/s4tf6/tensorflow-swift-apis/Sources/Tensor/Utilities.swift' and '/usr/local/google/home/saeta/src/s4tf6/tensorflow-swift-apis/Sources/TensorFlow/Core/Utilities.swift'
> <unknown>:0: note: filenames are used to distinguish private declarations with the same name
> ```

I reproduced the error. Fixed in:
* https://github.com/tensorflow/swift-apis/pull/546
* https://github.com/apple/swift/pull/28124

Shashi456:
@saeta, if this is being done systematically and some order of files being shifted from tensorflow to tensor and deeplearning, I'd love to help as well !

saeta:
Hi @Shashi456 ! Thanks for volunteering. That would be super helpful. I'm currently thinking through what makes sense to do next. As I come up with a plan, I'll be sure to let you know! :-) (There are some other considerations I'm trying to think through, so it might be a little bit.) Thanks again! -Brennan

PR #541, opened 2019-11-03
Move Dense and Dropout layers to separate files

rickwierenga:
For details see: https://groups.google.com/a/tensorflow.org/forum/m/#!topic/swift/-I-P8wjhsTg

### DIFF
--- a/Sources/TensorFlow/Layers/Core.swift
+++ b/Sources/TensorFlow/Layers/Core.swift
@@ -12,22 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-/// A layer that encloses a custom differentiable function.
-public struct Function<Input: Differentiable, Output: Differentiable>: ParameterlessLayer {
-    public typealias Body = @differentiable (Input) -> Output
-
-    @noDerivative public let body: Body
-
-    public init(_ body: @escaping Body) {
-        self.body = body
-    }
-
-    @differentiable
-    public func callAsFunction(_ input: Input) -> Output {
-        body(input)
-    }
-}
-
 /// A flatten layer.
 ///
 /// A flatten layer flattens the input when applied without affecting the batch size.
@@ -82,3 +66,18 @@ public struct Reshape<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
     }
 }
 
+/// A layer that encloses a custom differentiable function.
+public struct Function<Input: Differentiable, Output: Differentiable>: ParameterlessLayer {
+    public typealias Body = @differentiable (Input) -> Output
+
+    @noDerivative public let body: Body
+
+    public init(_ body: @escaping Body) {
+        self.body = body
+    }
+
+    @differentiable
+    public func callAsFunction(_ input: Input) -> Output {
+        body(input)
+    }
+}
--- a/Sources/TensorFlow/Layers/Core.swift
+++ b/Sources/TensorFlow/Layers/Core.swift
@@ -12,67 +12,19 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-fileprivate extension Tensor where Scalar: TensorFlowFloatingPoint {
-     /// Computes dropout given a probability.
-     // TODO: Remove the underscore once `droppingOut(probability:)` has been removed.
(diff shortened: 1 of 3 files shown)


rickwierenga:
@googlebot I signed it!

rickwierenga:
I just noticed (too late) that this PR also includes the precondition described [here](https://groups.google.com/a/tensorflow.org/forum/m/#!topic/swift/Hi42LFqe_s4). Is this a problem?

Shashi456:
@rickwierenga, it's ideal if we keep the splitting of core.swift into separate files and this precondition as separate PRs

Sources/TensorFlow/Layers/Core.swift
```diff
@@ -12,67 +12,19 @@
...
-    @noDerivative public let probability: Double
-
-    /// Creates a dropout layer.
-    ///
-    /// - Parameter probability: The drop probability.
-    public init(probability: Double) {
-        self.probability = probability
-    }
+/// A layer that encloses a custom differentiable function.
+public struct Function<Input: Differentiable, Output: Differentiable>: ParameterlessLayer {
```
saeta: Could you keep the ordering of the layers the same as before? (This will help make the history easier to follow.)

Sources/TensorFlow/Layers/Dropout.swift
```diff
@@ -0,0 +1,82 @@
...
+public struct Dropout<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
+    @noDerivative public let probability: Double
+
+    /// Creates a dropout layer.
+    ///
+    /// - Parameter probability: The drop probability.
+    /// - Precondition: probability must be a value between 0 and 1 (inclusive). 
+    public init(probability: Double) {
+        precondition(0...1 ~= probability,
+            "Probability must be a value between 0 and 1 (inclusive) but is \(probability)")
```
saeta: FWIW, although it's generally better to have one logical change per PR, given that this is just one statement, I think it's probably okay to leave in. 

Note: this change would break users who relied upon probability outside the range acting as if it's clamped to the range. I think this is okay, however, as this likely indicates a bug. That said, if we wanted to, we could provide a convenience initializer that would do the clamping to 0...1 and then delegate to this initializer. (This could be done in a follow-on PR.)

rickwierenga: > FWIW, although it's generally better to have one logical change per PR, given that this is just one statement, I think it's probably okay to leave in.

You are absolutely right. I forgot to exclude it in the commit--will look more closely in the future.

> Note: this change would break users who relied upon probability outside the range acting as if it's clamped to the range. I think this is okay, however, as this likely indicates a bug. That said, if we wanted to, we could provide a convenience initializer that would do the clamping to 0...1 and then delegate to this initializer. (This could be done in a follow-on PR.)

The convenience initializer ia a great idea. I think it should show some kind of message though indicating the input value will be clamped. Should this be a permanent or temporary initializer (with deprecation warning)?

rickwierenga:
Just fixed the ordering issue. I will leave the changes in Dropout.swift for the next PR.

Sources/TensorFlow/Layers/Dropout.swift
```diff
@@ -0,0 +1,82 @@
...
+    public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+        switch Context.local.learningPhase {
+        case .training:
+            return applyingTraining(to: input)
+        case .inference:
+            return applyingInference(to: input)
+        }
+    }
+}
+
```
dan-zheng: Please remove empty lines.

Sources/TensorFlow/Layers/Dense.swift
```diff
@@ -0,0 +1,91 @@
...
+        weightInitializer: ParameterInitializer<Scalar> = glorotUniform(),
+        biasInitializer: ParameterInitializer<Scalar> = zeros()
+    ) {
+        self.init(
+            weight: weightInitializer([inputSize, outputSize]),
+            bias: biasInitializer([outputSize]),
+            activation: activation)
+    }
+}
+
```
dan-zheng: Please remove empty lines here, too.

Issue #528, opened 2019-10-06
conv1D missing as a global function


SamPruden:
`Conv1D` is available as a layer which is implemented via `conv2D`, but `conv1D` is not available as a global function. I'm using convolution for data processing outside of the net, so it would be useful to have. Is there interest in getting this added?

I've just nabbed and tweaked the implementation from `Conv2D` for now.

```Swift
@differentiable(wrt: (input, filter))
func conv1D<Scalar: TensorFlowFloatingPoint>(
  _ input: Tensor<Scalar>,
  filter: Tensor<Scalar>,
  stride: Int = 1,
  padding: Padding = .valid,
  dilation: Int = 1
) -> Tensor<Scalar> {
  return conv2D(
    input.expandingShape(at: 1),
    filter: filter.expandingShape(at: 0),
    strides: (1, 1, stride, 1),
    padding: padding,
    dilations: (1, 1, dilation, 1)
  ).squeezingShape(at: 1)
}
```

I'm pretty new to playing with Swift and TensorFlow and I'm definitely not a linear algebra expert so I don't want to take responsibility for claiming this is a good implementation, but it looks like it could just be dropped into `Operators/NN.swift` with appropriate comments and that would be it done. I can't see any test coverage for the other similar global functions.

If we want this done and that's all there is to it I could make a pull, but I'd be happier for somebody else to just drop it in. I don't even have a proper Swift setup here at the moment (I'm just using Colab) and I feel a little uncomfortable putting a pull together without even confirming it builds, so if somebody else wants to do it that would be great.

ocampor:
I will take a look to this one.

saeta:
I believe this is now fixed due to #549. Thanks! -Brennan

PR #549, opened 2019-11-08
Extract conv1D functionality from Conv1D Layer

ocampor:
### Description
As described in #528, the layer `Conv1D` is doing two things at a time (1) calculating a one-dimensional convolution and (2) applying the activation function. If we split this functionality into two, we enable clients to only use the conv1D. This will allow them to write code that is easier to read.

### Solution
Extract the `conv1D` functionality into `NN.swfit`

### DIFF
--- a/Sources/TensorFlow/Operators/NN.swift
+++ b/Sources/TensorFlow/Operators/NN.swift
@@ -72,6 +72,35 @@ public extension Padding {
     }
 }
 
+/// Returns a 1-D convolution with the specified input, filter, stride, and padding.
+///
+/// - Parameters:
+///   - input: The input.
+///   - filter: The convolution filter.
+///   - stride: The stride of the sliding filter.
+///   - padding: The padding for the operation
+///   - dilation: The dilation factor.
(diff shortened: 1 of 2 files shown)


ocampor:
Thanks @saeta, I do agree that conv1D should have it's own unit test; however, I found that all the NN functions are tested in `LayerTests.swift`. I analyzed the test and the functionality that is tested is the convolution and not the layer. May be it would ge a good idea to extract those tests from `LayerTests.swift` and create a `NNTests.swift` in the future?.

Sources/TensorFlow/Operators/NN.swift
```diff
@@ -72,6 +72,35 @@ public extension Padding {
     }
 }
 
+/// Returns a 1-D convolution with the specified input, filter, stride, and padding.
+///
+/// - Parameters:
+///   - input: The input.
+///   - filter: The convolution filter.
+///   - stride: The stride of the sliding filter.
+///   - padding: The padding for the operation
```
Shashi456:
```suggestion
///   - padding: The padding for the operation.
```

Issue #545, opened 2019-11-06
Make `Tensor.product(squeezingAxes:)` differentiable


dan-zheng:
[`Tensor.product(squeezingAxes: Tensor<Int32>)`](https://github.com/tensorflow/swift-apis/blob/30a1071d9cdf0322c171899e9c17423fa240e980/Sources/TensorFlow/Operators/Math.swift#L1817) is [currently not `@differentiable`](https://github.com/tensorflow/swift-apis/blob/30a1071d9cdf0322c171899e9c17423fa240e980/Sources/TensorFlow/Operators/Math.swift#L1815).

We should do the following:
* Add the attribute `@differentiable(wrt: self, vjp: _vjpProduct(squeezingAxes:) where Scalar: TensorFlowFloatingPoint)` to `Tensor.product(squeezingAxes: Tensor<Int32>)`.
* Define `func _vjpProduct(squeezingAxes axes: Tensor<Int32>)`.
* Mark `Tensor.product(squeezingAxes: [Int])` and `Tensor.product(squeezingAxes: Int)` as `@differentiable`. No registered derivative is needed because they call `Tensor.product(squeezingAxes: Tensor<Int32>)`.

---

* [TensorFlow reference implementation of `ProdGrad`](https://github.com/tensorflow/tensorflow/blob/72d0facc37e377d89d8034bb825bf0e93a4f810d/tensorflow/cc/gradients/math_grad.cc#L882).
* [Use case from mailing list](https://groups.google.com/a/tensorflow.org/d/msg/swift/9ss6t-neuGE/XqMOA-CQAAAJ).

dan-zheng:
cc @kongzii

kongzii:
Continuing the discussion at [Use case from mailing list](https://groups.google.com/a/tensorflow.org/d/msg/swift/9ss6t-neuGE/XqMOA-CQAAAJ), I will start working on this :)

dan-zheng:
Done in https://github.com/tensorflow/swift-apis/pull/550.

PR #550, opened 2019-11-09
Make `Tensor.product(squeezingAxes:)` differentiable

kongzii:
Solving issue at https://github.com/tensorflow/swift-apis/issues/545

While creating `_vjpProduct` I used [C++ ProdGrad](https://github.com/tensorflow/tensorflow/blob/72d0facc37e377d89d8034bb825bf0e93a4f810d/tensorflow/cc/gradients/math_grad.cc#L882 ) and [Python ProdGrad](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/math_grad.py#L276) for reference.

I could not find function `reducedShape`, which is available in Python TF. So I created it in `Sources/TensorFlow/Operators/Basic.swift`. Perhaps it should be somewhere else or it could be implemented better, I will gladly accept suggestions and edit it if needed.

### DIFF
--- a/Sources/TensorFlow/Core/TensorShape.swift
+++ b/Sources/TensorFlow/Core/TensorShape.swift
@@ -110,6 +110,20 @@ public extension TensorShape {
         get { return TensorShape(dimensions[bounds]) }
         set { dimensions[bounds] = ArraySlice(newValue.dimensions) }
     }
+
+    /// Returns the reduced shape along the given axes. For each axis, the dimension is set to 1.
+    ///
+    /// Adapted from `reduced_shape` in Python TensorFlow:
+    /// https://github.com/tensorflow/tensorflow/blob/8023e0736756b93308cbe847e5e6afbfb625f8d8/tensorflow/python/ops/math_ops.py#L3490
+    func reducedShape(axes: [Int32]) -> TensorShape {
+        var reducedShapeScalars = dimensions.map { Int($0) }
+
+        for axe in axes {
+            reducedShapeScalars[Int(axe)] = 1
+        }
+
+        return TensorShape(reducedShapeScalars)
+    }
 }
 
 extension TensorShape: Equatable {
--- a/Sources/TensorFlow/Core/TensorShape.swift
+++ b/Sources/TensorFlow/Core/TensorShape.swift
@@ -110,20 +110,6 @@ public extension TensorShape {
         get { return TensorShape(dimensions[bounds]) }
         set { dimensions[bounds] = ArraySlice(newValue.dimensions) }
     }
-
-    /// Returns the reduced shape along the given axes. For each axis, the dimension is set to 1.
-    ///
-    /// Adapted from `reduced_shape` in Python TensorFlow:
-    /// https://github.com/tensorflow/tensorflow/blob/8023e0736756b93308cbe847e5e6afbfb625f8d8/tensorflow/python/ops/math_ops.py#L3490
-    func reducedShape(axes: [Int32]) -> TensorShape {
-        var reducedShapeScalars = dimensions.map { Int($0) }
-
-        for axe in axes {
-            reducedShapeScalars[Int(axe)] = 1
-        }
-
-        return TensorShape(reducedShapeScalars)
-    }
 }
 
 extension TensorShape: Equatable {
--- a/Sources/TensorFlow/Operators/Basic.swift
+++ b/Sources/TensorFlow/Operators/Basic.swift
@@ -148,6 +148,18 @@ public extension Tensor {
         reshaped(toShape: other.shapeTensor)
     }
 
+    /// Helper function for reduction ops.
+    /// Returns 1D Tensor, the output shape as if keepDims were set to True.
+    func reducedShape(axes: Tensor<Int32>) -> TensorShape {
+        var reducedShapeScalars = self.shapeTensor.scalars
+
+        for axe in axes.scalars {
+            reducedShapeScalars[Int(axe)] = 1
+        }
+
+        return TensorShape(reducedShapeScalars.map { Int($0) })
+    }
+
     /// Reshape to the specified shape.
     /// - Precondition: The number of scalars matches the new shape.
     @inlinable
--- a/Sources/TensorFlow/Operators/Basic.swift
+++ b/Sources/TensorFlow/Operators/Basic.swift
@@ -148,18 +148,6 @@ public extension Tensor {
         reshaped(toShape: other.shapeTensor)
     }
 
-    /// Helper function for reduction ops.
-    /// Returns 1D Tensor, the output shape as if keepDims were set to True.
-    func reducedShape(axes: Tensor<Int32>) -> TensorShape {
-        var reducedShapeScalars = self.shapeTensor.scalars
-
-        for axe in axes.scalars {
-            reducedShapeScalars[Int(axe)] = 1
-        }
-
-        return TensorShape(reducedShapeScalars.map { Int($0) })
-    }
-
     /// Reshape to the specified shape.
     /// - Precondition: The number of scalars matches the new shape.
     @inlinable
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -1812,8 +1812,8 @@ public extension Tensor where Scalar: Numeric {
     ///
     /// - Parameter axes: The dimensions to reduce.
     /// - Precondition: Each value in `axes` must be in the range `-rank...rank`.
-    // TODO: Make this @differentiable.
     @inlinable
+    @differentiable(wrt: self, vjp: _vjpProduct(squeezingAxes:) where Scalar: TensorFlowFloatingPoint)
     func product(squeezingAxes axes: Tensor<Int32>) -> Tensor {
         _Raw.prod(self, reductionIndices: axes, keepDims: false)
     }
@@ -1823,6 +1823,7 @@ public extension Tensor where Scalar: Numeric {
     /// - Parameter axes: The dimensions to reduce.
     /// - Precondition: Each value in `axes` must be in the range `-rank...rank`.
     @inlinable
+    @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
     func product(squeezingAxes axes: [Int]) -> Tensor {
         // TODO(TF-433): Remove workaround for differentiating `map`.
         let axes = {axes.map(Int32.init)}()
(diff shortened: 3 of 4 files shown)


Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -148,6 +148,18 @@ public extension Tensor {
         reshaped(toShape: other.shapeTensor)
     }
 
+    /// Helper function for reduction ops.
+    /// Returns 1D Tensor, the output shape as if keepDims were set to True.
+    func reducedShape(axes: Tensor<Int32>) -> TensorShape {
```
dan-zheng: It seems like `reducedShape` is ported from [`def reduced_shape(input_shape, axes)`](https://github.com/tensorflow/tensorflow/blob/8023e0736756b93308cbe847e5e6afbfb625f8d8/tensorflow/python/ops/math_ops.py#L3490).

- Since `reducedShape` is a helper function, please make it private rather than public.
- Since `reducedShape` is an operation on tensor shapes, please change it to take `TensorShape` directly instead of `Tensor`.
- Since `axes` is always 1-D, please change its type to `[Int32]`. (It should be more efficient to perform shape operations on non-`Tensor` types, like `[Int32]` and `Tensor`, because `Tensor` operations have more overhead.)

How about redefining `reducedShape` as a private method on `TensorShape`:
```suggestion
private extension TensorShape {
  /// Returns the reduced shape along the given axes. For each axis, the dimension is set to 1.
  ///
  /// Adapted from `reduced_shape` in Python TensorFlow:
  /// https://github.com/tensorflow/tensorflow/blob/8023e0736756b93308cbe847e5e6afbfb625f8d8/tensorflow/python/ops/math_ops.py#L3490
  func reducedShape(alongAxes axes: [Int32]) -> TensorShape { ... }
}
```

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -2224,6 +2227,49 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+
+    @inlinable
+    func _vjpProduct(squeezingAxes axes: Tensor<Int32>) -> (Tensor, (Tensor) -> Tensor) {
+        // The gradient can be expressed by dividing the product by each entry of the
+        // input tensor, but this approach can't deal with zeros in the input.
+        // Here, we avoid this problem by composing the output as a product of two cumulativeProduct operations.
+        let result = product(squeezingAxes: axes)
+        return (result, { v in
+            // Reshape reduction indices for the case where the parameter is a scalar
+            let reductionIndices = axes.reshaped(to: TensorShape(-1))
```
dan-zheng: Rather than defining `let reductionIndicesSafe` below, how about using `var reductionIndices` and mutating it?
```suggestion
            var reductionIndices = axes.reshaped(to: TensorShape(-1))
```

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -2224,6 +2227,49 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+
+            // Expand grad to full input shape
+            let outputShapeKeptDims = self.reducedShape(axes: reductionIndices)
+            let vReshaped = v.reshaped(to: outputShapeKeptDims)
+
+            let vBroadcasted = vReshaped.broadcasted(to: self.shape)
+
+            // Normalize any negative indices in the reduction_axes to positive values
+            let reductionIndicesSafe = (reductionIndices + Int32(self.rank)) % Int32(self.rank)
```
dan-zheng: Define `reductionIndices` as a `var`:
```suggestion
            reductionIndices = (reductionIndices + Int32(self.rank)) % Int32(self.rank)
```

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -2224,6 +2227,49 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
             ) / self
         })
     }
+
+    @inlinable
```
dan-zheng: Please add attribution comment:
```
    /// Adapted from `_ProdGrad` in Python TensorFlow:
    /// https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/math_grad.py
```

Sources/TensorFlow/Operators/Basic.swift
kongzii: Sure, just a little problem over here:

If I define it as `private`, I get `error: 'reducedShape' is inaccessible due to 'fileprivate' protection level`.

If I try it with `internal`, I get error: `instance method 'reducedShape(axes:)' is internal and cannot be referenced from an '@inlinable' function`.

I added reducedShape as public extension in `Sources/TensorFlow/Core/TensorShape.swift` for now. Can you help me with the `private` problem? Thanks.

Sources/TensorFlow/Operators/Math.swift
kongzii: Changed.

Sources/TensorFlow/Operators/Math.swift
kongzii: Changed.

Sources/TensorFlow/Operators/Math.swift
kongzii: Added. :)

Sources/TensorFlow/Core/TensorShape.swift
```diff
@@ -110,6 +110,20 @@ public extension TensorShape {
         get { return TensorShape(dimensions[bounds]) }
         set { dimensions[bounds] = ArraySlice(newValue.dimensions) }
     }
+
+    /// Returns the reduced shape along the given axes. For each axis, the dimension is set to 1.
+    ///
+    /// Adapted from `reduced_shape` in Python TensorFlow:
+    /// https://github.com/tensorflow/tensorflow/blob/8023e0736756b93308cbe847e5e6afbfb625f8d8/tensorflow/python/ops/math_ops.py#L3490
+    func reducedShape(axes: [Int32]) -> TensorShape {
```
dan-zheng: Actually, since the logic of `reducedShape` is so simple, how about inlining its defining in `_vjpProduct`?
Sorry for the thrash. Adding `reducedShape` as an extension method on `TensorShape` is rather invasive.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -2224,6 +2227,51 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+        /// Here, we avoid this problem by composing the output as a product of two cumulativeProduct operations.
+        /// Adapted from `_ProdGrad` in Python TensorFlow:
+        /// https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/math_grad.py
+        let result = product(squeezingAxes: axes)
+        return (result, { v in
+            // Reshape reduction indices for the case where the parameter is a scalar
+            var reductionIndices = axes.reshaped(to: TensorShape(-1))
+
+            // Expand grad to full input shape
+            let outputShapeKeptDims = self.shape.reducedShape(axes: reductionIndices.scalars)
```
dan-zheng: See other comment - please inline `reducedShape`.
```suggestion
            var outputShape = self.shape
            for axis in reductionIndices.scalars {
                outputShape[Int(axis)] = 1
            }
```

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -2224,6 +2227,51 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+
+    @inlinable
+    func _vjpProduct(squeezingAxes axes: Tensor<Int32>) -> (Tensor, (Tensor) -> Tensor) {
+        /// The gradient can be expressed by dividing the product by each entry of the
+        /// input tensor, but this approach can't deal with zeros in the input.
+        /// Here, we avoid this problem by composing the output as a product of two cumulativeProduct operations.
+        /// Adapted from `_ProdGrad` in Python TensorFlow:
+        /// https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/math_grad.py
+        let result = product(squeezingAxes: axes)
+        return (result, { v in
```
dan-zheng: Currently, the `{ v in ... }` pullback captures all of `self`, which is not efficient.
Ideally, we should add a capture list to capture only the minimal necessary values:
```swift
        return (result, { [rank = self.rank, shape = self.shape, ...] v in
```
However, there's one tricky `self` capture: `let permuted = self.transposed(permutation: perm)`.

I wonder what is the size of `permuted` compared with `self`? If `permuted` is significantly smaller than `self`, then it may be worthwhile to compute it outside of the pullback and to add it to the capture list.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -2224,6 +2227,51 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+            let idx = Tensor<Int32>(0 ..< Int32(self.rank))
+
+            let other = Tensor<Int32>(Array(Set(idx.scalars).symmetricDifference(reductionIndices.scalars)))
+
+            let perm = reductionIndices.concatenated(with: other)
+
+            let reducedNum = Int(self.shapeTensor.gathering(atIndices: reductionIndices).product().scalars[0])
+            let otherNum = Int(self.shapeTensor.gathering(atIndices: other).product().scalars[0])
+
+            let permutated = self.transposed(permutation: perm)
```
```suggestion
            let permuted = self.transposed(permutation: perm)
```

Sources/TensorFlow/Operators/Basic.swift
dan-zheng: Sorry for the thrash! Let's actually inline `reducedShape`: https://github.com/tensorflow/swift-apis/pull/550#discussion_r344675367.

Sources/TensorFlow/Core/TensorShape.swift
kongzii: Yeah, no problem.

Sources/TensorFlow/Operators/Math.swift
kongzii: What do you mean by 

> If permuted is significantly smaller than self

Isn't it the same Tensor as self, just with permuted dimensions, so they should be equal in the size?

I tried 

```
print("size self", MemoryLayout.size(ofValue: self))  // output: 16
print("size permuted", MemoryLayout.size(ofValue: permuted)) // output: 16
```

Or perhaps I just don`t understand correctly.

dan-zheng: Oh yes, you're right - `transposed(permutation:)` always returns a tensor with the same size (scalar count).
In that case, let's just capture `self` instead of introducing a capture list.

dan-zheng:
Hi @kongzii: could you please enable "Allow edits from maintainers"?
image
I'd like to push some minor style edits to your branch, if you don't mind. It's more efficient than leaving a bunch of GitHub suggestions. 🙂

kongzii:
> Hi @kongzii: could you please enable "Allow edits from maintainers"?
> image
> I'd like to push some minor style edits to your branch, if you don't mind. It's more efficient than leaving a bunch of GitHub suggestions. 🙂

It should be enabled 🤔 

obrazek

dan-zheng:
> It should be enabled 🤔

Thanks! I pushed https://github.com/tensorflow/swift-apis/pull/550/commits/de3fcb2650337f22b53fd2e73c80ce9c1d22051b with minor edits:
* I added a `_vjpProduct` test for negative reduction indices. This requires normalizing `reductionIndices` early in `_vjpProduct` before it's used to avoid out-of-bounds indexing crashes.
* Please limit lines to 100 columns. [`swift-format`](https://github.com/apple/swift-format) can help.

PR #556, opened 2019-11-15
Add vjps that don't use slicing or gather on shapes.

pschuh:
For _vjpMean(alongAxes axes: [Int]),
We can do:
let count = axes.map { myShape[$0] }.reduce(1, *)
instead of dispatching an op to do the shape computations.

### DIFF
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -1890,7 +1890,7 @@ public extension Tensor where Scalar: Numeric {
     /// - Parameter axes: The dimensions to reduce.
     /// - Precondition: Each value in `axes` must be in the range `-rank...rank`.
     @inlinable
-    @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
+    @differentiable(wrt: self, vjp: _vjpMean(squeezingAxes:) where Scalar: TensorFlowFloatingPoint)
     func mean(squeezingAxes axes: [Int]) -> Tensor {
         // TODO(TF-433): Remove workaround for differentiating `map`.
         let axes = {axes.map(Int32.init)}()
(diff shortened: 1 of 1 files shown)


Sources/TensorFlow/Operators/Math.swift
```diff
@@ -2201,6 +2201,25 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
         })
     }
 
+    @inlinable
+    func _vjpMean(alongAxes axes: [Int]) -> (Tensor, (Tensor) -> Tensor) {
+        let value = mean(alongAxes: axes)
+        let myShape = shape
```
dan-zheng: Is the temporary binding `let myShape = shape` necessary?
If so, could you please add a comment explaining why it's needed? Future readers of the code then won't be tempted to remove the line.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -2201,6 +2201,25 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
         })
     }
 
+    @inlinable
```
dan-zheng: Could you please add a comment explaining that these VJPs are defined to avoid `Tensor` operations on `axes`?
```swift
    // Note: this optimized derivative is defined in addition to `_vjpMean(alongAxes axes: Tensor<Int32>)`
    // to avoid `Tensor` operations on `axes`.
```
The comment may sound a bit overly detailed, but I think it's good to explain in code that these are optimized derivatives (not strictly necessary).

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -2201,6 +2201,31 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
         })
     }
 
+    // Specialization to avoid _Raw.gather on shapes when axes is known to be
+    // [Int].
+    @inlinable
+    func _vjpMean(alongAxes axes: [Int]) -> (Tensor, (Tensor) -> Tensor) {
+        let value = mean(alongAxes: axes)
+        // Cache shape because it is a computed property.
+        let myShape = shape
```
```suggestion
        let cachedShape = shape
```

PR #557, opened 2019-11-15
Add more preconditions for Basic.swift module

ocampor:
- [x] Precondition for `unstacked`
- [x] Preconditions for `split` for count and axis
- [x] Precondition for `tiled`
- [x] Preconditions for all `reshaped`
- [x] Preconditions for `gathering`

As requested in #517

### DIFF
--- a/Sources/TensorFlow/Operators/Basic.swift
+++ b/Sources/TensorFlow/Operators/Basic.swift
@@ -188,7 +188,23 @@ public extension Tensor {
     @inlinable
     @differentiable(wrt: self, vjp: _vjpReshaped(toShape:) where Scalar: TensorFlowFloatingPoint)
     func reshaped(toShape newShape: Tensor<Int32>) -> Tensor {
-        _Raw.reshape(self, shape: newShape)
+        let totalNegative = newShape.scalars.filter({$0 == -1}).count
+        let positiveShapeSizes = newShape.scalars.filter({$0 > 0})
+        let newShapeScalarCount = positiveShapeSizes.reduce(1, {$0 * $1})
+
+        precondition(totalNegative <= 1, "Only one input size may be -1.")
+
+        if totalNegative == 1 {
+            precondition(
+                scalarCount % Int(newShapeScalarCount) == 0,
+                "The number of scalars must be a multiple of the new shape.")
+        } else {
+            precondition(
+                scalarCount == newShapeScalarCount,
+                "The number of scalars must match the new shape.")
+        }
+
+        return _Raw.reshape(self, shape: newShape)
     }
 
     /// Return a copy of the tensor collapsed into a 1-D `Tensor`, in row-major order.
--- a/Sources/TensorFlow/Operators/Basic.swift
+++ b/Sources/TensorFlow/Operators/Basic.swift
@@ -102,6 +102,9 @@ public extension Tensor {
     @differentiable(vjp: _vjpSplit(count:alongAxis:) where Scalar: TensorFlowFloatingPoint)
     func split(count: Int, alongAxis axis: Int = 0) -> [Tensor] {
         preconditionAxis(axis)
+        precondition(
+            shape[axis] % count == 0,
+            "Number of ways to split should evenly divide the split dimension.")
         return _Raw.split(splitDim: Tensor<Int32>(Int32(axis)), value: self, numSplit: Int64(count))
     }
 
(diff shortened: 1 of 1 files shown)


ocampor:
While I was implementing these preconditions, I thought that it would be a good idea to add some tests to avoid bugs in this newly added code. May be it is too much?

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -88,7 +101,11 @@ public extension Tensor {
     @inlinable
     @differentiable(vjp: _vjpSplit(count:alongAxis:) where Scalar: TensorFlowFloatingPoint)
     func split(count: Int, alongAxis axis: Int = 0) -> [Tensor] {
-        _Raw.split(splitDim: Tensor<Int32>(Int32(axis)), value: self, numSplit: Int64(count))
+        preconditionAxis(axis)
+        precondition(
+            shapeTensor[axis].scalarized() % Int32(count) == 0,
```
saeta: Note, I think this call to `.scalarized()` will truncate lazy-tensor style traces, as well as prevent dynamic graph extraction. I believe @bgogul was working on some "Swift-side" shape tracking. Do you know the status of this @bgogul ? Could that help us here?

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -162,7 +188,23 @@ public extension Tensor {
     @inlinable
     @differentiable(wrt: self, vjp: _vjpReshaped(toShape:) where Scalar: TensorFlowFloatingPoint)
     func reshaped(toShape newShape: Tensor<Int32>) -> Tensor {
-        _Raw.reshape(self, shape: newShape)
+        let totalNegative = newShape.scalars.filter({$0 == -1}).count
```
saeta: Unfortunately, these will also inhibit optimizations. :-( That said, I think it'd be really good to catch issues as early as possible and have good quality error messages. I believe we can eventually get good error messages in the fully dynamic cases as we upgrade our underlying runtime.

How about the following short-term compromise between usability and performance of: adding the preconditions to the `TensorShape` overload of `reshaped` above, where we can eagerly check the shape errors. WDYT?

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -408,7 +450,8 @@ public extension Tensor {
         atIndices indices: Tensor<Index>,
         alongAxis axis: Int = 0
     ) -> Tensor {
-        _Raw.gatherV2(params: self, indices: indices, axis: Tensor<Int32>(Int32(axis)))
+        preconditionAxis(axis)
```
saeta: This is great!

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -34,6 +34,18 @@ public extension TensorFlowScalar {
 }
 
 public extension Tensor {
+    /// Helper function that assess if `axis` is in the range `[-rank, rank)`, where `rank` is the rank of
+    /// the provided tensors.
+    @usableFromInline
+    internal func preconditionAxis(_ axis: Int) {
```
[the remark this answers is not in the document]
ocampor: This is a helper function to validate that the axis is in the range of -rank,rank. I don’t think it would add much to the public API. Although, I would need to check how useful will be in other modules.

Sources/TensorFlow/Operators/Basic.swift
ocampor: That is actually something I though. While I was testing the methods I found the error messages were quite understandable. I think the issue was created because there are some methods that just break the applications without any error. May be we can modify #517 to find the methods that crash without human readable error error and trace?

Sources/TensorFlow/Operators/Basic.swift
bgogul: The swift-side shape tracking is already in: https://github.com/tensorflow/swift-apis/blob/275fe8837f9a6e8e5af53dea8052be87a8a5a39d/Sources/TensorFlow/Core/LazyTensorOperation.swift#L78

This should not adversely affect the lazy tensor trace extraction.

Sources/TensorFlow/Operators/Basic.swift
saeta: I checked with @pschuh and @asuhan and the new tracing system that's coming soon, and the performance concerns I was worried about are not important. Looking forward to getting this PR merged!

PR #558, opened 2019-11-15
Move TensorHandle related code out of ShapedArray.

pschuh:
This decouples ShapedArray from TensorHandle and allows for alternative implementations. The current implementation uses an enum to specify the backing storage. Switching to virtual dispatch allows moving all the TensorHandle related logic to a separate file.

### DIFF
--- a/Sources/TensorFlow/Core/ShapedArray.swift
+++ b/Sources/TensorFlow/Core/ShapedArray.swift
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 import Swift
-import CTensorFlow
 
 //===------------------------------------------------------------------------------------------===//
 // TensorBuffer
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/Core/ShapedArray.swift
```diff
@@ -25,101 +24,43 @@ import CTensorFlow
...
-            self.array = array
-        }
-    }
-
-    enum Allocation {
-        case native(BoxedArray)
-        case tensorFlow(CTensor)
-    }
+internal class TensorBuffer<Scalar> {
+    let count : Int
```
bgogul: Can you document this field?

Sources/TensorFlow/Core/ShapedArray.swift
```diff
@@ -25,101 +24,43 @@ import CTensorFlow
...
-            debugLog("Deallocating underlying buffer.")
-        case let .tensorFlow(cTensor):
-            debugLog("Deleting underlying tensor.")
-            TF_DeleteTensor(cTensor)
-        }
-        debugLog("Returning from deinit of TensorBuffer.")
+    func withUnsafeBufferPointer<R>(
+        _ body: (UnsafeBufferPointer<Scalar>) throws -> R
+    ) rethrows -> R {
+      fatalError("TensorBuffer should not be constructed directly.")
```
bgogul: The error message does not seem to match the semantics of `withUnsafeBufferPointer`?

Sources/TensorFlow/Core/ShapedArray.swift
```diff
@@ -25,101 +24,43 @@ import CTensorFlow
...
     }
 
-    init(allocation: Allocation, count: Int) {
-        self.allocation = allocation
-        self.count = count
+    func withUnsafeMutableBufferPointer<R>(
+        _ body: (inout UnsafeMutableBufferPointer<Scalar>) throws -> R
+    ) rethrows -> R {
+      fatalError("TensorBuffer should not be constructed directly.")
```
bgogul: Same as above. Error message does not match semantics of the `withUnsafeMutableBufferPointer`?

Sources/TensorFlow/Core/TensorHandle.swift
```diff
@@ -212,3 +212,94 @@ public struct VariantHandle {
...
+internal class CTensorTensorBuffer<Scalar> : TensorBuffer<Scalar> {
+    let cTensor: CTensor
+
+    /// Creates a local tensor buffer from a C `TF_Tensor*` value and takes ownership of the value.
+    init(owning cTensor: CTensor, count: Int) {
+        debugLog("Initializing TensorBuffer with a cTensor of \(count) elements.")
+        let actualCount = (0..<TF_NumDims(cTensor)).reduce(1) { accumulator, next in
+            accumulator * Int(TF_Dim(cTensor, next))
+        }
+        assert(actualCount == count)
```
bgogul: Can't we get rid of the `count` argument from the `init` as it is implied from `cTensor`?

pschuh: I believe the existing code wants to ensure that the count matches the shape as a consistency check.

PR #559, opened 2019-11-16
Remove TensorProtocol.

pschuh:
Unused and introduces extra dependencies on TensorHandle.


dan-zheng:
> This was only needed during the GPE days, I believe.

`TensorProtocol` was originally introduced as a unifying protocol when we had statically-ranked `Tensor{X}D` types in addition to the unranked `Tensor` type. We tried to generically define tensor operations on `TensorProtocol`, but in practice this led to ambiguity errors for the operations.

`TensorProtocol` became obsolete when we deleted the `Tensor{X}D` types!

PR #561, opened 2019-11-19
paddingIndex documentation for TransposedConv2D

Shashi456:


Shashi456:
Also I'd like to ask if I could start an Issue stating that we support NCWH format for layers as well, is it a wise idea to do that? or should we wait for all layers to be added first and then go onto this idea

burmako:
Thank you for the contribution! 

Speaking of working on NCWH format, let's start with a design document. How do you feel about putting together a writeup on what this work will entail, and then discussing that at one of the future open design meetings?

PR #571, opened 2019-11-28
Add new Array ops and reorganize Math.swift for linear algebra PRs

awav:
Hello everyone,

This PR has 3 parts:

1. `Math.swift` contains a mix of different functions. The file is already too long, so I decided to split it up on `Math.swift`, `Array.swift` and `Linalg.swift` as it is done in python TensorFlow.
   - `Math.swift` has elementary functions.
   - `Linalg.swift` has linear algebra operations
   - `Array.swift` has operations with matrix manipulations: diagonal, extract parts of a tensor and etc.
2. I added two more ops and their gradients: `diagonal` and `bandPart` and I also implemented gradients to the `diagonalPart` op.
3. The `diagonalPart` behavior was wrong and it didn't match [tf.linalg.diag_part](https://www.tensorflow.org/api_docs/python/tf/linalg/diag_part), so, I fixed it as well.

PS: this PR is preparation for linear algebra operations.

Thanks!


awav:
@googlebot I signed it!

Sources/TensorFlow/Operators/Array.swift
```diff
@@ -0,0 +1,103 @@
...
+    /// For example:
+    ///
+    /// ```
+    /// // 't' is [[1, 0, 0, 0]
+    /// //         [0, 2, 0, 0]
+    /// //         [0, 0, 3, 0]
+    /// //         [0, 0, 0, 4]]
+    /// t.diagonalPart()
+    /// // [1, 2, 3, 4]
+    ///
```
dan-zheng: minor: missing end backticks:
```suggestion
    /// ```
```

Sources/TensorFlow/Operators/Array.swift
```diff
@@ -0,0 +1,103 @@
...
+    /// ```
+    /// // 't' is [1, 2, 3, 4]
+    ///
+    /// t.diagonal()
+    /// // [[1, 0, 0, 0]
+    /// //  [0, 2, 0, 0]
+    /// //  [0, 0, 3, 0]
+    /// //  [0, 0, 0, 4]]
+    ///
```
dan-zheng:
```suggestion
    /// ```
```

Sources/TensorFlow/Operators/Array.swift
```diff
@@ -0,0 +1,103 @@
...
+    /// //  [-1,  0,  1, 2]
+    /// //  [ 0, -1,  0, 1]
+    /// //  [ 0,  0, -1, 0]]
+    ///
+    /// t.bandPart(2, 1)
+    /// // [[ 0,  1,  0, 0]
+    /// //  [-1,  0,  1, 0]
+    /// //  [-2, -1,  0, 1]
+    /// //  [ 0, -2, -1, 0]]
+    ///
```
dan-zheng:
```suggestion
    /// ```
```

Tests/TensorFlowTests/OperatorTests/LinalgTests.swift
```diff
@@ -0,0 +1,51 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import XCTest
+@testable import TensorFlow
+
+
+final class LinalgTests: XCTestCase {
```
dan-zheng:
```suggestion
final class LinearAlgebraTests: XCTestCase {
```

awav:
@dan-zheng, @Shashi456 

I renamed `Linalg.swift` and moved `Array.swift` content to the `LinearAlgebra.swift` file. I have mixed feelings though about making functions without arguments in field-like style:

```swift
    var diagonal: Tensor {
        _Raw.matrixDiag(diagonal: self)
    }
```

Many other similar functions in `Math.swift` are proper functions, e.g. `squared()`.

_Another question: what do you think about Cholesky as an extension to the Tensor, similar to QR decomposition?_

PS: https://github.com/tensorflow/swift-apis/commit/9548ea2fec10bb83ca8364ce8bcf82306dad9009 doesn't work for me, `derivative` is not recognized.

Shashi456:
@awav, are you on the latest nightly? `derivative` not working might be an issue because of the same.

dan-zheng:
> PS: [9548ea2](https://github.com/tensorflow/swift-apis/commit/9548ea2fec10bb83ca8364ce8bcf82306dad9009) doesn't work for me, `derivative` is not recognized. I had to revert it back.

If you're on macOS, please use the [latest 2019-11-27 toolchain](https://storage.googleapis.com/swift-tensorflow/mac/swift-tensorflow-DEVELOPMENT-2019-11-27-a-osx.pkg). We [released](https://github.com/tensorflow/swift/pull/332) that macOS toolchain just for the `@differentiating -> @derivative(of:)` renaming.

dan-zheng:
> I renamed `Linalg.swift` and moved `Array.swift` content to the `LinearAlgebra.swift` file. I have mixed feelings though about making functions without arguments in field-like style:
> 
> ```swift
>     var diagonal: Tensor {
>         _Raw.matrixDiag(diagonal: self)
>     }
> ```
> 
> Many other similar functions in `Math.swift` are proper functions, e.g. `squared()`.

Thanks for bringing up this good point!

Settling on a consistent style ("argument-less instance method" vs "computed property" vs "top-level function") for math functions like `Tensor.squared()` and `Tensor.diagonal` is important.

As you pointed out, almost of these math functions are currently defined as argument-less methods (or top-level functions) - computed properties are reserved for basic shape properties, like `Tensor.rank`, `Tensor.shape`, and `Tensor.scalarCount`.

How about reverting to `Tensor.diagonal()` and `Tensor.diagonalPart()` as instance functions for now, for consistency with existing APIs? Sorry for the thrash!

> _Another question: what do you think about Cholesky as an extension to the Tensor, similar to QR decomposition?_

Another interesting question. How about exposing both `cholesky` and `qrDecomposition` as top-level functions taking a `Tensor` argument, rather than `Tensor` instance methods?

My thought process:
* Instance methods are fluent. `Tensor` instance methods represent basic operations that that are conceptually simple enough to be expressed via a single verb: "*apply basic operation* to a tensor".
  * `Tensor.transposed()` -> "transpose"
  * `Tensor.mean()` -> "mean"
  * Some cannot be expressed as a single verb, but are still conceptually simple: `Tensor.diagonalPart()` -> "diagonal part" .
* Top-level functions are not fluent. These are used for basic elementwise operations (`abs(_:)`, `exp(_:)`, `mean(_:)`) - let's ignore these for now. Top-level functions can also represent more complex, domain-specific operations: "*perform non-trivial operation* to a tensor" or "*compute non-trivial value* from a tensor".
  * I think cholesky decomposition and QR decomposition both fit in this category.

Tests/TensorFlowTests/OperatorTests/LinearAlgebraTests.swift
```diff
@@ -0,0 +1,67 @@
...
+        let permutations = [[1, 0], [0, 2, 1], [0, 1, 3, 2]] // To avoid permuting batch dimensions.
+        for (shape, permutation) in zip(shapes, permutations) {
+            let a = Tensor<Float>(randomNormal: TensorShape(shape))
+            let x = matmul(a, a.transposed(permutation: permutation)) // Make `a` positive-definite.
+            let l = cholesky(x)
+            let xReconstructed = matmul(l, l.transposed(permutation: permutation))
+            assertEqual(xReconstructed, x, accuracy: 1e-5)
+        }
+
+        // The expected value of the gradient was computed using the following Python code:
```
Shashi456: @dan-zheng Do you think we should separate the test for cholesky function and its gradient? IMO unit testing is better when you test just for one thing. and if the cholesky function implementation undergoes any change in the future, such distinction can help locate any downstream errors.

dan-zheng: I think not separating is fine! I don't feel a strong need to have separate files or a massive file just for testing derivatives.

awav:
> If you're on macOS, please use the latest 2019-11-27 toolchain. We released that macOS toolchain just for the @differentiating -> @derivative(of:) renaming.

This worked, thanks!

> How about reverting to Tensor.diagonal() and Tensor.diagonalPart() as instance functions for now, for consistency with existing APIs? Sorry for the thrash!

No problem, I reverted it.

> As you pointed out, most of these math functions are currently defined as argument-less methods (or top-level functions) - computed properties are reserved for basic shape properties, like Tensor.rank, Tensor.shape, and Tensor.scalarCount

> Another interesting question. How about exposing both Cholesky and QR decomposition as top-level functions taking a Tensor argument, rather than Tensor instance methods?

Should I leave the changes as-is for now? I don't think it a scope of this PR.
In short, my initial thoughts about the Tensor API.
* I do like the variable-like style for argument-less functions.
* Cholesky for matrices is the same op as square root for scalars, so it makes sense to make Cholesky as a tensor instance method. Similar logic applies to many decompositions and other linear algebra operations.
* How about overloading functions for the new Tensor classes. So that new implementation works without changing the code.

dan-zheng:
> * Cholesky for matrices is the same op as square root for scalars, so it makes sense to make Cholesky as a tensor instance method. Similar logic applies to many decompositions and other linear algebra operations.

Standardizing on instance methods sounds fine to me!

> * How about overloading functions for the new Tensor classes. So that new implementation works without changing the code.

I’m not sure what’s meant by this, could you please clarify?

awav:
> I’m not sure what’s meant by this, could you please clarify?

The question is more about what is the best option: function overloading vs function overriding in a user perspective. Let's say a user creates custom tensor so that it behaves exactly like a standard one, but there is a more efficient operation, e.g. Cholesky for this object. And the user wants to override it without modifying the existing code. Is that better to have Cholesky as an instance-based method or top-level function in the first place?

@dan-zheng, @Shashi456 any other comments about PR's changes?

Thanks!

Shashi456:
LGTM.

awav:
@dan-zheng, ping :)

dan-zheng:
> The question is more about what is the best option: function overloading vs function overriding in a user perspective. Let's say a user creates custom tensor so that it behaves exactly like a standard one, but there is a more efficient operation, e.g. Cholesky for this object. And the user wants to override it without modifying the existing code. Is that better to have Cholesky as an instance-based method or top-level function in the first place?

Sorry! I'm still not sure I understand your question, perhaps I'm lacking some context.

`Tensor` cannot be subclassed because it is a Swift `struct`, not a `class`. We currently don't have a story for "custom tensor implementations" or "multiple `Tensor` backends", so `Tensor` operations currently use `TensorFlow._Raw` operations as the implementation.

dan-zheng:
Thanks for the ping!

As discussed, could you please change `cholesky(_:)` to an instance method `Tensor.cholesky()` (and standardizing on instance methods for `Tensor` math operations)? After that, I think this patch is ready to merge. We'll try to review follow-up PRs more quickly!

awav:
> Sorry! I'm still not sure I understand your question, perhaps I'm lacking some context.

Long story short. I would like to implement `https://github.com/getkeops/keops`, to do that I need a custom tensor. Could you advise where I can post the question about it?

> As discussed, could you please change `cholesky(_:)` to an instance method `Tensor.cholesky()` (and standardizing on instance methods for Tensor math operations)? After that, I think this patch is ready to merge. We'll try to review follow-up PRs more quickly!

Only Cholesky or together with QR decomposition?

awav:
@dan-zheng, after some thinking, guess, we shouldn't rush with it. I would keep the same Cholesky as is and make a poll or ask other people about the change. E.g. PyTorch, TensorFlow, Scipy, Numpy -  these frameworks implement linear algebra operations as top-level functions.

dan-zheng:
> Long story short. I would like to implement `https://github.com/getkeops/keops`, to do that I need a custom tensor. Could you advise where I can post the question about it?

I see. Please send questions to the [Swift for TensorFlow mailing list](https://groups.google.com/a/tensorflow.org/forum/#!forum/swift)!

dan-zheng:
> @dan-zheng, after some thinking, guess, we shouldn't rush with it. I would keep the same Cholesky as is and make a poll or ask other people about the change. E.g. PyTorch, TensorFlow, Scipy, Numpy - these frameworks implement linear algebra operations as top-level functions.

That sounds fine to me! This PR then leaves `cholesky(_:)` and `Tensor.qrDecomposition(fullMatrices:)` unchanged and only changes their file location.

awav:
Thanks, @dan-zheng, @Shashi456!

awav:
Is that ready for merge? I have a couple of PR waiting for this :)

dan-zheng:
> Is that ready for merge? I have a couple of PR waiting for this :)

Yes! Merging now.

awav:
Awesome! Thanks!

SumanSudhir:
`newbie`
Can you please guide me on how to use this implemented trace function now in implementing another linear algebra operator and test it.

PR #335, opened 2019-07-08
[Initializers] Refactor random and variance-scaling initializers

jon-tow:
# Summary 
### Random Initializers
* Adds tests, `testRandomUniform()`, `testRandomNormal()`, and `testTruncatedNormal()`.
### Variance-Scaling Initializers
* Refactor computation of fans, `fanIn` and `fanOut`.
* Add paper references to `glorot` initializers.
* Update `glorotNormal` to use `truncatedNormal` initialization.
* Add tests, `testGlorotUniform()` and `testGlorotNormal()`.

PR #576, opened 2019-12-06
Correcting calculation of Glorot uniform for Tensors

BradLarson:
During the refactoring of random initializers in PR #335, the limits on the Glorot uniform were missing a sqrt() around their values. This led to an incorrect distribution being calculated. This was missed in the tests introduced in that PR due to a too-large absolute value for the threshold.

This has been fixed, and the threshold is now able to be set per-test, so it is set to a tight enough value to catch a regression. Additionally, the previous values of the RNN and LSTM layer tests were correct, and are reverted here to what they were before they were changed in the above PR. The Sequential test values have been changed to match the results given by this new distribution, but I can't independently verify those.

I believe this fixes [TF-1029](https://bugs.swift.org/browse/TF-1029).


brettkoonce:
@BradLarson testing, thanks again for looking into this!

brettkoonce:
@BradLarson tested locally, works for me!  👍

PR #577, opened 2019-12-07
Fix deprecation warnings for method-style differential operators.

dan-zheng:
Before:
```
/swift-models/Gym/CartPole/main.swift:182:13: warning: 'gradient(in:)' is deprecated: Method-style differential operators are deprecated and will be removed; use top-level
function 'TensorFlow.gradient(at:in:)' instead
        net.gradient { net -> Tensor<Float> in
            ^
```

After:
```
/swift-models/Gym/CartPole/main.swift:182:13: warning: 'gradient(in:)' is deprecated: Method-style differential operators are deprecated and will be removed; use top-level function 'TensorFlow.gradient(at:in:)' instead
        net.gradient { net -> Tensor<Float> in
            ^
```

### DIFF
--- a/Tests/TensorFlowTests/LayerTests.swift
+++ b/Tests/TensorFlowTests/LayerTests.swift
@@ -1173,7 +1173,7 @@ final class LayerTests: XCTestCase {
             let inputs: [Tensor<Float>] = Array(repeating: x, count: 4)
             let rnn = RNN(LSTMCell<Float>(inputSize: 4, hiddenSize: 4))
             withTensorLeakChecking {
-                let (outputs, _) = rnn.valueWithPullback(at: inputs) { rnn, inputs in
+                let (outputs, _) = valueWithPullback(at: rnn, inputs) { rnn, inputs in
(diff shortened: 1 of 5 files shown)


Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -1204,7 +1204,7 @@ final class LayerTests: XCTestCase {
           biasInitializer: zeros())
         )
         withTensorLeakChecking {
-            let (outputs, _) = rnn.valueWithPullback(at: inputs) { rnn, inputs in
+            let (outputs, _) = valueWithPullback(at: rnn, inputs) { rnn, inputs in
```
Shashi456: @dan-zheng you missed a `valueWithPullback` on line 1176.

dan-zheng: Thanks for the catch! Done in https://github.com/tensorflow/swift-apis/pull/577/commits/0e75b1b91296c823238640a6c51f2a5129b23054.

PR #578, opened 2019-12-07
[Linear Algebra] Add eye and svd functions

Shashi456:
~So I've not written a test for svd because, I'm not sure how to calculate the permutation for a standard transpose for any given matrix to invoke `.tranposed`
Since for a tolerance check, i'll have to calculate the reconstructed matrix as `u*diag(s) * transpose(v)`.~ Will have the test uploaded soon. 

I took inspiration for the eye function from [here](https://github.com/tensorflow/tensorflow/blob/1cf0898dd4331baf93fe77205550f2c2e6c90ee5/tensorflow/python/ops/linalg_ops_impl.py#L33). 
The issue with this implementation though is that `numCols`, `batchShape` need to be optional, but there is no convenient way for optional parameters in swift. and declaring it as `numCols: Int? = nil` doesn't exactly compile. So i wanted to ask if i should just overload the function/have multiple function declarations. 

Finally, added the `setDiagonal` function, it's functionality is different from `set_matrix_diag` of tensorflow because that implementation uses `setMatrixDiagV2` which is not available in the rawopsgenerated file.

### DIFF
--- a/Sources/TensorFlow/Operators/LinearAlgebra.swift
+++ b/Sources/TensorFlow/Operators/LinearAlgebra.swift
@@ -56,12 +56,12 @@ public extension Tensor where Scalar: TensorFlowNumeric {
         _Raw.matrixDiag(diagonal: self)
     }
 
-    /// Returns a batched matrix tensor with new batched diagonal values.
-    /// Given the input tensor and diagonal, this operation returns a tensor with the same
-    /// shape and values as the input, except for the specified diagonals of the innermost matrices
-    /// which will be overwritten by the values in diagonal.
+    /// Returns `self` with new diagonal values, given that `self` is an optionally batched matrix.
     ///
-    /// Parameter diagonal: A tensor with rank `rank - 1`.
+    /// The returned tensor has the same shape and values as `self`, except for the specified
+    /// diagonals of the innermost matrices which are overwritten by the values in `diagonal`.
+    ///
+    /// Parameter diagonal: A tensor with rank `rank - 1` representing the new diagonal values.
     @inlinable
     func withDiagonal(_ diagonal: Tensor<Scalar>) -> Tensor {
         _Raw.matrixSetDiag(self, diagonal: diagonal)
@@ -145,7 +145,6 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
 ///   - rowCount: The number of rows in each batch matrix.
 ///   - columnCount: The number of columns in each batch matrix.
 ///   - batchShape: The leading batch dimensions of the returned tensor.
-/// TODO: Make columnCount and batchShape optional.
 public func eye<Scalar: Numeric>(
     rowCount: Int,
     columnCount: Int? = nil,
@@ -153,21 +152,14 @@ public func eye<Scalar: Numeric>(
 ) -> Tensor<Scalar> {
     let columnCount = columnCount ?? rowCount
     let diagonalSize = min(rowCount, columnCount)
-    var diagShape = [diagonalSize]
-    if(!batchShape.isEmpty) {
-        diagShape = batchShape + diagShape
-    }
-    let diagonalOnes = Tensor<Scalar>(ones: TensorShape(diagShape))
+    let diagonalShape = batchShape + [diagonalSize]
+    let diagonalOnes = Tensor<Scalar>(ones: TensorShape(diagonalShape))
     if rowCount == columnCount {
         return diagonalOnes.diagonal()
-    } else {
-        var shape = [rowCount, columnCount]
-        if(!batchShape.isEmpty) {
-        shape = batchShape + shape
-        }
-        let zeroMatrix = Tensor<Scalar>(zeros: TensorShape(shape))
-        return zeroMatrix.withDiagonal(diagonalOnes)
     }
+    let shape = batchShape + [rowCount, columnCount]
+    let zeroMatrix = Tensor<Scalar>(zeros: TensorShape(shape))
+    return zeroMatrix.withDiagonal(diagonalOnes)
 }
 
 /// Computes the trace of an optionally batched matrix.
@@ -250,7 +242,7 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
     ///     computed, which can be significantly faster.
     ///   - fullMatrices:  If `true`, compute full-sized `u` and `v`. If `false`, compute only the
     ///     leading `min(shape[rank - 1], shape[rank - 2])` singular vectors. Ignored if
-    //      `computeUv` is `false`.
+    //      `computeUV` is `false`.
     @inlinable
     func svd(computeUV: Bool = true, fullMatrices: Bool = false
     ) -> (s: Tensor<Scalar>, u: Tensor<Scalar>?, v: Tensor<Scalar>?) {
--- a/Sources/TensorFlow/Operators/LinearAlgebra.swift
+++ b/Sources/TensorFlow/Operators/LinearAlgebra.swift
@@ -199,10 +199,10 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
     //      `computeUv` is `false`.
     @inlinable
     func svd(computeUv: Bool = true, fullMatrices: Bool = false
-    ) -> (s: Tensor<Scalar>, u: Tensor<Scalar>, v: Tensor<Scalar>) {
+    ) -> (s: Tensor<Scalar>, u: Tensor<Scalar>?, v: Tensor<Scalar>?) {
         let (s, u, v) = _Raw.svd(self, computeUv: computeUv, fullMatrices: fullMatrices)
         if !computeUv {
-            return s
+            return (s, nil, nil)
         }
         return (s, u, v)
     }
--- a/Sources/TensorFlow/Operators/LinearAlgebra.swift
+++ b/Sources/TensorFlow/Operators/LinearAlgebra.swift
@@ -231,21 +231,34 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
         _Raw.qr(self, fullMatrices: fullMatrices)
     }
 
-    /// Returns the singular value decompositions of one or more matrices.
+    /// Returns the singular value decomposition of `self`, given that `self` is an optionally
+    /// batched matrix.
     ///
-    /// Computes the SVD of each inner matrix in the `input` tensor such that
-    /// `input[..., :, :] = u[..., :, :] * diag(s[..., :, :]) * transpose(v[..., :, :])`
+    /// The singular value decomposition (SVD) of the optionally batched matrix `self` is values
+    /// `s`, `u`, and `v`, such that:
+    ///
+    ///     self[..., :, :] = u[..., :, :] • s[..., :, :].diagonal() • v[..., :, :].transposed()`
+    ///
+    /// self` must be a tensor with shape `[..., M, N]`. Let `K = min(M, N)`.
     ///
     /// - Parameters:
-    ///   - computeUV: If `true`, left and right singular vectors will be computed and
-    ///     returned as `u` and `v`, respectively. Otherwise, only the singular values will be
-    ///     computed, which can be significantly faster.
-    ///   - fullMatrices:  If `true`, compute full-sized `u` and `v`. If `false`, compute only the
-    ///     leading `min(shape[rank - 1], shape[rank - 2])` singular vectors. Ignored if
-    //      `computeUV` is `false`.
+    ///   - computeUV: If `true`, the left and right singular vectors are computed and returned as
+    ///     `u` and `v`, respectively. If `false`, `nil` values are returned as `u` and `v`.
+    ///   - fullMatrices: If `true`, `u` and `v` respectively have shapes `[..., M, M]` and
+    ///     `[..., N, N]`. If `false`, `u` and `v` respectively have shapes `[..., M, K]` and
+    ///     `[..., K, N]`. Ignored when `computeUV` is false.
+    ///
+    /// - Returns:
+    ///   - s: The singular values, with shape `[..., K]`. Within each vector, the singular values
+    ///     are sorted in descending order.
+    ///   - u: The left singular vectors.
+    ///   - v: The right singular vectors.
+    ///
+    /// - Precondition: `self` must be a tensor with shape `[..., M, N]`.
     @inlinable
-    func svd(computeUV: Bool = true, fullMatrices: Bool = false
-    ) -> (s: Tensor<Scalar>, u: Tensor<Scalar>?, v: Tensor<Scalar>?) {
+    func svd(computeUV: Bool = true, fullMatrices: Bool = false) -> (
+        s: Tensor<Scalar>, u: Tensor<Scalar>?, v: Tensor<Scalar>?
+    ) {
         let (s, u, v) = _Raw.svd(self, computeUv: computeUV, fullMatrices: fullMatrices)
         if !computeUV {
             return (s, nil, nil)
--- a/Sources/TensorFlow/Operators/LinearAlgebra.swift
+++ b/Sources/TensorFlow/Operators/LinearAlgebra.swift
@@ -30,7 +30,7 @@ public extension Tensor where Scalar: TensorFlowNumeric {
     /// // [1, 2, 3, 4]
     /// ```
     @inlinable
-    @differentiable(wrt: self, vjp: _vjpDiagonalPart where Scalar: TensorFlowFloatingPoint)
+    @differentiable(where Scalar: TensorFlowFloatingPoint)
     func diagonalPart() -> Tensor {
         precondition(rank >= 2, "The tensor must have at least rank 2.")
         return _Raw.matrixDiagPart(self)
@@ -51,11 +51,12 @@ public extension Tensor where Scalar: TensorFlowNumeric {
     /// //  [0, 0, 0, 4]]
     /// ```
     @inlinable
-    @differentiable(wrt: self, vjp: _vjpDiagonal where Scalar: TensorFlowFloatingPoint)
+    @differentiable(where Scalar: TensorFlowFloatingPoint)
     func diagonal() -> Tensor {
         _Raw.matrixDiag(diagonal: self)
     }
 
+<<<<<<< HEAD
     /// Returns a batched matrix tensor with new batched diagonal values.
     /// Given the input tensor and diagonal, this operation returns a tensor with the same
     /// shape and values as the input, except for the specified diagonals of the innermost matrices
@@ -67,6 +68,15 @@ public extension Tensor where Scalar: TensorFlowNumeric {
         _Raw.matrixSetDiag(self, diagonal: diagonal)
     }
 
+||||||| merged common ancestors
+=======
+    @available(*, deprecated, renamed: "bandPart(subdiagonalCount:superdiagonalCount:)")
+    @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
+    func bandPart(_ subdiagonalCount: Int, _ superdiagonalCount: Int) -> Tensor {
+        return bandPart(subdiagonalCount: subdiagonalCount, superdiagonalCount: superdiagonalCount)
+    }
+
+>>>>>>> 7330b8d7106489c906f85b92c8f5471f6959b10e
     /// Returns a copy of a innermost tensor defined by a central band boundaries.
     /// The output is a tensor of the same shape as the instance `[..., :, :]`.
     ///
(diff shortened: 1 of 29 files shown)


Shashi456:
Is there any neat of doing this in swift: 
```python
    m = u.shape[-1]
    n = v.shape[-1]
    if m <= n:
      v = v[..., :m]
    else:
      u = u[..., :n]
```

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -106,6 +117,31 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
...
 
+/// Returns an identity matrix or a batch of matrices.
+///
+/// - Parameters:
+///   - numRows: A non-negative integer giving the number of rows in each batch matrix.
+///   - numCols: A non-negative integer giving the number of rows in each batch matrix.
+///   - batchShape: A list of integers. The returned tensor will have leading batch dimensions
+///     of this shape.
+public func eye<Scalar: Numeric>(
+    numRows: Int,
```
dan-zheng: Variables related to the "number of things" are named `xxxCount` in Swift, e.g. [`Array.count`](https://developer.apple.com/documentation/swift/array/2943906-count) and [`Tensor.scalarCount`](https://www.tensorflow.org/swift/api_docs/Structs/Tensor#scalarcount).
```suggestion
    rowCount: Int,
```

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -106,6 +117,31 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+    numCols: Int,
```
```suggestion
    columnCount: Int,
```

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -56,6 +56,17 @@ public extension Tensor where Scalar: TensorFlowNumeric {
...
     }
 
+    /// Returns a batched matrix tensor with new batched diagonal values.
+    /// Given the input tensor and diagonal, this operation returns a tensor with the same
+    /// shape and values as the input, except for the specified diagonals of the innermost matrices
+    /// which will be overwritten by the values in diagonal.
+    ///
+    /// Parameter diagonal: A tensor with rank k.
+    @inlinable
+    func setDiagonal(diagonal: Tensor<Scalar>) -> Tensor {
```
dan-zheng: No `diagonal:` argument label is necessary because the method name already contains "diagonal".
```suggestion
    func setDiagonal(_ diagonal: Tensor<Scalar>) -> Tensor {
```

Shashi456:
@dan-zheng could you comment on two things : 
1) The optionality of no of columns and no of batches in the `eye` function
2) If there's a neat way of doing the following with tensors, since  array indexing is not very easy. Although i wrote a reconstruction test, it would only work for symmetric matrices 
``` python
    m = u.shape[-1]
    n = v.shape[-1]
    if m <= n:
      v = v[..., :m]
    else:
      u = u[..., :n]
```

dan-zheng:
> 1. The optionality of no of columns and no of batches in the `eye` function

Making both column count and batch shape optional sounds good, matching [`tf.eye`](https://www.tensorflow.org/api_docs/python/tf/eye):

```swift
public func eye<Scalar: Numeric>(
    rowCount: Int,
    columnCount: Int? = nil,
    batchShape: [Int]? = []
) {
    let columnCount = columnCount ?? rowCount
    ...
}  
```

> 2. If there's a neat way of doing the following with tensors, since  array indexing is not very easy. Although i wrote a reconstruction test, it would only work for symmetric matrices
> 
> ```python
>     m = u.shape[-1]
>     n = v.shape[-1]
>     if m <= n:
>       v = v[..., :m]
>     else:
>       u = u[..., :n]
> ```

I believe this is equivalent:
```swift
import TensorFlow
var u = Tensor<Float>(ones: [3, 5, 7])
var v = Tensor<Float>(ones: [3, 5])
let m = u.shape.dimensions.last!
let n = v.shape.dimensions.last!
if m <= n {
    v = v[TensorRange.ellipsis, ..<m]
} else {
    u = u[TensorRange.ellipsis, ..<n]
}
```

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -106,6 +117,31 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+///     of this shape.
+public func eye<Scalar: Numeric>(
+    numRows: Int,
+    numCols: Int,
+    batchShape: [Int]
+) -> Tensor<Scalar> {
+    let diagonalSize = min(numRows, numCols)
+    let diagShape = batchShape + [diagonalSize]
+    let diagonalOnes = Tensor<Scalar>(ones: TensorShape(diagShape))
+    if numRows == numCols{
```
awav:
```suggestion
    if numRows == numCols {
```

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -106,6 +117,31 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+        return diagonalOnes.diagonal()
+    }
+    else{
```
awav:
```suggestion
    else {
```

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -148,4 +184,26 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+    ///
+    /// - Parameters:
+    ///   - computeUv: If `true` then left and right singular vectors will be computed and
+    ///     returned in `u` and `v`, respectively. Otherwise, only the singular values will be
+    ///     computed, which can be significantly faster.
+    ///   - fullMatrices:  If `true`, compute full-sized `u` and `v`. If `false` compute only the
+    ///     leading `min(shape[rank - 1], shape[rank - 2])` singular vectors. Ignored if
+    //      `computeUv` is `false`.
+    @inlinable
+    func svd(computeUv: Bool = true, fullMatrices: Bool = false
```
awav: Can you rename `computeUv` to `computeUV`, `uv` or something more readable? It might follow the language style, but it is slightly obscure, what is meant by `computeUv`.

Shashi456: I think I took this variable name from the raw ops generated. I do agree that it sounds sort of vague. 

But would like @dan-zheng's idea on this

Shashi456:
I'll add the tests and other changes today.

Shashi456:
Will write eye and set diagonal functions next.

Shashi456:
I'm currently looking at optionality in swift for the eye function, Will write the test as soon as i find a convenient solution.

Shashi456:
@dan-zheng added the tests and todo for eye function optionality. I'll get around to that soon, but did not want to block this PR.

SumanSudhir:
Don't you think  that it would be better if you write eye function like this
` func eye(_ num_rows: Int, _ num_columns: Int = 0)`

if we give it input eye(3) or eye(3,3) then it will give the same output 
`[[1.0, 0.0, 0.0],` 
 `[0.0, 1.0, 0.0],`   
 `[0.0, 0.0, 1.0]]`    

but if you give it input as eye(3,2) then output will be
`[[1.0, 0.0],`
` [0.0, 1.0],`
` [0.0, 0.0]]`

There is no batch concept in either numpy or pytorch in eye function but if it is needed then it can be added as.
`` func eye(_ num_rows: Int, _ num_columns: Int = 0, batch_shape: [Int])``
and in a function we can use it as
`if(!batch_shape.isEmpty){`
`}`

I have implemented it to contribute to swift and tested it on different test cases but since I am new to open source I don't know how to contribute full function code from scratch.

Shashi456:
@SumanSudhir the eye function trying to be replicated is that of tensorflow python. I'll take a look at your approach and implement it tomorrow.

SumanSudhir:
Hey, @Shashi456 if you don't mind then can I pull request for `eye` function. I have tested it and it is working as expected, it can take input with  `number of columns` and `batch shape` as an optional parameter and give output in any datatype format as required

Some examples of tested input and output 

`let x: Tensor<Int32> = eye(4)`
`[[1, 0, 0, 0],`
 `[0, 1, 0, 0],`
 `[0, 0, 1, 0],`
 `[0, 0, 0, 1]]`

`let x: Tensor<Int32> = eye(4,5)`
`[[1, 0, 0, 0, 0],`
 `[0, 1, 0, 0, 0],`
 `[0, 0, 1, 0, 0],`
 `[0, 0, 0, 1, 0]]`

`let x: Tensor<Int32> = eye(4,batch_shape: [1])`
`[[[1, 0, 0, 0],`
  `[0, 1, 0, 0],`
  `[0, 0, 1, 0],`
  `[0, 0, 0, 1]]]`

`let x: Tensor<Int32> = eye(4,5,batch_shape: [1])`
`[[[1, 0, 0, 0, 0],`
  `[0, 1, 0, 0, 0],`
  `[0, 0, 1, 0, 0],`
 ` [0, 0, 0, 1, 0]]]`

`let x: Tensor<Float32> = eye(4,5,batch_shape: [1])`
`[[[1.0, 0.0, 0.0, 0.0, 0.0],`
  `[0.0, 1.0, 0.0, 0.0, 0.0],`
  `[0.0, 0.0, 1.0, 0.0, 0.0],`
  `[0.0, 0.0, 0.0, 1.0, 0.0]]]`

`let x: Tensor<Float32> = eye(4,5,batch_shape: [1,2])`
`[[[[1.0, 0.0, 0.0, 0.0, 0.0],`
   `[0.0, 1.0, 0.0, 0.0, 0.0],`
   `[0.0, 0.0, 1.0, 0.0, 0.0],`
   `[0.0, 0.0, 0.0, 1.0, 0.0]],`
`                                       `
  `[[1.0, 0.0, 0.0, 0.0, 0.0],`
  ` [0.0, 1.0, 0.0, 0.0, 0.0],`
  ` [0.0, 0.0, 1.0, 0.0, 0.0],`
   `[0.0, 0.0, 0.0, 1.0, 0.0]]]]`

Shashi456:
@SumanSudhir could you update this function after this PR is merged? That sounds like a better idea.

Shashi456:
@dan-zheng could you review this once more? Also happy new year :)

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -128,6 +139,32 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+    columnCount: Int,
+    batchShape: [Int]
+) -> Tensor<Scalar> {
+    let diagonalSize = min(rowCount, columnCount)
+    let diagShape = batchShape + [diagonalSize]
+    let diagonalOnes = Tensor<Scalar>(ones: TensorShape(diagShape))
+    if rowCount == columnCount {
+        return diagonalOnes.diagonal()
+    }
+    else {
```
dan-zheng: Please move `else` to previous line: `} else {`.

Tests/TensorFlowTests/OperatorTests/MatrixTests.swift
```diff
@@ -112,10 +121,23 @@ final class MatrixTests: XCTestCase {
         let expectedGrad = 2 * t3.bandPart(subdiagonalCount: 0, superdiagonalCount: -1)
         XCTAssertEqual(computedGrad, expectedGrad)
     }
+
+    func testEye() {
+        // Test for symmetric identity matrix.
+        var identity: Tensor<Float> = eye(rowCount: 2, columnCount: 2, batchShape:[1])
```
dan-zheng:
```suggestion
        var identity: Tensor<Float> = eye(rowCount: 2, columnCount: 2, batchShape: [1])
```

Tests/TensorFlowTests/OperatorTests/MatrixTests.swift
```diff
@@ -112,10 +121,23 @@ final class MatrixTests: XCTestCase {
...
+        XCTAssertEqual(identity, [[[1.0, 0.0],
+                                   [0.0, 1.0]]])
+        // Test for non-symmetric identity matrix.
+        identity = eye(rowCount: 2, columnCount: 3, batchShape:[1])
```
dan-zheng:
```suggestion
        identity = eye(rowCount: 2, columnCount: 3, batchShape: [1])
```

Tests/TensorFlowTests/OperatorTests/MatrixTests.swift
```diff
@@ -112,10 +121,23 @@ final class MatrixTests: XCTestCase {
         let expectedGrad = 2 * t3.bandPart(subdiagonalCount: 0, superdiagonalCount: -1)
         XCTAssertEqual(computedGrad, expectedGrad)
     }
+
+    func testEye() {
+        // Test for symmetric identity matrix.
+        var identity: Tensor<Float> = eye(rowCount: 2, columnCount: 2, batchShape:[1])
+        XCTAssertEqual(identity, [[[1.0, 0.0],
```
dan-zheng: Minor: the trailing `.0` can be omitted from floating-point literals.
```suggestion
        XCTAssertEqual(identity, [[[1, 0],
```

Same below please.

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -56,6 +56,17 @@ public extension Tensor where Scalar: TensorFlowNumeric {
...
     }
 
+    /// Returns a batched matrix tensor with new batched diagonal values.
+    /// Given the input tensor and diagonal, this operation returns a tensor with the same
+    /// shape and values as the input, except for the specified diagonals of the innermost matrices
+    /// which will be overwritten by the values in diagonal.
+    ///
+    /// Parameter diagonal: A tensor with rank k.
+    @inlinable
+    func setDiagonal(_ diagonal: Tensor<Scalar>) -> Tensor {
```
dan-zheng: `withDiagonal` is a more idiomatic Swift name for this operation.
`setDiagonal` sounds like the name of a mutating function.

```suggestion
    func withDiagonal(_ diagonal: Tensor<Scalar>) -> Tensor {
```

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -56,6 +56,17 @@ public extension Tensor where Scalar: TensorFlowNumeric {
         _Raw.matrixDiag(diagonal: self)
     }
 
+    /// Returns a batched matrix tensor with new batched diagonal values.
+    /// Given the input tensor and diagonal, this operation returns a tensor with the same
+    /// shape and values as the input, except for the specified diagonals of the innermost matrices
+    /// which will be overwritten by the values in diagonal.
+    ///
+    /// Parameter diagonal: A tensor with rank k.
```
```suggestion
    /// Parameter diagonal: A tensor with rank `rank - 1`.
```

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -196,4 +233,26 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
...
         _Raw.qr(self, fullMatrices: fullMatrices)
     }
+
+    /// Returns the singular value decompositions of one or more matrices.
+    ///
+    /// Computes the SVD of each inner matrix in the `input` tensor such that
+    /// `input[..., :, :] = u[..., :, :] * diag(s[..., :, :]) * transpose(v[..., :, :])`
+    ///
+    /// - Parameters:
+    ///   - computeUv: If `true` then left and right singular vectors will be computed and
```
```suggestion
    ///   - computeUv: If `true`, left and right singular vectors will be computed and
```

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -196,4 +233,26 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+    ///     returned in `u` and `v`, respectively. Otherwise, only the singular values will be
```
```suggestion
    ///     returned as `u` and `v`, respectively. Otherwise, only the singular values will be
```

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -196,4 +233,26 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+    ///     computed, which can be significantly faster.
+    ///   - fullMatrices:  If `true`, compute full-sized `u` and `v`. If `false` compute only the
```
```suggestion
    ///   - fullMatrices:  If `true`, compute full-sized `u` and `v`. If `false`, compute only the
```

Shashi456:
@dan-zheng sorry for the trivial formatting errors, I'll have the changes up in a few hours.

Shashi456:
@dan-zheng done.

Shashi456:
@dan-zheng ping. Built and tested locally.

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -128,6 +139,37 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+    let columnCount = columnCount ?? rowCount
+    let diagonalSize = min(rowCount, columnCount)
+    var diagShape = [diagonalSize]
+    if(!batchShape.isEmpty) {
+        diagShape = batchShape + diagShape
+    }
+    let diagonalOnes = Tensor<Scalar>(ones: TensorShape(diagShape))
+    if rowCount == columnCount {
+        return diagonalOnes.diagonal()
+    } else {
```
dan-zheng: After an early `return` in an `if` statement, no `else` clause is necessary. Please move code to the same scope as the `if` statement to avoid unnecessary indentation.

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -128,6 +139,37 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
     }
 }
 
+/// Returns an identity matrix or a batch of matrices.
+///
+/// - Parameters:
+///   - rowCount: The number of rows in each batch matrix.
+///   - columnCount: The number of columns in each batch matrix.
+///   - batchShape: The leading batch dimensions of the returned tensor.
+/// TODO: Make columnCount and batchShape optional.
```
dan-zheng: Please remove obsolete comment:
```suggestion
```

Aside: `//` should be used for `TODO` comments. `///` is reversed for API documentation comments.

Sources/TensorFlow/Operators/LinearAlgebra.swift
dan-zheng: I think `computeUV` is the better name, since `Uv` is more easily misinterpreted as one camelcased entity than `UV`. Could you please update the parameter name and doc comments?

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -196,4 +238,26 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+    ///   - computeUv: If `true`, left and right singular vectors will be computed and
+    ///     returned as `u` and `v`, respectively. Otherwise, only the singular values will be
+    ///     computed, which can be significantly faster.
+    ///   - fullMatrices:  If `true`, compute full-sized `u` and `v`. If `false`, compute only the
+    ///     leading `min(shape[rank - 1], shape[rank - 2])` singular vectors. Ignored if
+    //      `computeUv` is `false`.
+    @inlinable
+    func svd(computeUv: Bool = true, fullMatrices: Bool = false
+    ) -> (s: Tensor<Scalar>, u: Tensor<Scalar>?, v: Tensor<Scalar>?) {
+        let (s, u, v) = _Raw.svd(self, computeUv: computeUv, fullMatrices: fullMatrices)
```
dan-zheng: Is it possible to return the output of `_Raw.svd` directly? That simplifies the body to one line.
```suggestion
        return _Raw.svd(self, computeUv: computeUv, fullMatrices: fullMatrices)
```
The `if !computeUv { ... }` seems redundant to me, if `_Raw.svd` returns `nil` values for `u` and `v` when `computeUV` is `false`.

Sources/TensorFlow/Operators/LinearAlgebra.swift
dan-zheng: A semantic name like `computeLeftAndRightSingularVectors` may be more idiomatic than `computeUV`. I'm not sure what's the best semantic name, so using `computeUV` for now sounds good to me, to be consistent with [`numpy.svd`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.linalg.svd.html) and [`tf.linalg.svd`](https://www.tensorflow.org/api_docs/python/tf/linalg/svd).

Sources/TensorFlow/Operators/LinearAlgebra.swift
dan-zheng: Edit: the current logic seems necessary and fine!

Returning the output of `_Raw.svd` directly doesn't work. When `computeUV` is false, the non-optional results `u` and `v` are undefined:

```swift
import TensorFlow
let x = Tensor<Float>(ones: [2, 2, 2])
print(_Raw.svd(x, computeUv: false))
```

```console
$ swift svd.swift
(s: [[1.9999999,       0.0],
 [1.9999999,       0.0]], u: 0.0, v: 0.0)
```

They seem to be zero tensors for this example.

Shashi456:
@dan-zheng, will follow suit for linear algebra functions. Since it was my first time contributing to this area, I wasn't entirely sure of formatting. Sorry!

PR #579, opened 2019-12-12
[WIP] Equations for losses

joaogui1:
As discussed in the google group I'm adding equations for the losses, but I'm afraid that without LaTeX support it may be a little bit unreadable, what do you think? 
I have added equations for the most common losses as of now, after feedback I can write the rest.


Sources/TensorFlow/Loss.swift
```diff
@@ -13,6 +13,9 @@
 // limitations under the License.
 
 /// Returns the L1 loss between predictions and expectations.
+/// Given the x and y vectors L1 loss computes:
```
BradLarson:
```suggestion
/// Given the x and y vectors, the L1 loss is computed as follows:
```

Sources/TensorFlow/Loss.swift
```diff
@@ -13,6 +13,9 @@
...
+///  `reduction([l_1, l_2, ..., l_N]^T)`, where `l_i = |x_i - y_i|`
+///   and `N` is yhe batch size
```
BradLarson:
```suggestion
///   and `N` is the batch size.
```

Sources/TensorFlow/Loss.swift
```diff
@@ -28,6 +31,9 @@ public func l1Loss<Scalar: TensorFlowFloatingPoint>(
 }
 
 /// Returns the L2 loss between predictions and expectations.
+/// Given the x and y vectors L1 loss computes:
```
BradLarson:
```suggestion
/// Given the x and y vectors, the L2 loss is computed as follows:
```

Sources/TensorFlow/Loss.swift
```diff
@@ -28,6 +31,9 @@ public func l1Loss<Scalar: TensorFlowFloatingPoint>(
...
+///  `reduction([l_1, l_2, ..., l_N]^T)`, where `l_i = (x_i - y_i)^2`
+///   and `N` is yhe batch size
```
BradLarson:
```suggestion
///   and `N` is the batch size.
```

Sources/TensorFlow/Loss.swift
```diff
@@ -43,6 +49,7 @@ public func l2Loss<Scalar: TensorFlowFloatingPoint>(
 }
 
 /// Returns the mean absolute error between predictions and expectations.
+/// Applies the mean reduction to the l1 loss
```
BradLarson: Does this line add significant information that is missing in the line above? If so, this may need to be rephrased so that it flows better with the existing comment. Additionally, a period is needed.

The same can apply to the added comment just below this one.

Sources/TensorFlow/Loss.swift
```diff
@@ -69,6 +77,9 @@ public func meanSquaredError<Scalar: TensorFlowFloatingPoint>(
 }
 
 /// Returns the mean squared logarithmic error between predictions and expectations.
+/// Given the x and y vectors the mean squared logarithmic is:
```
BradLarson:
```suggestion
/// Given the x and y vectors, the mean squared logarithmic error is:
```

Sources/TensorFlow/Loss.swift
```diff
@@ -69,6 +77,9 @@ public func meanSquaredError<Scalar: TensorFlowFloatingPoint>(
...
+///  `(l_1, l_2, ..., l_N)/N`, where `l_i = (log(x_i + 1) - log(y_i + 1))^2`
+///  and `N` is yhe batch size
```
BradLarson:
```suggestion
///  and `N` is the batch size.
```

Sources/TensorFlow/Loss.swift
```diff
@@ -228,6 +239,7 @@ func _vjpSoftmaxCrossEntropyHelper<Scalar: TensorFlowFloatingPoint>(
 }
 
 /// Returns the softmax cross entropy (categorical cross entropy) between logits and labels.
+/// Given the logits and probabilites cross entropy computes `reduction(-logits * log(p))`.
```
BradLarson:
```suggestion
/// Given the logits and probabilities, the softmax cross entropy loss computes `reduction(-logits * log(p))`.
```

Also, is this calculation correct if the softmax is applied along with the cross entropy loss?

Sources/TensorFlow/Loss.swift
```diff
@@ -13,6 +13,9 @@
 // limitations under the License.
 
 /// Returns the L1 loss between predictions and expectations.
+/// Given the x and y vectors L1 loss computes:
+///  `reduction([l_1, l_2, ..., l_N]^T)`, where `l_i = |x_i - y_i|`
```
eaplatanios: What does the `^T` represent?

joaogui1: It's transpose, just to denote the shape of the vector

Sources/TensorFlow/Loss.swift
joaogui1: Well, it explains a little bit better the computation used, but maybe it's superfluous?

Sources/TensorFlow/Loss.swift
joaogui1: Yeah, sorry about that, I believe I fixed it in the last commit

joaogui1:
Ok, so I guess that's all of the non-categorical losses, but I'm a bit confused about how to describe the categorical hinge and categorical crossentropy, can someone shine a light on this?

8bitmp3:
Hi @joaogui1 @eaplatanios @BradLarson Just discussed this thread and the one from the Swift Google group about API docs with @saeta at an event we're attending offline. 

I think this is awesome. I would like to suggest something to improve user experience for the growing community. Perhaps, a logical approach is to follow very closely or word-for-word, where applicable, both descriptions and equations in the TF 2.x API docs, if the exist already.

Otherwise, reinventing the wheel can introduce inconsistency, which TF/S4TF users will notice, since www.tensorflow.org has one search.

_Example:_

In TF 2.0 tf.keras.losses docs, the squared hinge loss is defined as follows ([source](https://github.com/tensorflow/tensorflow/blob/r2.0/tensorflow/python/keras/losses.py#L564-L595)):
```python
class SquaredHinge(LossFunctionWrapper):
"""Computes the squared hinge loss between `y_true` and `y_pred`.
  `loss = square(maximum(1 - y_true * y_pred, 0))`
  `y_true` values are expected to be -1 or 1. If binary (0 or 1) labels are
  provided we will convert them to -1 or 1.
...
```

vs what is in one of the [latest commits](https://github.com/tensorflow/swift-apis/pull/579/commits/ddb8de3f01b5b9183a2b8d5d9222575ade23538e) here, which 

```swift
/// Returns the squared hinge loss between predictions and expectations.
/// Given the predicted and expected, the Hinge loss is computed as follows:
///  `reduction(max(0, 1 - predicted * expected)^2)` 
///
/// - Parameters:
/// ...
public func squaredHingeLoss<Scalar: TensorFlowFloatingPoint>(
...) ...
}
```

The use is `y_true` and `y_pred` is pretty consistent and understandable across the TF 2.x APIs and because the S4TF docs are hosted on the same .org site and under `/tensorflow/` on github, I think we should reuse what's been reviewed and edited by @lamberta and the rest of the docs team. If something is more Swifty rather than Pythonic, we could certainly amend it on a case-by-case basis.

Maybe I'm wrong, let me know what you all think. cc @dynamicwebpaige 

Can I also draw the S4TF community's attention to the TensorFlow docs style guide
https://www.tensorflow.org/community/contribute/docs_style as well as the Google dev docs style guide https://developers.google.com/style

joaogui1:
I changed the names to be compatible, (so it's y_true and y_pred now) but didn't copy the documentation, do you guys think it would be better to have identical docs? 
Also there's another difference, since we have a combined softmax + crossentropy loss calling the logits y_pred would be misleading (the real y_preds should be softmax(logits), do you think the name should be changed anyway?

joaogui1:
Pinging @bartchr808 @8bitmp3

Sources/TensorFlow/Loss.swift
bartchr808: I agree with @BradLarson, and I think maybe we could have something like `Returns the mean absolute error (L1) between...` to make it more clear that `l1Loss<Scalar: TensorFlowFloatingPoint>` is being used and we are taking the mean of it. But even that may not be necessary as L1 is defined to be the absolute error. So I'm personally leaning towards not including this addition. Same for the added comment below for mean L2 loss.

Sources/TensorFlow/Loss.swift
```diff
@@ -115,6 +127,8 @@ public func hingeLoss<Scalar: TensorFlowFloatingPoint>(
 }
 
 /// Returns the squared hinge loss between predictions and expectations.
+/// Given the `y_pred` and `y_true`, the Hinge loss is computed as follows:
```
bartchr808:
```suggestion
/// Given the `y_pred` and `y_true`, the hinge loss is computed as follows:
```

Sources/TensorFlow/Loss.swift
```diff
@@ -100,6 +110,8 @@ public func meanAbsolutePercentageError<Scalar: TensorFlowFloatingPoint>(
 }
 
 /// Returns the hinge loss between predictions and expectations.
+/// Given the `y_pred` and `y_true`, the Hinge loss is computed as follows:
```
bartchr808:
```suggestion
/// Given the `y_pred` and `y_true`, the hinge loss is computed as follows:
```

Sources/TensorFlow/Loss.swift
```diff
@@ -195,7 +213,7 @@ public func kullbackLeiblerDivergence<Scalar: TensorFlowFloatingPoint>(
 }
 
 /// Returns the softmax cross entropy (categorical cross entropy) between logits and labels.
-///
+///
```
bartchr808:
```suggestion
///
```

Sources/TensorFlow/Loss.swift
```diff
@@ -228,6 +246,8 @@ func _vjpSoftmaxCrossEntropyHelper<Scalar: TensorFlowFloatingPoint>(
 }
 
 /// Returns the softmax cross entropy (categorical cross entropy) between logits and labels.
+/// Given the logits and probabilites, the softmax cross entropy computes `reduction(-softmax(logits) * log(p))`
```
bartchr808:
```suggestion
/// Given the logits and probabilities, the softmax cross entropy computes `reduction(-softmax(logits) * log(p))`
```

Sources/TensorFlow/Loss.swift
```diff
@@ -228,6 +246,8 @@ func _vjpSoftmaxCrossEntropyHelper<Scalar: TensorFlowFloatingPoint>(
...
+/// Where softmax(x) = `exp(x)/sum(exp(x))`
```
bartchr808: For consistency since you added spaces b/w the operators in changes above:
```suggestion
/// Where softmax(x) = `exp(x) / sum(exp(x))`
```

Sources/TensorFlow/Loss.swift
```diff
@@ -262,6 +282,8 @@ func _vjpSoftmaxCrossEntropyHelper<Scalar: TensorFlowFloatingPoint>(
 }
 
 /// Returns the sigmoid cross entropy (binary cross entropy) between logits and labels.
+/// Given the logits and probabilites, the sigmoid cross entropy computes `reduction(-sigmoid(logits) * log(p))`
+/// Where sigmoid(x) = `1/(1 + exp(-x))`
```
bartchr808: Same here:
```suggestion
/// Where sigmoid(x) = `1 / (1 + exp(-x))`
```

bartchr808:
Firstly thanks @joaogui1 for doing this and @8bitmp3 for bringing up the idea of "follow[ing] very closely...both descriptions and equations in the TF 2.x API docs"! 😄 

> do you guys think it would be better to have identical docs?

Personally, I agree with @8bitmp3 and  I feel trying to have as close of a 1:1  matching between documentation would be ideal. As with code style, Python and Swift have fairly different styles in the way you define functions. Thus, our API naming cannot be 1:1 in order to stay _**Swifty**_. However, when it comes to writing the documentation, I think this is where we could try and match the documentation whenever possible as the exiting docs are generally TensorFlow specific and language agnostic. So then, it can make anyone transitioning from Python to Swift (either a straight switch or progressively through Python interop) have an easier time possibly recalling what certain functions do (e.g. you Google search some sentence from some documentation in Python TensorFlow, and it shows matches in Swift for TensorFlow docs).

Additionally, I think this can make maintaining docs much easier for the community, @lamberta and the rest of the docs team.

As a small disclaimer, I've been back and busy with school since September, so I haven't been able to follow along with the progress being made in the community, so I may sometimes miss out on some details here and there! 😆

BradLarson:
The general consensus seems to be that we would prefer to closely mirror the documentation from the Python implementations, but replace the variables such that they reflect the Swift code. For example, in the case that @8bitmp3 cites above, that would turn into:

```swift
/// Computes the squared hinge loss between `expected` and `predicted`.
///   `loss = square(maximum(1 - expected * predicted, 0))`
```

(I'm not sure if the subsequent two lines in the Keras implementation about the parameters apply to how we handle this, but if they do, we could bring those over too with parameter names replaced.)

8bitmp3:
Thumbs up and cheers @bartchr808 @BradLarson for your feedback and thanks @joaogui1 for kick-starting this process. To sum up, so that we're on the same page:

- Mirror TensorFlow API docs as much as possible to help Python users and maintain consistency.
- Amend variables and other bits, where applicable, to keep it _Swifty_.

And, for additional guidance, these Google docs/code style guides should also be useful:
- TensorFlow docs style guide: https://www.tensorflow.org/community/contribute/docs_style
- Google dev docs style guide: https://developers.google.com/style
- Google Swift style guide: https://google.github.io/swift/

(I'll also conduct some deep exploration since the project is already at v0.6 and it's very cold outside.)

image

8bitmp3:
@saeta since you mentioned it in https://groups.google.com/a/tensorflow.org/forum/#!topic/swift/tQKFYL1Ykac

I can add a CONTRIBUTING.md if y'all don't mind

BradLarson:
It's been a little while since we last discussed this, are you still interested in working on these documentation additions? We're revisiting this after the holidays, and wanted to check back in.

joaogui1:
Hi, the idea was to finish it after the holidays, but I've been quite sick. If it's not a problem I will finish it once I've recovered

BradLarson:
Please take your time, hope you feel better soon. We were just checking in on the list of pending pull requests to make sure we weren't neglecting anything. Whenever you feel ready to work on this, we'll be glad to review.

rahulbhalley:
Great to see that API documentation is now being discussed seriously! Very much 👍 for this effort.

Sources/TensorFlow/Loss.swift
```diff
@@ -262,6 +282,8 @@ func _vjpSoftmaxCrossEntropyHelper<Scalar: TensorFlowFloatingPoint>(
 }
 
 /// Returns the sigmoid cross entropy (binary cross entropy) between logits and labels.
+/// Given the logits and probabilites, the sigmoid cross entropy computes `reduction(-sigmoid(logits) * log(p))`
+/// Where sigmoid(x) = `1 / (1 + exp(-x))`
```
rahulbhalley: Will equations like these be shown as LaTex does?

dan-zheng: No, code enclosed in backticks appear with inline code formatting, just like in Markdown.
The [Jazzy API documentation generator](https://github.com/realm/jazzy#linux) doesn't support LaTex formatting.

joaogui1:
So, there are some points worthy discussing about the hinge losses:

1. In tf/keras it seems they convert the values from binary to -1/1 if the input is not -1/1, and swift doesn't do that, should we do it?

2. Our implementation of CategoricalHingeLoss seems to be both the opposite and somewhat different from[ tf/keras,](https://github.com/tensorflow/tensorflow/blob/340387d8b428c6c2080426fcfd08d43e3b7e98fa/tensorflow/python/keras/losses.py#L598) to clarify swift's positive is keras' negative and swift's negative is a little bit different from keras' positive, but pretty similar. Should we worry?

joaogui1:
Oks, sorry again for the delay, I believe that's about it, feedback is welcome as always.

Sources/TensorFlow/Loss.swift
8bitmp3: I think having no LaTex—at least for now—is good for user experience. @fchollet wrote an entire (best-selling) [book](https://www.manning.com/books/deep-learning-with-python) on deep learning without any LaTex: 

> "...we’ll steer away from mathematical notation, which can be off-putting for those without any mathematics background and isn’t strictly necessary to explain things well."

joaogui1:
Pinging @BradLarson

Sources/TensorFlow/Loss.swift
```diff
@@ -284,10 +306,10 @@ public func sigmoidCrossEntropy<Scalar: TensorFlowFloatingPoint>(
     return reduction(maxLogitsWithZero - logits * labels + log1p(exp(-negAbsLogits)))
 }
 
-/// Returns the Huber loss between predictions and expectations.
+/// Computes the Huber loss between `predicted` and `expected`.
 ///
-/// For each value `x` in the difference `expected - predicted`, the loss is:
-/// - `0.5 * x^2` if `abs(x) <= δ`.
+/// For each value `x` in `errpr = expected - predicted`:
```
BradLarson:
```suggestion
/// For each value `x` in `error = expected - predicted`:
```

Sources/TensorFlow/Loss.swift
rahulbhalley: @8bitmp3 you just made me feel nervous 😬 because my book will include equations.

8bitmp3: #suntorytime

joaogui1:
> So, there are some points worthy discussing about the hinge losses:
> 
> 1. In tf/keras it seems they convert the values from binary to -1/1 if the input is not -1/1, and swift doesn't do that, should we do it?
> 2. Our implementation of CategoricalHingeLoss seems to be both the opposite and somewhat different from[ tf/keras,](https://github.com/tensorflow/tensorflow/blob/340387d8b428c6c2080426fcfd08d43e3b7e98fa/tensorflow/python/keras/losses.py#L598) to clarify swift's positive is keras' negative and swift's negative is a little bit different from keras' positive, but pretty similar. Should we worry?

And what do you guys think of this @8bitmp3 @BradLarson ?

PR #581, opened 2019-12-13
Avoid scalars and immutable Parameter<> fields.

pschuh:
Scalars participating in vjps become "scalarized()" which become host transfers.
let Parameter values cannot be written to via key-path-iterable.

### DIFF
--- a/Sources/TensorFlow/Layers/Normalization.swift
+++ b/Sources/TensorFlow/Layers/Normalization.swift
@@ -33,9 +33,9 @@ public struct BatchNorm<Scalar: TensorFlowFloatingPoint>: Layer {
     /// The variance epsilon value.
     @noDerivative public let epsilon: Scalar
     /// The running mean.
-    @noDerivative public let runningMean: Parameter<Scalar>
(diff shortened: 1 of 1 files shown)


Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -85,12 +85,13 @@ public struct BatchNorm<Scalar: TensorFlowFloatingPoint>: Layer {
           var normalizedAxes = Array(0..<input.rank)
           normalizedAxes.remove(at: positiveAxis)
           let moments = input.moments(alongAxes: normalizedAxes)
-          runningMean.value += (moments.mean - runningMean.value) * (1 - momentum)
-          runningVariance.value += (moments.variance - runningVariance.value) * (1 - momentum)
-          let inv = rsqrt(moments.variance + epsilon) * scale
+          let decayMomentum = Tensor<Scalar>(1 - momentum)
```
dan-zheng: Could you please try removing the explicit `<Scalar>` generic parameter for consistency with other code?
I believe it can be inferred.
```suggestion
          let decayMomentum = Tensor(1 - momentum)
```

PR #583, opened 2019-12-13
Add categorical intializer for Tensor

xihui-wu:
This PR introduces capability of drawing samples from a categorical distribution into Tensor's intializer.

### DIFF
--- a/Sources/TensorFlow/Initializers.swift
+++ b/Sources/TensorFlow/Initializers.swift
@@ -453,6 +453,31 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
     }
 }
 
+public extension Tensor where Scalar: TensorFlowIndex {
+    /// Creates a tensor by drawing samples from a categorical distribution.
+    ///
+    /// - Parameters:
+    ///     - logits: 2-D Tensor with shape `[batch_size, num_classes]`.  Each slice `[i, :]`
+    ///         represents the unnormalized log probabilities for all classes.
+    ///     - num_samples: 0-D.  Number of independent samples to draw for each row slice.
+    ///
+    /// - Attrs:
+    ///     - seed: If either seed or seed2 is set to be non-zero, the internal random number
+    ///         generator is seeded by the given seed.  Otherwise, a random seed is used.
+    ///     - seed2: A second seed to avoid seed collision.
+    ///
+    /// - Output: 2-D Tensor with shape `[batch_size, num_samples]`.  Each slice `[i, :]`
+    ///     contains the drawn class labels with range `[0, num_classes)`.
+    init<T: TensorFlowNumeric>(
+        logits: Tensor<T>,
+        numSamples: Tensor<Int32>,
+        seed: Int64 = 0,
+        seed2: Int64 = 0
+    ) {
+        self = _Raw.multinomial(logits: logits, numSamples: numSamples, seed: seed, seed2: seed2)
+    }
+}
+
 //===------------------------------------------------------------------------------------------===//
 // Variance Scaling
 //===------------------------------------------------------------------------------------------===//
--- a/Sources/TensorFlow/Initializers.swift
+++ b/Sources/TensorFlow/Initializers.swift
@@ -457,24 +457,22 @@ public extension Tensor where Scalar: TensorFlowIndex {
     /// Creates a tensor by drawing samples from a categorical distribution.
     ///
     /// - Parameters:
-    ///     - logits: 2-D Tensor with shape `[batch_size, num_classes]`.  Each slice `[i, :]`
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/Initializers.swift
```diff
@@ -453,6 +453,31 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+    ///
+    /// - Attrs:
+    ///     - seed: If either seed or seed2 is set to be non-zero, the internal random number
+    ///         generator is seeded by the given seed.  Otherwise, a random seed is used.
+    ///     - seed2: A second seed to avoid seed collision.
+    ///
+    /// - Output: 2-D Tensor with shape `[batch_size, num_samples]`.  Each slice `[i, :]`
+    ///     contains the drawn class labels with range `[0, num_classes)`.
+    init<T: TensorFlowNumeric>(
+        logits: Tensor<T>,
```
marcrasi: I suggest naming this label `randomCategorialLogits`, so that it describes what the initializer does with the logits.

Sources/TensorFlow/Initializers.swift
```diff
@@ -453,6 +453,31 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+        numSamples: Tensor<Int32>,
```
marcrasi: The standard in swift is to always use "count" instead of "num". I suggest `sampleCount`.

Also, since this is always an integer scalar, I suggest making it a more swift-native type like `Int`, and make the implementation convert it to the appropriate tensor type before passing it to the op.

Sources/TensorFlow/Initializers.swift
```diff
@@ -453,6 +453,31 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+        seed: Int64 = 0,
```
marcrasi: The other `random` initializers in this file use `TensorFlowSeed` with default value `Context.local.randomSeed`. This should do the same thing with this one, for consistency.

Sources/TensorFlow/Initializers.swift
```diff
@@ -453,6 +453,31 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+        seed2: Int64 = 0
```
marcrasi: Do you know why there are 2 seeds? Would it make sense to also make this one `seed2: TensorFlowSeed = Context.local.randomSeed`?

Sources/TensorFlow/Initializers.swift
```diff
@@ -453,6 +453,31 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
     }
 }
 
+public extension Tensor where Scalar: TensorFlowIndex {
+    /// Creates a tensor by drawing samples from a categorical distribution.
+    ///
+    /// - Parameters:
```
marcrasi: This should use the same documentation comment formatting as the other doc comments in this repo. In particular:
* Put all the parameters under "Parameters". Swift doesn't have a concept of "Attrs".
* Name the return section "Returns" instead of "Output".
* Rename the `batch_size`, `num_classes`, `num_samples` to match swift naming conventions (camel case and "count" instead of "num").

Sources/TensorFlow/Initializers.swift
xihui-wu: "An second seed to avoid seed collision" as per the doc. But as we synced offline, going to pass in one TensorFlowSeed and call statelessMultinomial inside.

Tests/TensorFlowTests/InitializerTests.swift
```diff
@@ -170,6 +170,17 @@ final class InitializerTests: XCTestCase {
...
 
+    func testCategoricalFromLogits() {
+        let probabilities = Tensor<Float>([[0.5, 0.3, 0.2], [0.6, 0.3, 0.1]])
+        let logits = log(probabilities)
+        let t = Tensor<Int32>(randomCategorialLogits: logits, sampleCount: 1)
+
+        XCTAssertEqual(TensorShape([2, 1]), t.shape)
+        // Test all elements are in range of [0, 3)
+        XCTAssertEqual(1, (Tensor<Int32>(t .>= Tensor<Int32>([[0], [0]])).product()).scalar!)
+        XCTAssertEqual(1, (Tensor<Int32>(t .< Tensor<Int32>([[3], [3]])).product()).scalar!)
```
marcrasi: It would be a bit simpler to use `.all()` instead of `.product()`, like this: `XCTAssertTrue((t .>= Tensor<Int32>([[0], [0]])).all())`

Sources/TensorFlow/Initializers.swift
marcrasi: Oh sorry, didn't notice this earlier: The indentation should follow the conventions at https://google.github.io/swift/#parameter-returns-and-throws-tags

Issue #562, opened 2019-11-20
Add linear algebra operations and their derivatives


dan-zheng:
Swift for TensorFlow should provide standard linear algebra operations from [`tf.linalg`](https://www.tensorflow.org/api_docs/python/tf/linalg)/[`np.linalg`](https://docs.scipy.org/doc/numpy-1.13.0/reference/routines.linalg.html).

Some of these operations are already accessible via [TensorFlow bindings](https://github.com/tensorflow/swift-apis/blob/master/Sources/TensorFlow/Bindings/RawOpsGenerated.swift): `_Raw.cholesky`, `_Raw.choleskyGrad`.

Reference TensorFlow derivative implementations: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/linalg_grad.py

Mirror of [TF-980](https://bugs.swift.org/browse/TF-980).

awav:
This is not a full list of operations, but it would be good to start with this:

- [ ] Norm
- [ ] Trace
- [ ] Cholesky
- [ ] Triangular Solve
- [ ] Cholesky Solve
- [ ] Diag (make a [batch] tensor from [batch] diagonal)
- [ ] Diag Part (extract [batch] diagonal)
- [ ] Einsum
- [ ] Eye
- [ ] Logdet
- [ ] Eigh

awav:
@dan-zheng , how can I add a new binding [here](https://github.com/tensorflow/swift-apis/blob/master/Sources/TensorFlow/Bindings/RawOpsGenerated.swift)?

Shashi456:
@awav you can't add new bindings in that file, You can check where how its generated [here](https://github.com/tensorflow/swift-apis/tree/master/Sources/TensorFlow/Bindings).

awav:
@Shashi456 , yeah, I've checked it. But, it is not clear why for e.g. "TriangularSolve" is not reflected in the [RawOpsGenerated.swift](https://raw.githubusercontent.com/tensorflow/swift-apis/master/Sources/TensorFlow/Bindings/RawOpsGenerated.swift) file, but "TriangularSolve" is in the [ops.pbtxt](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/ops.pbtxt). Does S4TF uses different protobuf file?

jon-tow:
Hey @awav. I believe `triangular_solve` is aliased under `matrixTriangularSolve`.

dan-zheng:
[...]

PR #586, opened 2019-12-19
[Linear Algebra] Trace operator

awav:
The PR adds `trace` linear operator https://github.com/tensorflow/swift-apis/issues/562.

### DIFF
--- a/Sources/TensorFlow/Operators/LinearAlgebra.swift
+++ b/Sources/TensorFlow/Operators/LinearAlgebra.swift
@@ -107,11 +107,10 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
 }
 
 /// Computes the trace of an optionally batched matrix.
-/// The trace is the the sum along the main diagonal of an optionally batched matrix.
+/// The trace is the the sum along the main diagonal of each inner-most matrix.
 ///
 /// The input is a tensor with shape `[..., M, N]`.
-/// The output is a tensor with shape `[...]`. The main diagonal of a matrix has
-/// `min(M, N)` length, in cases when `N` does not equal `M`.
+/// The output is a tensor with shape `[...]`.
 ///
 /// - Parameter matrix: A tensor of shape `[..., M, N]`.
 /// - Precondition: `matrix` must be a tensor with shape `[..., M, N]`.
--- a/Sources/TensorFlow/Operators/LinearAlgebra.swift
+++ b/Sources/TensorFlow/Operators/LinearAlgebra.swift
@@ -110,15 +110,15 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
 /// Returns the sum along the main diagonal of [batched] matrix.
 ///
 /// The input has to be [batched] square matrix.
-/// When the input is a tensor of shape `[..., M, M]` then the output is a tensor
+/// When the input is a tensor of shape `[..., M, N]` then the output is a tensor
 /// with dimensions `[...]`
 ///
-/// - Parameter matrix: A tensor of shape `[..., M, M]`
+/// - Parameter matrix: A tensor of shape `[..., M, N]`
+/// - Precondition: The [batched] matrix rank must be greater or equal 2.
 @inlinable
 @differentiable(wrt: matrix where T: TensorFlowFloatingPoint)
 public func trace<T: TensorFlowNumeric>(_ matrix: Tensor<T>) -> Tensor<T> {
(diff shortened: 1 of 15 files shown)


Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -106,6 +106,23 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
...
 
+/// Computes the trace of an input tensor.
+/// Returns the sum along the main diagonal of [batched] matrix.
+///
+/// The input has to be [batched] square matrix.
+/// When the input is a tensor of shape `[..., M, M]` then the output is a tensor
+/// with dimensions `[...]`
+///
+/// - Parameter matrix: A tensor of shape `[..., M, M]`
+
```
saeta: Nit: Could you please remove the extra newline here?

eaplatanios:
Thanks @awav! You could also add the preconditions in the doc string using `- Precondition:`. We have not been consistent in doing that so far, but I think it helps to have them there.

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -106,6 +106,22 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+/// The input has to be [batched] square matrix.
+/// When the input is a tensor of shape `[..., M, M]` then the output is a tensor
+/// with dimensions `[...]`
+///
+/// - Parameter matrix: A tensor of shape `[..., M, M]`
+@inlinable
+@differentiable(wrt: matrix where T: TensorFlowFloatingPoint)
+public func trace<T: TensorFlowNumeric>(_ matrix: Tensor<T>) -> Tensor<T> {
+    precondition(matrix.rank >= 2, "The tensor must have at least rank 2.")
+    precondition(matrix.shape[matrix.rank - 1] == matrix.shape[matrix.rank - 2], "Squared [batched] matrix expected")
```
dan-zheng: Python TensorFlow doesn't seem to have a "square matrix" requirement:
```python
import tensorflow as tf
print(tf.__version__) # 2.0.0-beta1
x = tf.constant([[1, 2], [3, 4], [5, 6]])
print(x.shape) # (3, 2), not square
print(tf.linalg.trace(x)) # tf.Tensor(5, shape=(), dtype=int32)
```
[The documentation for `tf.linalg.trace`](https://www.tensorflow.org/api_docs/python/tf/linalg/trace) suggests the same: inputs with shape `[..., M, N]` are supported.

Did you add this precondition for a reason? I think lifting the precondition and matching TensorFlow's behavior would be good, as a rule of thumb.

awav: From, https://en.wikipedia.org/wiki/Trace_(linear_algebra)

> **The trace is only defined for a square matrix (n × n).**

TensorFlow is wrong :)

dan-zheng: This brings us to the question of whether we should match the behavior of Python TensorFlow (for exact reproducibility, which is an important use case and may affect model convergence), or pursue different, more sensible/correct behavior.

In this case, adding an additional precondition seems "safe", because it doesn't lead to silently different computation results compared with Python TensorFlow. There aren't yet users of `trace` that want it to work for non-square matrices. So adding the precondition sounds good to me!

---

Another scenario: I recall a previous discussion with @eaplatanios about using a different (more mathematically correct?) derivative implementation for a particular function than Python TensorFlow. We decided against that change because it does lead to different computation results compared with Python TensorFlow, breaking exact reproducibility.

awav: Apparently, the numpy does the same mistake... or a feature... Should I remove that condition then?

dan-zheng: > Apparently, the numpy does the same mistake... or a feature... Should I remove that condition then?

Ah. In that case, could you please remove the precondition and update doc comments?
(Mentions of "square matrix" should be removed, and `M, M` should be replaced with `M, N`.)

awav: done.

dan-zheng: Thanks! I made minor edits in https://github.com/tensorflow/swift-apis/pull/586/commits/c1407b673a5e89a3078c57f27facf9bb1d7b713d, hope you don't mind.

PR #587, opened 2019-12-19
Improved 'logSumExp' numerical stability.

eaplatanios:
For some reason the "finiteness" checks in the implementation of `logSumExp` keep causing numerical stability issues for me when working with Bayesian models (where such issues become frequent and important). Removing the checks fixes these problems for me. I actually also don't really see why we need them and why the Python TensorFlow API uses them.


Tests/TensorFlowTests/OperatorTests/MathTests.swift
```diff
@@ -346,6 +346,14 @@ final class MathOperatorTests: XCTestCase {
         let ySmall = xSmall.logSumExp()
         let expectedYSmall = Tensor<Float>(-235.28029)
         assertEqual(ySmall, expectedYSmall, accuracy: 0.0001)
+
+        let xLarge = Tensor<Float>([
```
marcrasi: Does this test case fail before you make the change? That would be surprising because it doesn't look like anything should be infinite in this test case. If it does really fail before you make the change, I'd suspect that there is some bug in one of the calls you removed in this PR, like `withoutDerivative(at:)`, `Tensor(zerosLike:)`, `.replacing`, or `.isFinite`. And if that is the case, then we should fix that bug instead of removing the finiteness check.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -2394,12 +2394,7 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
     @inlinable
     @differentiable(wrt: self)
     func logSumExp(squeezingAxes axes: Tensor<Int32>) -> Tensor {
-        let rawMax = max(alongAxes: axes)
-        let offset = withoutDerivative(at: rawMax) { rawMax in 
-            Tensor<Scalar>(zerosLike: rawMax).replacing(
-                with: rawMax,
-                where: rawMax.isFinite)
-        }
+        let offset = max(alongAxes: axes)
```
marcrasi: This PR also removes the `withoutDerivative(at:)`, which is not part of the finiteness check. I don't know what it's for, but the python tf logic also does it, so I assume that it's important for some reason.

marcrasi: For reference, tf python logic is here: https://github.com/tensorflow/tensorflow/blob/r2.0/tensorflow/python/ops/math_ops.py#L2517

Tests/TensorFlowTests/OperatorTests/MathTests.swift
eaplatanios: This test case would fail if the `logSumExp` was implemented without subtracting the max first. I believe the instability in my use case comes from differentiating the current `logSumExp` implementation (not sure which part or why though). I will try to create a test case for that (it's a bit difficult because I only encounter the issue in while performing inference in a Bayesian model). In general though, I don't know why that finiteness check is necessary. The result should be the same if certain elements are non-finite irrespective of whether we have that check. I know it's there in the Python implementation but I can't really understand why it's necessary.

Sources/TensorFlow/Operators/Math.swift
eaplatanios: I believe the gradients are equivalent whether or not the `withoutDerivative(at:)` is added. I am also not sure why that was there in the first place. Presumably because `zerosLike` is not differentiable?

marcrasi:
@eaplatanios is this blocking anything you're doing? This seems important but I'm not comfortable changing the implementation to be different from the py tf implementation without understanding why the py tf implementation is like it is, and without having a test case that exposes the failure that we're fixing.

I could try to play around with this and ask some tf people about the py implementation.

eaplatanios:
@marcrasi sorry for not responding earlier. There is definitely a problem somewhere that is avoided with this PR. I'm not quite sure if it has to do with the implementation of `logSumExp` or with something else (e.g., auto-diff related) as I didn't have time to investigate it. I'd still be curious to know about why the py tf implementation has that check for the max.

marcrasi:
I do want to investigate this myself some time soon because it does look important! Hopefully I'll have a bit of time in the next few days to experiment and come to some conclusion.

marcrasi:
I did some experiments and got interesting results.

I have no conclusion yet on the best thing to do. There are tradeoffs. Maybe some of this info will help you understand where your numerical instability issues were coming from.

I'll come back in a few days and make some suggestion after sleeping on it.

**Results on finite inputs**

Both variants produce the same values on finite inputs, so it's only interesting to compare the gradients.

There are at least two desirable properties of logSumExp gradients, but we can't have both of them at the same time:

 1. The elements sum to 1.
 2. Each element is the closest floating point number to the true value.

Neither variant achieves either property all the time, though the modified variant in this PR seems to achieve property (1) more often than the original variant that's currently checked in.

```
print(gradient(at: Tensor<Float>([0, 0.625])) { logSumExpOriginal($0, squeezingAxes: [0]) }.sum()) // 1.0000001
print(gradient(at: Tensor<Float>(ones: [100])) { logSumExpOriginal($0, squeezingAxes: [0]) }.sum()) // 1.0000001

print(gradient(at: Tensor<Float>([0, 0.625])) { logSumExpModified($0, squeezingAxes: [0]) }.sum()) // 1.0
print(gradient(at: Tensor<Float>(ones: [100])) { logSumExpModified($0, squeezingAxes: [0]) }.sum()) // 1.0000001
```

The reason why the modified variant achieves property (1) more often is that the AD-calculated gradient of `offset` is `1 - sumOfAllElementsOfGradient`. The gradient of the original variant throws this gradient away because of the `withoutDerivative`, while the gradient of the modified variant adds this to the gradient of the max element. This "correction term" makes the elements more likely to sum to 1. (Note that under infinite precision math, this "correction term" is 0). See the "hand backpropagated gradients" section for an explicit implementation of this correction term.

As for property (2), I suspect that we can do better on that axis using a manual gradient because all the AD-generated gradients do something like `exp(x - offset) * (1 / sumExp)` while a manual gradient can do `exp(x - offset) / sumExp` which avoids throwing away some accuracy. I haven't yet done any experiments to verify that this gradient is better at (2) though.

**Results on infinite inputs**

The original variant produces fewer nans than the modified variant.

```
print(valueWithGradient(at: Tensor<Float>([Float.infinity])) { logSumExpOriginal($0, squeezingAxes: [0]) }) // (value: inf, gradient: [-nan])
print(valueWithGradient(at: Tensor<Float>([0, Float.infinity])) { logSumExpOriginal($0, squeezingAxes: [0]) }) // (value: inf, gradient: [ 0.0, -nan])
print(valueWithGradient(at: Tensor<Float>([10, Float.infinity])) { logSumExpOriginal($0, squeezingAxes: [0]) }) // (value: inf, gradient: [ 0.0, -nan])

print(valueWithGradient(at: Tensor<Float>([-Float.infinity])) { logSumExpOriginal($0, squeezingAxes: [0]) }) // (value: -inf, gradient: [-nan])
print(valueWithGradient(at: Tensor<Float>([0, -Float.infinity])) { logSumExpOriginal($0, squeezingAxes: [0]) }) // (value: 0.0, gradient: [1.0, 0.0])
print(valueWithGradient(at: Tensor<Float>([10, -Float.infinity])) { logSumExpOriginal($0, squeezingAxes: [0]) }) // (value: 10.0, gradient: [1.0, 0.0])

print(valueWithGradient(at: Tensor<Float>([Float.infinity])) { logSumExpModified($0, squeezingAxes: [0]) }) // (value: -nan, gradient: [nan])
print(valueWithGradient(at: Tensor<Float>([0, Float.infinity])) { logSumExpModified($0, squeezingAxes: [0]) }) // (value: -nan, gradient: [-nan, -nan])
print(valueWithGradient(at: Tensor<Float>([10, Float.infinity])) { logSumExpModified($0, squeezingAxes: [0]) }) // (value: -nan, gradient: [-nan, -nan])

print(valueWithGradient(at: Tensor<Float>([-Float.infinity])) { logSumExpModified($0, squeezingAxes: [0]) }) // (value: -nan, gradient: [nan])
print(valueWithGradient(at: Tensor<Float>([0, -Float.infinity])) { logSumExpModified($0, squeezingAxes: [0]) }) // (value: 0.0, gradient: [1.0, 0.0])
print(valueWithGradient(at: Tensor<Float>([10, -Float.infinity])) { logSumExpModified($0, squeezingAxes: [0]) }) // (value: 10.0, gradient: [1.0, 0.0])
```

**Functions that I ran the experiments on**

```
@differentiable
func logSumExpOriginal(_ t: Tensor<Float>, squeezingAxes axes: Tensor<Int32>) -> Tensor<Float> {
    let rawMax = t.max(alongAxes: axes)
    let offset = withoutDerivative(at: rawMax) { rawMax in
        Tensor<Float>(zerosLike: rawMax).replacing(
                with: rawMax,
                where: rawMax.isFinite)
    }
    let result = TensorFlow.log(TensorFlow.exp(t - offset).sum(squeezingAxes: axes))
    let resultShape = withoutDerivative(at: result.shapeTensor)
    return result + offset.reshaped(toShape: resultShape)
}

@differentiable
func logSumExpModified(_ t: Tensor<Float>, squeezingAxes axes: Tensor<Int32>) -> Tensor<Float> {
    let offset = t.max(alongAxes: axes)
    let result = TensorFlow.log(TensorFlow.exp(t - offset).sum(squeezingAxes: axes))
    let resultShape = withoutDerivative(at: result.shapeTensor)
    return result + offset.reshaped(toShape: resultShape)
}
```

**Hand backpropagated gradients**

Note that neither of these do the finiteness checks, so they're only correct for finite values.

```
func gradientLogSumExpOriginal(_ t: Tensor<Float>, squeezingAxes axes: Tensor<Int32>) -> Tensor<Float> {
    let offset = t.max(alongAxes: axes)
    let exps = TensorFlow.exp(t - offset)
    let sumExps = exps.sum(squeezingAxes: axes)
    let inverseSumExps = 1 / sumExps
    return exps * inverseSumExps
}

func gradientLogSumExpModified(_ t: Tensor<Float>, squeezingAxes axes: Tensor<Int32>) -> Tensor<Float> {
    let originalGradient = gradientLogSumExpOriginal(t, squeezingAxes: [0])

    let offset = t.max(alongAxes: axes)
    let indicators = Tensor<Float>(t .== offset)
    let indicatorsCount = indicators.sum(alongAxes: axes)
    let adjustment = (1 - originalGradient.sum(alongAxes: axes)) * indicators / indicatorsCount

    return originalGradient + adjustment
}
```

I verified that these give the same results as the corresponding AD-generated gradients over a range of values, using:

```
func gradientsSame(_ f: (Tensor<Float>) -> Tensor<Float>, _ g: (Tensor<Float>) -> Tensor<Float>) -> Bool {
    for x in 0...40 {
        for y in 0...40 {
            let t = Tensor([Float(x)/8, Float(y)/8])
            if f(t) != g(t) {
                return false
            }
        }
    }
    return true
}`
```

**Manual gradient**

Note that this does not do the finiteness check, so we'd probably want to add that before using it for real.

```
func gradientManual(_ t: Tensor<Float>, squeezingAxes axes: Tensor<Int32>) -> Tensor<Float> {
    let offset = t.max(alongAxes: axes)
    let exps = TensorFlow.exp(t - offset)
    let sumExps = exps.sum(squeezingAxes: axes)
    return exps / sumExps
}
```

If you run this through `gradientsSame`, you get that it's not the same as the AD-generated gradient of the original or modified variants!

dan-zheng:

---

dabrahams:
@marcrasi to follow up with more specific suggestions.

marcrasi:
Closing for now since I haven't made progress on this for a while and since I feel reasonably confident that there isn't a big issue in the current logSumExp derivative. Of course it could be made a bit better so we can revisit this eventually.

PR #588, opened 2019-12-20
Add `subdiagonalCount:superdiagonalCount:` argument labels to `Tensor.bandPart`.

dan-zheng:
`subdiagonalCount:superdiagonalCount:` argument labels improve clarity.
Parameter descriptions copied from https://www.tensorflow.org/api_docs/python/tf/linalg/band_part.

cc @awav

### DIFF
--- a/Sources/TensorFlow/Operators/LinearAlgebra.swift
+++ b/Sources/TensorFlow/Operators/LinearAlgebra.swift
@@ -56,6 +56,12 @@ public extension Tensor where Scalar: TensorFlowNumeric {
         _Raw.matrixDiag(diagonal: self)
     }
 
+    @available(*, deprecated, renamed: "bandPart(subdiagonalCount:superdiagonalCount:)")
+    @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
+    func bandPart(_ subdiagonalCount: Int, _ superdiagonalCount: Int) -> Tensor {
+        return bandPart(subdiagonalCount: subdiagonalCount, superdiagonalCount: superdiagonalCount)
+    }
+
     /// Returns a copy of a innermost tensor defined by a central band boundaries.
     /// The output is a tensor of the same shape as the instance `[..., :, :]`.
     ///
--- a/Sources/TensorFlow/Operators/LinearAlgebra.swift
+++ b/Sources/TensorFlow/Operators/LinearAlgebra.swift
@@ -79,12 +79,18 @@ public extension Tensor where Scalar: TensorFlowNumeric {
     /// //  [-2, -1,  0, 1]
     /// //  [ 0, -2, -1, 0]]
     /// ```
+    ///
+    /// - Parameters:
+    ///   - subdiagonalCount: The number of subdiagonals to keep. If negative, keep entire lower
+    ///     triangle.
+    ///   - superdiagonalCount: The number of superdiagonals to keep. If negative, keep entire upper
+    ///     triangle.
     @inlinable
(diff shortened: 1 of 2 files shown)


dan-zheng:
`swift-models` needs to be updated after this patch:
```
/swift-models/Transformer/Model.swift:95:29: error: missing argument labels 'subdiagonalCount:superdiagonalCount:' in call
    let mask = ones.bandPart(-1, queryTimeSteps - keyTimeSteps)
                            ^
                             subdiagonalCount:  superdiagonalCount:
```

saeta:
> One quick nit, however: could you add a deprecated no-label version to help make it a smoother migration for users?

I suspect this will also help get CI to pass (https://source.cloud.google.com/results/invocations/5b144b04-f116-41f9-b435-574cd360de47/targets/s4tf%2Fapis%2Fubuntu%2Fpresubmit/log)

dan-zheng:
> One quick nit, however: could you add a deprecated no-label version to help make it a smoother migration for users?

Sure! Done in https://github.com/tensorflow/swift-apis/pull/588/commits/71db9e55d52586db97109e7150662d5957021102.

awav:
@dan-zheng, Agree. The verbosity is an important thing and it is good to have the naming to avoid confusion. I don't have a strong opinion about how the arguments should be called in `bandPart` case. Although, I must ask, what do you think about slightly shorter version `subdiagCount` and `superdiagCount`?

dan-zheng:
> @dan-zheng, Agree. The verbosity is an important thing and it is good to have the naming to avoid confusion. I don't have a strong opinion about how the arguments should be called in `bandPart` case.

I'm glad we all agree!

The Swift API design guidelines clearly (haha) [advocate clarity](https://swift.org/documentation/api-design-guidelines/#fundamentals):

> * **Clarity at the point of use** is your most important goal. Entities such as methods and properties are declared only once but used repeatedly. Design APIs to make those uses clear and concise. When evaluating a design, reading a declaration is seldom sufficient; always examine a use case to make sure it looks clear in context.
> 
> * **Clarity is more important than brevity.** Although Swift code can be compact, it is a non-goal to enable the smallest possible code with the fewest characters. Brevity in Swift code, where it occurs, is a side-effect of the strong type system and features that naturally reduce boilerplate.

---

> Although, I must ask, what do you think about slightly shorter version `subdiagCount` and `superdiagCount`?

The Swift API design guidelines recommend [avoiding abbrevations](https://swift.org/documentation/api-design-guidelines/#avoid-abbreviations):

> * **Avoid abbreviations.** Abbreviations, especially non-standard ones, are effectively terms-of-art, because understanding depends on correctly translating them into their non-abbreviated forms.
> > The intended meaning for any abbreviation you use should be easily found by a web search.

"Easily found by a web search" is a concrete measure that we can verify:
* https://lmgtfy.com/?q=subdiag
* https://lmgtfy.com/?q=superdiag

The top results for "subdiag" and "superdiag" are not related to diagonals from linear algebra: "diag" refers to something else, as in "subdiagram" or "super diagnostics".

Later results do reference subdiagonals and superdiagonals, showing that "subdiag" and "superdiag" are common in-domain abbreviations. They're very reasonable as local variable names.

To adhere to the API design guidelines and to maintain consistency with the rest of the codebase, I'd recommend that we continue to avoid abbreviations.

dan-zheng:
Review feedback has been addressed.
Merging to avoid conflicts with pending patch for TF-1076: using `@derivative` for derivative registration.

PR #590, opened 2019-12-21
Fix derivatives for min reduction ops.

dan-zheng:
Fix VJPs for:
- `Tensor.min(alongAxes:)`
- `Tensor.min(squeezingAxes:)`

Previously, the VJPs returned the result of `max` as the original value.
Now, the result of `min` is correctly returned.

Add tests, checking against Python TensorFlow.

### DIFF
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -1396,6 +1396,8 @@ public func min<T>(_ lhs: Tensor<T>, _ rhs: T) -> Tensor<T> where T: Numeric & C
     min(lhs, Tensor(rhs))
 }
 
+// Note: adapted from `_MinOrMaxGrad`:
+// https://github.com/tensorflow/tensorflow/blob/r2.1/tensorflow/python/ops/math_grad.py#L223.
 @inlinable
 internal func _vjpMinMaxHelper<T: TensorFlowFloatingPoint>(
     _ x: Tensor<T>,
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1706,35 +1708,73 @@ public extension Tensor where Scalar: Numeric & Comparable {
 }
 
 internal extension Tensor where Scalar: TensorFlowFloatingPoint {
-  @inlinable
-  func _vjpMinOrMax(squeezingAxes axes: Tensor<Int32>) -> (Tensor, (Tensor) -> Tensor) {
-    let result = max(squeezingAxes: axes)
```
dan-zheng: Note: here's the issue in the old implementation.

`_vjpMinOrMax(squeezingAxes:)` is registered as the derivative of `min(squeezingAxes:)`, but it returns `max(squeezingAxes: axes)` as the original result.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1706,35 +1708,73 @@ public extension Tensor where Scalar: Numeric & Comparable {
...
 
-      return gradientUnsqueezed.broadcasted(toShape: self.shapeTensor) * indicators / selectedCount
-    })
-  }
+        return gradientUnsqueezed.broadcasted(toShape: self.shapeTensor) * indicators / selectedCount
+    }
 
-  @inlinable
-  func _vjpMinOrMax(alongAxes axes: Tensor<Int32>) -> (Tensor, (Tensor) -> Tensor) {
-    let result = max(alongAxes: axes)
```
dan-zheng: Note: here's the issue in the old implementation.

`_vjpMinOrMax(alongAxes:)` is registered as the derivative of `min(alongAxes:)`, but it returns `let result = max(alongAxes: axes)` as the original result.

eaplatanios:
Good catch Dan! Thanks! :)

PR #591, opened 2019-12-21
[AutoDiff] Use `@derivative` for derivative registration.

dan-zheng:
Rewrite all derivative registration using `@differentiable(jvp:vjp:)` with
`@derivative`. This has no functional impact.

Keep original `@differentiable` attributes so that derivative functions are
publicly exposed.

When [retroactive derivative registration]() is complete:
- `@differentiable(jvp:vjp:)` will be deprecated.
- `@derivative` attribute will be the canonical way to register derivatives.

Resolves TF-1076.

---

No remaining uses of `@differentiable(vjp:)` for derivative registration:
```console
$ grep -nr "vjp:" *
# No remaining uses.
```

---

Example:

```swift
// Before: `@differentiable(vjp:)`.
extension Tensor where Scalar: Numeric {
    @differentiable(vjp: _vjpAdd where Scalar: TensorFlowFloatingPoint)
    public static func + (lhs: Tensor, rhs: Tensor) -> Tensor { ... }
}

    @inlinable
    static func _vjpAdd(lhs: Tensor, rhs: Tensor) -> (
        value: Tensor, pullback: (Tensor) -> (Tensor, Tensor)
    ) { ... }
}
```

```swift
// After: `@derivative`.
extension Tensor where Scalar: Numeric {
    @differentiable(where Scalar: TensorFlowFloatingPoint)
    public static func + (lhs: Tensor, rhs: Tensor) -> Tensor { ... }
}

    @inlinable
    @derivative(of: +)
    static func _vjpAdd(lhs: Tensor, rhs: Tensor) -> (
        value: Tensor, pullback: (Tensor) -> (Tensor, Tensor)
    ) { ... }
}
```

### DIFF
--- a/Sources/TensorFlow/Core/Tensor.swift
+++ b/Sources/TensorFlow/Core/Tensor.swift
@@ -125,7 +125,7 @@ public extension Tensor {
     /// Reshape to scalar.
     /// - Precondition: The tensor has exactly one scalar.
     @inlinable
-    @differentiable(wrt: self, vjp: _vjpScalarized where Scalar: TensorFlowFloatingPoint)
+    @differentiable(where Scalar: TensorFlowFloatingPoint)
     func scalarized() -> Scalar {
         precondition(shape.contiguousSize == 1,
            "This tensor must have exactly one scalar but contains \(shape.contiguousSize).")
@@ -135,7 +135,8 @@ public extension Tensor {
 
 internal extension Tensor where Scalar: TensorFlowFloatingPoint {
     @inlinable
-    func _vjpScalarized() -> (Scalar, (Scalar) -> Tensor) {
+    @derivative(of: scalarized)
+    func _vjpScalarized() -> (value: Scalar, pullback: (Scalar) -> Tensor) {
         return (scalarized(), { v in Tensor(v) })
     }
 }
@@ -162,7 +163,7 @@ public extension Tensor {
     }
 
     @inlinable
-    @differentiable(vjp: _vjpScalars where Scalar: TensorFlowFloatingPoint)
+    @differentiable(where Scalar: TensorFlowFloatingPoint)
     var scalars: [Scalar] {
         return array.scalars
     }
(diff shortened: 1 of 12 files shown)


Tests/TensorFlowTests/TensorAutoDiffTests.swift
```diff
@@ -575,16 +567,6 @@ final class TensorAutoDiffTests: XCTestCase {
         XCTAssertEqual(pb(Tensor(ones: [3, 3])), Tensor(repeating: 5.9604645e-08, shape: [3, 3]))
     }
 
-    // SR-9345
```
saeta: QQ: What happened to these tests? Any reason why they were deleted?

dan-zheng: I decided to delete the test because it's no longer meaningful: it was added in https://github.com/apple/swift/pull/20788, but differentiation implementation details have significantly changed since. The test is also not inherently tied to `Tensor` - the AutoDiff test suite at `apple/swift` covers such tests.

The test calls a top-level function `foo` with a `@differentiable(vjp vjpFoo:)` attribute. It's possible to remove the attribute and add `@derivative(of: foo)` to `func vjpFoo`, there's no issue there.

However, defining a vague top-level function `foo` isn't nice. Moving `func foo` and `func vjpFoo` to a local context would be better, but isn't possible because [`@derivative` is disabled in local contexts](https://github.com/apple/swift/pull/28483).

Sources/TensorFlow/Core/Tensor.swift
```diff
@@ -184,24 +186,26 @@ extension Tensor where Scalar: TensorFlowFloatingPoint {
...
         self.init(shape: [], scalars: [value], on: device)
     }
 }
 
 internal extension Tensor where Scalar: TensorFlowFloatingPoint {
     @inlinable
-    static func _vjpScalarInit(_ value: __owned Scalar, on device: Device = Device.getDefault
-    ) -> (Tensor, (Tensor) -> Scalar) {
+    @derivative(of: init(_:on:))
+    static func _vjpScalarInit(_ value: __owned Scalar, on device: Device = Device.getDefault) -> (
```
burmako: @dan-zheng Perhaps it would make sense to rename _vjpXXX functions to _derivativeXXX?

dan-zheng: I don't feel strongly! The best name is `func _` (a private anonymous function), which is used in the [differentiable programming manifesto](http://bit.ly/swift-autodiff-manifesto) but not yet implemented.

When linear functions and transposition are done, users will register only primitive JVP functions returning a `@differentiable(linear)` differential function, and primitive VJP functions will be removed. At that point, `@derivative` functions will unambiguously be JVP functions, so `derivativeXXX` names become appropriate and `vjpXXX` names become obsolete.

PR #593, opened 2019-12-22
NFC: garden `Device`.

dan-zheng:
- `Device.getDefault` -> `Device.default`
- API documentation edits.

### DIFF
--- a/Sources/TensorFlow/Core/Device.swift
+++ b/Sources/TensorFlow/Core/Device.swift
@@ -1,16 +1,16 @@
-/// Represents a device where a tensor logically resides.
-/// Currently, this is just a stub because tensorflow will transfer
-/// tensors between devices on demand.
+/// A device on which `Tensor`s can be allocated.
+///
+/// Currently, this is a stub because TensorFlow transfers tensors between devices on demand.
 public struct Device {
-  public static var getDefault: Device { Device() }
+    public static var `default`: Device { Device() }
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/Core/Device.swift
```diff
@@ -1,16 +1,16 @@
-/// Represents a device where a tensor logically resides.
-/// Currently, this is just a stub because tensorflow will transfer
-/// tensors between devices on demand.
+/// A device on which `Tensor`s can be allocated.
+///
+/// Currently, this is a stub because TensorFlow transfers tensors between devices on demand.
 public struct Device {
-  public static var getDefault: Device { Device() }
+    public static var `default`: Device { Device() }
```
asuhan: Does `swift-format` do this? If not, I'd rather leave it alone.

dan-zheng: [`swift-format` for this repository](https://github.com/tensorflow/swift-apis/blob/master/.swift-format) is configured to use 4-space indentation and 100 column width. [Here's how `swift-format` configurations work](https://github.com/apple/swift-format#configuration).

I'd say that 4-space indentation and [100 column width](https://google.github.io/swift/#column-limit) is important for consistency with the [Google Swift Style Guide](https://google.github.io/swift). The rest of the codebase adheres to it.

PR #598, opened 2019-12-26
GlobalMinPool layers

kongzii:
I noticed that there are `Max` and `Avg` global pool layers but no `Min`. So here are their implementations.

### DIFF
--- a/Tests/TensorFlowTests/LayerTests.swift
+++ b/Tests/TensorFlowTests/LayerTests.swift
@@ -859,6 +859,16 @@ final class LayerTests: XCTestCase {
         let layer = GlobalMinPool1D<Float>()
         let input = Tensor(shape: [2, 2, 2], scalars: (0..<8).map(Float.init))
         let computedGradient = gradient(at: input, layer) { $1($0).sum() }
+        // The expected value of the gradient was computed using the following Python code:
+        // ```
+        // import tensorflow as tf
+        // from tensorflow.python.keras import backend
+        // x = tf.reshape(tf.range(8, dtype=tf.float32), [2, 2, 2])
+        // with tf.GradientTape() as tape:
+        // tape.watch(x)
+        // y = tf.math.reduce_sum(backend.min(x, axis=[1]))
+        // print(tape.gradient(y, x))
+        // ```
         XCTAssertEqual(computedGradient.0,
                        [[[1.0, 1.0],
                          [0.0, 0.0]],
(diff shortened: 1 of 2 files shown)


Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -847,6 +847,69 @@ final class LayerTests: XCTestCase {
...
+        let input = Tensor(shape: [1, 10, 1], scalars: (0..<10).map(Float.init))
+        let output = layer.inferring(from: input)
+        let expected = Tensor<Float>([[0]])
+        XCTAssertEqual(output, expected)
+    }
+
+    func testGlobalMinPool1DGradient() {
+        let layer = GlobalMinPool1D<Float>()
+        let input = Tensor(shape: [2, 2, 2], scalars: (0..<8).map(Float.init))
+        let computedGradient = gradient(at: input, layer) { $1($0).sum() }
```
Shashi456: Please add a tensorflow 2.0 python equivalent code.

Shashi456:
Could you mention the paper these layers were introduced in? And if they are available in pytorch or tensorflow?
Also, since the layer itself does mostly look like a wrapper,  we also have a `function` layer which is used in situations like these.

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -859,6 +859,16 @@ final class LayerTests: XCTestCase {
         let layer = GlobalMinPool1D<Float>()
         let input = Tensor(shape: [2, 2, 2], scalars: (0..<8).map(Float.init))
         let computedGradient = gradient(at: input, layer) { $1($0).sum() }
+        // The expected value of the gradient was computed using the following Python code:
+        // ```
+        // import tensorflow as tf
+        // from tensorflow.python.keras import backend
+        // x = tf.reshape(tf.range(8, dtype=tf.float32), [2, 2, 2])
+        // with tf.GradientTape() as tape:
+        // tape.watch(x)
```
Shashi456: I don't think this is aligned properly.

Tests/TensorFlowTests/LayerTests.swift
kongzii: Added.

Tests/TensorFlowTests/LayerTests.swift
kongzii: Yea, I noticed it right after pushing. Fixed.

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -847,6 +847,99 @@ final class LayerTests: XCTestCase {
...
+    func testGlobalMinPool1DGradient() {
+        let layer = GlobalMinPool1D<Float>()
+        let input = Tensor(shape: [2, 2, 2], scalars: (0..<8).map(Float.init))
+        let computedGradient = gradient(at: input, layer) { $1($0).sum() }
+        // The expected value of the gradient was computed using the following Python code:
+        // ```
+        // import tensorflow as tf
+        // from tensorflow.python.keras import backend
+        // x = tf.reshape(tf.range(8, dtype=tf.float32), [2, 2, 2])
+        //     with tf.GradientTape() as tape:
```
Shashi456: Wrongly indented here.

Shashi456:
LGTM, after the indentation. But I'm still not entirely sold if we need a global min pooling layer or not. Maybe @dan-zheng or @eaplatanios can comment on that.

Tests/TensorFlowTests/LayerTests.swift
kongzii: Weird, looked ok in my ide. Fixed.

kongzii:
I can see that there is no GlobalMinPool in Python TF either. Is there a reason for that? I came across it while implementing https://ttic.uchicago.edu/~kgimpel/papers/he+etal.emnlp15.pdf, also, if there is layer for `Avg` and `Max`, why not `Min`?

cgarciae:
Since the most common activation function is relu, a Global Min Pooling layer would most likely yield a tensor full of 0s which might not be useful.

dan-zheng:
> Since the most common activation function is relu, a Global Min Pooling layer would most likely yield a tensor full of 0s which might not be useful.

That's actually a great point that I overlooked. I think @Shashi456 hinted at the same earlier.
(Global) max/average pooling layers seem standard in Keras/PyTorch, but I didn't find any mentions of min pooling layers.

@kongzii: I wonder if you had use cases in mind for min pooling layers? If not, let's close this issue.

PR #599, opened 2019-12-27
[Linear Algebra] Triangular solve

awav:
This PR adds `triangularSolve` operation with its gradient. Unfortunately, TF triangularSolve function doesn't do automatic broadcasting, so I had to implement it. For broadcasting case, I needed an implementation for extracting leading dimension to input tensors - in the code, it is called `extractLeadingDims`. This function can be useful in general.

 PR requires some polishing and open for discussion.

Thanks!


Sources/TensorFlow/Core/TensorShape.swift
```diff
@@ -94,6 +94,18 @@ public extension TensorShape {
     var endIndex: Int {
         return dimensions.endIndex
     }
+    
+    static func +  (lhs: TensorShape, rhs: TensorShape) -> TensorShape {
```
awav: This was surprising that TensorShape doesn't behave as a standard Array or a List. That can be useful in many cases!

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -197,3 +197,103 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
         _Raw.qr(self, fullMatrices: fullMatrices)
     }
 }
+
+
+// MARK: Solvers
+
+
+///
```
awav: I will add docstrings later

Tests/TensorFlowTests/OperatorTests/LinearAlgebraTests.swift
```diff
@@ -111,13 +111,78 @@ final class LinearAlgebraTests: XCTestCase {
...
+    }
+    
     static var allTests = [
         ("testCholesky", testCholesky),
         ("testQRDecompositionApproximation", testQRDecompositionApproximation),
         ("testTrace", testTrace),
         ("testTraceGradient", testTraceGradient),
         ("testLogdet", testLogdet),
-        ("testLogdetGradient", testLogdetGradient)
+        ("testLogdetGradient", testLogdetGradient),
```
Shashi456: Please do add `testTriangularSolve` to the tests as well.

Tests/TensorFlowTests/OperatorTests/LinearAlgebraTests.swift
```diff
@@ -111,13 +111,78 @@ final class LinearAlgebraTests: XCTestCase {
...
+            [1, 0, 0],
+            [2, 1, 0],
+            [3, 2, 1]
+        ])
+        let y = Tensor<Float>([1, 1, 3]).reshaped(to: [-1, 1])
+        let y2 = Tensor<Float>([y, y])
+        let x = triangularSolve(matrix: a, rhs: y2)
+    }
+    
+    /// Value of the gradient was computed using the following code:
```
Shashi456: I think it would be better if you could shift the python TensorFlow 2.0 code within the test itself.

awav: @Shashi456, yes, I do that but for different shapes to check that broadcasting in triangular solve works correctly.

awav:
@dan-zheng, @Shashi456 I updated PR

Shashi456:
@awav could you add a test for `extractLeadingDims` ?

awav:
@Shashi456, I added tests.

awav:
@dan-zheng, @Shashi456 Hello all, any comments on the code? :)

Sources/TensorFlow/Core/TensorShape.swift
dan-zheng: The proper way to fix this is to conform `TensorShape` to `Collection`: https://github.com/tensorflow/swift-apis/issues/609.

`RangeReplaceableCollection` provides a [default implementation](https://github.com/apple/swift/blob/02d8d83658033d88c8ec99db2ee37cff99292b6b/stdlib/public/core/RangeReplaceableCollection.swift#L967) for [`+(_:_:)`](https://developer.apple.com/documentation/swift/rangereplaceablecollection/2965327).

If `TensorShape` conforms to `RangeReplaceableCollection`, manually defining `static func +` becomes 
unnecessary. I'd prefer for someone to implement the `TensorShape: RangeReplaceableCollection` conformance instead of manually defining `static func +`: would you be interested?

awav:
@dan-zheng, 

> Sorry for the late reply!

No problem! 

> Would you be interested in exploring this in a separate PR? I'd say avoiding manual definitions for functions that have default implementations is sufficiently desirable that #609 is a blocker for this PR.

Yes, I can do it. Although, I need a bit more input on how to do it properly or just a simple example of how it can be done.

dan-zheng:
> Yes, I can do it. Although, I need a bit more input on how to do it properly or just a simple example of how it can be done.

Awesome!

You can start by adding an empty `extension TensorShape: Collection {}` conformance to `Sources/TensorFlow/Core/TensorShape.swift`. If you try to compile it, the compiler will produce error messages prompting you to implement the missing `Collection` protocol requirements. Xcode can even fill in dummy protocol requirement skeletons for you.

Do the same for `RangeReplaceableCollection`: `extension TensorShape: RangeReplaceableCollection { ... }`. When that conformance is done, you can use the [`RangeReplaceableCollection.+(_:_:)`](https://github.com/apple/swift/blob/02d8d83658033d88c8ec99db2ee37cff99292b6b/stdlib/public/core/RangeReplaceableCollection.swift#L967) default implementation!

awav:
@dan-zheng, I cannot assign myself to the task in Github's swift-apis. Guess, I'm not the project member.

According to the https://github.com/apple/swift/blob/02d8d83658033d88c8ec99db2ee37cff99292b6b/stdlib/public/core/RangeReplaceableCollection.swift#L967 link, `+` uses the append method, that means that TensorShape should be modifiable. Right now, TensorShape is immutable, which is not a bad thing.

dan-zheng:
> @dan-zheng, I cannot assign myself to the task in Github's swift-apis. Guess, I'm not the project member.

Ah, that's right. Would you like me to assign the issue to you?

> Also, according to [your link](https://github.com/apple/swift/blob/02d8d83658033d88c8ec99db2ee37cff99292b6b/stdlib/public/core/RangeReplaceableCollection.swift#L967), `+` uses the append method, that means that TensorShape should be modifiable. Right now, TensorShape is immutable, which is not a bad thing.

I think you may be confusing a few concepts. In Swift, we don't typically say that types like `TensorShape` as mutable or immutable. Instead, we categorize them into [value types and reference types](https://www.swiftbysundell.com/basics/value-and-reference-types). Value types (structs and enums) have value semantics, while reference types (classes) have reference semantics. I'd highly recommend checking out the article linked above to learn more, if you're not familiar with these concepts!

Value types like `TensorShape` can define `mutating` methods like [`RangeReplaceableCollection.append(_:)`](https://developer.apple.com/documentation/swift/rangereplaceablecollection/3018249-append). These methods can only be applied to mutable values (e.g. values declared with `var`).

---

Short example of `mutating` methods:

```swift
// A value type.
struct TensorShape {
  mutating func mutate() {}
}
var mutable = TensorShape()
mutable.mutate() // okay

let immutable = TensorShape()
immutable.mutate() // not okay
```
```console
$ swift test.swift
test.swift:9:11: error: cannot use mutating member on immutable value: 'immutable' is a 'let' constant
immutable.mutate() // not okay
~~~~~~~~~ ^
test.swift:8:1: note: change 'let' to 'var' to make it mutable
let immutable = TensorShape()
^~~
var
```

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -266,3 +266,122 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
         return (s, u, v)
     }
 }
+
+
+// MARK: Solvers
+
+
+/// Solves optionally batched systems of linear equations with upper or lower triangular matrices by backsubstitution.
```
dan-zheng: Line exceeds 100 columns, please wrap lines to 100 columns.

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -266,3 +266,122 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
         return (s, u, v)
     }
 }
+
+
+// MARK: Solvers
+
```
dan-zheng: minor: please remove extra empty line (use one max)
```suggestion
```

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -266,3 +266,122 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+
+/// Solves optionally batched systems of linear equations with upper or lower triangular matrices by backsubstitution.
+/// Returns solution to the system `A x = b`. Shape of return matches `b`.
+///
+/// - Parameters:
+///     - matrix: A batched matrix tensor.
+///     - rhs: A batched vector tensor.
```
dan-zheng: Please add parameter doc comments for all parameters: missing for `lower` and `adjoint`.

Tests/TensorFlowTests/OperatorTests/LinearAlgebraTests.swift
```diff
@@ -141,14 +141,121 @@ final class LinearAlgebraTests: XCTestCase {
         let computedGradient = gradient(at: input) { logdet($0).sum() }
         assertEqual(computedGradient, expectedGradient, accuracy: 1e-5)
     }
+
```
dan-zheng: minor: please remove extra empty line
```suggestion
```

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -266,3 +266,122 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+/// `[N, M, K]` and `[N, M]`, it will raise an error.
+///
+/// - Parameters:
+///     - left: Left tensor.
+///     - right: Right tensor.
+///     - ignoreLast: Number of dimensions from the end to ignore. Default value is zero.
+/// - Precondition: Left tensor rank cannot be lower than `ignoreLast`.
+/// - Precondition: Right tensor rank cannot be lower than `ignoreLast`.
+@inlinable
+public func extractLeadingDims<T: TensorFlowNumeric>(
```
dan-zheng: Please make both `extractLeadingDims` on functions `internal`. We don't want to expose these functions to users - please keep that in mind when deciding whether to make any function `public`.

I know these functions are used in `LinearAlgebraTests.swift` as well, but that test has `@testable import TensorFlow`, which should allow it to access internal functions. That's the proper way to handle "functions that should be private but need to called from tests".

awav: The initial thinking was that these functions could be useful for users and other parts of the project. But, I don't mind to make them internal. If needed they can be exported after.

dan-zheng: > The initial thinking was that these functions could be useful for users and other parts of the project. But, I don't mind to make them internal. If needed they can be exported after.

It's not appropriate to expose these APIs with their current non-idiomatic naming. Public APIs should be orthogonal (and somewhat simple, we should try hard not to expose multiple APIs with overlapping functionality), named appropriately, and defined at an appropriate location.

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -266,3 +266,121 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+///
+/// - Parameters:
+///     - left: Left tensor shape.
+///     - right: Right tensor shape.
+///     - ignoreLast: Number of dimensions from the end to ignore. Default value is zero.
+/// - Precondition: Left tensor rank cannot be lower than `ignoreLast`.
+/// - Precondition: Right tensor rank cannot be lower than `ignoreLast`.
+@inlinable
+public func extractLeadingDims(_ left: TensorShape, _ right: TensorShape, ignoreLast: Int = 0) -> TensorShape {
+    precondition(left.rank >= ignoreLast, "The left tensor must have at least rank 2.")
```
dan-zheng: I feel like this implementation can be vastly simplified. Something like:
* Assert that zipped leading dimensions are equal.
* Return `smallerShape.prefix`.

I don't have time to thoroughly review right now, but I'll try to review before this is merged.

awav: > Assert that zipped leading dimensions are equal.

Not sure, I understand what you mean by that.

> Return `smallerShape.prefix`.

Did you mean `largerShape`?

There are 3 cases for input tensors A and B:

1. A.shape > B.shape, let's say A.shape = [A, B, C, D, F] and b.shape = [D, F], then `extractLeadingDims` returns [A, B, C]
2. 1. A.shape < B.shape, `extractLeadingDims` returns same [A, B, C]
3. A.shape == B.shape, and `extractLeadingDims` returns []

```suggestion
    precondition(left.rank >= ignoreLast, "The left tensor must have at least rank `ignoreLast`.")
```
Same for the `right.rank` precondition.

dan-zheng: The `right.rank` precondition wasn't updated.

Sources/TensorFlow/Operators/LinearAlgebra.swift
dan-zheng: The `lower` and `adjoint` parameter doc comments are missing a description. Please don't neglect those.

awav: done

Sources/TensorFlow/Operators/LinearAlgebra.swift
awav: done

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -266,3 +266,121 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+    precondition(right.rank >= ignoreLast, "The right tensor must have at least rank 2.")
+    let (smallerShape, largerShape) = left.rank > right.rank ? (right, left) : (left, right)
+    let smaller = smallerShape.dropLast(ignoreLast)
+    let larger = largerShape.dropLast(ignoreLast)
+    let smallerEndIndex = smaller.count - 1
```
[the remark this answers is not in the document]
awav: I refactored the function.

awav:
@dan-zheng, I addressed your comments. The PR should be ready for review/merge. Thanks!

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -266,3 +266,128 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+/// - `Left.shape == Right.shape`, and `extractLeadingDims` returns `[]`
+///
+/// - Parameters:
+///     - left: Left tensor.
+///     - right: Right tensor.
+///     - ignoreLast: Number of dimensions from the end to ignore. Default value is zero.
+/// - Precondition: Left tensor rank cannot be lower than `ignoreLast`.
+/// - Precondition: Right tensor rank cannot be lower than `ignoreLast`.
+@inlinable
+internal func extractLeadingDims<T: TensorFlowNumeric>(
```
dan-zheng: `internal` is the default access level and can be omitted.
```suggestion
func extractLeadingDims<T: TensorFlowNumeric>(
```

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -266,3 +266,128 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+/// - `Left == Right`, and `extractLeadingDims` returns `[]`
+///
+/// - Parameters:
+///     - left: Left tensor shape.
+///     - right: Right tensor shape.
+///     - ignoreLast: Number of dimensions from the end to ignore. Default value is zero.
+/// - Precondition: Left tensor rank cannot be lower than `ignoreLast`.
+/// - Precondition: Right tensor rank cannot be lower than `ignoreLast`.
+@inlinable
+internal func extractLeadingDims(_ left: TensorShape, _ right: TensorShape, ignoreLast: Int = 0) -> TensorShape {
```
```suggestion
func extractLeadingDims(_ left: TensorShape, _ right: TensorShape, ignoreLast: Int = 0) -> TensorShape {
```

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -266,3 +266,128 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+        if broadcastMatrix {
+            return (triMatrixGrad.unbroadcasted(to: matrix.shape), rhsGrad)
+        } else if broadcastRhs {
+            return (triMatrixGrad, rhsGrad.unbroadcasted(to: rhs.shape))
+        }
+        return (triMatrixGrad, rhsGrad)
+    }
+    return (x, triangularSolveGrad)
+}
+
```
```suggestion
```

Sources/TensorFlow/Operators/LinearAlgebra.swift
awav: Interesting, why all "func _vjp*" methods use `internal` then?

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -266,3 +266,131 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+///   E.g. the left shape equals `[D, F]` and the right shape equals `[A, B, C, D, F]`, then
+///   the `extractLeadingDims` returns `[A, B, C]`.
+/// - When both shape sizes for left tensor and right tensor equal, the `extractLeadingDims` returns `[]`.
+///
+/// - Parameters:
+///     - left: Left tensor.
+///     - right: Right tensor.
+///     - ignoreLast: Number of dimensions from the end to ignore. Default value is zero.
+/// - Precondition: Left tensor rank cannot be lower than `ignoreLast`.
+/// - Precondition: Right tensor rank cannot be lower than `ignoreLast`.
```
awav: I updated the documentation once more time :)

awav: I updated the documentation once more time :)

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -266,3 +266,131 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+@differentiable
+public func triangularSolve<T: TensorFlowFloatingPoint>(
+    matrix: Tensor<T>,
+    rhs: Tensor<T>,
+    lower: Bool = true,
+    adjoint: Bool = false
+) -> Tensor<T> {
+    precondition(matrix.rank >= 2, "The matrix tensor must have at least rank 2.")
+    precondition(rhs.rank >= 2, "The rhs tensor must have at least rank 2.")
+    if matrix.rank < rhs.rank {
```
dan-zheng: I wonder if you used some reference implementation and tests for `triangularSolve` and `_vjpTriangularSolve` in this PR? I'm particularly curious about the `extractLeadingDims` helper function.

Does `triangularSolve` match [`tf.linalg.triangular_solve`](https://www.tensorflow.org/api_docs/python/tf/linalg/triangular_solve) in behavior, or does it have different/additional functionality?

awav: The `trisolve` from TensorFlow doesn't work for batched cases. https://github.com/tensorflow/tensorflow/issues/26204#issuecomment-498989256. I will make same contribution to the TF2.0 soon as well.

awav: @dan-zheng, the trisolve gradient you can find here https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/ops/linalg_grad.py#L607

awav:
@dan-zheng, okay, the good thing is that `trisolve` supports broadcasting in TensorFlow master.

awav:
@dan-zheng, @Shashi456, @asuhan Hello everyone, can someone take a look at this PR? I think it is almost there and needs a small push :) Thanks!

asuhan:
@dan-zheng Do we still need to check the `trisolve` broadcasting semantics?

dan-zheng:
> @dan-zheng Do we still need to check the `trisolve` broadcasting semantics?

@awav mentioned that batched input support for `_Raw.matrixTriangularSolve` was added in https://github.com/tensorflow/tensorflow/commit/b105944eb6c563849a085a1765d6700ee2c0f35c.

I verified that the behavior is consistent with the Swift implementation added in this patch: https://gist.github.com/dan-zheng/6c1404149aa16a4c7082047ce3890ee2

---

Currently, our version of `tensorflow` is [pinned at `v2.1.0-rc1`](https://github.com/apple/swift/blob/f7055f562fb66cfa1676bb956c8445cf4aff6589/utils/update_checkout/update-checkout-config.json#L329).

We should be able to remove the custom Swift broadcasting code in `func triangularSolve` after we:
- Upgrade the `tensorflow` version to a commit past https://github.com/tensorflow/tensorflow/commit/b105944eb6c563849a085a1765d6700ee2c0f35c in [`update-checkout-config.json`](https://github.com/apple/swift/tree/tensorflow/utils/update_checkout/update-checkout-config.json).
- [Regenerate Swift bindings for TensorFlow.](https://github.com/tensorflow/swift-apis/tree/master/Sources/TensorFlow/Bindings)

`func triangularSolve` should then directly forward to `_Raw.matrixTriangularSolve`.

PR #600, opened 2019-12-27
A scale layer for "mean network"

kongzii:
We already have a `Dropout` layer, but in the paper [Improving neural networks by preventingco-adaptation of feature detectors](https://arxiv.org/pdf/1207.0580.pdf) they mention `At test time, we use the “mean network” that contains all of the hidden units but with their outgoing  weights  halved  to  compensate  for  the  fact  that  twice  as  many  of  them  are  active.` But I did not find layer that could accomplish it. 

I am not sure about name of this layer and where to put it in source codes, so comments are welcome. 

Or perhaps layer like this is not wanted in swift-apis?

### DIFF
--- a/Sources/TensorFlow/Layers/Dropout.swift
+++ b/Sources/TensorFlow/Layers/Dropout.swift
@@ -68,3 +68,39 @@ public struct Dropout<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
         }
     }
 }
+
+/// A scale layer with LearningPhase distinguishing.
+///
+/// Consists of multiplying all nodes by a given value, but only in provided phase. 
+/// Used after Dropout layer to compensate for learning with dropped nodes.
+///
+/// Reference: [Improving neural networks by preventing co-adaptation of feature detectors]
+/// (https://arxiv.org/pdf/1207.0580.pdf).
+@frozen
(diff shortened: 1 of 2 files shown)


eaplatanios:
@kongzii the `Dropout` layer is already handling the scaling---it just does so in the opposite way. As you can see [here](https://github.com/tensorflow/swift-apis/blob/7330b8d7106489c906f85b92c8f5471f6959b10e/Sources/TensorFlow/Layers/Dropout.swift#L23), at training time the values are scaled by `1 / keepProbability`. So, if your probability is set to `0.5` this means that at train time the values are doubled. You can also think of this as keeping them fixed at training time and halving them at inference time (which is not exactly equivalent but serves the same purpose here).

kongzii:
> @kongzii the `Dropout` layer is already handling the scaling---it just does so in the opposite way. As you can see [here](https://github.com/tensorflow/swift-apis/blob/7330b8d7106489c906f85b92c8f5471f6959b10e/Sources/TensorFlow/Layers/Dropout.swift#L23), at training time the values are scaled by `1 / keepProbability`. So, if your probability is set to `0.5` this means that at train time the values are doubled. You can also think of this as keeping them fixed at training time and halving them at inference time (which is not exactly equivalent but serves the same purpose here).

Nice. I missed this, thanks.

PR #601, opened 2020-01-01
added precondition of matmul

SumanSudhir:
Listed in #517

- [ ]  precondition for matmul

### DIFF
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -2750,7 +2750,7 @@ public func matmul<Scalar: Numeric>(
 ) -> Tensor<Scalar> {
     precondition(lhs.rank >= 2, "The input must have rank greater than or equal to 2.")
     precondition(rhs.rank >= 2, "The input must have rank greater than or equal to 2.")
-    precondition(lhs.shape[1] == rhs.shape[0] , "Dimesion must be equal,but are \(lhs.shape[1]) and \(rhs.shape[0]).")
+    precondition(lhs.shape[1] == rhs.shape[0] , "Dimensions must be equal,but are \(lhs.shape[1]) and \(rhs.shape[0]).")
     if lhs.rank > 2 || rhs.rank > 2 {
         // TODO(TF-629): Conjugate to make compatible with the adjoint.
         return _Raw.batchMatMulV2(lhs, rhs, adjX: transposeLhs, adjY: transposeRhs)
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -2751,11 +2751,11 @@ public func matmul<Scalar: Numeric>(
     precondition(lhs.rank >= 2, "Input tensors must have at least rank 2.")
     precondition(rhs.rank >= 2, "Input tensors must have at least rank 2.")
     precondition(
-      lhs.shape[1] == rhs.shape[0],
-      """
-      Dimension mismatch: lhs.shape[1] is not equal to rhs.shape[0] \
-      (\(lhs.shape) vs \(rhs.shape)).
-      """
+        lhs.shape[1] == rhs.shape[0],
+        """
+        Dimension mismatch: lhs.shape[1] is not equal to rhs.shape[0] \
+        (\(lhs.shape) vs \(rhs.shape)).
+        """
     )
     if lhs.rank > 2 || rhs.rank > 2 {
         // TODO(TF-629): Conjugate to make compatible with the adjoint.
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -2748,15 +2748,8 @@ public func matmul<Scalar: Numeric>(
     _ rhs: Tensor<Scalar>,
     transposed transposeRhs: Bool = false
 ) -> Tensor<Scalar> {
-    precondition(lhs.rank >= 2, "Input tensors must have at least rank 2.")
-    precondition(rhs.rank >= 2, "Input tensors must have at least rank 2.")
-    precondition(
-        lhs.shape[1] == rhs.shape[0],
-        """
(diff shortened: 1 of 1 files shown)


Sources/TensorFlow/Operators/Math.swift
```diff
@@ -2748,6 +2748,9 @@ public func matmul<Scalar: Numeric>(
     _ rhs: Tensor<Scalar>,
     transposed transposeRhs: Bool = false
 ) -> Tensor<Scalar> {
+    precondition(lhs.rank >= 2, "The input must have rank greater than or equal to 2.")
+    precondition(rhs.rank >= 2, "The input must have rank greater than or equal to 2.")
+    precondition(lhs.shape[1] == rhs.shape[0] , "Dimesion must be equal,but are \(lhs.shape[1]) and \(rhs.shape[0]).")
```
Shashi456: The spelling of dimension and it would be plural in this case.

SumanSudhir: resolved

SumanSudhir:
After adding precondition time taken by function matmul is getting increased by around 60% when I ran it for 1M iterations as compared to matmul without precondition.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -2748,6 +2748,9 @@ public func matmul<Scalar: Numeric>(
     _ rhs: Tensor<Scalar>,
     transposed transposeRhs: Bool = false
 ) -> Tensor<Scalar> {
+    precondition(lhs.rank >= 2, "The input must have rank greater than or equal to 2.")
+    precondition(rhs.rank >= 2, "The input must have rank greater than or equal to 2.")
+    precondition(lhs.shape[1] == rhs.shape[0] , "Dimensions must be equal,but are \(lhs.shape[1]) and \(rhs.shape[0]).")
```
Shashi456: Nit: space after comma.

SumanSudhir: Does space have any effect on output??

Shashi456: Not exactly, that's just a formatting suggestion.

SumanSudhir:
@dan-zheng can you please review and merge it so, that I can start working on adding another issue

Sources/TensorFlow/Operators/Math.swift
dan-zheng: Please keep lines under 100 columns:

```suggestion
    precondition(
      lhs.shape[1] == rhs.shape[0],
      """
      Dimension mismatch: lhs.shape[1] is not equal to rhs.shape[0] \
      (\(lhs.shape) vs \(rhs.shape)).
      """
    )
```

dan-zheng:
> After adding precondition time taken by function matmul is getting increased by around 60% when I ran it for 1M iterations as compared to matmul without precondition.

Did you benchmark `matmul` without optimizations (`-Onone`)?

I think `-Onone` performance regressions due to adding preconditions are acceptable. Preconditions can be disabled with the `-Ounchecked` optimization mode, e.g. `swift build -c release -Xswiftc -Ounchecked`.

[This blog post](https://blog.krzyzanowskim.com/2015/03/09/swift-asserts-the-missing-manual/) has more info on which optimization modes disable assertions/preconditions.

SumanSudhir:
I run matmul for 1M iterations with and without precondition and I found that there was an increase of 60% in run time with precondition. Here is link to [gist](https://gist.github.com/SumanSudhir/486a169b23351479d051be6d536a576c)

dan-zheng:
It seems there's a test failure:
```
Test Case 'LayerTests.testSequential' started at 2020-01-03 23:05:26.311
Precondition failed: Dimension mismatch: lhs.shape[1] is not equal to rhs.shape[0] ([4, 1] vs [4, 1]).: file /swift-apis/Sources/TensorFlow/Operators/Math.swift, line 2753
```

That's interesting because this patch only adds preconditions without changing functional behavior. @SumanSudhir: could you please help investigate?

SumanSudhir:
Yaa ok, I will

SumanSudhir:
with input `([4,1] and [1,4])` even matmul without precondition is showing error


`
Fatal error: In[0] is not a matrix. Instead it has shape [2]: file /swift-base/tensorflow-swift-apis/Sources/TensorFlow/Bindings/EagerExecution.swift, line 300`

SumanSudhir:
can you please try by removing the last dimension mismatch precondition as the matmul function takes 4 arguments with two-argument as optional. This optional argument might affect the dimension mismatch condition.

dan-zheng:
> Can you please try by removing the last dimension mismatch precondition as the matmul function takes 4 arguments with two-argument as optional. This optional argument might affect the dimension mismatch condition.

Good point about `transposeLhs` and `transposeRhs` complicating the "dimension mismatch" precondition.

Conservatively dropping that precondition for now sounds good to me. For reference, [here's](https://github.com/tensorflow/tensorflow/blob/30b98cc73a8b046c8ce0915f6c0c7398a4947192/tensorflow/core/framework/common_shape_fns.cc#L432-L445) TensorFlow's logic for checking `matmul` inner/outer dimensions.

SumanSudhir:
Can it be merged now?

dan-zheng:
> Can it be merged now?

Yep!

SumanSudhir:
Can I PR for eye function? I have written code for eye and it is similar to as in TensorFlow. I have also described it [here](https://github.com/tensorflow/swift-apis/pull/578#issuecomment-570098153)

dan-zheng:
> Can I PR for eye function? I have written code for eye and it is similar to as in TensorFlow. I have also described it [here](https://github.com/tensorflow/swift-apis/pull/578#issuecomment-570098153)

The pending patch https://github.com/tensorflow/swift-apis/pull/578 by @Shashi456 already implements `eye` with functionality similar to your implementation. Let's try to make progress on that PR instead.

Feel free to look into [other issues](https://github.com/tensorflow/swift-apis/issues) to prevent duplicate work!

PR #603, opened 2020-01-02
Update year in license headers to 2020

Shashi456:
Happy new year and decade to everyone :)


BradLarson:
I might be mistaken, but I don't think we need to update these to the current date. The copyright notice should reflect date of original publication of the file, so I think 2019 is fine for all of these. New files will have 2020 in them.

Shashi456:
We did update most of them from 2018 to 2019 last year, @bradlarson. I'll close the PR in that case.

PR #592, opened 2019-12-22
[Linear Algebra] Add `logdet` operation

param087:
Adds support for `logdet` linear operation.
#562

PR #604, opened 2020-01-04
[Linear Algebra] Add det and slogdet

Shashi456:


Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -143,6 +143,24 @@ public func trace<T: TensorFlowNumeric>(_ matrix: Tensor<T>) -> Tensor<T> {
     return matrix.diagonalPart().sum(squeezingAxes: -1)
 }
 
+/// Computes the determinant of one or more square matrices.
+/// 
+/// - Parameter matrix: A tensor of shape `[..., M, M]`.
+/// - Returns: A tensor containing the determinants of all input submatrices.
+@inlinable 
+func det<T:TensorFlowFloatingPoint>(_ matrix: Tensor<T>) -> Tensor<T> {
```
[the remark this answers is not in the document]
Shashi456: Although the check makes sense, is there any way to avoid these precondition overheads o.O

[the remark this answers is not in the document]
awav: I think `precondition` itself is lightweight because the `det` operation and any other linalg operation are much more expensive in terms of clock time.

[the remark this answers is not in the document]
dan-zheng: `-Onone` performance regressions due to adding `precondition`s are not a big concern.
`precondition`s can be disabled with `-Ounchecked`, e.g. `swift build -c release -Xswiftc -Ounchecked`.

[This blog post](https://blog.krzyzanowskim.com/2015/03/09/swift-asserts-the-missing-manual/) has more info on which optimization modes disable assertions/preconditions.

All of these details are in the PR description of https://github.com/tensorflow/swift-apis/issues/517.

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -143,6 +143,24 @@ public func trace<T: TensorFlowNumeric>(_ matrix: Tensor<T>) -> Tensor<T> {
...
+    _Raw.matrixDeterminant(matrix)
+}
+
+/// Computes the sign and the log of the absolute value of the determinant of
+/// one or more square matrices.
+///
+/// - Parameter matrix: A tensor of shape `[..., N, M, M]`.
+@inlinable
+func slogdet<T:TensorFlowFloatingPoint>(_ matrix: Tensor<T>) -> (Tensor<T>, Tensor<T>) {
```
```suggestion
func slogdet<T: TensorFlowFloatingPoint>(_ matrix: Tensor<T>) -> (Tensor<T>, Tensor<T>) {
```

Sources/TensorFlow/Operators/LinearAlgebra.swift
```suggestion
func det<T: TensorFlowFloatingPoint>(_ matrix: Tensor<T>) -> Tensor<T> {
```

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -143,6 +143,24 @@ public func trace<T: TensorFlowNumeric>(_ matrix: Tensor<T>) -> Tensor<T> {
     return matrix.diagonalPart().sum(squeezingAxes: -1)
 }
 
+/// Computes the determinant of one or more square matrices.
```
```suggestion
/// Computes the determinant of an optionally batched square matrix.
```

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -143,6 +143,24 @@ public func trace<T: TensorFlowNumeric>(_ matrix: Tensor<T>) -> Tensor<T> {
...
+/// 
+/// - Parameter matrix: A tensor of shape `[..., M, M]`.
+/// - Returns: A tensor containing the determinants of all input submatrices.
+@inlinable 
+func det<T:TensorFlowFloatingPoint>(_ matrix: Tensor<T>) -> Tensor<T> {
+    _Raw.matrixDeterminant(matrix)
+}
+
+/// Computes the sign and the log of the absolute value of the determinant of
+/// one or more square matrices.
```
```suggestion
/// an optionally batched square matrix.
```

Shashi456:
@dan-zheng I'll take care of the doc comments like you mentioned in the SVD PR. Thanks for the precondition performance explanation.

dan-zheng:
> @dan-zheng I'll take care of the doc comments like you mentioned in the SVD PR. Thanks for the precondition performance explanation.

https://github.com/tensorflow/swift-apis/pull/578 is actually already merged. The doc comments for `Tensor.svd` are already polished:
```swift
    /// Returns the singular value decomposition of `self`, given that `self` is an optionally
    /// batched matrix.
    /// ...
    func svd(computeUV: Bool = true, fullMatrices: Bool = false) -> ( ... }
```

Sources/TensorFlow/Operators/LinearAlgebra.swift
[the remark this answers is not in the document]
dan-zheng: Not adding a `precondition` for now sounds fine.

There are other operations that expect `Tensor` arguments to have specific shapes that are also missing `precondition`s; `precondition`s can be added later at once.

Sources/TensorFlow/Operators/LinearAlgebra.swift
dan-zheng: Done.

Sources/TensorFlow/Operators/LinearAlgebra.swift
dan-zheng: Done.

Sources/TensorFlow/Operators/LinearAlgebra.swift
dan-zheng: Done.

Sources/TensorFlow/Operators/LinearAlgebra.swift
dan-zheng: Done.

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -177,6 +177,25 @@ public func trace<T: TensorFlowNumeric>(_ matrix: Tensor<T>) -> Tensor<T> {
...
+    _Raw.matrixDeterminant(matrix)
+}
+
+/// Computes the sign and the natural logarithm of the absolute value of the determinant of an
+/// optionally batched square matrix.
+///
+/// - Parameter matrix: A tensor of shape `[..., N, M, M]`.
+/// - Returns: A tensor containing the determinants of all input submatrices.
+@inlinable
+func slogdet<T: TensorFlowFloatingPoint>(_ matrix: Tensor<T>) -> (Tensor<T>, Tensor<T>) {
```
dan-zheng: When evaluating API names, I pretend that I'm encountering the API for the first time:
* Is the function name easily understandable?
* If the function returns a tuple of values, is the meaning of each value clear?
  * I think tuple element labels are often useful for providing additional meaning.

---

I think the name `slogdet` has room for improvement:
* The name is domain-specific and doesn't make sense to people unfamiliar with `numpy.linalg.slogdet` or `tf.linalg.slogdet`.
* The returned tuple has no element labels, so the meaning of each tuple element is not clear.

---

How about (make sure to wrap to 100 columns):
```suggestion
func signAndLogDeterminant<T: TensorFlowFloatingPoint>(_ matrix: Tensor<T>) -> (sign: Tensor<T>, logDeterminant: Tensor<T>) {
```

dan-zheng: Is it possible to define the derivative of `slogdet`?

---

Motivation: `slogdet` overlaps with the previously defined `func logdet` below:

```swift
@differentiable(wrt: matrix where T: TensorFlowFloatingPoint)
func logdet<T: TensorFlowFloatingPoint>(_ matrix: Tensor<T>) -> Tensor<T> {
    return 2.0 * log(cholesky(matrix).diagonalPart()).sum(squeezingAxes: -1) 
}
```

(Aside: `func logdet` should be renamed to `func logDeterminant`.)

`func logdet` is differentiable, while `func slogdet` is not. If we can define a derivative for `func slogdet`, we could consider changing the implementation of `func logdet` to call `func slogdet` instead of custom computation.

dan-zheng: cc @param087, who implemented `func logdet` in https://github.com/tensorflow/swift-apis/pull/592.
I wonder what you used as the reference implementation for `func logdet`? I'm curious why it uses custom logic instead of forwarding to a `_Raw` op.

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -177,6 +177,25 @@ public func trace<T: TensorFlowNumeric>(_ matrix: Tensor<T>) -> Tensor<T> {
     return matrix.diagonalPart().sum(squeezingAxes: -1)
 }
 
+/// Computes the determinant of an optionally batched matrix.
+/// 
+/// - Parameter matrix: A tensor of shape `[..., M, M]`.
+/// - Returns: A tensor containing the determinants of all input submatrices.
+@inlinable 
+func det<T: TensorFlowFloatingPoint>(_ matrix: Tensor<T>) -> Tensor<T> {
```
```suggestion
func determinant<T: TensorFlowFloatingPoint>(_ matrix: Tensor<T>) -> Tensor<T> {
```

Sources/TensorFlow/Operators/LinearAlgebra.swift
dan-zheng: @Shashi456: to make progress on this, could you please rename the existing `func logdet` to `func logDeterminant`? We can try to unify or code share `func logDeterminant` and `func signAndLogDeterminant` in a follow-up.

Shashi456:
@dan-zheng will do so.

Sources/TensorFlow/Operators/LinearAlgebra.swift
dan-zheng: Actually, how about just the name `func logDeterminant` instead of `func signAndLogDeterminant`? The result element labels sufficiently distinguish the sign and log det results.

awav: @dan-zheng I would recommend against this renaming. `slogdet` and `logdet` are well known and used by every single scientific language I know - Python (numpy), R, Julia, Matlab. The `det` abbreviation is also broadly used in the literature as well.

awav: @Shashi456 ^^^

Shashi456: @dan-zheng since I mostly take inspiration from the tensorflow-python API and check numpy linear algebra library for additional consistency. I initially named the functions slogdet and det.

awav:
Continuing my comment https://github.com/tensorflow/swift-apis/pull/604#discussion_r365685758,

@dan-zheng, in general, I would prefer short and concise name for functions, because often math-like code has long and nested expressions which by itself are really hard to read, https://github.com/GPflow/GPflow/blob/develop/gpflow/expectations/squared_exponentials.py#L79

> The name is domain-specific and doesn't make sense to people unfamiliar with numpy.linalg.slogdet or tf.linalg.slogdet

It is not about numpy only, these abbreviations are used everywhere Python (numpy, tensorflow, pytorch), R, matlab, Julia and just common literature about linear algebra.

Also, for me renaming of `det` to `determinant` looks like renaming `log` to `logarithm`.
Please, please, don't do this!

dan-zheng:
> > The name is domain-specific and doesn't make sense to people unfamiliar with numpy.linalg.slogdet or tf.linalg.slogdet
> 
> It is not about numpy only, these abbreviations are used everywhere: Python (numpy, tensorflow, pytorch), R, matlab, Julia and just commonly in the literature about linear algebra.
> 
> Also, for me renaming of `det` to `determinant` looks like renaming `log` to `logarithm`.
> Please, please, don't do this!

Okay, that's fair! For every API, we could try to find an idiomatic Swift name or to use an existing domain-specific term-of-art name.

In some cases, we chose new idiomatic Swift names:
* Reduction APIs: `alongAxes:` and `squeezingAxes:` vs `keep_dims=True` and `keeps_dims=False`
* Non-standard APIs: [`Tensor.bandPart(subdiagonalCount:superdiagonalCount:`)](https://github.com/tensorflow/swift-apis/pull/588) instead of argumentless `tf.linalg.band_part`

In this case, I feel using an existing term-of-art name is reasonable! `det`, `logdet`, and `slogdet` do seem like standard operations. Let's go with those names for name, but let's still add `(sign: Tensor<T>, logDeterminant: Tensor<T>` result tuple element labels for `slogdet`.

Thanks for your feedback!

Sources/TensorFlow/Operators/LinearAlgebra.swift
dan-zheng: Obsolete. Let's continue to use `det` after discussion in https://github.com/tensorflow/swift-apis/pull/604#issuecomment-573653157.

Sources/TensorFlow/Operators/LinearAlgebra.swift
dan-zheng: I feel reusing existing term-of-art names here is reasonable. Could you please add `(sign:logDeterminant:)` result tuple element labels to `slogdet` please?

```swift
func slogdet<T: TensorFlowFloatingPoint>(_ matrix: Tensor<T>) -> (sign: Tensor<T>, logDeterminant: Tensor<T>) {
```

dan-zheng: Obsolete. Let's continue to use `slogdet` after discussion in https://github.com/tensorflow/swift-apis/pull/604#issuecomment-573653157.

Shashi456:
@dan-zheng WDYT?

Sources/TensorFlow/Operators/LinearAlgebra.swift
param087: > cc @param087, who implemented `func logdet` in #592.
> I wonder what you used as the reference implementation for `func logdet`? I'm curious why it uses custom logic instead of forwarding to a `_Raw` op.

@dan-zheng, I had taken reference from https://github.com/tensorflow/tensorflow/blob/r2.1/tensorflow/python/ops/linalg/linalg_impl.py#L72-L103 to implement `logdet`.  Considering the python implementation of `logdet` and making it differentiable, I found custom logic is more promising than that of forwarding to a `_Raw` op.

Tests/TensorFlowTests/OperatorTests/MatrixTests.swift
```diff
@@ -121,7 +120,7 @@ final class MatrixTests: XCTestCase {
         let expectedGrad = 2 * t3.bandPart(subdiagonalCount: 0, superdiagonalCount: -1)
         XCTAssertEqual(computedGrad, expectedGrad)
     }
-
+
```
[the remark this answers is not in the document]
Shashi456: I didn't touch this file. I have no idea where this popped out from xD. Will remove this.

Sources/TensorFlow/Operators/LinearAlgebra.swift
```diff
@@ -177,6 +177,29 @@ public func trace<T: TensorFlowNumeric>(_ matrix: Tensor<T>) -> Tensor<T> {
...
+/// optionally batched square matrix.
+///
+/// - Parameter matrix: A tensor of shape `[..., N, M, M]`.
+/// - Returns: 
+///   - sign: A tensor with shape `[N]`, representing the signs of the natural logarithms of the 
+///     determinants of input submatrices.
+///   - logAbsDeterminant: A tensor with shape `[N]`, representing the natural logarithms of the 
+///     absolute values of the determinants of input submatrices.
+@inlinable
+func slogdet<T: TensorFlowFloatingPoint>(_ matrix: Tensor<T>) -> (sign: Tensor<T>, logAbsDeterminant: Tensor<T>) {
```
dan-zheng: Please remember to wrap lines to 100 columns. I'll fix this now.

Tests/TensorFlowTests/OperatorTests/LinearAlgebraTests.swift
```diff
@@ -115,6 +115,22 @@ final class LinearAlgebraTests: XCTestCase {
         assertEqual(computedGradient, expectedGradient, accuracy: 1e-16)
     }
 
+    func testDet() {
+        let matrix = Tensor<Float>(shape: [1, 4, 4], scalars: (0..<16).map(Float.init))
+        let computedDet = det(matrix)
+        let expectedDet = Tensor<Float>([0])
```
dan-zheng: Could you please at one more test for `det` and `slogdet`, with different expected values?
I'd recommend using input tensors with shape `[2, 2, 4, 4]`.

Shashi456:
tests locally pass.

PR #608, opened 2020-01-08
Handle `C` imports for Windows

compnerd:
Windows does not use `Glibc` but rather `ucrt` and `MSVCRT` for the libc
module.  Add a platform specific case.


compnerd:
CC: @marcrasi @dan-zheng

PR #611, opened 2020-01-09
Runtime: correct a thread allocation leak

compnerd:
`pthread_key_create` takes a destructor which is supposed to *release*
the resources associated with the key.  The code was accidentally
retaining the value instead, which would increment the reference count
before invalidating the key.  Release the value instead.


compnerd:
CC: @rxwei @saeta

compnerd:
For additional context:

`static var local` in `_ThreadLocalState` does the following:

```
pthread_setspecific(key, Unmanaged.passRetained(state).toOpaque())
```

`Unmanaged.passRetained` will return a +1'ed object which requires manual release.  The destructor will release the value by means of the `release` balancing the +1 in the allocation.

pschuh:
Note. This is clearer, but I think that the original code is also correct.

PR #612, opened 2020-01-09
Adding He and LeCun initializers

BradLarson:
This adds the He (also referred to as Kaiming) and LeCun initializers for tensor values, along with helper functions for layer initialization (including one for the existing Glorot normal initializer).

Tests have been provided for these new initializers, and seeds have been set for the various initializer tests to stabilize test values.


Sources/TensorFlow/Initializers.swift
```diff
@@ -552,6 +552,94 @@ public extension Tensor where Scalar: TensorFlowFloatingPoint {
...
+    /// It draws random samples from a uniform distribution between `-limit` and `limit`
+    /// generated by the default random number generator, where `limit` is
+    /// `sqrt(6 / fanIn)` and `fanIn` represents the number of input features multiplied by the 
+    /// receptive field size.
+    ///
+    /// Reference: ["Delving Deep into Rectifiers: Surpassing Human-Level Performance on ImageNet 
+    /// Classification"](https://www.cv-foundation.org/openaccess/content_iccv_2015/papers/He_Delving_Deep_into_ICCV_2015_paper.pdf)
+    ///
+    /// - Parameters:
+    ///   - shape: The dimensions of the tensor.
```
dan-zheng: Could you please add a parameter doc comment for `seed`? Some other initializers have it.
```swift
    ///   - seed: The seed value.
```
Same elsewhere please.

BradLarson: This was also missing from the Glorot uniform initializers, so I've added it into the new ones as well as the old.

Shashi456:
I wanted to know your opinions @dan-zheng and @bradlarson, on maybe having type aliases or like wrapper call with the alternative names for initializers? I know typealiases aren't the best way to go about doing this. But since both he/kaiming initialization and Xavier/glorot are popular, maybe we could have a couple?

dan-zheng:
> I wanted to know your opinions @dan-zheng and @BradLarson, on maybe having type aliases or like wrapper call with the alternative names for initializers? I know typealiases aren't the best way to go about doing this. But since both he/kaiming initialization and Xavier/glorot are popular, maybe we could have a couple?

I wonder what do you have in mind exactly? Is the purpose of adding typealiases to make code shorter?

If so, I'd push back because:
* Initializer argument labels already seem quite short, e.g. `Tensor.init(leCunUniform:seed:)`. They seem clear and as short as possible to me.
* Exposing multiple names for public APIs in a library is not good. It's better to have one canonical name for things.
* The existing API names are very clear, and clarity is more important than brevity. Check out the [Swift API design guidelines](https://swift.org/documentation/api-design-guidelines/#fundamentals).

Shashi456:
@dan-zheng, the purpose would be make the api accessible/user-friendly in a sense. For example, tensorflow uses `glorot` as the standard name while pytorch uses `xavier` instead. It's probably just a preference, but down the line having the flexibility might help.

dan-zheng:
> @dan-zheng, the purpose would be make the api accessible/user-friendly in a sense. For example, tensorflow uses `glorot` as the standard name while pytorch uses `xavier` instead. It's probably just a preference, but down the line having the flexibility might help
> 
> Your points make sense though. Thanks!

I see, thanks for the clarification. The motivation is that different libraries have different names for the same initializers (e.g. `glorot` in TensorFlow, `xavier` in PyTorch), so it may be nice to support all names. (To be clear, this would involve extra declarations, not typealiases.)

I still feel that exposing multiple names for public APIs in a library is not good: it's better to have one canonical name for things.

If some people strongly prefer different names for a given API, it seems more productive to discuss the best single name for the API rather than supporting multiple names.

BradLarson:
@Shashi456 - Generally, I've only seen frameworks pick one variation or another and not support both. We've been following the [TensorFlow](https://www.tensorflow.org/api_docs/python/tf/keras/initializers) / [Keras](https://keras.io/initializers/) conventions where these are specified by someone's surname, but I know that [JAX](https://github.com/google/jax/blob/master/jax/nn/initializers.py#L73) / [PyTorch](https://pytorch.org/docs/stable/nn.init.html) / [Caffe](https://github.com/BVLC/caffe/issues/1537) use someone's given name for these. I don't know that I've seen an implementation where both were used as synonyms, and I personally prefer the use of surname for these.

I figured that as long as the alternative that someone might be familiar with was called out in documentation (as I've done for the He (Kaiming) initializer and could do for the Glorot (Xavier) one as well), that might make it easy to see what to use when adapting code.

Shashi456:
Yes alright. @bradlarson yeah, that's a nice suggestion. I was exactly thinking from a code adaption perspective, since we port models from one framework to another. 
The comments idea sounds good to me. 

The difference of names is sometimes confusing.

BradLarson:
@Shashi456 - I updated the documentation on the Glorot initializers to at least mention the Xavier synonym, which should hopefully help when people search for this in the process of porting models over.

Shashi456:
@bradlarson, are there any other initializers that we need to support?

BradLarson:
@Shashi456 - I can't think of obvious cases off the top of my head. These two were important for replicating network designs in benchmarks and tests, and should bring parity to what JAX, Keras, PyTorch, and others provide by default. We needed them for building out apples-to-apples comparisons of networks in other frameworks (I believe JAX uses LeCun initialization by default, and He initialization is common in many places).

There's [a great walkthrough of how different initialization schemes were arrived at](https://towardsdatascience.com/weight-initialization-in-neural-networks-a-journey-from-the-basics-to-kaiming-954fb9b47c79) that makes the case that maybe we should be using the He (Kaiming) initialization as our default instead of Glorot (Xavier) for the common cases involving deep networks using ReLU activations, but that's a separate discussion.

PR #614, opened 2020-01-11
added more precondition

SumanSudhir:
#517  
**Precondiiton for function:**
- [ ] replacing
- [ ] all
- [ ] any
- [ ] max
- [ ] min
- [ ] argmax
- [ ] argmin
- [ ] sum
- [ ] product
- [ ] mean
- [ ] variance
- [ ] cumulativeSum
- [ ] cumulativeProduct

### DIFF
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -2249,7 +2249,7 @@ public extension Tensor where Scalar: Numeric {
         reverse: Bool = false
     ) -> Tensor {
         precondition(axis.rank == 0, "Axis must have rank 0.")
-        precondition(areAxesInRange(axis), "All axis must be in the range `[-rank, rank)`.")
+        precondition(areAxesInRange(axis), "Axis must be in the range `[-rank, rank)`.")
         return _Raw.cumsum(self, axis: axis, exclusive: exclusive, reverse: reverse)
     }
 
@@ -2323,7 +2323,7 @@ public extension Tensor where Scalar: Numeric {
         exclusive: Bool = false,
         reverse: Bool = false
     ) -> Tensor {
-        precondition(areAxesInRange(axis), "All axis must be in the range `[-rank, rank)`.")
+        precondition(areAxesInRange(axis), "Axis must be in the range `[-rank, rank)`.")
         return _Raw.cumprod(self, axis: axis, exclusive: exclusive, reverse: reverse)
     }
 }
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -1543,7 +1543,8 @@ public extension Tensor {
     @inlinable
     @differentiable(wrt: (self, other) where Scalar: TensorFlowFloatingPoint)
     func replacing(with other: Tensor, where mask: Tensor<Bool>) -> Tensor {
-        _Raw.select(condition: mask, t: other, e: self)
+        precondition(self.shape == other.shape, "`self` and `other` must have the same shape.")
+        return _Raw.select(condition: mask, t: other, e: self)
     }
 }
 
@@ -1590,6 +1591,7 @@ public extension Tensor where Scalar == Bool {
     /// - Precondition: Each value in `axes` must be in the range `-rank..<rank`.
     @inlinable
     func all(squeezingAxes axes: Int...) -> Tensor {
+        precondition(areAxesInRange(axes), "All axes must be in the range `[-rank, rank)`.")
         let axes = axes.map(Int32.init)
         return _Raw.all(self, reductionIndices: Tensor<Int32>(axes), keepDims: false)
     }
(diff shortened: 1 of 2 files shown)


SumanSudhir:
@dan-zheng I have change precondition of `areAxesInRange()` as it is not necessary that the input axes tensor must have `rank 1`. Eg: `cumlativeSum()` and `moments`

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -2229,7 +2248,9 @@ public extension Tensor where Scalar: Numeric {
         exclusive: Bool = false,
         reverse: Bool = false
     ) -> Tensor {
-        _Raw.cumsum(self, axis: axis, exclusive: exclusive, reverse: reverse)
+        precondition(axis.rank == 0, "Axis must have rank 0.")
+        precondition(areAxesInRange(axis), "All axis must be in the range `[-rank, rank)`.")
```
Shashi456: In `cumprod` and `cumsum`, if the axis is plural then I think the documentation must read, `All axes must be in range ...,` Like the previous conditions.  
If it's a singular value, is calling the function necessary instead of just doing the check locally ?

SumanSudhir: Previously it was given axis instead of axes and also the same comment was written above that `axis` must be in the range `-rank..<rank`. So I followed the same pattern in writing the precondition. Changing the input variable from axis to axes may conflict with some written test cases.

Shashi456: Oh then I think, you don't have to right `all axis`, it's just referring to a singular axis. You can start the statement with `Axis must be in range...`

SumanSudhir: I think `rank 0` tensor means that there is only one value in axis tensor. So previously it was written as singular.
`var input: Tensor<Int32> = Tensor(5)`
` input.rank` is giving `0` as output

SumanSudhir:
@dan-zheng can you please review it?

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -2229,7 +2248,9 @@ public extension Tensor where Scalar: Numeric {
         exclusive: Bool = false,
         reverse: Bool = false
     ) -> Tensor {
-        _Raw.cumsum(self, axis: axis, exclusive: exclusive, reverse: reverse)
+        precondition(axis.rank == 0, "Axis must have rank 0.")
+        precondition(areAxesInRange(axis), "Axis must be in the range `[-rank, rank)`.")
```
dan-zheng: > @dan-zheng I have change precondition of `areAxesInRange()` as it is not necessary that the input axes tensor must have `rank 1`. Eg: `cumlativeSum()` and `moments` the former require `rank 0` tensor while later takes `rank 1` tensor as input

Thanks for the clarification!

The correct fix is actually to add another precondition helper method called `isAxisInRange` taking a `Tensor<Int32>`, not to call `areAxesInRange(_: Tensor<Int32>)` on a scalar `Tensor<Int32>` representing a single axis.

`isAxisInRange(_: Tensor<Int32>` should have a precondition:
```swift
precondition(axis.rank == 0, "Axis must have rank 0.")
```

And `areAxesInRange(_: Tensor<Int32>` should have the old precondition:

```swift
precondition(axes.rank == 0, "Axis must have rank 1.")
```

```suggestion
        precondition(isAxisInRange(axis), "Axis must be in the range `[-rank, rank)`.")
```

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -2294,6 +2314,7 @@ public extension Tensor where Scalar: Numeric {
     ///   - exclusive: Indicates whether to perform an exclusive cumulative product.
     ///   - reverse: Indicates whether to perform the cumulative product in reversed order.
     /// - Returns: Result of the cumulative product operation.
+    /// - Precondition: `axis.rank` must be `0`.
```
```suggestion
    /// - Precondition: `axis` must have rank `0`.
```

dan-zheng: Done in https://github.com/tensorflow/swift-apis/pull/614/commits/e1b6b405d96f412c4c284757ac823b6f1cf5534a.

dan-zheng:
Weirdly, there's a test failure:
```
/swift-apis/Tests/TensorFlowTests/OperatorTests/LinearAlgebraTests.swift:77: error: LinearAlgebraTests.testSVD : XCTAssertEqual failed: ("-3.0785367") is not equal to ("-3.0785267") +/- ("1e-05") -
/swift-apis/Tests/TensorFlowTests/OperatorTests/LinearAlgebraTests.swift:89: error: LinearAlgebraTests.testSVD : XCTAssertEqual failed: ("-3.0785367") is not equal to ("-3.0785267") +/- ("1e-05") -
Test Case 'LinearAlgebraTests.testSVD' failed (0.012 seconds)
```

Is this a flaky test? Rerunning tests now.

Issue #609, opened 2020-01-09
Conform `TensorShape` to `Collection` protocols


dan-zheng:
`TensorShape` represents the shape of a tensor. It wraps an `[Int]` and provides additional utilities like `var rank: Int` and `var contiguousSize: Int`.

`TensorShape` should conform to `Collection` and related protocols:
- `Collection`
- `MutableCollection`
- `RandomAccessCollection`
- `RangeReplaceableCollection`

`Collection` protocol requirement implementations should forward to the underlying implementation from `[Int]`.

---

### First steps

You can start by adding empty conformances to `Sources/TensorFlow/Core/TensorShape.swift`, then seeing what compilation errors arise:
```swift
extension TensorShape: Collection {}
```

If you try to compile it, the compiler will produce error messages prompting you to implement the missing `Collection` protocol requirements. Xcode can even fill in protocol requirement skeletons for you.

Use the same technique to implement conformances to other `Collection` protocols.

awav:
Hello @dan-zheng, please assign me to this task. I've started working on it. Thanks!

dan-zheng:
Done in https://github.com/tensorflow/swift-apis/pull/615. Thanks @awav!
https://github.com/tensorflow/swift-apis/pull/599 should be unblocked now: no need for manual `TensorShape.+(_:_:)` definitions.

PR #615, opened 2020-01-11
Conform TensorShape to Collection-like protocols

awav:
PR addresses #609

### DIFF
--- a/Sources/TensorFlow/Core/TensorShape.swift
+++ b/Sources/TensorFlow/Core/TensorShape.swift
@@ -159,7 +159,7 @@ extension TensorShape: RangeReplaceableCollection {
     @inlinable
     public mutating func replaceSubrange<C>(
         _ subrange: Range<Index>, with newElements: C
-    ) where C : Collection, Element == C.Element {
+    ) where C: Collection, Element == C.Element {
         dimensions.replaceSubrange(subrange, with: newElements)
     }
 }
--- a/Sources/TensorFlow/Core/TensorShape.swift
+++ b/Sources/TensorFlow/Core/TensorShape.swift
@@ -73,21 +73,9 @@ public struct TensorShape: ExpressibleByArrayLiteral {
     }
 }
 
-extension TensorShape: Collection, MutableCollection, RandomAccessCollection, RangeReplaceableCollection {
+extension TensorShape: Collection, MutableCollection {
     public typealias Element = Int
     public typealias Index = Int
-    public typealias Indices = Range<Int>
-    public typealias SubSequence = Self
-    
-    @inlinable
-    public init() {
-        self.init([])
-    }
-    
-    @inlinable
-    public func index(after i: Index) -> Index {
-        return dimensions.index(after: i)
-    }
     
     /// The rank of the shape (i.e. the number of dimensions).
     @inlinable
@@ -109,13 +97,18 @@ extension TensorShape: Collection, MutableCollection, RandomAccessCollection, Ra
     public var endIndex: Index {
         return dimensions.endIndex
     }
-
-    /// Access the size of the i-th dimension.
-    /// - Parameter index: The index of a dimension.
+    
     @inlinable
-    public subscript(index: Index) -> Element {
-        _read { yield dimensions[index] }
-        _modify { yield &dimensions[index] }
+    public func index(after i: Index) -> Index {
+        return dimensions.index(after: i)
+    }
+    
+    /// Access the size of the i-th dimension.
+    /// - Parameter position: The index of a dimension.
+    @inlinable
+    public subscript(position: Index) -> Element {
+        _read { yield dimensions[position] }
+        _modify { yield &dimensions[position] }
     }
 
     /// Access the size of the i-th dimension.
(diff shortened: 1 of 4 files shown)


Shashi456:
Just a question, what's the use case of an append function in a tensor shape protocol ?

dan-zheng:
> Just a question, what's the use case of an append function in a tensor shape protocol ?

In practice, defining `append` is necessary because it's a [`RangeReplaceableCollection`](https://developer.apple.com/documentation/swift/rangereplaceablecollection) protocol requirement. Conforming `TensorShape` to `RangeReplaceableCollection` enables the `RangelReplaceableCollection.+(_:_:)` functions for concatenation.

Semantically, `TensorShape.append(_:)` is basically just a mutating version of concatenation (of one element). It has essentially the same use cases as `+`, i.e. concatenating a shape with a dimension.

Sources/TensorFlow/Core/TensorShape.swift
```diff
@@ -73,43 +73,78 @@ public struct TensorShape: ExpressibleByArrayLiteral {
     }
 }
 
-public extension TensorShape {
+extension TensorShape: Collection, MutableCollection, RandomAccessCollection, RangeReplaceableCollection {
```
dan-zheng: This is minor, but would you mind separating conformances to different protocols into separate extensions? Each conformance extension should define just the requirement implementations for that conformance : for example, `extension TensorShape: RandomAccessCollection { .. }` should contain just `RandomAccessCollection` requirement implementations. This "correct separation" isn't enforced by the compiler, so please take care!

Motivation: separating conformances is good for separation of concerns.

awav:
@dan-zheng, is there a way to speed up swift compilation? Any changes in files cause whole project rebuild.

dan-zheng:
> @dan-zheng, is there a way to speed up swift compilation? Any changes in files cause whole project rebuild.

Slow `TensorFlow` compilation time is really frustrating, we've noticed the same. https://github.com/tensorflow/swift-apis/issues/618 tracks improving the situation, though it doesn't have concrete action items yet. Let's continue compile-time improvement discussions there.

Sources/TensorFlow/Core/TensorShape.swift
```diff
@@ -73,45 +73,98 @@ public struct TensorShape: ExpressibleByArrayLiteral {
     }
 }
 
-public extension TensorShape {
+extension TensorShape: Collection, MutableCollection {
```
dan-zheng: Please separate `extension TensorShape: Collection` and `extension TensorShape: MutableCollection`. Make each extension contain only protocol requirement implementations for that protocol, please.

Sources/TensorFlow/Core/TensorShape.swift
awav: @dan-zheng, I left Collection and MutableCollection together in single extension block, MutableCollection requires same subscript methods. RandomAccessCollection and RangeReplaceableCollection are now separate extensions.

dan-zheng: > @dan-zheng, I left Collection and MutableCollection together in single extension block, MutableCollection requires same subscript methods.

Ah, I see. `MutableCollection` requires `subscript` with additional `set` accessor? If so, that sounds reasonable to me! Thanks for the context.

Sources/TensorFlow/Core/TensorShape.swift
dan-zheng: Never mind, resolved: https://github.com/tensorflow/swift-apis/pull/615#discussion_r365906757

Sources/TensorFlow/Core/TensorShape.swift
```diff
@@ -73,45 +73,98 @@ public struct TensorShape: ExpressibleByArrayLiteral {
...
+    public typealias SubSequence = Self
+
+    @inlinable
+    public init() {
+        self.init([])
+    }
+
+    @inlinable
+    public mutating func append(_ newElement: Element) {
+        dimensions.insert(newElement, at: endIndex)
```
dan-zheng: Does this not work?
```suggestion
        dimensions.append(newElement)
```

Tests/TensorFlowTests/TensorTests.swift
```diff
@@ -66,12 +66,32 @@ final class TensorTests: XCTestCase {
         XCTAssertFalse(tensor == tensor.reshaped(to: [2, 3]))
         XCTAssertTrue(tensor != tensor.reshaped(to: [2, 3]))
     }
+    
+    func testTensorShapeCollectionConformance() {
```
dan-zheng: minor:
```suggestion
    func testTensorShapeCollectionOperations() {
```

Sources/TensorFlow/Core/TensorShape.swift
awav: > Ah, I see. MutableCollection requires subscript with additional set accessor?

That's correct!

Sources/TensorFlow/Core/TensorShape.swift
awav: yes, it works!

Sources/TensorFlow/Core/TensorShape.swift
```diff
@@ -73,45 +73,98 @@ public struct TensorShape: ExpressibleByArrayLiteral {
...
     @inlinable
-    subscript(bounds: Range<Int>) -> TensorShape {
+    public subscript(bounds: Range<Int>) -> TensorShape {
         get { return TensorShape(dimensions[bounds]) }
         set { dimensions[bounds] = ArraySlice(newValue.dimensions) }
     }
 }
 
+extension TensorShape: RandomAccessCollection {
+    public typealias Indices = Range<Int>
```
dan-zheng: Please move `typealias Indices = Range<Int>` to the `extension TensorShape: Collection, MutableCollection` above, since `Indices` is an associated type of `Collection`.

Sources/TensorFlow/Core/TensorShape.swift
```diff
@@ -73,45 +73,98 @@ public struct TensorShape: ExpressibleByArrayLiteral {
...
+
+    @inlinable
+    public mutating func append<S : Sequence>(contentsOf newElements: S) where Element == S.Element {
+        dimensions.append(contentsOf: newElements)
+    }
+
+    @inlinable
+    public mutating func replaceSubrange<C>(
+        _ subrange: Range<Index>, with newElements: C
+    ) where C : Collection, Element == C.Element {
```
dan-zheng:
```suggestion
    ) where C: Collection, Element == C.Element {
```

Sources/TensorFlow/Core/TensorShape.swift
awav: done

Tests/TensorFlowTests/OperatorTests/LinearAlgebraTests.swift
```diff
@@ -15,6 +15,7 @@
 import XCTest
 @testable import TensorFlow
 
+
```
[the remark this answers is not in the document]
dan-zheng: Done.

PR #616, opened 2020-01-12
TensorFlow: additional cases of float explicit math ops

compnerd:
Use the floating point variant of the math operations.  This is needed
to build on Windows.  It additionally reduces the unnecessary type
conversions necessary to convert single to double precision values and
vice-versa.


compnerd:
CC: @dan-zheng @marcrasi

dan-zheng:
I wonder if `ucrt` on Windows defines only `sqrtf`, not `sqrt`?

compnerd:
The implementation is not type generic, sqrt is for double only.  Overloads are available for C++.  

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/sqrt-sqrtf-sqrtl?view=vs-2019

PR #620, opened 2020-01-13
docs: clarify the CMake dependency version information

compnerd:
The documented instructions rely on CMake 3.16 for the support for
specifying the build tree.  The older releases did not emit the file
output map in the correct location with the option specification.


compnerd:
CC: @dan-zheng

Issue #619, opened 2020-01-13
Regenerate RawOpsGenerated.swift


dan-zheng:
`Sources/TensorFlow/Bindings/RawOpsGenerated.swift` needs regeneration based on [`tensorflow/tensorflow` `v2.1.0-rc1`](https://github.com/tensorflow/tensorflow/tree/v2.1.0-rc1). Instructions are at https://github.com/tensorflow/swift-apis/tree/master/Sources/TensorFlow/Bindings.

---

Note: the version of `tensorflow` used should always match the version specified in [`utils/update_check/update-checkout-config.json`](https://github.com/apple/swift/blob/0478407e87b01aa34a3c4f809067e97101322a17/utils/update_checkout/update-checkout-config.json#L299). It's currently `v2.1.0-rc1`.

Shashi456:
@dan-zheng Been trying to generate the files, i'm getting this error : 
```python
Cannot load api def for Dequantize: Attribute dtype not defined in base api for Dequantize
Cannot load api def for QuantizeAndDequantizeV2: Attribute axis not defined in base api for QuantizeAndDequantizeV2
Cannot generate code for AccumulatorApplyGradient: has ref-valued input
Cannot generate code for AccumulatorNumAccumulated: has ref-valued input
Cannot generate code for AccumulatorSetGlobalStep: has ref-valued input
Cannot generate code for AccumulatorTakeGradient: has ref-valued input
Cannot generate code for ApplyAdaMax: has ref-valued input
Cannot generate code for ApplyAdadelta: has ref-valued input
Cannot generate code for ApplyAdagrad: has ref-valued input
Cannot generate code for ApplyAdagradDA: has ref-valued input
Cannot generate code for ApplyAdagradV2: has ref-valued input
Cannot generate code for ApplyAdam: has ref-valued input
Cannot generate code for ApplyAddSign: has ref-valued input
Cannot generate code for ApplyCenteredRMSProp: has ref-valued input
Cannot generate code for ApplyFtrl: has ref-valued input
Cannot generate code for ApplyFtrlV2: has ref-valued input
Cannot generate code for ApplyGradientDescent: has ref-valued input
Cannot generate code for ApplyMomentum: has ref-valued input
Cannot generate code for ApplyPowerSign: has ref-valued input
Cannot generate code for ApplyProximalAdagrad: has ref-valued input
Cannot generate code for ApplyProximalGradientDescent: has ref-valued input
[...]

PR #621, opened 2020-01-14
Generate Raw ops from Tensorflow2.1.0rc1

Shashi456:
Fixes: #619
Execute has to be overloaded up to 14 variables due to the existence of `parseSequenceExampleV2`.

Updated `generatewrappers.swift` TensorFlow import command. 

High-arity return statement functions in `RawOpsGenerated.swift` have to be broken down by using let variables.

The accuracy of svd had to be changed for the tests to pass. 

Build and test passes.


Sources/TensorFlow/Bindings/EagerExecution.swift
```diff
@@ -563,4 +563,243 @@ internal struct TFE_Op: TFTensorOperation {
         return result
     }
 
+    @inlinable @inline(__always)
```
dan-zheng: Please update `EagerExecution.swift.gyb` to generate up to arity 14 `execute` methods. Change [this line](https://github.com/tensorflow/swift-apis/blob/0f8bd19e812ee26f3d7cb68536172656b818138b/Sources/TensorFlow/Bindings/EagerExecution.swift.gyb#L311) to:
```
%for n in range(1, 14 + 1):
```

Tests/TensorFlowTests/OperatorTests/LinearAlgebraTests.swift
```diff
@@ -74,7 +74,7 @@ final class LinearAlgebraTests: XCTestCase {
             }
             let aReconstituted = matmul(u!, matmul(s.diagonal(), 
                                         transposed: false, v!, transposed: true))
-            assertEqual(aReconstituted, a, accuracy: 1e-5)
+            assertEqual(aReconstituted, a, accuracy: 1e-3)
```
dan-zheng: > The accuracy of svd had to be changed for the tests to pass.

What error happens if you don't lower the accuracy? 1e-3 is a low accuracy value, which seems concerning - perhaps there's an implementation bug (mismatch between Swift and Python TensorFlow).

Shashi456: @dan-zheng /mnt/g/code/swift-apis/Tests/TensorFlowTests/OperatorTests/LinearAlgebraTests.swift:77: error: LinearAlgebraTests.testSVD : XCTAssertEqual failed: ("2.3218951") is not equal to ("2.3218844") +/- ("1e-05") -
/mnt/g/code/swift-apis/Tests/TensorFlowTests/OperatorTests/LinearAlgebraTests.swift:77: error: LinearAlgebraTests.testSVD : XCTAssertEqual failed: ("-3.1908154") is not equal to ("-3.1908028") +/- ("1e-05") -
/mnt/g/code/swift-apis/Tests/TensorFlowTests/OperatorTests/LinearAlgebraTests.swift:89: error: LinearAlgebraTests.testSVD : XCTAssertEqual failed: ("2.3218951") is not equal to ("2.3218844") +/- ("1e-05") -
/mnt/g/code/swift-apis/Tests/TensorFlowTests/OperatorTests/LinearAlgebraTests.swift:89: error: LinearAlgebraTests.testSVD : XCTAssertEqual failed: ("-3.1908154") is not equal to ("-3.1908028") +/- ("1e-05") -

The values seem very close to me. maybe i overdid it by decreasing it to 1e-3

dan-zheng: Thanks for sharing! Always using the tightest accuracy possible (`3e-5`) sounds great!

Shashi456:
@dan-zheng is build time on kokoro really on the order of 30 minutes and more?

dan-zheng:
> @dan-zheng is build time on kokoro really on the order of 30 minutes and more?

Could be. CI tests more than just [tensorflow/swift-apis](https://github.com/tensorflow/swift-apis), also [tensorflow/swift-models](https://github.com/tensorflow/swift-models) and other related repositories to ensure they're not broken.

Shashi456:
@dan-zheng Changed the error bound and checked locally, the tests pass. i think this can be merged now

dan-zheng:
> We should probably also change the commit hash in the tensorflow branch, since this is an api update?

Yep, feel free to submit a PR to [apple/swift](https://github.com/apple/swift/tree/tensorflow). Here's an example: https://github.com/apple/swift/pull/29079

PR #622, opened 2020-01-15
TensorFlow: abstract Thread interfaces

compnerd:
This rewrites the `Thread(perform:)` to be more amenable to supporting
alternate threading types (e.g. Windows).  This reduces the conditionals
and adds some vertical whitespace to make it easier to read at a glance.

### DIFF
--- a/Sources/TensorFlow/Core/Threading.swift
+++ b/Sources/TensorFlow/Core/Threading.swift
@@ -22,43 +22,83 @@ import Glibc
 
 // A thread that runs the provided body.
 class Thread {
-    var thread: pthread_t
-    init(perform body: @escaping () -> ()) {
-        class ThreadArg {
-        var body : () -> ()
-            init(body : @escaping () -> ()) {
-                self.body = body
-            }
-        }
-#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @dan-zheng @pschuh

Sources/TensorFlow/Core/Threading.swift
```diff
@@ -22,43 +22,83 @@ import Glibc
...
+    typealias Handle = HANDLE
+#else
+#if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
+    typealias Handle = pthread_t?
+#else
+    typealias Handle = pthread_t
+#endif
+#endif
+
+    static func _invalid_handle() -> Handle {
```
dan-zheng: Camelcase please: `static func _invalidHandle() -> Handle { ... }`.

Sources/TensorFlow/Core/Threading.swift
```diff
@@ -22,43 +22,83 @@ import Glibc
...
+#if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
+      return nil
+#else
+      return pthread_t()
+#endif
 #endif
-        let threadBody: ThreadBody = { arg in
+    }
+
+    class Procedure {
```
dan-zheng: @compnerd: why is a `Procedure` abstraction necessary? Is `typealias Procedure = () -> ()` sufficient instead?

PR #623, opened 2020-01-15
TensorFlow: abstract TLS handling

compnerd:
This abstracts out the TLS handling behind an interface to allow us to
change the implementation on different platforms.  In particular, it
allows handling of TLS through fibres on Windows.

### DIFF
--- a/Sources/TensorFlow/Core/Threading.swift
+++ b/Sources/TensorFlow/Core/Threading.swift
@@ -16,9 +16,51 @@
 import Darwin
 #elseif os(Windows)
 import ucrt
+import WinSDK
 #else
 import Glibc
 #endif
+
+struct ThreadLocalStorage {
+    struct Key {
+#if os(Windows)
+        var _value: DWORD
+#else
+        var _value: pthread_key_t
+#endif
+
+#if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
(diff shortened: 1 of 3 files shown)


compnerd:
CC: @pschuh @dan-zheng

Sources/TensorFlow/Core/Threading.swift
```diff
@@ -16,10 +16,51 @@
...
+    typealias Key = pthread_key_t
+#endif
+
+#if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
+    typealias KeyDestructor = @convention(c) (UnsafeMutableRawPointer) -> Void
+#else
+    typealias KeyDestructor = @convention(c) (UnsafeMutableRawPointer?) -> Void
+#endif
+
+    public static func KeyCreate(_ destructor: KeyDestructor?) -> Key {
```
dan-zheng: [Swift API design guidelines](https://swift.org/documentation/api-design-guidelines/#begin-factory-name-with-make):
> Begin names of factory methods with "make", e.g. `x.makeIterator()`.

Please also add the `destructor:` argument label. Callers passing trailing closures should be unaffected.

```suggestion
    public static func makeKey(destructor: KeyDestructor?) -> Key {
```

compnerd: I think that I would rather make `Key` a proper abstraction.  I was thinking of doing that in a follow up anyway.

PR #624, opened 2020-01-15
TensorFlow: abstract Mutex interfaces

compnerd:
Abstract out the mutex handling APIs to enable support for Windows'
threading primitives.  This enables mutexes on Windows via the SRWLOCK
(slim reader/writer locks).  The full set of the threading interfaces
is now fully isolated into Core/Threading.swift with cross-platform
implementations.

### DIFF
--- a/Sources/TensorFlow/Core/Runtime.swift
+++ b/Sources/TensorFlow/Core/Runtime.swift
@@ -185,7 +185,7 @@ public final class _ExecutionContext {
     @usableFromInline let status: CTFStatus = TF_NewStatus()
 
     /// The mutex for preventing potential concurrent access.
-    private var mutex: pthread_mutex_t = pthread_mutex_t()
+    private var mutex: Mutex = Mutex()
 
     /// Initializes a new execution context by initializing available devices.
     @usableFromInline
@@ -294,8 +294,6 @@ public final class _ExecutionContext {
             debugLog("Device \(deviceId) has type \(deviceType) and name \(deviceName).")
             deviceNames.append(deviceName)
         }
-
-        pthread_mutex_init(&mutex, nil)
     }
 
     deinit {
@@ -304,7 +302,6 @@ public final class _ExecutionContext {
         TFE_DeleteContext(eagerContext)
         TF_DeleteBuffer(tensorFlowConfig)
         TF_DeleteStatus(status)
-        pthread_mutex_destroy(&mutex)
     }
 }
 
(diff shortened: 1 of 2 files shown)


compnerd:
CC: @dan-zheng @pschuh

Sources/TensorFlow/Core/Threading.swift
```diff
@@ -146,6 +146,51 @@ class Thread {
     }
 }
 
+class Mutex {
```
dan-zheng: Could you please add a simple documentation comment? Even though this is an internal API.

Sources/TensorFlow/Core/Runtime.swift
```diff
@@ -404,13 +401,15 @@ internal extension _ExecutionContext {
...
-        let lockStatus = pthread_mutex_lock(&mutex)
+        let lockStatus = mutex.acquire()
         internalConsistencyCheck(lockStatus == 0)
         defer {
-            let unlockStatus = pthread_mutex_unlock(&mutex)
+            let unlockStatus = mutex.release()
             internalConsistencyCheck(unlockStatus == 0)
+#if !os(Windows)
             // Create a cancellation point.
             pthread_testcancel()
```
dan-zheng: Ignorant question: does it not make sense to move `pthread_testcancel()` to `Mutex.release`? It seems unfortunate that callers of `Mutex.release` still need an extra `#if !os(Windows)` directive with `pthread_testcancel()`.

compnerd: Im entirely happy to do that.  The reasoning behind not doing so immediately is that should every release be associated with a cancellation point?

Sources/TensorFlow/Core/Threading.swift
compnerd: Sure!

Sources/TensorFlow/Core/Runtime.swift
dan-zheng: > The reasoning behind not doing so immediately is that should every release be associated with a cancellation point?

I'm clueless. Maybe @pschuh has an opinion?

compnerd: Okay, did a bit of code archeology with @pschuh ... I think that we came to the conclusion that we do not have a strong reason to keep this functionality at this time.  I can do a follow up change to remove the cancellation point support.

PR #627, opened 2020-01-15
build: adjust flags for inlining and partial specializations

compnerd:
This replicates some of the flags from the tensorflow branch of
apple/swift into the repository.  These control optimizations that we
would like to always enable.


compnerd:
CC: @dan-zheng

compnerd:
Do we want to force these flags always?  Is there a JIRA issue that we should reference as to why we are enabling these forcefully?

PR #631, opened 2020-01-15
TensorFlow: remove cancellation point handling

compnerd:
The cancellation point handling is of questionable value.  Furthermore,
the concept is not portable to all platforms (e.g. Windows).  Until we
have a clear reason and proper abstraction for cancellation points,
remove them to make the paths more uniform.


compnerd:
It is a part of the POSIX standard.  See Section 2.9.5 (Thread Cancellation): https://pubs.opengroup.org/onlinepubs/009695399/functions/xsh_chap02_09.html.

PR #632, opened 2020-01-16
TensorFlow: provide `setenv` on Windows

compnerd:
Windows does not have `setenv` but does provide `_putenv_s` for certain
environments.  This can allow us to emulate the `setenv` as this just
does not allow optional replacement.  Since this is used for IPC with
TensorFlow, providing a racy version is not a big deal.

### DIFF
--- a/Sources/TensorFlow/Core/Runtime.swift
+++ b/Sources/TensorFlow/Core/Runtime.swift
@@ -39,6 +39,17 @@ import Glibc
 
 import CTensorFlow
 
+#if os(Windows)
+// NOTE(abdulras) although the function is racy, we do not really care as the
+// usage here is to not override the value if the user specified one before
+// creating the process.
+@discardableResult
+func setenv(_ variable: String, _ value: String, _ `override`: Int) -> Int {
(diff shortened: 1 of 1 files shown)


Sources/TensorFlow/Core/Runtime.swift
```diff
@@ -39,6 +39,17 @@ import Glibc
 
 import CTensorFlow
 
+#if os(Windows)
+// NOTE(abdulras) although the function is racy, we do not really care as the
```
dan-zheng: Minor comment: is this note something that you plan to take action on sometime? If not, I'd recommend removing the `abdulras` reference (especially since it's a `NOTE` and not a `TODO`).
```suggestion
// NOTE: although the function is racy, we do not really care as the
```

compnerd: I don't think that there is anything that is actionable there (at least for the short term).  I think that moving away from the environment variables and into proper context to the function calls is better, but that would be a longer term thing and need to work out the APIs with the other side.

PR #635, opened 2020-01-16
build: add a `USE_BUNDLED_CTENSORFLOW` option

compnerd:
This allows us to build with a toolchain which has the `CTensorFlow`
module bundled with it as well as with a toolchain which does not.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,9 @@ list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/modules)
 
 set(CMAKE_Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift)
 
+option(USE_BUNDLED_CTENSORFLOW
+  "Use the CTensorFlow module bundled in the active Swift toolchain" OFF)
+
 find_package(TensorFlow REQUIRED)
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -6,6 +6,9 @@ list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/modules)
 
 set(CMAKE_Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift)
 
+option(USE_BUNDLED_CTENSORFLOW
+  "Use the CTensorFlow module bundled in the toolchain" OFF)
```
dan-zheng: Sorry, I'm reading this out of context and the meaning isn't fully clear to me. What is "the toolchain", in `Use the CTensorFlow module bundled in the toolchain`? Perhaps the wording can be edited to improve clarity.

compnerd: Happy to reword it if you have ideas.  The toolchain is the toolchain image that you build from apple/swift's tensorflow branch.

dan-zheng: We chatted offline. How about:
```suggestion
  "Use the CTensorFlow module bundled in the active Swift toolchain" OFF)
```

compnerd: Updated after offline conversation

PR #646, opened 2020-01-26
Add the swish function

Shashi456:
Can someone check the vjp for the function?

Inspired from [here](https://github.com/tensorflow/tensorflow/blob/7d9144dc5f7725f9b4fbf005234d95fa7ded4baa/tensorflow/python/ops/nn_impl.py#L521).

### DIFF
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -1341,6 +1341,10 @@ public func swish<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
     x * sigmoid(x)
 }
 
+/// Note: A custom vjp function for swish is required to avoid excessive 
+/// tensor memory consumption due to storing both `x` and `sigmoid(x)` for 
+/// backprop. This vjp recomputes `sigmoid(x)` during backprop, so that 
+/// the `sigmoid(x)` expression can be freed during the forward pass.
 @inlinable
 @derivative(of: swish)
 func _vjpSwish<T: TensorFlowFloatingPoint>(
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -1347,7 +1347,8 @@ func _vjpSwish<T: TensorFlowFloatingPoint>(
     _ x: Tensor<T>
 ) -> (value: Tensor<T>, pullback: (Tensor<T>) -> Tensor<T>) {
     return (swish(x), { v in 
-        let grad = sigmoid(x) * (1.0 + x * (1 - sigmoid(x)))
+        let sigmoidFeatures = sigmoid(x)
+        let grad = sigmoidFeatures * (1.0 + x * (1 - sigmoidFeatures))
         return grad * v
       })
 }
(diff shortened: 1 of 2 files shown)


Tests/TensorFlowTests/OperatorTests/MathTests.swift
```diff
@@ -250,6 +250,13 @@ final class MathOperatorTests: XCTestCase {
         assertEqual(y, expectedY, accuracy: 1e-5)
     }
 
+    func testSwish() {
```
awav: Looks good! Could you test the gradient as well? Thanks!

Shashi456: I have another PR testing gradients for a lot of these functions, I'll have that up in a few days.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1330,6 +1330,28 @@ func _vjpSelu<T: TensorFlowFloatingPoint>(
...
+    x * sigmoid(x)
+}
+
+@inlinable
+@derivative(of: swish)
+func _vjpSwish<T: TensorFlowFloatingPoint>(
+    _ x: Tensor<T>
+) -> (value: Tensor<T>, pullback: (Tensor<T>) -> Tensor<T>) {
+    return (swish(x), { v in 
+        let grad = sigmoid(x) * (1.0 + x * (1 - sigmoid(x)))
```
awav:
```suggestion
        let sigmoid_features = sigmoid(x)
        let grad = sigmoid_features * (1.0 + x * (1 - sigmoid_features))
```

awav: I think it does make sense to compute it explicitly once when x tensor is large.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1330,6 +1330,28 @@ func _vjpSelu<T: TensorFlowFloatingPoint>(
...
+/// https://arxiv.org/abs/1710.05941
+@inlinable
+@differentiable
+public func swish<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
+    x * sigmoid(x)
+}
+
+@inlinable
+@derivative(of: swish)
+func _vjpSwish<T: TensorFlowFloatingPoint>(
```
marcrasi: We shouldn't need a custom vjp at all here. The compiler should be able to synthesize this automatically! Can you remove it?

Tests/TensorFlowTests/OperatorTests/MathTests.swift
marcrasi: Once we have these tests, they'll verify that the compiler indeed does compute a correct vjp for the function.

Sources/TensorFlow/Operators/Math.swift
marcrasi: Ah, I see a justification for the custom vjp at [the link you posted](https://github.com/tensorflow/tensorflow/blob/7d9144dc5f7725f9b4fbf005234d95fa7ded4baa/tensorflow/python/ops/nn_impl.py#L521).

That's a good justification, and this custom vjp does do the same thing.

You should add a comment explaining why there is a custom vjp.

PR #648, opened 2020-01-28
Update README with Sequential API

rickwierenga:
This also shows commits of [an earlier PR](https://github.com/tensorflow/swift-apis/pull/560) for some reason.


Shashi456:
IMO, I think showcasing the common way of declaring models is still the better choice. `Sequential` exists for people who want to use it, but we still mostly declare models by extending the layer protocol generally.

pschuh:
The older PR's changes are due to the previous PR being squashed and merged. You can do a git rebase in order to make your history clean.

saeta:
Hi @rickwierenga ! First of all, thank you so much for this PR. Documentation is super important and I really appreciate you putting in this effort.

I would like our "front-door" README.md to document the "canonical" way to define layers, as this is where people start in their introduction. Because `Sequential` is fancy / magical syntactic sugar, it might give people the wrong starting point. (Sequential cannot be the canonical way to write layers, because it does not support skip connections, and other sorts of advanced structure.) That said, it'd be great to document `Sequential` better, and it looks like there's [room for improvement](https://github.com/tensorflow/swift-apis/blob/fb339cb865f1ecfc13f855689e61739756ff9c0d/Sources/TensorFlow/Layers/Sequential.swift#L15). Do you think you could instead turn this change into an example added onto the `Sequential` type? (Note: you'll want to [add it to the .gyb file](https://github.com/tensorflow/swift-apis/blob/master/Sources/TensorFlow/Layers/Sequential.swift.gyb) so your improvements won't be lost.)

To simplify history, I'm going to close this PR, and look forward to your new PR! (Feel free to reference this PR in your new PR.)

All the best,
-Brennan

PR #653, opened 2020-01-29
Add documentation to Sequential

rickwierenga:
See https://github.com/tensorflow/swift-apis/pull/648. The autoencoder example is copied from [here](https://github.com/tensorflow/swift-models/blob/master/Autoencoder/main.swift). I trust that is OK with you since the source is also a TensorFlow library.

### DIFF
--- a/Sources/TensorFlow/Layers/Sequential.swift
+++ b/Sources/TensorFlow/Layers/Sequential.swift
@@ -12,7 +12,37 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-/// A layer that sequentially composes two other layers.
+/// A layer that sequentially composes two or more other layers.
+///
+/// ### Usage Example: ###
+///
+/// - Build a simple 2-layer perceptron model for MNIST:
+///
+/// ````
+/// let inputSize = 28 * 28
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/Layers/Sequential.swift
```diff
@@ -12,7 +12,37 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-/// A layer that sequentially composes two other layers.
+/// A layer that sequentially composes two or more other layers.
+///
+/// ### Usage Example: ###
```
8bitmp3: For consistency, I'd recommend using `Example` instead of `Usage Example` to be more in line with non-Swift API docs. For instance: https://github.com/tensorflow/tensorflow/blob/r2.1/tensorflow/python/keras/engine/sequential.py#L43-L391. This is to do with the fact that S4TF and TF share the same search on the main .org site.

8bitmp3: Or maybe in your case you should use `Examples` (plural) 😊

rickwierenga: Good suggestions, thanks!

PR #654, opened 2020-01-29
Fix argument documentation

sgugger:
This decay is not weight decay, so should be renamed to not be confusing.

### DIFF
--- a/Sources/TensorFlow/Optimizers/SGD.swift
+++ b/Sources/TensorFlow/Optimizers/SGD.swift
@@ -43,7 +43,7 @@ public class SGD<Model: Differentiable>: Optimizer
     ) {
         precondition(learningRate >= 0, "Learning rate must be non-negative")
         precondition(momentum >= 0, "Momentum must be non-negative")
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/Optimizers/SGD.swift
```diff
@@ -25,7 +25,7 @@ public class SGD<Model: Differentiable>: Optimizer
     /// The momentum factor. It accelerates stochastic gradient descent in the relevant direction
     /// and dampens oscillations.
     public var momentum: Float
-    /// The weight decay.
+    /// The learning rate decay.
```
dan-zheng: In `Sources/TensorFlow/Optimizers/MomentumBased.swift`, there's another occurrence of `/// The weight decay.` that should be renamed to `/// The learning rate decay.`, if you'd like to update that as well!

sgugger: Changed this one as well.

sgugger:
Also changed the precondition failure messages.

PR #655, opened 2020-01-29
Core: support `PythonKit` as an alternative to `Python`

compnerd:
On master, `PythonKit` is an optional package to build.  This option
will allow building `TensorFlow` on master with `PythonKit`.

### DIFF
--- a/Sources/TensorFlow/Core/PythonConversion.swift
+++ b/Sources/TensorFlow/Core/PythonConversion.swift
@@ -12,8 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#if canImport(Python)
+#if canImport(Python) || canImport(PythonKit)
+
+#if canImport(PythonKit)
+import PythonKit
+#else
 import Python
+#endif
+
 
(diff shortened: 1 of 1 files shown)


Sources/TensorFlow/Core/PythonConversion.swift
```diff
@@ -12,13 +12,23 @@
...
+#else
 import Python
+#endif
+
 
 /// The `numpy` Python module.
 /// Note: Global variables are lazy, so the following declaration won't produce
 // a Python import error until it is first used.
+#if canImport(PythonKit)
+private let np = PythonKit.import("numpy")
```
dan-zheng: The `Python` in `Python.import` here refers to [a global variable](https://github.com/pvieito/PythonKit/blob/c699071262161e2f2d9fa081d9cc8ef31ca4ec5f/PythonKit/Python.swift#L640), not a module name.

```suggestion
private let np = Python.import("numpy")
```

Same below: `PythonKit.isinstance` -> `Python.isinstance`.

I believe the only necessary change in this patch is the top-level ifdef'd import declaration!

PR #659, opened 2020-02-02
Bug fix in 'Tensor.batchGathering'

eaplatanios:
`Tensor.batchGathering` was not properly handling negative axis arguments.


Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -478,22 +478,22 @@ public extension Tensor {
                 "'batchDimensionCount' must be less than the tensor's rank.")
         }
 
+        // Adjust axis to be positive.
+        let axis = axis < 0 ? axis + rank : axis
```
Shashi456: Just a question, wouldn't adding the precondition solve this issue?

eaplatanios: A precondition would completely disallow negative axis arguments. However, these are supported throughout the Python API and most of the S4TF API and so it’s probably better consistently support them everywhere as many users are accustomed to them.

t-ae: Related PR: #641 
Converting negative axis to positive axis is common operation.
Can we have that as a method?

Shashi456: @eaplatanios, if that is the case, removing the to-do for precondition makes sense right, atleast the `axis<0` part?

eaplatanios: You're right. I'll remove that. Thanks for noticing that. Actually, that TODO was there because autodiff was having trouble with the precondition in the past but now we can remove it altogether as you suggested.

eaplatanios: @t-ae we could also it is a simple one-liner so I'm not sure if that would help.

t-ae: @eaplatanios 
My opinion:

1. It appears 6 times
1 in this PR. 4 in my PR. 1 I mentioned in my PR.
I think 6 is considerable number. And there can be more.

2. Subtle difference in implementation
In this PR: `axis < 0 ? axis + rank : axis`
In my PR: `($0 + self.rank) % self.rank`
These are same because of `precondition` range check.
Future contributers won't be confused if there's standard way defined.

3. Assertion
We can have axis range check in that method.

eaplatanios: @t-ae I see your point. I can look into adding that in a separate PR, but let's first get this one reviewed and merged.

t-ae:
@asuhan 
I think ternary operator is more desirable.
Even if the axis is not less than `rank` (so it is not in range `[-rank, rank)`), modulo operator wraps it into` [0, rank)`.

PR #660, opened 2020-02-03
Fix derivative of `RNN.callAsFunction(_:initialState:)`.

dan-zheng:
Previously, `RNN._vjpcallAsFunction(_:initialState:)` incorrectly used a zero
initial state. Now, it uses `initialState` as the initial state.

Add `RNN` gradient tests for `SimpleRNNCell`, `LSTMCell`, and `GRUCell`.
Todo: verify that gradients are correct using a reference implementation.

### DIFF
--- a/Tests/TensorFlowTests/LayerTests.swift
+++ b/Tests/TensorFlowTests/LayerTests.swift
@@ -1139,10 +1139,9 @@ final class LayerTests: XCTestCase {
     func testRNN() {
         let x = Tensor<Float>(rangeFrom: 0.0, to: 0.4, stride: 0.1).rankLifted()
         let inputs: [Tensor<Float>] = Array(repeating: x, count: 4)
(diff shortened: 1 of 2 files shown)


Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -1190,22 +1189,62 @@ final class LayerTests: XCTestCase {
                      [ 0.074910110, 0.021107012, -0.049724963, -0.069670826],
                      [ 0.078670055, 0.022462710, -0.051899005, -0.075331904]],
                     accuracy: 1e-6)
+                let (𝛁lstm, _) = pullback(.init(inputs.map { LSTMCell<Float>.State(cell: $0, hidden: $0) }))
+                // TODO: Verify that LSTM gradients are correct using a reference implementation.
```
marcrasi: I'm interested in doing this. Should we do something like check some TF python code into this repository that reproduces the weights?

dan-zheng: That sounds good!

PR #663, opened 2020-02-06
add initialState wrt parameter for RNN, and tests

marcrasi:
Adds an `initialState` wrt parameter for RNN derivatives, so that we can train "decoder" models where the model creates an initial state for a "decoder RNN" that produces the final output sequence.

Updates the LSTM test to test gradients wrt this new parameter.

Also changes the LSTM test to be based on data from a reference keras implementation.


marcrasi:
It would be cool if there was some utility I could use to save these weights to a file and then load them up with fewer lines of code. I wonder if the model serialization stuff that @BradLarson is working on would help. I won't block this PR on it though. I or someone can refactor it later.

Issue #596, opened 2019-12-24
Make conv2DBackpropInput public


t-ae:
`Conv*D` layers have corresponding public functions `conv*D`.
With them we can easily create customized convolutional layers.

On the other hand, `TansposedConv*D` is implemented with internal function `conv2DBackpropInput`.
It makes creating customized transposed convolutional layers hard.

I propose to make it public, or add new `transposedConv2D` function as alias.

dan-zheng:
Sorry for the late reply. I think renaming `conv2DBackpropInput` and making it a public API is reasonable!

---

[`func matmul`](https://github.com/tensorflow/swift-apis/blob/c7afd2dc7c55735e8449256e6e3efc11f6dc5576/Sources/TensorFlow/Operators/Math.swift#L2774) has `transposed: Bool` arguments for transposing `lhs` or `rhs`:
```swift
/// Performs matrix multiplication with another tensor and produces the result.
@inlinable
@differentiable(where Scalar: TensorFlowFloatingPoint)
public func matmul<Scalar: Numeric>(
    _ lhs: Tensor<Scalar>,
    transposed transposeLhs: Bool = false,
    _ rhs: Tensor<Scalar>,
    transposed transposeRhs: Bool = false
) -> Tensor<Scalar> {
    precondition(lhs.rank >= 2, "Input tensors must have at least rank 2")
    precondition(rhs.rank >= 2, "Input tensors must have at least rank 2")
    if lhs.rank > 2 || rhs.rank > 2 {
        // TODO(TF-629): Conjugate to make compatible with the adjoint.
        return _Raw.batchMatMulV2(lhs, rhs, adjX: transposeLhs, adjY: transposeRhs)
    }
    return _Raw.matMul(lhs, rhs, transposeA: transposeLhs, transposeB: transposeRhs)
}
```

Adding a `transposed: Bool = false` argument (or arguments, for both input and filter?) to the existing function `public func conv2D` sounds ideal to me, if possible. @t-ae: would you be interested in looking into this?

Increasing the API surface area with `public func transposedConv2D` is less desirable.

t-ae:
I think `public func transposedConv2D` is worth adding.

1. Both TensorFlow and PyTorch have separate function for transposed conv2D.

TensorFlow:
- https://www.tensorflow.org/api_docs/python/tf/nn/conv2d?version=stable
- https://www.tensorflow.org/api_docs/python/tf/nn/conv2d_transpose?version=stable

PyTorch
[...]

PR #672, opened 2020-02-13
Add transposedConv2D (#596)

t-ae:
`TansposedConv2D` layer is using internal API `conv2DBackpropInput`.
It makes creating custom transposed convolutional layers outside the library hard.

This PR exposes that API with new name `transposedConv2D`.

### DIFF
--- a/Sources/TensorFlow/Operators/NN.swift
+++ b/Sources/TensorFlow/Operators/NN.swift
@@ -169,8 +169,10 @@ public func transposedConv2D<Scalar: TensorFlowFloatingPoint>(
     padding: Padding = .valid,
     dilations: (Int, Int, Int, Int) = (1, 1, 1, 1)
 ) -> Tensor<Scalar> {
-    conv2DBackpropInput(input, shape: shape, filter: filter,
-                        strides: strides, padding: padding, dilations: dilations)
+    precondition(input.shape.rank == 4, "The input must have rank 4.")
+    precondition(filter.shape.rank == 4, "The filter must have rank 4.")
+    return conv2DBackpropInput(input, shape: shape, filter: filter,
+                               strides: strides, padding: padding, dilations: dilations)
 }
 
 /// TensorFlow builtin conv2d gradient helper for the input.
@@ -184,8 +186,6 @@ func conv2DBackpropInput<Scalar: TensorFlowFloatingPoint>(
     padding: Padding = .valid,
     dilations: (Int, Int, Int, Int) = (1, 1, 1, 1)
 ) -> Tensor<Scalar> {
-    precondition(x.shape.rank == 4, "The input must have rank 4.")
-    precondition(filter.shape.rank == 4, "The filter must have rank 4.")
     return _Raw.conv2DBackpropInput(
         inputSizes: shape,
         filter: filter,
--- a/Sources/TensorFlow/Operators/NN.swift
+++ b/Sources/TensorFlow/Operators/NN.swift
@@ -142,8 +142,8 @@ func _vjpConv2D<Scalar: TensorFlowFloatingPoint>(
 ) -> (value: Tensor<Scalar>, pullback: (Tensor<Scalar>) -> (Tensor<Scalar>, Tensor<Scalar>)) {
     let value = conv2D(x, filter: filter, strides: strides, padding: padding, dilations: dilations)
     return (value, { v in
-        (transposedConv2D(v, shape: x.shapeTensor, filter: filter,
-                          strides: strides, padding: padding, dilations: dilations),
+        (conv2DBackpropInput(v, shape: x.shapeTensor, filter: filter,
+                             strides: strides, padding: padding, dilations: dilations),
          conv2DBackpropFilter(v, input: x, filterSizes: filter.shapeTensor,
                               strides: strides, padding: padding, dilations: dilations))
     })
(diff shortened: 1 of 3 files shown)


Sources/TensorFlow/Operators/NN.swift
```diff
@@ -142,46 +142,56 @@ func _vjpConv2D<Scalar: TensorFlowFloatingPoint>(
...
                               strides: strides, padding: padding, dilations: dilations))
     })
 }
 
-/// TensorFlow builtin conv2d gradient helper for the input.
-@differentiable(wrt: (x, filter))
-@usableFromInline
-func conv2DBackpropInput<Scalar: TensorFlowFloatingPoint>(
-    _ x: Tensor<Scalar>,
+/// Returns a 2-D transposed convolution with the specified input, filter, strides, and padding.
```
Shashi456: I'm not really for this renaming of gradient helper for the conv operation. It would be better if a separate wrapper for `transposedConv2D` is written which calls `conv2DBackpropInput`. This current renaming would also confuse people who check the source code about as to why transposed functions are being called for a convolution operation.

Sources/TensorFlow/Operators/NN.swift
```diff
@@ -142,46 +142,56 @@ func _vjpConv2D<Scalar: TensorFlowFloatingPoint>(
...
     padding: Padding = .valid,
     dilations: (Int, Int, Int, Int) = (1, 1, 1, 1)
 ) -> Tensor<Scalar> {
+    precondition(input.shape.rank == 4, "The input must have rank 4.")
+    precondition(filter.shape.rank == 4, "The filter must have rank 4.")
     return _Raw.conv2DBackpropInput(
         inputSizes: shape,
         filter: filter,
-        outBackprop: x,
+        outBackprop: input,
```
Shashi456: Do tests pass after this change? Afaik all convolution operations are checked with their gradients as well.

Sources/TensorFlow/Operators/NN.swift
t-ae: I originally proposed to add alias for it in #596, and @dan-zheng proposed to just rename it.
Adding alias is less confusing for today's contributors, but it won't have much meaning for future contributors, so I decided to take rename way.

Sources/TensorFlow/Operators/NN.swift
t-ae: It's merely a renaming of argument name.
I thought it should have same signature as `conv2D`.
Of course tests are passing.

Sources/TensorFlow/Operators/NN.swift
Shashi456: The Tensorflow `Conv2DTranspose` calls the generated op as a wrapper, [here](https://github.com/tensorflow/tensorflow/blob/964a78f3834a17b44fb8898f3d3be95883fa0007/tensorflow/python/ops/nn_ops.py#L2218). Similarly done for `conv1dTranspose` and `conv3dTranpose`, While the conv2d [calls](https://github.com/tensorflow/tensorflow/blob/964a78f3834a17b44fb8898f3d3be95883fa0007/tensorflow/python/ops/nn_ops.py#L925), `input` and `filter` separately. 
A separation of concerns, with separate functions make sense and make for a cleaner api, than renaming `conv2dBackpropInput` as `transposedConv2D`.

Sources/TensorFlow/Operators/NN.swift
Shashi456: @t-ae oh sorry, overlooked the renaming!

Sources/TensorFlow/Operators/NN.swift
t-ae: And there's separate `conv2d_backprop_input` [here](https://github.com/tensorflow/tensorflow/blob/964a78f3834a17b44fb8898f3d3be95883fa0007/tensorflow/python/ops/nn_ops.py#L2078).
`conv2d_transpose_v2` has extra operations (name_scope and data_format part), so they necessarily should be separated.
In S4TF there's no such operations. I'm not sure yet we should separate them.

Shashi456: The `name_scope` and `data_format` isn't exactly an operation as they are related to input type. We currently only support the NHWC format, I think once we are thoroughly done implementing layers, We should start looking to support the NCWH format.

t-ae:
I restored `conv2DBackpropInput` and added `transposedConv2D`.
Now `TransposedConv*D` layers use `transposedConv2D` and `_vjpConv2D*` funcs use `conv2DBackpropInput`.

> Could you please add some tests for transposedConv2D and its derivative? You can adapt existing tests for conv2D.

It looks there's no tests for `conv2D`. But there are tests for `Conv2D` instead.
Also `TransposedConv2D` has its test, but it didn't have test of its gradients, so I just added it.


By the way I got two questions.

1. `_vjpConv2DBackpropInput` and some other functions are `@usableFromInline`.
Is there any reason to prefer `@usableFromInline` to `@inlinable`?

2. `transposedConv2D` has argument `shape: Tensor<Int32>`. It seems we can change its type to `TensorShape`.
Do you think it's good idea?

Shashi456:
@t-ae all shape variables in the operators file for convolutional operator are `Tensor<Int32>` because the backend generated ops need them in that format, so making it a `TensorShape` is not a good idea.

t-ae:
> all shape variables in the operators file for convolutional operator are Tensor<Int32> 

They are internal but `transposedConv2D` is public.
We'd better decide its signature from the point of view of users.

According to [tf.nn.conv2d_transpose](https://www.tensorflow.org/api_docs/python/tf/nn/conv2d_transpose) documentation,

> output_shape: A 1-D Tensor representing the output shape of the deconvolution op.

`TensorShape` is always 1-D so user can't pass wrong parameter.

I don't know if there's case rank of `shape` is not 4. If there's no such case, it should be rather `(Int, Int, Int, Int)` than `TensorShape`.

t-ae:
@dan-zheng 
The discussion about the type of `shape` is left over.
Do you have any comments?

PR #673, opened 2020-02-13
Improve precondition error messages.

### LINKED ISSUE #642: Duplication of axes in reduction causes several problems
It seems reduction functions are supporting axes with duplication like `[0, 0]` (`_Raw.*` handle it under the hood).
But `_vjp*` don't consider it. So gradient computation can be incorrect or cause crash.

```swift 
import TensorFlow

let tensor = Tensor<Float>(zeros: [3])

print(tensor.mean(alongAxes: 0)) // [0, 0]
print(tensor.mean(alongAxes: 0, 0)) // [0, 0]

print(gradient(at: tensor) { tensor in tensor.mean(alongAxes: 0).scalarized() }) // [0.33333334, 0.33333334, 0.33333334]
// Incorrect gradient
print(gradient(at: tensor) { tensor in tensor.mean(alongAxes: 0, 0).scalarized() }) // [0.11111111, 0.11111111, 0.11111111]


print(tensor.sum(squeezingAxes: 0)) // 0.0
print(tensor.sum(squeezingAxes: 0, 0)) // 0.0

print(gradient(at: tensor) { tensor in tensor.sum(squeezingAxes: 0).scalarized() }) // [1.0, 1.0, 1.0]
// This causes crash
// print(gradient(at: tensor) { tensor in tensor.sum(squeezingAxes: 0, 0).scalarized() })
```

saeta:
This PR improves error messages regarding ranks and shapes for a number of operations on `Tensor`, especially regarding mis-specified axes.


saeta:
@t-ae do you think this will be a good fix for #642 ? Thanks!

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1239,3 +1239,25 @@ extension Tensor {
...
+/// treated as performance sensitive. For a quick benchmark comparison between
+/// a naive implementation and this implementation, check out
+/// https://colab.research.google.com/drive/1xPS4JkU9YwgKFDUNgCvrqquUUe1R0cVS#scrollTo=kZRlD4utdPuX
+/// which shows a ~10x performance difference.
+@usableFromInline
+internal func ensureNonDuplicated<T: BinaryInteger>(_ axes: [T]) -> Bool {
+    guard axes.count < 5 else { return Set(axes).count == axes.count }  // Slow path.
+    for i in 0..<axes.count {
+        for j in (i+1)..<axes.count {
+            guard axes[i] != axes[j] else { return false }
```
t-ae: It seems negative axes are not in consideration.

t-ae:
This patch seems just banning axes duplication.
But TensorFlow is supporting that.
Shouldn't we make it same behavior? (I don't think axes duplication doesn't help, though.)

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1229,7 +1229,7 @@ extension Tensor {
     /// Returns `true` if all given axes are in the range `[-rank, rank)`.
     @usableFromInline
     internal func areAxesInRange<T: BinaryInteger>(_ axes: [T]) -> Bool {
-        return !axes.contains(where: { !isAxisInRange($0) })
+        return !axes.contains(where: { !isAxisInRange($0) }) && ensureNonDuplicated(axes)
```
dabrahams: This line's change doesn't match what the commit message says you're changing, and appears to give the method new semantics that clash with its name.    If this PR is about improving error messages, the changes should just do that.

dabrahams: This line's change appears to give the method new semantics that clash with its name and its doc comment

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -2363,6 +2366,13 @@ internal extension Tensor where Scalar: TensorFlowFloatingPoint {
         return (value, { [shape = shapeTensor] in $0.broadcasted(toShape: shape) })
     }
 
+    @inlinable
+    @derivative(of: sum(alongAxes:))
+    func _vjpSum(alongAxes axes: [Int]) -> (value: Tensor, pullback: (Tensor) -> Tensor) {
+        let value = sum(alongAxes: axes)
+        return (value, { [shape = shapeTensor] in $0.broadcasted(toShape: shape) })
+    }
```
dabrahams: Does this part of the change match the PR description?

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1239,3 +1239,25 @@ extension Tensor {
...
+///
+/// Note: this function is on the critical path of operations, and so should be
+/// treated as performance sensitive. For a quick benchmark comparison between
+/// a naive implementation and this implementation, check out
+/// https://colab.research.google.com/drive/1xPS4JkU9YwgKFDUNgCvrqquUUe1R0cVS#scrollTo=kZRlD4utdPuX
+/// which shows a ~10x performance difference.
+@usableFromInline
+internal func ensureNonDuplicated<T: BinaryInteger>(_ axes: [T]) -> Bool {
+    guard axes.count < 5 else { return Set(axes).count == axes.count }  // Slow path.
+    for i in 0..<axes.count {
```
dabrahams: 'long as we're prematurely optimizing, do you want to go `0..<(axes.count - 1)`?

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1239,3 +1239,25 @@ extension Tensor {
         return areAxesInRange(axes.scalars)
     }
 }
+
+/// Checks to ensure elements within an array are distinct.
+///
```
dabrahams: I think there's a precondition missing here, that the array is nonempty.  IIUC, your implementation will trap if it is violated.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -2882,3 +2892,37 @@ public extension Tensor where Scalar: Numeric {
...
 }
+
+//===------------------------------------------------------------------------------------------===//
+// Precondition helpers.
+//===------------------------------------------------------------------------------------------===//
+
+internal extension Tensor {
+    @usableFromInline
+    func preconditionCheck(
+        axes: Tensor<Int32>,
```
dabrahams: More of a general question than a comment on this change: do we end up with these two overloads (on `Tensor<Int32>/[Int]` of many functions?  Should we be generic over `Collection where Element : BinaryInteger` or something?

dan-zheng:

---

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1618,7 +1618,7 @@ public extension Tensor where Scalar == Bool {
     /// - Precondition: Each value in `axes` must be in the range `-rank..<rank`.
     @inlinable
     func all(squeezingAxes axes: Int...) -> Tensor {
-        precondition(areAxesInRange(axes), "All axes must be in the range `[-rank, rank)`.")
+        preconditionCheck(axes: axes)
```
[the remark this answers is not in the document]
saeta: Done!

Sources/TensorFlow/Operators/Math.swift
saeta: Perhaps, but I'm currently unconvinced. For how, I'll leave this as-is, as this follows the current existing patterns.

Sources/TensorFlow/Operators/Basic.swift
saeta: That's an excellent point. Given that the initial purpose of this PR was to improve the error messages, I've removed this functionality in this PR.

Sources/TensorFlow/Operators/Basic.swift
saeta: hehe, excellent point! :-)

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1239,3 +1239,25 @@ extension Tensor {
...
+/// specified axes are not duplicated.
+///
+/// Note: this function is on the critical path of operations, and so should be
+/// treated as performance sensitive. For a quick benchmark comparison between
+/// a naive implementation and this implementation, check out
+/// https://colab.research.google.com/drive/1xPS4JkU9YwgKFDUNgCvrqquUUe1R0cVS#scrollTo=kZRlD4utdPuX
+/// which shows a ~10x performance difference.
+@usableFromInline
+internal func ensureNonDuplicated<T: BinaryInteger>(_ axes: [T]) -> Bool {
+    guard axes.count < 5 else { return Set(axes).count == axes.count }  // Slow path.
```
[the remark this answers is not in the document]
saeta: Fair point. I've removed this section of code from this PR for now, but I appreciate the comment!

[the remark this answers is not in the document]
saeta: ack.

Sources/TensorFlow/Operators/Basic.swift
saeta: This is a great point. I think that this is important, and deserves its own PR, so I'm removing this functionality for now, and I or someone else can add it in a subsequent patch correctly. Thank you for the review @t-ae!

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1239,3 +1239,25 @@ extension Tensor {
         return areAxesInRange(axes.scalars)
     }
 }
+
+/// Checks to ensure elements within an array are distinct.
```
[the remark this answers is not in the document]
saeta: First of all, thank you very much for the detailed review. I appreciate it. Since this functionality isn't part of the original motivation for this PR, I've removed it, and will send it along (having (hopefully) learned from your comments) in a subsequent PR.

saeta:
Note: I've updated the PR description; previous:

> This change makes 2 changes:
> It improves the error messages for mis-specified axes.
> It ensures the same axis isn't specified multiple times. (Fixes #642.)

Sources/TensorFlow/Operators/Math.swift
dabrahams: Well of course follow the existing patterns until we make a more general decision.

saeta: SGTM; thanks!

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1334,8 +1334,12 @@ extension Tensor {
 
   /// Returns `true` if the given scalar tensor is in the range `[-rank, rank)`.
   @usableFromInline
-  internal func isAxisInRange(_ axis: Tensor<Int32>) -> Bool {
-    precondition(axis.rank == 0, "Axis must have rank 0.")
+  internal func isAxisInRange(
```
dabrahams: I realize you didn't add this name, but the `isSomeThingSomeAdjective` style of predicate naming comes from Cocoa and is not really part of the Swift style (even though many people still do it); `someThingIsSomeAdjective` reads much more naturally.  In this case, `axisIsInRange(a)` or `isInRange(axis: a)` would be more appropriate.  Because `Tensor<Int32>` is weak type information, I might opt for the 2nd version, which is consistent with what you did for `ensureValid(axes:)`

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1334,8 +1334,12 @@ extension Tensor {
 
   /// Returns `true` if the given scalar tensor is in the range `[-rank, rank)`.
```
dabrahams: I realize this isn't part of your change, but a few things about the comment:
- Should “the given scalar tensor” be consider sufficient to imply the precondition?  I think it kinda buries the conditions for using the function in an adjective, so it might be better to have an explicit precondition clause.
- It's more descriptive to say “Returns `true` if `axis` is in…”
- “`[-rank, rank)`” implies that it's valid code. If you want to use that notation, I think you want “[`-rank`, `rank`)” but very frankly I think `-rank..<rank` is more appropriate here (and you can leave out  “the range”)
- It's even more descriptive to say something about what it returns that's less like code and more like its high-level meaning, e.g. something analogous to “refers to an axis of `self`”

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1347,8 +1351,16 @@ extension Tensor {
...
   /// Returns `true` if all scalars of the given 1-D tensor are in the range `[-rank, rank)`.
   @usableFromInline
-  internal func areAxesInRange(_ axes: Tensor<Int32>) -> Bool {
-    precondition(axes.rank == 1, "Axes must have rank 1.")
+  internal func areAxesInRange(
+    _ axes: Tensor<Int32>,
+    file: StaticString = #file,
+    line: UInt = #line
+  ) -> Bool {
+    precondition(
```
dabrahams: I thought you were going to stick to improving error messages with this commit?

The comment still implies the old precondition was correct; you've expanding the method to mean that it works on "scalar tensors," so you need to update the comment to reflect that.  If you do that, consider whether the convenience offered by the expansion is enough to outweigh the complexity added by complicating the semantics.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -2996,3 +2996,37 @@ extension Tensor where Scalar: Numeric {
     matmul(lhs, rhs)
   }
 }
+
+//===------------------------------------------------------------------------------------------===//
+// Precondition helpers.
+//===------------------------------------------------------------------------------------------===//
+
+internal extension Tensor {
+    @usableFromInline
```
dabrahams: Can I get a doc comment please!

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -2996,3 +2996,37 @@ extension Tensor where Scalar: Numeric {
...
+    func ensureValid(
+        axes: Tensor<Int32>,
+        function: StaticString = #function,
+        file: StaticString = #file,
+        line: UInt = #line
+    ) {
+        precondition(
+            areAxesInRange(axes, file: file, line: line),
+            "All axes must be in the range `[-rank, rank)` when calling \(function) (rank is: \(rank), axes: \(axes))",
```
dabrahams: What are the backticks doing in here?

Sources/TensorFlow/Operators/Basic.swift
saeta: Follow on: #756 

FWIW: I'm not sure I appreciate the difference between `isAxisInRange(_:)` vs `isInRange(axis:)`; they seem approximately equivalently awkward, and equivalently comprehensible when encountering them. :-D

dabrahams: There's a difference in comprehensibility: `isAxisInRange(q)` could be asking about whether some unseen axis is in the range `q`, rather than whether some axis `q` is in some unseen range.  

Regarding awkwardness, we don't say "if is (the) dog white, do something"  we say "if (the) dog is white, do something." 

"is dog pilot Snoopy" is more analogous to what's going on with `isAxisInRange(a)`; "axis/dog" are not serving a purpose in this scenario other than to make up for weak type information, so it should get out of the primary (base) name and clearly indicate the entity to which it applies.

Make sense?

PR #674, opened 2020-02-13
Make bias optional in Dense and all Conv layers

sgugger:
This PR makes bias optional in the dense layer and all convolutional layers. This is necessary since most models (ResNets and variants) use convolutionla layers with no bias.

Since AD does not support optionals yet, this is implemented behind the scene using a boolean flag. That flag can be removed once AD deals properly with optionals, and we can keep the same inits.

### DIFF
--- a/Sources/TensorFlow/Layers/Convolutional.swift
+++ b/Sources/TensorFlow/Layers/Convolutional.swift
@@ -49,7 +49,7 @@ public struct Conv1D<Scalar: TensorFlowFloatingPoint>: Layer {
     ///   - dilation: The dilation factor for the temporal dimension.
     public init(
         filter: Tensor<Scalar>,
-        bias: Tensor<Scalar>?,
+        bias: Tensor<Scalar>? = nil,
         activation: @escaping Activation = identity,
         stride: Int = 1,
         padding: Padding = .valid,
@@ -163,7 +163,7 @@ public struct Conv2D<Scalar: TensorFlowFloatingPoint>: Layer {
     ///     (dilation height, dilation width).
     public init(
         filter: Tensor<Scalar>,
-        bias: Tensor<Scalar>?,
+        bias: Tensor<Scalar>? = nil,
         activation: @escaping Activation = identity,
         strides: (Int, Int) = (1, 1),
         padding: Padding = .valid,
@@ -286,7 +286,7 @@ public struct Conv3D<Scalar: TensorFlowFloatingPoint>: Layer {
     ///   - dilations: The dilation factor for spatial/spatio-temporal dimensions.
     public init(
         filter: Tensor<Scalar>,
-        bias: Tensor<Scalar>?,
+        bias: Tensor<Scalar>? = nil,
         activation: @escaping Activation = identity,
         strides: (Int, Int, Int) = (1, 1, 1),
         padding: Padding = .valid,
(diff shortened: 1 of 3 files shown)


Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -137,6 +142,8 @@ public struct Conv2D<Scalar: TensorFlowFloatingPoint>: Layer {
     @noDerivative public let padding: Padding
     /// The dilation factor for spatial dimensions.
     @noDerivative public let dilations: (Int, Int)
+    /// Workaround optionals not being handled by AD
```
dan-zheng: TF-1004 tracks enum differentiation support.
Could you please add references to TF-1004 in comments explaining that `useBias` is a workaround?
```suggestion
    // Note: `useBias` is a workaround for TF-1004: enum differentiation support.
```

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -47,18 +49,19 @@ public struct Conv1D<Scalar: TensorFlowFloatingPoint>: Layer {
     ///   - dilation: The dilation factor for the temporal dimension.
     public init(
         filter: Tensor<Scalar>,
-        bias: Tensor<Scalar>,
+        bias: Tensor<Scalar>?,
```
dan-zheng: How about adding a default argument value  `nil` to all `bias` arguments?
```suggestion
        bias: Tensor<Scalar>? = nil,
```

Sources/TensorFlow/Layers/Convolutional.swift
sgugger: Does enum differentiation imply optional differentiation? (Sorry if this is a dumb question, just want to make sure.)

Sources/TensorFlow/Layers/Convolutional.swift
sgugger: Sure, I can add that on all inits.

Sources/TensorFlow/Layers/Convolutional.swift
dan-zheng: Good point - enum differentiation support doesn't strictly imply `Optional` differentiation support.

I created a dedicated issue TF-1153 to track differentiation support for `Optional` operations, please reference that instead!

sgugger: Done, let me know if the phrasing is okay.

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -104,14 +108,15 @@ public extension Conv1D where Scalar.RawSignificand: FixedWidthInteger {
         padding: Padding = .valid,
         dilation: Int = 1,
         activation: @escaping Activation = identity,
+        useBias: Bool = true,
```
Shashi456: Can a private variable be used in an initializer like this? I mean will we be able to set it to false externally?

sgugger: Here `useBias` is not the same as the internal flag, it's just to know if it has to pass a bias or nil to the main init.

Sources/TensorFlow/Layers/Convolutional.swift
dan-zheng: Wording looks good!

PR #675, opened 2020-02-14
Update the AdaGrad optimizer class API docs 📒

8bitmp3:
This PR updates the AdaGrad optimizer class API docs, following similar work on Adam.

- Expanded the description of AdaGrad ↔ TF 2.x/Keras docs.
- Defined a bunch of parameters from the existing comments.
- Added a small LeNet example adapted from /swift-models/ (tested separately on MNIST).

I'm continuing going through the rest of the commonly used APIs. More PRs on their way.

Thanks team.


8bitmp3:
(hope you don't mind the commit's name - a late night copy/paste error - or else I can undo and redo with a proper name)

Sources/TensorFlow/Optimizers/MomentumBased.swift
```diff
@@ -65,11 +65,63 @@ public class RMSProp<Model: Differentiable>: Optimizer
...
+/// var leNetClassifier = Sequential {
+///     Conv2D<Float>(filterShape: (5, 5, 1, 6), padding: .same, activation: relu)
+///     AvgPool2D<Float>(poolSize: (2, 2), strides: (2, 2))
+///     ...
+///     Dense<Float>(inputSize: 84), outputSize: 10)
+/// }
+/// // Define the AdaGrad optimizer with a learning rate set to 1e-3, rho - to 0.99,
+/// // and epsilon: 1e-8
+/// let adaGradOptimizer = AdaGrad(for: leNetClassifier, learningRate: 1e-3, 
+///                                rho: 0.99, epsilon: 1e-8)
```
saeta: Tiny style nit: we might want to have each argument (and label) be on their own line:

e.g.:
```swift
/// let adaGradOptimizer = AdaGrad(
///     for: leNetClassifier,
///     learningRate: 1e-3, 
///     rho: 0.99,
///     epsilon: 1e-8)
```

(For context, the "rule" is spelled out in https://google.github.io/swift/#line-wrapping rule no. 2 that starts "Comma-delimited lists are only laid out in one direction: horizontally or vertically.")

8bitmp3: Good point @saeta. Old Python habits? Bookmarking that style guide.

8bitmp3: Meanwhile, @saeta if you can find out why there's a fancy rho param in this AdaGrad implementation that would be awesome. I went through various literature/research papers but couldn't find anything.

8bitmp3:
the 2nd commit dealing with line-wrapping was probably not necessary @saeta but it removed the `kokoro:run`

PR #679, opened 2020-02-14
Update the RMSProp optimizer class API docs

8bitmp3:
This PR updates the RMSProp optimizer class API docs. :

- Added a new description: sourced from Alex Graves's 2013 [paper](https://arxiv.org/abs/1308.0850) and Hinton's [lecture](http://www.cs.toronto.edu/~tijmen/csc321/slides/lecture_slides_lec6.pdf) from 2012. The latter is said to be the first public mention of the optimizer. The description is pretty much word for word, hope that's ok. Feel free to modify it.
- Added a new parameter section with the default values from the class initializer.
  - Added a definition for the `rho` param. It was taken from Lasagne lib API [docs](https://lasagne.readthedocs.io/en/latest/modules/updates.html#lasagne.updates.rmsprop). (Lasagne was built for Theano.) Another description was used from TF 2.x API docs.
  - Updated the TODO code comment for `rho` (it's a 0.9 constant (Hinton) and may be referred by other Greeks in other literature).
  - Since the default params are in a separate params section now, I removed a sentence in the old description, which recommended which default values to use.
- Modified the reference using Graves' reference style from the 2013 [paper](https://arxiv.org/pdf/1308.0850.pdf) that followed the first "unpublished" introduction of RMSProp (or _rmsprop_) ("T. Tieleman and G. Hinton. Lecture 6.5 - rmsprop:..." in 2012).
- Added an example code snippet - an image classifier on Cifar-10 adapted from /swift-models/ - with comments (tested in a notebook separately to make sure it works.)

Continuing doing more work on API docs.

TGIF.


dan-zheng:

---

PR #680, opened 2020-02-15
Math protocols

awav:
The `swift-apis` defines a shaped array protocol and has implementations for many math operations. However, with all power that Swift has, it is still not that easy to extend math operations with user-defined shaped structures, advanced arrays such as sparse, banded and symmetric matrices, maybe even NumPy arrays. To address the extendability issue, we need to come up with standardised math protocols, as well as default gradient implementations. In many cases, the user could even leave the default gradient unchanged for custom shaped structures. 

_By this PR, I want to establish a blueprint for math protocols that could be applied universally outside of the TensorFlow ecosystem._

The task is closely related to [swift-numerics](https://github.com/apple/swift-numerics) project. Ideally, math protocols should settle there, and yet `swift-apis` is more mature for experimenting with actual algorithms, and therefore `swift-apis` will benefit from protocols right off the bat.

[tf.linalg.LinearOperator](https://www.tensorflow.org/api_docs/python/tf/linalg/LinearOperator) and [tfp.bijectors.Bijector](https://www.tensorflow.org/probability/api_docs/python/tfp/bijectors/Bijector) are examples where math protocols might prove useful. The Swift could adapt operators like  `LinearOperatorLowerTriangular` and `LinearOperatorToeplitz` as special tensors, sort of LowerTriangularTensor and ToeplitzTensor that conform math protocols and also ensure integration with other shaped structures.

The first stage would be to clean interfaces and create stubs for protocols.

I would appreciate any comments and feedback from everyone, especially @dan-zheng, @BradLarson and other members of the S4TF team. Thanks!


8bitmp3:
@awav this is cool. Not an expert in any of this but just wondering if https://github.com/tensorflow/swift-apis/pull/680/commits/4adfa4e109cf4be64c903f3239d245f7ab824e28 is different from generic elementary functions in https://github.com/tensorflow/swift-apis/blob/master/Sources/TensorFlow/Operators/Math.swift

awav:
@8bitmp3, I've updated the description to the PR.

> Not an expert in any of this but just wondering if 4adfa4e is different from generic elementary functions in https://github.com/tensorflow/swift-apis/blob/master/Sources/TensorFlow/Operators/Math.swift

It is different in the sense that math functions defined for a broader class of shaped structures including scalars and shaped arrays.

awav:
> These APIs sound interesting!
> 
> tf.linalg.LinearOperator seems like a unified matrix interface that avoids materialization
> 
> tfp.bijectors.Bijector seems like a "smooth covering map" interface (defining forward, inverse, log_det_jacobian methods), useful for probabilistic programming.
> 

@dan-zheng, @dabrahams Things like `LinearOperator` and `Bijector` play a crucial role in building all sort of algorithms in machine learning.

* Bijector abstraction is an important part of constrained parameter. The simplest example is `variance` of Gaussian distribution - it can take only positive values. Let's say you use a non-constrained gradient optimisation, like BFGS or SGD, for fitting your model that uses variance as a parameter, something like Gaussian mixture model. The variance in the model is used as a positive value only, but gradients must be computed w.r.t. unconstrained representation. For that purpose, Swift could have a tensor-like structure that conforms math protocols and works smoothly with other tensors. That tensor-like structure could accept Bijector object in constructor and keep unconstrained value hidden inside. When that parameter is used in a computation, it transforms implicitly unconstrained tensor with `Bijector.forward` to constrained space. Here is an example from GPflow how it works in Python:

```python
In [11]: p = gpflow.Parameter(0.1, transform=tfp.bijectors.Exp())  # `0.1` is in constrained space
In [12]: p
Out[12]: <gpflow.Parameter 'Variable:0' dtype=float64 unconstrained-shape=() unconstrained-value=-2.3025850929940455 constrained-shape=() constrained-value=0.10000000000000002>
In [13]: p + tf.convert_to_tensor(1.0, dtype=p.dtype)
Out[13]: <tf.Tensor: shape=(), dtype=float64, numpy=1.1>
```

[Normalizing Flows](https://arxiv.org/abs/1912.02762) is another excellent example where bijector abstraction is used.

* `LinearOperator` fits perfectly for use in distributions with different parameterisation and same realisations for pdf, CDF, sampling procedures. Example: multivariate Gaussian distribution can have different covariance matrix structure, it might be [diagonal](https://www.tensorflow.org/versions/r1.15/api_docs/python/tf/contrib/distributions/MultivariateNormalDiag), [lower triangular](https://www.tensorflow.org/versions/r1.15/api_docs/python/tf/contrib/distributions/MultivariateNormalTriL) or the MVN can be parameterised with a precision matrix and have banded structure.


> I think exploring these APIs in-depth could be a productive direction, if you're interested! These are use cases that may motivate public protocol additions (e.g. a LinearOperator protocol or Bijector protocol).

@dan-zheng, Yes, I'm interested! Do you have anything particular in mind what I should do with it? Do you want to make a proposal about bijectors in S4TF?

> Default protocol derivative implementations is a great idea! It enables protocol requirements like AdditiveArithmetic.+ or ElementaryFunctions.sqrt to be differentiable by default, making differentiable programming work seamlessly with protocol-oriented programming.
>
> We've thought about this for a while and it actually requires some more compiler support, tracked by TF-982.

I thought that I was doing something wrong with gradients when they didn't work in my experiments with the code. I would say this is a **must have** feature.

> I agree that swift-numerics is the right place for math protocols to settle and become standardised. As you may know, protocols like ElementaryFunctions and RealFunctions are already being developed there.
>
> When it makes sense, I think tensorflow/swift-apis can become a user of swift-numerics.

I was going to implement these ideas in swift-apis first and then bring them to swift-numerics.

> I don't feel that a new duplicated protocol should be added to tensorflow/swift-apis.

Sorry for the confusion. I didn't mean to duplicate the code; the intention was to make clear interfaces for custom tensor structures. The current state isn't reviewable yet. I will brush up the code and remove duplications.

dabrahams:
@dan-zheng to follow up.

texasmichelle:
Remind to @dan-zheng about follow-up

dan-zheng:
So sorry for the late reply. Thank you @awav for providing more context!

> > I think exploring these APIs in-depth could be a productive direction, if you're interested! These are use cases that may motivate public protocol additions (e.g. a LinearOperator protocol or Bijector protocol).
> 
> @dan-zheng, Yes, I'm interested! Do you have anything particular in mind what I should do with it? Do you want to make a proposal about bijectors in S4TF?

I think prototyping a library for `LinearOperator` or `Bijector` with working use cases would be productive! For example, you could port TensorFlow Probability APIs to Swift and use them to implement some probabilistic programming applications, like the ones you linked. Use cases are absolutely critical for evaluating APIs.

I'd recommend doing this work in a new SwiftPM package using [Swift for TensorFlow toolchains](https://github.com/tensorflow/swift/blob/master/Installation.md). For example, you could start by creating a [`Bijector` protocol](https://www.tensorflow.org/probability/api_docs/python/tfp/bijectors/Bijector#mathematical_details_2) and [implementing some conforming types like `Exp`](https://www.tensorflow.org/probability/api_docs/python/tfp/bijectors/Bijector#example_bijectors_2).

I'd also recommend reusing existing math protocols (like `ElementaryFunctions`) instead of duplicating them or adding new general math protocols, if possible. I feel the bar for creating new public protocols is high: protocols should enable generic algorithms and be well-motivated by use cases, and we should minimize the number of public protocols to reduce API surface. You could start by adding a `Bijector` or `LinearOperator` protocol, then seeing what other functionality is needed to implement use cases.

Feel free to continue discussing here or on the [Swift for TensorFlow mailing list](https://groups.google.com/a/tensorflow.org/forum/#!forum/swift)!

PR #681, opened 2020-02-15
Update the AdaDelta optimizer class API docs 📗

8bitmp3:
Hi all. Updating the AdaDelta optimizer class API docs here:

- Description: expanded, based on Zeiler's 2012 original paper, Keras and TF 2.x docs.
- New parameters section:
    - (Note: TF 2.x AdaDelta API has different default LR and epsilon values defined in the class in `/optimizer_v2/` compared with another `optimizer.py`, the research paper, and Keras' implementation. It's not a new issue but I raised it just in case: https://github.com/tensorflow/tensorflow/issues/36785.)
- Reference: updated (author, year of publishing).
- New example code snippet: it's an autoencoder adapted from /swift-models/ with comments. It was tested in Colab on FashionMNIST separately:
    - The 9th epoch input/output looked like this, if anyone is curious:
        input: image output: image
- To keep things consistent with the new params section, I expanded the definition of `ρ` in the code comments (the words are from the original paper.) Hope you don't mind that.

Working on the rest of the API docs.

Enjoy your weekend.


dan-zheng:

---

Sources/TensorFlow/Optimizers/MomentumBased.swift
```diff
@@ -105,20 +105,69 @@ public class AdaGrad<Model: Differentiable>: Optimizer
...
+/// ````
 public class AdaDelta<Model: Differentiable>: Optimizer
     where Model.TangentVector: VectorProtocol & PointwiseMultiplicative & ElementaryFunctions,
           Model.TangentVector.VectorSpaceScalar == Float {
     public typealias Model = Model
     /// The learning rate.
     public var learningRate: Float
-    /// The decay factor, corresponding to fraction of gradient to keep at each time step.
+    /// The decay factor of the previous parameter update/a fraction of gradient
+    /// to keep at each time step.
```
dan-zheng, line 170: I wonder what's the rationale behind the documentation change for `var rho`?
The current doc comment matches https://keras.io/optimizers/#adadelta:
> **rho**: float >= 0. Adadelta decay factor, corresponding to fraction of gradient to keep at each time step.

dan-zheng: I reverted this change temporarily, since the rationale isn't clear to me.

PR #684, opened 2020-02-20
Support batched Dense without bias

t-ae:
This PR add support for `Dense` with `batched == true && useBias == false` overlooked in #674.


sgugger:
Oh thanks for fixing what I missed! I agree this looks good, but it seems the tests fail for an unrelated reason.

PR #689, opened 2020-02-21
Refine tiled

t-ae:
This PR Adds

- New API `tiled(multiples: [Int])`
- Precondition for `tiled`
- Tests for `tiled`

### DIFF
--- a/Sources/TensorFlow/Operators/Basic.swift
+++ b/Sources/TensorFlow/Operators/Basic.swift
@@ -132,6 +132,22 @@ public extension Tensor {
             splitDim: Tensor<Int32>(Int32(axis)),
             numSplit: Int64(sizes.shape[0]))
     }
+    
+    /// Returns a tiled tensor, constructed by tiling this tensor.
+    ///
+    /// This constructor creates a new tensor by replicating this tensor `multiples` times. The
+    /// constructed tensor's `i`'th dimension has `self.shape[i] * multiples[i]` elements, and the
+    /// values of this tensor are replicated `multiples[i]` times along the `i`'th dimension. For
+    /// example, tiling `[a b c d]` by `[2]` produces `[a b c d a b c d]`.
+    ///
+    /// - Precondition: The expected `rank` of multiples must be `1`.
+    /// - Precondition: The shape of `multiples` must be `[tensor.rank]`.
+    @inlinable
+    @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
+    func tiled(multiples: [Int]) -> Tensor {
+        // TODO(TF-433): Remove workaround for differentiating `map`.
+        tiled(multiples: Tensor<Int32>({multiples.map(Int32.init)}()))
+    }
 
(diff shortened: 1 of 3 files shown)


Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -132,6 +132,23 @@ public extension Tensor {
...
+    /// values of this tensor are replicated `multiples[i]` times along the `i`'th dimension. For
+    /// example, tiling `[a b c d]` by `[2]` produces `[a b c d a b c d]`.
+    ///
+    /// - Precondition: The expected `rank` of multiples must be `1`.
+    /// - Precondition: The shape of `multiples` must be `[tensor.rank]`.
+    /// - Precondition: All scalars in `multiples` must be non-negative.
+    @inlinable
+    @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
+    func tiled(multiples: [Int]) -> Tensor {
+        // TODO(TF-433): Remove workaround for differentiating `map`.
```
saeta: Flagging this for @dan-zheng / @marcrasi .

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -142,13 +159,15 @@ public extension Tensor {
...
     /// - Precondition: The shape of `multiples` must be `[tensor.rank]`.
+    /// - Precondition: All scalars in `multiples` must be non-negative.
     @inlinable
     @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
     func tiled(multiples: Tensor<Int32>) -> Tensor {
         precondition(multiples.rank == 1, "The expected rank of multiples must be 1.")
         precondition(
             rank == multiples.shapeTensor.scalarized(),
             "The shape of multiples must be [tensor.rank].")
+        precondition((multiples .>= 0).all(), "All scalars in multiples must be non-negative.")
```
saeta: At a high level, it's relatively "cheap" to do preconditions against Swift values, and Tensor metadata (e.g. shape, rank), but it's much more expensive (performance wise) to make assertions about the values of Tensors. (This is because Tensors logically represent an asynchronously computed "future" value (i.e. coming from an accelerator) that might not be available yet. Asserting preconditions about the values of tensors forces a device synchronization step, which is expensive for performance.)

You're right that both implementations of tiled should have this precondition, but for performance reasons, could we only have this precondition in the implementation above, and make the assertion on `[Int]`? (e.g. using `multiples.allSatisfy { $0 >= 0 }` or something like that?)

Aside: if an element in multiples is equal to 0, doesn't that mean that the resulting tensor contains zero elements? Do we want this as a possibility, or should we have the precondition be `$0 > 0`? Or am I confused?

t-ae: I followed your suggestion. Thank you for pointing out.
By the way I wonder if `tile(multiple: Tensor<Int32>)` is worth being public.


> Do we want this as a possibility, or should we have the precondition be $0 > 0?

I tried `tf.tile`. It supports 0 for multiplier.

PR #694, opened 2020-02-23
[Dockerfile] Use `rm -f`.

dan-zheng:
Use `rm -f`, which succeeds even when the argument file doesn't exist.

---

Uncovered during `apple/swift` build changes https://github.com/apple/swift/pull/30018:
```console
Step 14/36 : RUN rm /swift-tensorflow-toolchain/usr/lib/swift/linux/x86_64/TensorFlow.swiftinterface
 ---> Running in 02dc699a647d
[91mrm: cannot remove '/swift-tensorflow-toolchain/usr/lib/swift/linux/x86_64/TensorFlow.swiftinterface': No such file or directory
[0mThe command '/bin/sh -c rm /swift-tensorflow-toolchain/usr/lib/swift/linux/x86_64/TensorFlow.swiftinterface' returned a non-zero code: 1
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,10 +27,10 @@ RUN cmake --build /BinaryCache/tensorflow-swift-apis --verbose
 
 # Clean out existing artifacts.
 # TODO: move into bash scripts...
-RUN rm /swift-tensorflow-toolchain/usr/lib/swift/linux/x86_64/TensorFlow.swiftinterface
-RUN rm /swift-tensorflow-toolchain/usr/lib/swift/linux/x86_64/TensorFlow.swiftdoc
-RUN rm /swift-tensorflow-toolchain/usr/lib/swift/linux/x86_64/TensorFlow.swiftmodule
-RUN rm /swift-tensorflow-toolchain/usr/lib/swift/linux/libswiftTensorFlow.so
+RUN rm -f /swift-tensorflow-toolchain/usr/lib/swift/linux/x86_64/TensorFlow.swiftinterface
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -27,10 +27,10 @@ RUN cmake --build /BinaryCache/tensorflow-swift-apis --verbose
 
 # Clean out existing artifacts.
 # TODO: move into bash scripts...
-RUN rm /swift-tensorflow-toolchain/usr/lib/swift/linux/x86_64/TensorFlow.swiftinterface
-RUN rm /swift-tensorflow-toolchain/usr/lib/swift/linux/x86_64/TensorFlow.swiftdoc
-RUN rm /swift-tensorflow-toolchain/usr/lib/swift/linux/x86_64/TensorFlow.swiftmodule
-RUN rm /swift-tensorflow-toolchain/usr/lib/swift/linux/libswiftTensorFlow.so
+RUN rm -f /swift-tensorflow-toolchain/usr/lib/swift/linux/x86_64/TensorFlow.swiftinterface
```
dan-zheng: Note: I believe this logic was added when `TensorFlow` was built as part of the Swift standard library.
Its purpose is to delete `TensorFlow` toolchain artifacts so that the toolchain can be used to build [tensorflow/swift-apis](https://github.com/tensorflow/swift-api).

This may no longer be relevant when `TensorFlow` is no longer built as part of the standard library (ongoing effort), so we can simply remove these lines.

dan-zheng:
I think this patch is not controversial (effectively NFC).
Merging to unblock progress on fixing Swift for TensorFlow toolchain builds.

PR #697, opened 2020-02-25
Public make op

s1ddok:
This patch exposes makeOp function for custom op creation. Details are in[ custom op issue of tensorflow/swift](https://github.com/tensorflow/swift/issues/391)

### DIFF
--- a/Sources/TensorFlow/Core/Runtime.swift
+++ b/Sources/TensorFlow/Core/Runtime.swift
@@ -326,6 +326,13 @@ public final class _ExecutionContext {
   }
 }
 
+@available(
+  *, deprecated, message: "makeOp will go away in favor of directly dispatching custom ops."
+)
+public func _makeOp(_ name: String, _ nOutputs: Int) -> TFTensorOperation {
+  _ExecutionContext.makeOp(name, nOutputs)
+}
+
 extension _ExecutionContext {
   // The execution mode is effectively encoded in the TensorOperation.
   // We can use this to switch between different execution modes.


s1ddok:
@googlebot I signed it!

dan-zheng:

---

s1ddok:
@dan-zheng @pschuh I've updated the repo to comply with SwiftFormat

pschuh:
Hi, just thinking about this pr in general. I've been stalling here because there are unnecessary overheads involved in calling through the makeOp infrastructure that don't really make sense when writing a new op from scratch. If we were, instead, to provide some c headers and then expose the tensorflow::Tensor c++ type itself, then you could write your cuda op directly against that. Then, there would be no need to pack and unpack the arguments into the dictionary needed for makeOp() and everything would be much closer to the overhead of a pure function call. Do you have a particular reason why you want to plug into the existing tf op infrastructure (compatibility with existing tf programs) other than the fact that this is available?

s1ddok:
@pschuh yes, part of me wants to just get up and running quickly to introduce my own data-augmentation operations. another thing is that I want to avoid synchronizing CUDA context between TF and my code, so I want the whole management to be done under the hood. Custom ops seem like a good instruments for that since they abstract away everything outside actual implementation against provided arguments. Another thing is that I think custom ops may be even faster because you don't have to convert your data back and forth between Tensor and say array of raw floats.

pschuh:
Ok, it is just expedient. We're trying to avoid just breaking public apis and since we don't really want this to be public, I would add a public wrapper with an _(so it doesn't just show up in the docs) (or at least just a warning):

@available(*, deprecated, message: "makeOp will go away in favor of directly dispatching custom ops.")
public func _makeOp(_ name: String, _ nOutputs: Int) -> TFTensorOperation {
  makeOp(name, nOutputs)
}

Your concerns about efficiency do not apply to what I was thinking (gpu data would never be materialized into a swift Array<Float>). My plan would be to allow extracting the c++ tensorflow::Tensor type from the swift Tensor type. This should have no performance overhead except avoiding the dispatch overhead of tensorflow op construction and dispatching.

s1ddok:
I personally think that the underscore idea is better than deprecation message. I'll introduce this addition.

Maybe I just don't know enough of TF internal (sorry I'm a very beginner in this), but from what I was thinking there is no way to manually use any custom op code, you have to construct ops with TF so they can co-exist with all other operations and pipelines aren't broken.

Ideally I would love to be able to write raw GPU ops (maybe not even in CUDA, but with Vulkan for instance) but from what I understand memory management is done on TF side, so you can't really mix-and-match without custom ops construction. 

Sorry for my ignorance on the topic, it's fairly hard to find any info on how to do it the right way.

pschuh:
I would like to strongly discourage this, so I want both the _ and the deprecation warning. TF has a bunch of infrastructure for calling ops from a graph represented as protobufs. One major reason for this infrastructure is that the python overhead and threading problems mean that it is necessary to "stage out" a program as an execution graph. "Eager mode" in TF is equivalent to building a single-op graph and dispatching it. This single op graph will call your underlying c++ code (which just takes tensorflow::Tensor, etc). All of this is unnecessary overhead for swift which should be able to just call the function directly.

s1ddok:
Okay, I will do both!

I still don't understand how the CUDA context management will be done with this single-function approach, but what you say sounds very appealing. I think I can learn how to do this once the process of replacing TF ops with Eigen kernels is started on GitHub. 

Thanks for the insights!

texasmichelle:
@pschuh to follow up

Issue #696, opened 2020-02-25
Add optimizer tests


t-ae:
Currently there's no tests for optimizers.
It leads bugs like https://github.com/tensorflow/swift-apis/pull/683#issuecomment-589523022.

I wrote simple convergence tests but I'm not sure if they are worth adding.
https://github.com/t-ae/swift-apis/blob/optimizer-tests/Tests/TensorFlowTests/OptimizerTests.swift
Is there better way to test them?

saeta:
Hi @t-ae ! That is roughly how I'd try and write a convergence test for the optimizers. Can you confirm that the test fails on Linux due to the keypath issue you identified? If so, I think it'd be great to get these tests added to the test suite! (Out of curiosity, how long do they take to run?)

t-ae:
> Can you confirm that the test fails on Linux due to the keypath issue you identified?

Yes. `AdaMax` and `AMSGrad` fails only on Linux. They are using `recursivelyAllWritableKeyPaths`.

> Out of curiosity, how long do they take to run?

It takes 5 secs for these 8 tests in my environment.
Only `AdaGrad` is slow to converge with default parameter.
I found [`tf.keras.optimizers.Adagrad`](https://www.tensorflow.org/api_docs/python/tf/keras/optimizers/Adagrad)'s default learning rate is 0.001 whereas [`keras.optimizers.Adagrad`](https://keras.io/optimizers/#adagrad)'s default learnign rate is 0.01.
I don't know what this difference means but with leraning rate 0.01 it converges as fast as the others.

vballoli:
Hi, in my previous PR https://github.com/tensorflow/swift-apis/pull/564  @dan-zheng suggested adding an OptimizerTest which I'll be submitting today for RAdam inspired from https://github.com/tensorflow/addons/blob/b4cea81897/tensorflow_addons/optimizers/rectified_adam_test.py. Hopefully that can serve as a starting point for all other optimizers

t-ae:
Now my OptimizerTests.swift is merged. But it seems to fail to catch the problem https://github.com/tensorflow/swift-apis/pull/700#discussion_r385097412.
Numerical test @vballoli introduced in #700 is better than my convergence test.
I'll look into it after #700 is merged.

dan-zheng:
Closing, as both optimizer convergence tests and numerical correctness tests have been added. Thanks!

PR #698, opened 2020-02-26
Add OptimizerTests

t-ae:
This PR adds tests for optimizers, mainly for tracking the bug I found in other PR.
https://github.com/tensorflow/swift-apis/pull/683#issuecomment-589523022, https://bugs.swift.org/browse/TF-1178

On macOS all tests pass.
On Linux `testAdaMax` and `testAMSGrad` fail.

### DIFF
--- a/Tests/TensorFlowTests/OptimizerTests.swift
+++ b/Tests/TensorFlowTests/OptimizerTests.swift
@@ -1,0 +1,113 @@
+// Copyright 2019 The TensorFlow Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


saeta:
Note: fixes #696

Shashi456:
@marcrasi @saeta I had removed `OptimizerTests.swift` in #484, since it was incomplete and also redundantly tested optimizers. Given this PR and that sequential tests fail, #683, Should we consider removing the file and making this file the home for optimizer tests?

saeta:
In order to have [focused orthogonal tests](https://github.com/tensorflow/swift-apis/wiki/Testing#focused-orthogonal-tests), I think we should test the optimizers as @t-ae has done in this PR, and we should test Sequential with a unit test (and a single optimizer) as well. But we should not conflate the tests. Does that make sense @Shashi456 ?

marcrasi:
Hmm, these tests are passing for me on my local Linux machine with the latest nightly builds. Maybe the problem has magically fixed itself since last time we ran the tests, or maybe the problem is platform-dependent enough that my Linux machine does not have the problem. I will now:
* rerun CI, and
* try out these tests on a different linux machine

marcrasi:
Aha, when I run on my linux machine with the older nightly toolchain that existed when we last ran these tests, it does fail. So something has indeed magically fixed the problem! I expect that CI will pass now, and I'll merge these tests as soon as CI does pass so that we prevent the problem from reoccurring. Thanks for these nice orthogonal tests!

marcrasi:
I'm also going to dig around in git history a bit to see if I can find any clue about what broke and what fixed this.

PR #700, opened 2020-02-27
Fix RAdam stepSize, Add: OptimizerTests for RAdam

vballoli:
The RAdam implementation was missing a factor `learningRate / (1 - beta1Power)`. As mentioned in an earlier PR, OptimizerTests.swift was added based on the addons implementation in Tensorflow.


Tests/TensorFlowTests/OptimizerTests.swift
```diff
@@ -0,0 +1,54 @@
...
+        // To obtain gradient of type Model.TangentVector 
+        var gradient = model.gradient { model -> Tensor<Float> in 
+            return meanAbsoluteError(predicted: model(w.tranposed()), expected: Tensor<Float>(1.0))
+        }
+        // Custom gradient passed to check optimizer validity
+        gradient.w = Tensor<Float>([[0.1], [0.2]])
+        for _ in 0..<1000 {
+            optimizer.update(&model, along: gradient)
+        }
+        let expected_gradient = Tensor<Float>([[0.5543607], [1.55286]])
```
dan-zheng: Where did this value come from? Could you please share a reference implementation (e.g. from Python)? This seems very important for this change.

Tests/TensorFlowTests/OptimizerTests.swift
```diff
@@ -0,0 +1,54 @@
...
+            public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+                return matmul(input, w) + b
+            }
+        }
+        let w = Tensor<Float>([[1.0], [2.0]])
+        let b = Tensor<Float>([0.0])
+        var model = Model(w: w, b: b)
+        let optimizer = RAdam(for: model, learningRate: 1e-3)
+        // To obtain gradient of type Model.TangentVector 
+        var gradient = model.gradient { model -> Tensor<Float> in
```
dan-zheng: Please use top-level differential operators instead of method-style ones, which are deprecated and already have been (?) removed:
```suggestion
        var gradient = gradient(at: model) { model -> Tensor<Float> in 
```
Reason for removing the method-style ones: they're confusing and duplicate the top-level ones. Let's standardize on the top-level ones.

Tests/TensorFlowTests/OptimizerTests.swift
vballoli: https://github.com/tensorflow/addons/blob/b4cea8189724f88663637daa1df82d035d97cba4/tensorflow_addons/optimizers/rectified_adam_test.py#L73

The values aren't exactly the same, but can be(?) approximated I suppose.

Tests/TensorFlowTests/OptimizerTests.swift
vballoli: Sure, will work on the changes now. Just to clarify, you're talking about `valueWithGradient()` right ?

Tests/TensorFlowTests/OptimizerTests.swift
Shashi456: @dan-zheng official pytorch implementation [here](https://github.com/LiyuanLucasLiu/RAdam/blob/master/radam/radam.py).

vballoli: Unfortunately the official PyTorch version did not come with any value based tests, so I had to take the tests from the tensorflow/addons implementation

dan-zheng: It doesn't matter where the reference implementation comes from, but checking it in code is important! Some shorter reference implementations are included in inline comments, others are checked in as Python files under [`Tools/ReferenceImplementations`](https://github.com/tensorflow/swift-apis/tree/master/Tools/ReferenceImplementations).

Could you please check in a reference implementation in code? When run, the reference implementation should print out the ~exact values used in the test.

Tests/TensorFlowTests/OptimizerTests.swift
dan-zheng: > Sure, will work on the changes now. Just to clarify, you're talking about `valueWithGradient()` right ?

No, I'm just talking about `gradient` actually. Does the exact suggested code change (`gradient(at: model) { ... }`) make sense to you? It does the same thing as `model.gradient { ... }`, except the latter has been removed to reduce API surface area.

Tests/TensorFlowTests/OptimizerTests.swift
vballoli: I have checked these values in a Colab notebook. Here's the gist: https://gist.github.com/vballoli/8f882fd2b5bbafbf98ff28d945ad2f2f#file-blank_swift-ipynb

Tests/TensorFlowTests/OptimizerTests.swift
vballoli: Oh alright, yeah that makes sense. I'll be pushing the changes in sometime.

Tests/TensorFlowTests/OptimizerTests.swift
vballoli: I've updated the GItHub gist link.

dan-zheng: Perhaps you misunderstood: I'm asking if you could please add or link a Python reference implementation (expected values) that checks the Swift RAdam implementation tweaks here (actual values). Does that make sense?

In other words: how do you know `let expected_gradient = Tensor<Float>([[0.5543607], [1.55286]])` is the expected gradient? Did you run some Python RAdam program to get those values? If yes, please add/link the Python code!

Shashi456: @vballoli, Similarly to layer gradient tests, For example [Conv1D gradient](https://github.com/tensorflow/swift-apis/blob/c3f38deb27932ee38af551c3e44738083c25b37d/Tests/TensorFlowTests/LayerTests.swift#L138), you have to run the pytorch/tensorflow code and add the equivalent code in the test in a similar format. Both the test and the reference code should result in the same output.

vballoli: Ah, I get it. Will work on that. I found there was an issue with RAdam since accuracy had dropped and training loss increased drastically after 50 epochs or so. So when I went back to the source code, I found the missing factor that wasn't multiplied.

dan-zheng:

---

vballoli:
> All Swift code in [tensorflow/swift-apis](https://github.com/tensorflow/swift-apis) was automatically formatted via [`swift-format`](https://github.com/apple/swift-format) in #711.
> 
> [Formatting configuration here](https://github.com/tensorflow/swift-apis/blob/master/.swift-format).
> One notable change is that we're now using 2-space indentation, which is consistent with:
> 
> * [Google Swift Style Guide](https://google.github.io/swift)
> * The Swift standard library style
> * The style of [apple/swift-format](https://github.com/apple/swift-format)
> 
> As a result, this pull request now has merge conflicts due to code formatting changes. Please resolve the conflicts if you'd like us to re-review your patch! If you haven't already, please consider installing [`swift-format`](https://github.com/apple/swift-format) so you can auto-format code by yourself in the future!

Definitely. I'm in the midst of my exams right now, will do it as soon as they're done.

vballoli:
@dan-zheng I've added the Python code in the test comments and shifted to the to-level `gradient` as suggested. But `swift-format` is failing multiple tests on Ubuntu 18.04 and I can't post an issue anywhere on the repo. Could you help me out on that ?

Tests/TensorFlowTests/OptimizerTests.swift
```diff
@@ -0,0 +1,74 @@
...
+        // grad_0 = tf.Variable([0.1, 0.2], dtype=tf.dtypes.float32)
+        // grads_and_vars = list(zip([grad_0], [var_0]))
+        // optimizer = RectifiedAdam(lr=1e-3, epsilon=1e-8)
+        // for _ in range(1000):
+        //   optimizer.apply_gradients(grads_and_vars)
+        // print(var_0.read_value())
+        // >>> [0.5553605, 1.5548599]
+        // Current implementation = [0.5543607, [1.55286]]
+        // Difference of [0.0009997, 0.0019999].
+        // ```
```
dan-zheng: > @dan-zheng I've added the Python code in the test comments and shifted to the to-level `gradient` as suggested.

Thanks! I verified that the Python code produces the results in the comments (`[0.5553605, 1.5548599]`). 

I wonder if you have ideas about why there's a `[0.0009997, 0.0019999]` difference from Swift results?

What's the difference before your patch to `Sources/TensorFlow/Optimizers/MomentumBased.swift`? If the difference is lower with this patch, then that's empirical proof that this patch is in the right direction.

vballoli: I'm really not sure about the difference. I was thinking that there won't be any error since the original test in `addons` repo was using `eps=1e-7` as their default parameter, but still an error of `~1e-3` with `eps=1e-8` is something to look for. Regarding swift-format, all the CLI arguments aren't recognized. I'll look into it right now with a fresh build.

t-ae: @vballoli 
It seems there's another bug.


You are ignoring $\hat{m}_t$.

In `tensorflow_addons.optimizers.RectifiedAdam` it appears as `m_corr_t`.
https://github.com/tensorflow/addons/blob/v0.7.1/tensorflow_addons/optimizers/rectified_adam.py#L179

t-ae: I noticed there are several differences between TensorFlow implementation and PyTorch implementation.
https://github.com/tensorflow/addons/blob/v0.7.1/tensorflow_addons/optimizers/rectified_adam.py#L146-L209
https://github.com/LiyuanLucasLiu/RAdam/blob/master/radam/radam.py#L28-L94
PyTorch implementation is official, but it seems TensorFlow version is following the paper more strictly.

Current implementation is based on PyTorch. No wonder the test result doesn't match.

t-ae: My comments above were wrong.
I thought $\hat{m}_t$ doesn't appear in PyTorch version, but it is there in different form.
https://github.com/LiyuanLucasLiu/RAdam/blob/master/radam/radam.py#L159-L161

vballoli: Hi @t-ae, yeah the `* step` wasn't part of the algorithm, my bad. Sorry for the inactivity on this PR, I've not been able to get swift-format working ? Are there any alternatives ?

dan-zheng: > Sorry for the inactivity on this PR, I've not been able to get swift-format working ? Are there any alternatives ?

I resolved merge conflicts for you!

Sorry that `swift-format` isn't building for you - the team is discussing whether to bundle `swift-format` in Swift for TensorFlow toolchains for easier access.

dan-zheng:
I noticed https://github.com/tensorflow/swift-apis/pull/732 also adds optimizer numerical tests. After related PRs are all merged, I'll merge changes into this branch and adapt this PR to use the same approach.

vballoli:
Thanks a lot @dan-zheng !! Sorry I couldn't help with the other issues. I'm looking forward to add all the current optimizers to the test cases with the relevant python code.

PR #702, opened 2020-02-28
delete SequentialTests, replace with SequencedTests

marcrasi:
SequentialTests is hard to maintain and debug, as discussed in https://github.com/tensorflow/swift-apis/pull/683. We already have tests covering most of the things that it covers. The only thing we don't have tests for is `sequenced`. Therefore, this PR deletes SequentialTests and adds some tests covering `sequenced`.

cc @t-ae


Tests/TensorFlowTests/SequencedTests.swift.gyb
```diff
@@ -0,0 +1,42 @@
...
+    }
+}
+
+final class SequencedTests: XCTestCase {
+    % for count in counts:
+    func testSequenced${count}() {
+        let input = Float(0)
+	let output = input.sequenced(
+	    through: ${', '.join(['AddOne()' for _ in range(count)])})
+	XCTAssertEqual(output, ${count})
```
t-ae: Indentation is broken.
```suggestion
    let input = Float(0)
    let output = input.sequenced(
        through: ${', '.join(['AddOne()' for _ in range(count)])})
    XCTAssertEqual(output, ${count})
```

dan-zheng:

---

marcrasi:
> I think there is still some issues with the indentation?

Fixed.

> Also, since this is supposed to be differentiable, could we have a test that checks the differentiability and ensures the gradients are good?

Done!

> Also, I haven't checked, but can you confirm that we have a test that covers the function builder Sequential implementation somewhere?

I found one but it's non-orthogonal and it only tests one of the aritys. I added a gybbed one that tests all the aritys.

PR #704, opened 2020-02-28
Add more device and precision apis.

pschuh:
Update Optimizer, comparison ops BatchNorm and DifferentialOperators to handle mixed precision and devices.
The BatchNorm change is needed to reduce the number of zeros in the vjps.

### DIFF
--- a/Sources/TensorFlow/Core/Device.swift
+++ b/Sources/TensorFlow/Core/Device.swift
@@ -14,3 +14,10 @@ extension Tensor {
         }
     }
 }
+
+extension _Raw {
+  static func toDevice<T: TensorFlowScalar>(_ x: Tensor<T>, _ device: Device) -> Tensor<T> {
+    // TODO: Actually copy to device...
+    return x
+  }
+}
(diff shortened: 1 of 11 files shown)


Sources/TensorFlow/Core/Runtime.swift
```diff
@@ -567,6 +567,15 @@ struct DeviceScopes {
     }
 }
 
+// Evaluate the pullback on a one.
+@usableFromInline
+func pullbackOfOneLikeY<T: TensorFlowFloatingPoint, R>(
```
dan-zheng: `pullbackOfOneLikeY` could use a better name. We can defer this.

Sources/TensorFlow/Core/Device.swift
```diff
@@ -14,3 +14,10 @@ extension Tensor {
         }
     }
 }
+
+extension _Raw {
+  static func toDevice<T: TensorFlowScalar>(_ x: Tensor<T>, _ device: Device) -> Tensor<T> {
```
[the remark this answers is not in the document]
pschuh: Acknowledged

Sources/TensorFlow/Core/Runtime.swift
pschuh: Acknowledged.

Issue #699, opened 2020-02-27
Conform `Array` to `Module` and `Layer`


dan-zheng:
Add a conditional conformance: `extension Array: Layer where Element: Layer`.

An `Array` of `Layer`s is a `Layer` that applies each element layer in sequence to an input to produce an output. The `callAsFunction` method calls [`differentiableReduce`](https://github.com/apple/swift/blob/7c34fa581b4c30df092190bbe067570a638ac2f0/stdlib/public/Differentiation/DifferentiationSupport.swift#L1279-L1285), with the input as the initial value.

I think this definition is correct, but I haven't tested it:
```swift
import TensorFlow

extension Array: Module where Element: Layer, Element.Input == Element.Output {
    public typealias Input = Element.Input
    public typealias Output = Element.Output

    @differentiable
    public func callAsFunction(_ input: Input) -> Output {
        differentiableReduce(input) { $1($0) }
    }
}

extension Array: Layer where Element: Layer, Element.Input == Element.Output {}
```
Please add this to the library and add tests! [Motivating mailing list thread here.](https://groups.google.com/a/tensorflow.org/g/swift/c/JfJYs0x6vpA)

If you'd like to work on this, please comment and I'll add you as the assignee.

dan-zheng:
Comment: it's pretty slick that you can make `Array` a `Layer` in just 9 lines.
The power of differentiable programming + protocol-oriented programming!

SumanSudhir:
@dan-zheng can I take up this issue as I need a similar module for one of my models?

rahulbhalley:
Oh s---! @SumanSudhir bro I wanted this! You been too quick to get assigned! 🤪

Just kidding. Looking towards your work. 😃

SumanSudhir:
Sorry, my exam was going on, so I didn't spend time on it that day

saeta:
Reassigning to @dabrahams who had some thoughts on design here. Thanks!

s1ddok:
Just tried the proposed approach by @dan-zheng - very convenient but unfortunately doesn't work for now. Using array for my resblocks results in strange errors like that:

```Fatal error: Inputs to operation ReluGrad of type ReluGrad must have the same size and shape.  Input 0: [] != input 1: [1,64,64,256]: file /swift-base/tensorflow-swift-apis/Sources/TensorFlow/Bindings/EagerExecution.swift, line 30
/bin/sh: 53: Fatal: not found
```

[...]

PR #708, opened 2020-02-29
Conform Array to Module and Layer

SumanSudhir:
#699


Sources/TensorFlow/Core/ShapedArray.swift
```diff
@@ -995,3 +995,14 @@ fileprivate extension _ShapedArrayProtocol where Scalar: Equatable {
         }
     }
 }
+
+// conditional conformance of array to module and layer
+extension Array: Module where Element: Layer, Element.Input == Element.Output {
```
dan-zheng: This is only the `Array: Module` conformance. Please add the `Array: Layer` conformance too, from https://github.com/tensorflow/swift-apis/issues/699.
```swift
extension Array: Layer where Element: Layer, Element.Input == Element.Output {}
```

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -1647,8 +1647,37 @@ final class LayerTests: XCTestCase {
         XCTAssertEqual(transformerTensor.shape, transformerResult.shape)
     }
 
+    func testConformArraytoLayer(){
+        var add_layer:[Dense<Float>] = []
```
dan-zheng: The name `add_layer` doesn't quite make sense. It sounds like the name of a function, but it's an array.
```suggestion
        var layers: [Dense<Float>] = []
```

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -1647,8 +1647,37 @@ final class LayerTests: XCTestCase {
...
+        let input = Tensor<Float>(shape: [5, 8], scalars: (0..<40).map(Float.init))
+
+        let output = add_layer(input)
+        let expected = input.sequenced(through:layer1,layer2,layer3)
+        assertEqual(output, expected,accuracy: 1e-5)
+    }
+
     static var allTests = [
         ("testSequential", testSequential),
+        ("testConformArraytoLayer", testConformArraytoLayer),
```
dan-zheng:
```suggestion
        ("testArrayLayerConformance", testArrayLayerConformance),
```

Shashi456:
Can the doc comments include, what this enables us to do?

SumanSudhir:
@dan-zheng I didn't get the motive behind `Array: Layer` conformance.  How  `Array: Module` conformance and `Array: Layer` conformance are related to each other ??

SumanSudhir:
> Can the doc comments include what this enables us to do?

@Shashi456 hope this

dan-zheng:
> @dan-zheng I didn't get the motive behind `Array: Layer` conformance. How `Array: Module` conformance and `Array: Layer` conformance are related to each other ??

We want `Array` to conform to `Layer`, not just `Module`. This requires two conformances, and you should add both. The `Array: Layer` conformance just is an additional one-liner: https://github.com/tensorflow/swift-apis/pull/708#discussion_r386057160

SumanSudhir:
@dan-zheng I have added it can you please review it.

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -1647,8 +1647,37 @@ final class LayerTests: XCTestCase {
         XCTAssertEqual(transformerTensor.shape, transformerResult.shape)
     }
 
+    func testArrayLayerConformance(){
```
dan-zheng: Actually, just the name `testArray` is sufficient. Since this is `LayerTests.swift`, it's implied that the `Array: Layer` conformance is being tested.
```suggestion
    func testArray() {
```

SumanSudhir:
@dan-zheng can it be merged now?

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -1647,8 +1647,37 @@ final class LayerTests: XCTestCase {
         XCTAssertEqual(transformerTensor.shape, transformerResult.shape)
     }
 
+    func testArray(){
+        var layers: [Dense<Float>] = []
+        let sizes = [(8, 7),(7, 6), (6, 5)]
```
saeta:
```suggestion
        let sizes = [(8, 7), (7, 6), (6, 5)]
```

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -1647,8 +1647,37 @@ final class LayerTests: XCTestCase {
...
+        for (inputSize, outputSize) in sizes {
+            let weight = Tensor<Float>(shape: [inputSize, outputSize], scalars: (0..<inputSize*outputSize).map(Float.init))
```
saeta:
```suggestion
            let weight = Tensor<Float>(shape: [inputSize, outputSize], scalars: (0..<Float(inputSize*outputSize)))
```

Does this work? If so, seems clearer.

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -1647,8 +1647,37 @@ final class LayerTests: XCTestCase {
...
+            let bias = Tensor<Float>(shape: [1, outputSize], scalars: (0..<outputSize).map(Float.init))
```
saeta: Same here and below. Thanks!

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -1647,8 +1647,37 @@ final class LayerTests: XCTestCase {
...
+        bias = Tensor<Float>(shape: [1, outputSize], scalars: (0..<outputSize).map(Float.init))
+        let layer2 = Dense<Float>(weight: weight, bias: bias, activation: identity)
+        (inputSize, outputSize) = sizes[2]
+        weight = Tensor<Float>(shape: [inputSize, outputSize], scalars: (0..<inputSize*outputSize).map(Float.init))
+        bias = Tensor<Float>(shape: [1, outputSize], scalars: (0..<outputSize).map(Float.init))
+        let layer3 = Dense<Float>(weight: weight, bias: bias, activation: identity)
+        let input = Tensor<Float>(shape: [5, 8], scalars: (0..<40).map(Float.init))
+
+        let output = layers(input)
+        let expected = input.sequenced(through:layer1,layer2,layer3)
```
saeta:
```suggestion
        let expected = input.sequenced(through: layer1, layer2, layer3)
```

Sources/TensorFlow/Core/ShapedArray.swift
saeta: Marking as resolved, as this has now been added. Thanks @dan-zheng !

Tests/TensorFlowTests/LayerTests.swift
dan-zheng: I don't think it works actually.

The current code maps a `Range<Int>` to `[Float]`, which is accepted by `Tensor.init(shape:scalars)`.

The new code tries to construct a `Range<Float>`, which works but cannot be passed to `Tensor.init(shape:scalars)` because it doesn't fulfill generic requirements. (`Range<Float>` is a "continuous" range, not a discrete one.)

SumanSudhir: Yaa It is not working. It is throwing an error 
`error: initializer 'init(shape:scalars:on:)' requires that 'Float' conform to 'SignedInteger' let weight = Tensor<Float>(shape: [inputSize, outputSize], scalars: (0..<Float(inputSize*outputSize)))`

saeta: Yeah, okay. Thanks for trying. Could we factor out a helper method for creating a Float tensor of consecutive scalars of a given size? This pattern is used a lot in this test, and it might be easier to read if there was a helper method instead.

SumanSudhir: Are you saying different function to create tensor of specific size consisting of consecutive scalars?

saeta: Yeah, I'm thinking something like the following:

```swift
func makeConsecutiveTensor(x: Int, y: Int) -> Tensor<Float> {
  // ...
}
```

Note: suggestions for alternate names for this function are welcome!

SumanSudhir: It would be better if for creating consecutive integer we provide range function similar to [numpy.arange](https://docs.scipy.org/doc/numpy/reference/generated/numpy.arange.html) and [tf.range](https://www.tensorflow.org/api_docs/python/tf/range) and then reshape the output tensor as per our need.

```swift
func arange<Scalar: TensorFlowFloatingPoint> (_ x: Int, _ y: Int) -> Tensor<Scalar> {
  return Tensor<Scalar>(shape: [1, y-x], scalars: (x..<y).map(Scalar.init))
}
```
```swift
let x: Tensor<Float> = arange(2,8)
output: [[2.0, 3.0, 4.0, 5.0, 6.0, 7.0]]

x.reshaped(to:[3,2])
output: 
[[2.0, 3.0],
 [4.0, 5.0],
 [6.0, 7.0]]
```
It will also take care of data type of output variable that is needed

dan-zheng: > Could we factor out a helper method for creating a Float tensor of consecutive scalars of a given size? This pattern is used a lot in this test, and it might be easier to read if there was a helper method instead.

I think this `Tensor` initializer (using consecutive scalars) could be generally useful. If we are to write it, I think we should design something solid rather than adding a quick test utility.

I think we should consider this `Tensor` initializer after this PR, for separation of concerns! The current test code is fine.

SumanSudhir:
@dan-zheng I haven't made any major changes in the last commit but don't know why it is showing this branch has `conflict`?

dan-zheng:

---

SumanSudhir:
I am able to build swift-format but when I try to run any swift file then it is showing `command not found`
`$ swift-format Example.swift`
`swift-format: command not found  `

dan-zheng:
> I am able to build swift-format but when I try to run any swift file then it is showing `command not found`

You probably need to add `swift-format` to your `$PATH`.

If you directly built [apple/swift-format](https://github.com/apple/swift-format), I think you can copy the `swift-format` executable to `$HOME/bin` and add `$HOME/bin` to your path:
```console
$ cp .build/debug/swift-format $HOME/bin
$ export PATH="$HOME/bin:$PATH" # add this to ~/.bashrc (your shell rc file)
$ source ~/.bashrc # (your shell rc file)
$ swift-format # this should work
```

SumanSudhir:
@dan-zheng I have converted this to 2-space indentation but when I am adding `testArray()` it is showing conflict but when I remove it then it shows this can be merged. I am unable to understand why It is showing conflict after adding `testArray()`

t-ae:
@SumanSudhir 
You can merge `origin/master` and resolve conflicts on your local.

dabrahams:
@dabrahams to propose a more complete implementation of the idea.

SumanSudhir:
@dabrahams I am unable to [test it ](https://github.com/tensorflow/swift-apis/pull/708#pullrequestreview-377337672). 
It is showing error 
``` swift
error: <Cell 7>:3:8: error: type 'CollectionLayer<Base>' does not conform to protocol 'Layer'
struct CollectionLayer<Base: RandomAccessCollection>  : Layer where Base.Element: Layer, Base.Element.Input == Base.Element.Output {
       ^

error: <Cell 7>:3:8: error: type 'CollectionLayer<Base>' does not conform to protocol 'Layer'
struct CollectionLayer<Base: RandomAccessCollection>  : Layer where Base.Element: Layer, Base.Element.Input == Base.Element.Output {
       ^

error: <Cell 7>:3:8: error: type 'CollectionLayer<Base>.TangentVector' does not conform to protocol 'VectorProtocol'
struct CollectionLayer<Base: RandomAccessCollection>  : Layer where Base.Element: Layer, Base.Element.Input == Base.Element.Output {
       ^

error: <Cell 7>:3:8: error: type 'CollectionLayer<Base>' does not conform to protocol 'Module'
struct CollectionLayer<Base: RandomAccessCollection>  : Layer where Base.Element: Layer, Base.Element.Input == Base.Element.Output {
       ^

error: <Cell 7>:12:8: error: type 'CollectionModule<Base>.TangentVector' does not conform to protocol 'VectorProtocol'
struct CollectionModule<Base: RandomAccessCollection>  : Module where Base.Element: Layer, Base.Element.Input == Base.Element.Output {
       ^

error: <Cell 7>:12:8: error: type 'CollectionModule<Base>' does not conform to protocol 'Module'
struct CollectionModule<Base: RandomAccessCollection>  : Module where Base.Element: Layer, Base.Element.Input == Base.Element.Output {
       ^
```

dabrahams:
@SumanSudhir That wasn't intended to be complete; it was a sketch.  In particular, we need to fill in code in place of these comments:

```swift
   // Other Layer requirements implemented in terms of `base`.
...
    // Other Module requirements implemented in terms of `base`.
```

dabrahams:
I started to work on this, but https://bugs.swift.org/browse/TF-1267 needs to get addressed before I can make it fully general.  In the meantime, a wrapper just for Array should be pretty easy to write, and I'll try to get to it in the next few days.

dabrahams:
Actually https://bugs.swift.org/browse/TF-1267 is not a blocker.  Please see #955.  This should be wrapped up today.

Sources/TensorFlow/Layers/SequentialComposition.swift
```diff
@@ -0,0 +1,57 @@
...
+// limitations under the License.
+
+/// The sequential composition of the elements of some base collection of
+/// layers.
+///
+/// The first element of the base collection is applied first by the composite,
+/// so it is the inner call of the composition or the last in the sequence of
+/// composed layers in “f ∘ g” notation.
+public struct SequentialComposition<Base: Collection>
+  where Base.Element : Layer, Base.Element.Input == Base.Element.Output
```
dan-zheng, line 22: Currently, this patch makes it easy to apply a collection of `Layer`-conforming elements in sequence.

However, `SequentialComposition` is missing crucial conformances to `Differentiable` and `Layer`. `SequentialComposition.callAsFunction` should be `@differentiable(wrt: (self, input))`.

@dabrahams: would you like look into these changes?

dan-zheng: Here's a test case for the final desired functionality (code untested):
```swift
let (output, grad) =
  valueWithGradient(at: layers, input) { $0.sequentiallyComposed($1).sum() }
let (layersGrad, inputGrad) = grad

let (expectedOutput, expectedGrad) =
  valueWithGradient(at: input, layers[0], layers[1], layers[2]) { $0.sequenced(through: $1, $2, $3) }
let (expectedInputGrad, expectedLayer0Grad, expectedLayer1Grad, expectedLayer2Grad) = expectedGrad

XCTAssertEqual(output, expectedOutput)
XCTAssertEqual(inputGrad, expectedInputGrad)
XCTAssertEqual(layersGrad[0], expectedLayer0Grad)
XCTAssertEqual(layersGrad[1], expectedLayer1Grad)
XCTAssertEqual(layersGrad[2], expectedLayer2Grad)
```

dan-zheng:
Status update:

Rather than making concrete `Collection`-conforming types like `Array` conform to `Module` and `Layer`, it seems more generic and useful to create an "`Collection` + `Layer`" adaptor type usable by all `Collection`-conforming types.

This adaptor type is currently implemented in this PR as `SequentialComposition`:
```
public struct SequentialComposition<Base: Collection>
  where Base.Element : Layer, Base.Element.Input == Base.Element.Output {
  ...
}

```

dan-zheng:
I'm closing this PR, since directly conforming `Array` to `Layer` (https://github.com/tensorflow/swift-apis/issues/699) is no longer the direction we're pursuing.

A fresh issue https://github.com/tensorflow/swift-apis/issues/1021 tracks the new layer wrapper type approach and provides rationale.

---

To users: if you need to use an `Array: Layer` conformance in the short-term before #1021 is done, feel free to add it locally to your codebase. A sample definition is available at https://github.com/tensorflow/swift-apis/issues/699.

PR #709, opened 2020-03-02
Reversed tensor

t-ae:
Add `Tensor.reversed` APIs.


Tests/TensorFlowTests/OperatorTests/BasicTests.swift
```diff
@@ -707,6 +719,7 @@ final class BasicOperatorTests: XCTestCase {
         ("testConcatenation", testConcatenation),
         ("testVJPConcatenation", testVJPConcatenation),
         ("testTranspose", testTranspose),
+        ("testReverse", testReverse),
```
dan-zheng: Minor: it would be nice for tests to be named:
* `testTranspose` -> `testTransposed`
* `testReverse` -> `testReversed`

Tests/TensorFlowTests/OperatorTests/BasicTests.swift
```diff
@@ -707,6 +719,7 @@ final class BasicOperatorTests: XCTestCase {
         ("testConcatenation", testConcatenation),
         ("testVJPConcatenation", testVJPConcatenation),
```
dan-zheng: Minor, unrelated to this patch: I wonder why `testVJPConcatenation` isn't defined in `TensorAutoDiffTests.swift`. It should be!

Shashi456: Can we talk about the structure of where gradient tests should be placed at? In the case of matrix and linear algebra operators they are in there own test files. Same for layer tests, is `autodifftests`, supposed to be the place for all other gradient tests?

dan-zheng: Interesting point. I don't care strongly about how derivative test organization right now.

I do think that placing derivative tests right after "original function" tests is a bit nicer for locality. This means we could fully remove `TensorAutoDiffTests.swift`.

Maybe others feel differently though.

Tests/TensorFlowTests/OperatorTests/BasicTests.swift
t-ae: Also changed `testTile`, `testReshape`, `testFlatten`.
I'm not sure if I should rename some tests (`testReshapeToScalar`, `testBroadcast3x0To3x3`, etc) so I left them.

Shashi456: The renaming if necessary should happen in another PR. It's unrelated to this one.

t-ae: @Shashi456 
I reverted them except `testReversed`.
I'm glad if someone make other changes and PR (I'm not confident in my naming).

t-ae:
Added preconditions.
They are related to #673.

dan-zheng:
Before merging, I'd like to chat whether `reversed(axes:)` has the best argument label name.

Other methods, mostly reduction ops (e.g. `sum(alongAxes:)` and `sum(squeezingAxes:)`), use `alongAxes:` instead of `axes:`. I think shortening these labels to `along:` and `squeezing:` may improve usability at callers.

cc @dabrahams

saeta:
> Before merging, I'd like to chat whether `reversed(axes:)` has the best argument label name.
> 
> Other methods, mostly reduction ops (e.g. `sum(alongAxes:)` and `sum(squeezingAxes:)`), use `alongAxes:` instead of `axes:`. I think shortening these labels to `along:` and `squeezing:` may improve usability at callers.
> 
> cc @dabrahams

One related thought: I'm wondering whether it might be valuable to define a special `Axis` type. It would help resolve a number of performance concerns I have with https://github.com/tensorflow/swift-apis/pull/673. Additionally, it would make it clear that we can remove the "type name" from the argument label. Maybe a topic for a design review?

dan-zheng:
> > Before merging, I'd like to chat whether `reversed(axes:)` has the best argument label name.
> > Other methods, mostly reduction ops (e.g. `sum(alongAxes:)` and `sum(squeezingAxes:)`), use `alongAxes:` instead of `axes:`. I think shortening these labels to `along:` and `squeezing:` may improve usability at callers.
> > cc @dabrahams
> 
> One related thought: I'm wondering whether it might be valuable to define a special `Axis` type. It would help resolve a number of performance concerns I have with #673. Additionally, it would make it clear that we can remove the "type name" from the argument label. Maybe a topic for a design review?

Sure, we can chat about it!

I understand your points, but I feel the bar for adding a new public type is high, and I don't feel `Axis` meets that bar right now. I think removing "Axes" from `alongAxes:` and `squeezingAxes:` is ~reasonable without adding a new `Axis` type.

dan-zheng:

---

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -370,6 +370,37 @@ public extension Tensor {
             rangeFrom: 0, to: Int32(rank), stride: 1)
         return transposed(permutation: Tensor<Int32>(defaultPermutations))
     }
+    
+    /// Returns a tensor with specified dimensions reversed.
+    /// - Precondition: Each value in `axes` must be in the range `-rank..<rank`.
+    /// - Precondition: There must be no duplication in `axes`.
+    @inlinable
+    @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
+    func reversed(axes: Tensor<Int32>) -> Tensor {
```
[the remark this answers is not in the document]
pschuh: Yes, one type. Ideally something that is possible to add some sort of small vector optimization to (vs an actual Set or something). I personally think Int... are more trouble than they're worth as well. We can move there more easily by adding overloads to raw ops that that take [Int] and then call into the Tensor<Int32> versions. That way, no public apis would have Tensor<Int32> versions. I think we would probably need to convert and remove all of the dataset apis first because this would make tracing those functions probably not work.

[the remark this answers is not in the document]
saeta: Note: Ed Connell has been working on a fast SIMD-backed type for Tensor shapes in SwiftRT and that appears to be working very well.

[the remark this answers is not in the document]
dan-zheng: > Note: Ed Connell has been working on a fast SIMD-backed type for Tensor shapes in SwiftRT and that appears to be working very well.

That's cool! I see `TensorShape` and `AxisSet` as distinct types, but perhaps they can share some logic.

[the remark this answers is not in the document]
saeta: They absolutely should be separate types; just a pointer / inspiration for implementation.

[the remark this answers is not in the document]
t-ae: Renamed.

Tests/TensorFlowTests/TensorAutoDiffTests.swift
```diff
@@ -630,7 +645,7 @@ final class TensorAutoDiffTests: XCTestCase {
   }
 
   func testSideEffects() {
-    let foo: @differentiable (Tensor<Float>) -> Tensor<Float> = { x in
+    let foo: @differentiable(Tensor<Float>) -> Tensor<Float> = { x in
```
dan-zheng: Could you please revert this unrelated change? I think the original formatting `@differentiable (Tensor<Float>) -> Tensor<Float>` is correct.

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -378,6 +378,37 @@ extension Tensor {
...
+  /// Returns a tensor with specified dimensions reversed.
+  /// - Precondition: Each value in `axes` must be in the range `-rank..<rank`.
+  /// - Precondition: There must be no duplication in `axes`.
+  @inlinable
+  @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
+  func reversed(inAxes axes: [Int]) -> Tensor {
+      precondition(axes.count == Set(axes.map { $0 < 0 ? $0 + rank : $0 }).count,
+                   "There must be no duplication in axes.")
+      let axes = axes.map(Int32.init)
+      return reversed(inAxes: Tensor<Int32>(axes))
```
t-ae: I ran `swift-format` but some lines are not 2-space indented.
`swift-format --configuration .swift-format Sources/TensorFlow/Operators/Basic.swift -i`
What's happening?

Tests/TensorFlowTests/TensorAutoDiffTests.swift
t-ae: Fixed. `swift-format` did this change.
Perhaps my `swift-format` installation is wrong. Is there documentation describes how to install `swift-format` for S4TF?

Sources/TensorFlow/Operators/Basic.swift
pschuh: I wouldn't worry too much about the swift-format right now. I would prefer it to be some sort of presubmit check in which case there is no ambiguity about whose installation is official or not. When that happens, we'll have to re-run it.

Tests/TensorFlowTests/TensorAutoDiffTests.swift
```diff
@@ -16,7 +16,7 @@ import XCTest
 
 @testable import TensorFlow
 
-let cube: @differentiable (Tensor<Float>) -> Tensor<Float> = { ($0 * $0 * $0) }
+let cube: @differentiable(Tensor<Float>) -> Tensor<Float> = { ($0 * $0 * $0) }
```
dan-zheng: Hmm, did `swift-format` make this change? Maybe we can ask `swift-format` to fix this formatting (don't delete space after `@differentiable` in function types).
```suggestion
let cube: @differentiable (Tensor<Float>) -> Tensor<Float> = { ($0 * $0 * $0) }
```

Sources/TensorFlow/Operators/Basic.swift
dan-zheng: > `swift-format --configuration .swift-format Sources/TensorFlow/Operators/Basic.swift -i`

Note: if you run `swift-format` anywhere in the `swift-apis` directory (the top-level contains the `.swift-format` configuration file), there's no need to pass `--configuration`. `swift-format` can find it, looking recursively up the directory structure.

From https://github.com/apple/swift-format#configuration:
> For any source file being checked or formatted, `swift-format` looks for a JSON-formatted file named `.swift-format` in the same directory. If one is found, then that file is loaded to determine the tool's configuration. If the file is not found, then it looks in the parent directory, and so on.

saeta:
Thank you @t-ae !!!

Issue #685, opened 2020-02-20
Support Advanced Layers


Shashi456:
Now that we support a multitude of basic layers, I'd like this issue to serve as the discussion for supporting advanced layers. While #54 tracked basic layers and most have been implemented, We now need to consider which of the following to add in swift-apis. tensorflow/swift-models#231 added bert, with that also added support for transformers and attention layers. Having reviewed a variety of current frameworks, I've made the following list:

- [ ] Masking 
- [ ] Spatial Dropout 1D, 2D, 3D
- [ ] Cropping 1D, 2D, 3D
- [ ] Locally-Connected 1D, 2D
- [ ] ConvLSTM 1D, 2D
- [ ] Concatenate
- [ ] Gaussian Noise
- [ ] Gaussian Dropout
- [ ] Alpha Dropout 
- [ ] TimeDistributed
- [ ] Bidirectional 
- [ ] Dilation2D
- [ ] Erosion2D
- [ ] MaxUnpool 1D, 2D, 3D
- [ ] ReflectionPad, ReplicationPad and ConstantPadding
- [ ] GroupNorm, InstanceNorm
- [ ] Transformer, TransformerEncoder, TransformerDecoder
- [ ] PixelShuffle
- [ ] Attention
- [ ] LSTMCell & GRUCell
- [ ] Recusive Neural Nets #68 
- [ ] Neural Turing Machine #52 

If anyone else has any feature/layer requests please do add to this PR.

8bitmp3:
There’re Attention and MultiHeadAttention defined in `tensorflow/swift-models/Transformer`. Maybe those can graduate into ‘/swift-apis’ in some shape and form.

(see `struct Attention: ParameterlessLayer...`, for example)

Shashi456:
@8bitmp3 yeah that's the plan for the transformer and attention layers afaik, we stage the layers there while eventually moving the fleshed out versions to this repository.

brettkoonce:
could use a UpSampling2D to do an autoencoder demo.

Shashi456:
@brettkoonce Upsampling1D, 2D and 3D already exist.

brettkoonce:
@Shashi456 not sure how i missed that 😓

Shashi456:
@dan-zheng could you add a good first issue tag to this PR? And add any other comments if you have any.

sainikhit2k:
Hi, is this list updated? If so I would like to contribute

xanderdunn:
[...]

PR #710, opened 2020-03-02
Add depthToSpace/spaceToDepth

t-ae:
Add `depthToSpace`/`spaceToDepth` APIs.
They can be used to implement `PixelShuffle` listed in #685.


Shashi456:
@t-ae please add the tests to layer tests file. Neural networks ops if any are being tested there.

t-ae:
@Shashi456 
I added `NNTests.swift` because they are not `Layer`.
(IMHO all operators like `conv2D` should have their tests apart from `Layer`.)

dan-zheng:

---

Sources/TensorFlow/Operators/NN.swift
```diff
@@ -820,3 +820,45 @@ func _vjpAvgPool3D<Scalar: TensorFlowFloatingPoint>(
...
+
+@derivative(of: depthToSpace)
+func _vjpDepthToSpace<Scalar: TensorFlowFloatingPoint>(
+  _ input: Tensor<Scalar>,
+  size: Int
+) -> (value: Tensor<Scalar>, pullback: (Tensor<Scalar>) -> Tensor<Scalar>) {
+  (depthToSpace(input, size: size), { spaceToDepth($0, size: size) })
+}
+
+/// Rearranges blocks of spatial data, into depth.
```
dabrahams: Please see remarks on `depthToSpace`, which all apply here.

Tests/TensorFlowTests/OperatorTests/NNTests.swift
```diff
@@ -0,0 +1,87 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import XCTest
+import TensorFlow
+
+class NNTests: XCTestCase {
+  func testDepthToSpace() {
+    let input = Tensor<Float>([
```
dabrahams: These unit tests seem a little thin.  Are there are edge cases that should be stressed?  What about `size`s other than 2?

Sources/TensorFlow/Operators/NN.swift
```diff
@@ -820,3 +820,45 @@ func _vjpAvgPool3D<Scalar: TensorFlowFloatingPoint>(
     }
   )
 }
+
+//===------------------------------------------------------------------------------------------===//
+// Rearrange depth/space
+//===------------------------------------------------------------------------------------------===//
```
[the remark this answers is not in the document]
t-ae: The limit 100 is defined in `.swift-format`.
https://github.com/tensorflow/swift-apis/blob/master/.swift-format

Sources/TensorFlow/Operators/NN.swift
```diff
@@ -820,3 +820,45 @@ func _vjpAvgPool3D<Scalar: TensorFlowFloatingPoint>(
...
+
+/// Rearranges data from depth into blocks of spatial data.
+///
+/// - Precondition: The input must have rank 4.
+/// - Precondition: The size must be greater than 1.
```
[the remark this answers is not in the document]
t-ae: All other `Precondition` statements in this repository are English phrasing.
I think it's better to follow that custom in this PR.
There should be a issue to define the rule.

Sources/TensorFlow/Operators/NN.swift
```diff
@@ -820,3 +820,45 @@ func _vjpAvgPool3D<Scalar: TensorFlowFloatingPoint>(
     }
   )
 }
+
+//===------------------------------------------------------------------------------------------===//
+// Rearrange depth/space
+//===------------------------------------------------------------------------------------------===//
+
+/// Rearranges data from depth into blocks of spatial data.
```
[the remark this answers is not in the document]
t-ae: I'm flabbergasted to find I named it `size`😂
It must be `blockSize`!

> this would be much more idiomatic Swift as an extension method

I thought about it but this function is only for tensors with rank 4.
I think it's not good to have a method that cannot be used on most of tensors.

Tests/TensorFlowTests/OperatorTests/NNTests.swift
t-ae: I think there's no edge case.

t-ae:
@dabrahams 
Thank you for the review. All your comments are applied or replied.

Sources/TensorFlow/Operators/NN.swift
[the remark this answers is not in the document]
dabrahams: Hmm.  My first reaction is, “oh, that makes sense.”  But then after thinking more I see no fundamental difference between a method with preconditions and a function with preconditions.  Is your point that it might just be too easy to find an inapplicable method via code completion?

I'm working on a formulation that captures rank in the tensor type, so eventually we can solve these problems with conditional extensions.

Sources/TensorFlow/Operators/NN.swift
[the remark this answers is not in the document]
dabrahams: thanks.

Sources/TensorFlow/Operators/NN.swift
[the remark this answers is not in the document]
dabrahams: Hmm.  I appreciate the desire not to set new precedent and create inconsistency, but  we are trying to improve our API practices, and if we don't start doing some things better before, e.g., a whole-codebase revision of precondition documentation can occur, it will be hard to make progress.

@saeta your thoughts on how we should approach this would be appreciated. 
@dan-zheng shall we add this to the document we have in progress?

Tests/TensorFlowTests/OperatorTests/NNTests.swift
dabrahams: And there's no value in testing `blockSize != 2`?  I'll take your word for it if you say so.

Sources/TensorFlow/Operators/NN.swift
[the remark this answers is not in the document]
dan-zheng: Adopting common documentation practices sounds great to me!

> @dan-zheng shall we add this to the document we have in progress?

Sounds good!

Sources/TensorFlow/Operators/NN.swift
```diff
@@ -820,3 +820,51 @@ func _vjpAvgPool3D<Scalar: TensorFlowFloatingPoint>(
     }
   )
 }
+
+//===------------------------------------------------------------------------------------------===//
+// Rearrange depth/space
+//===------------------------------------------------------------------------------------------===//
+
+/// Returns `input` rearranged from depth into blocks of spatial data.
```
dabrahams: Unless “rearranged from depth into blocks of spatial data” is a very broadly-understood concept in this domain, I think we need an example, or at least a reference to documentation of a very similar operation.  I (a non-expert in the domain) can't understand what this summary means.

Sources/TensorFlow/Operators/NN.swift
```diff
@@ -820,3 +820,51 @@ func _vjpAvgPool3D<Scalar: TensorFlowFloatingPoint>(
...
+///
+/// - Parameters:
+///   - input: The tensor to be rearranged.
+///   - blockSize: The size of the spatial block.
+/// - Precondition: The input must have rank 4.
+/// - Precondition: The block size must be greater than 1.
+@differentiable(wrt: input where Scalar: TensorFlowFloatingPoint)
+public func depthToSpace<Scalar>(_ input: Tensor<Scalar>, blockSize: Int) -> Tensor<Scalar> {
```
dabrahams: This is an example where almost everything you need to know can be captured in the summary and a `- Parameters:` block just makes it harder to assimilate the meaning of the function

```suggestion
/// Returns `input` rearranged from depth into spatial data blocks of size `b`.
///
/// - Precondition: `input.rank == 4 && b > 1`.
@differentiable(wrt: input where Scalar: TensorFlowFloatingPoint)
public func depthToSpace<Scalar>(_ input: Tensor<Scalar>, blockSize b: Int) -> Tensor<Scalar> {
```

Sources/TensorFlow/Operators/NN.swift
t-ae: I copied examples from `_Raw.depthToSpace`/`_Raw.spaceToDepth`'s documentation.

Tests/TensorFlowTests/OperatorTests/NNTests.swift
t-ae: I looked into `depthtospace_op_test.py` and find there's one edge case I didn't expect.
https://github.com/tensorflow/tensorflow/blob/83aecde6387dd841fe3cad77b66e5c75f82bd882/tensorflow/python/kernel_tests/depthtospace_op_test.py
I added it in https://github.com/tensorflow/swift-apis/pull/710/commits/197f59a3fafc177f0f7c79f330475acd13a77ca2.

Sources/TensorFlow/Operators/NN.swift
t-ae: I updated the comment:
> Returns `input` rearranged from depth into blocks of spatial data of specified block size.

https://github.com/tensorflow/swift-apis/pull/710/commits/ecc981e235577162e538ad04e2f88a174497f315#diff-3aee58937ce4e8fc2ea9fcdbd5266e93R828
How about it?

dabrahams: Apple's Developer Publications imposed the “of the specified …” style  on all the doc comments we wrote in the Swift Standard Library to say, removing our direct references to parameter names  (actually the DevPubs guidelines also forbid `input` to be written in “code voice”), but it's actually suboptimal both for the person writing the comment (sometimes there's more than one of a thing in the parameter list and you have to come up with creative ways to refer to each one) and the person reading it (who has to mentally work around those creative and longwinded phrasings and associate the phrases with actual parameters).  Sometimes, the lack of clarity must be compensated for by adding a `- Parameters:` clause to the documentation, resulting in long and tedious descriptions of simple APIs whose meaning could be taken in at a glance. 

The parameters have names; let's use them, please.

dabrahams: Sorry, I forgot the <rant on> <rant off> delimiters ;-)

t-ae: > The parameters have names; let's use them, please.

OK.
Please send same suggestion again. I don't want to steal your credit.

dabrahams: Please don't worry about the credit, and if somebody else bothers you about it feel free to send them after me with blame ;-)

t-ae: OK. Updated with small change.
https://github.com/tensorflow/swift-apis/pull/710/commits/47a31206b07cf2f23e472e96db4d98e63694cb0f
Thank you for your help.

Sources/TensorFlow/Operators/NN.swift
dabrahams: That doesn't make it a good summary 😉

If you know what this thing is actually doing, and can come up with a better description, we'd really appreciate it.  I don't know what it's doing, or I'd suggest something myself.  If you can't think of anything better, I'll file an issue for improving these summaries.

Thanks

Sources/TensorFlow/Operators/NN.swift
dabrahams: LGTM, and thank you for your patience with all my very particular requests.

Sources/TensorFlow/Operators/NN.swift
t-ae: Updated: https://github.com/tensorflow/swift-apis/pull/710/commits/d8b6203045f19daa71973b98ebf0b101a026c287

From `_Raw.depthToSpace` documentation:
> Rearranges data from depth into blocks of spatial data.
This is the reverse transformation of SpaceToDepth. More specifically,
his op outputs a copy of the input tensor where values from the `depth`
dimension are moved in spatial blocks to the `height` and `width` dimensions.
The attr `block_size` indicates the input block size and how the data is moved.

I took most descriptive part (More specifically ~).
I afraid I can't do better. If it doesn't look good enough, please file an issue.

dabrahams: LGreatTM!

saeta:
Thank you @t-ae and everyone for getting this in!

PR #711, opened 2020-03-02
Run swift-format.

pschuh:
consolidate repos on 2-space now that swift-formatting works.


allevato:
> Generated files like `Sources/TensorFlow/Bindings/RawOpsGenerated.swift` seem auto-formatted too. This seems fine to me.

You can also add `// swift-format-ignore-file` at the top of these files if you don't want to deal with the churn (and the runtime cost of formatting huge generated files). Up to you.

Shashi456:
@dan-zheng @pschuh, will this also be done for swift-models?

dan-zheng:
> @dan-zheng @pschuh, will this also be done for swift-models?

Yep, when we get a chance.

dabrahams:
> A policy question is: shall we ensure that auto-formatting is maintained?
> Perhaps we could use GitHub hooks to auto-format PRs before merging.
> Or perhaps we could skip automation and periodically manually auto-format.

Personally I don't think the manual route is a good one, but I'd really like to know what @allevato recommends.  Tony, have you experimented with different integrations?

allevato:
> Tony, have you experimented with different integrations?

Nothing yet in Git, unfortunately. (We don't even enforce swift-format in swift-format yet 😬 )

The first thing that pops into my mind would be to use a git pre-commit hook that runs swift-format on the edited files and then re-adds them. But I don't believe there's a way to make the hook automatically present for any new user who clones the repo, and GitHub doesn't support server-side hooks that would let you enforce it there.

Combining the hook and CI might help provide users with guidance, though. You could configure CI to run swift-format in lint mode, and print the diagnostics and fail if there were any errors. Then as part of that script/phase, if there was a failure, provide instructions for users to install swift-format if they don't already have it and to configure the pre-commit hook for the repo so that future commits are auto-formatted (e.g., a link to where they can download the script and how to install it in their checkout).

But I'm by no means a Git expert, so I could be missing something that would make this easier.

saeta:
I think in order to make this a good experience, I'd like us to include a compatible version of `swift-format` with the S4TF toolchain. (This can also simplify the installation instructions for the git pre-commit hook, as we could assume that `swift-format` is already available / on the path / etc.) Note: this will require `swift-format` to start to stabilize; @allevato, do you think `swift-format` is ready for that yet?

joaqo:
Thought I'd share some short feedback from a happy S4TF user.

I've been working on a model using the swift-models repo, and after a bit I thought I'd do things properly and start using swift-format. Unfortunately, I kinda gave up after a few minutes of facing compatibility issues and not finding an easy S4TF oriented step by step guide on how to do so. This is probably my fault and I'm probably missing something, but yeah @saeta 's suggestion of integrating it into the toolchain would be ideal for someone like me. Thanks!

PR #712, opened 2020-03-03
NFC: rename `Tools` to `Utilities`.

dan-zheng:
Other Swift projects like apple/swift-package-manager use `Utilities`.
Consistency is a bit nice. `Utilities` is arguably more general and accurate.


dan-zheng:
Added "people who I thought might have an opinion" as reviewers.

PR #713, opened 2020-03-03
Core: remove Python compatibility

compnerd:
Remove the `Python` module compatibility.  When building on macOS, it
continues to conflict with the `Python` framework from CPython.


compnerd:
@dan-zheng yeah, I think we want to wait a little while before we merge this.

PR #716, opened 2020-03-04
Mark `Dense.init(weight:bias:activation)` as `@differentiable`.

dan-zheng:
Currently, `weight` is the only differentiability parameter.

Making `bias` a differentiability parameter requires `Optional` differentiation:
https://bugs.swift.org/browse/TF-1153.

### DIFF
--- a/Sources/TensorFlow/Layers/Dense.swift
+++ b/Sources/TensorFlow/Layers/Dense.swift
@@ -38,6 +38,11 @@ public struct Dense<Scalar: TensorFlowFloatingPoint>: Layer {
   /// The element-wise activation function type.
   public typealias Activation = @differentiable (Tensor<Scalar>) -> Tensor<Scalar>
 
+  /// Creates an instance from the given weight, optional bias, and activation function.
+  ///
+  /// - Note: currently, `weight` is the only differentiability parameter. `bias` can be made a
+  ///   differentiability parameter after `Optional` conditionally conforms to `Differentiable`:
+  ///   TF-499.
   @differentiable(wrt: weight)
   public init(
     weight: Tensor<Scalar>,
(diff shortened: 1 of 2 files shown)


marcrasi:
What's this for? Depending on what it's for, it might be extremely confusing to users why they can't differentiate with respect to bias. Could you make it differentiable wrt bias by defining two separate initializers `init(weight, bias, activation)` and `init(weight, activation)`?

Shashi456:
+1 on @marcrasi's suggestion. 

Perhaps mentioning in a comment that bias is not differentiable yet would also be useful wrt maintaining documentation.

dan-zheng:
> What's this for? Depending on what it's for, it might be extremely confusing to users why they can't differentiate with respect to bias. Could you make it differentiable wrt bias by defining two separate initializers `init(weight, bias, activation)` and `init(weight, activation)`?

Sorry for not providing context! @texasmichelle is working on https://github.com/texasmichelle/swift-models/tree/gpt2_example. She ran into a non-differentiability error because `Dense.init` is not `@differentiable`.
```console
$ swift build
/Users/danielzheng/swift-models/Models/Text/GPT2/Model.swift:319:13: error: expression is not differentiable
            Dense(weight: embedding.weight.transposed(), bias: Tensor(0.0), activation: { $0 })) }()
            ^
/Users/danielzheng/swift-models/Models/Text/GPT2/Model.swift:319:13: note: cannot differentiate functions that have not been marked '@differentiable' and that are defined in other files
            Dense(weight: embedding.weight.transposed(), bias: Tensor(0.0), activation: { $0 })) }()
            ^
```

I think marking `Dense.init` as `@differentiable` makes sense. It cannot be differentiable wrt `bias` currently because `bias` has type `Tensor<Scalar>?`. TF-499 tracks conforming `Optional` to `Differentiable`.

All said, does this patch make sense to you?

Shashi456:
@dan-zheng what if instead of making this change we considered adding a dense layer with non-optional bias to the GPT2 implementation?

marcrasi:
I think a doc comment makes this good enough. Users will get a static differentiability error and they'll be a bit confused but the comment should clarify it and I don't think it's likely to cause bugs.

PR #729, opened 2020-03-12
Use approximate comparison in testGRU

asuhan:


asuhan:
Thanks! I'll merge once Kokoro succeeds.

PR #730, opened 2020-03-13
Dedup DataTypes, Precision and Tensor using an #if define.

pschuh:


Sources/x10/CMakeLists.txt
```diff
@@ -74,6 +74,7 @@ target_sources(x10_tensor PRIVATE
   ../TensorFlow/Loss.swift
   ../TensorFlow/StdlibExtensions.swift)
 set_target_properties(x10_tensor PROPERTIES
+  DEFINE_SYMBOL USING_X10_BACKEND
```
compnerd: Please don't use `DEFINE_SYMBOL` for this.  This should be be:

```
target_compile_definitions(x10_tensor PRIVATE
  USING_X10_BACKEND)
```

`DEFINE_SYMBOL` is used to identify what module is being built.

asuhan:
Shouldn't Sources/x10/swift_bindings/TensorFlow/Core/Tensor.swift be removed as well?

pschuh:
I'll remove the extra files in a followup.

PR #732, opened 2020-03-13
Add numerical tests for optimizers

t-ae:
This PR adds new tests for `SGD`/`RMSProp`/`Adam` backed by TensorFlow code.
There already are convergence tests for them, but numerical tests are much better.

For other optimizers I tried adding similar tests but they didn't pass with small difference.

Related: #696


t-ae:
I missed the test for `AdaDelta` is also passing.
Added it.

PR #735, opened 2020-03-13
Add x10 continuous integration

asuhan:
~~Not ready, testing.~~ Working now.


asuhan:
Skips the full rebuild if x10 doesn't change, only rebuilds and runs x10 tests in that case - total time around 31-32 minutes (vs 28-29 minutes before). The full rebuild takes 71-72 minutes.

There's potential to improve build times with https://github.com/tensorflow/swift-apis/pull/793 (Bazel remote caching). I think this is still worth landing as is.

PR #736, opened 2020-03-13
build: clean up the install rules

compnerd:
Update the SwiftSupport to provide a wrapper `_install_target` to
provide the functionality of `install(TARGET ...)` until that properly
supports installation of Swift targets.  This isolates the logic into a
single location and will ease eventual migration to proper CMake
support.


compnerd:
Extended tests passed, merging.

PR #739, opened 2020-03-14
Compile TensorHandle into x10.

pschuh:


Sources/TensorFlow/Core/TensorHandle.swift
```diff
@@ -287,6 +289,7 @@ extension ShapedArray where Scalar: _TensorFlowDataTypeCompatible {
   }
 }
 
+#if !USING_X10_BACKEND
 // Tensor conversion.
 extension Tensor {
   public init(_ array: __owned ShapedArray<Scalar>) {
```
saeta: Out of curiosity, there should be an alternate implementation of this initializer that doesn't try a dynamic cast to CTensorTensorBuffer?

pschuh: That would be:
     self = array.buffer.withUnsafeBufferPointer { buffer in
        return Tensor(shape: TensorShape(array.shape), scalars: buffer)
      }

It exists in Tensor<>. I can merge them once this is in.

PR #742, opened 2020-03-16
Fix AMSGrad

t-ae:
`stepSize` computation in `AMSGrad` is wrong.
I also added test to ensure it behave identically to TensorFlow.


Tests/TensorFlowTests/OptimizerTests.swift
```diff
@@ -197,6 +197,30 @@ class OptimizerTests: XCTestCase {
     XCTAssertEqual(model.tensor, [0, 0.99998385, 1.9999841])
   }
 
+  func testAMSGradNumerical() {
+    // The expected value was computed using the following Python code:
+    // ```
```
dabrahams: Nice job including the expected value computation.

Markdown like backticks has no effect outside of doc comments (`///`).  I suggest instead setting the code off by a blank line and a couple of spaces of indentation, and adding `#!python` as a first line (as an extra hint for the reader).

dan-zheng: Note: I'm refactoring numerical optimizer tests, so this comment will go away soon.

PR #743, opened 2020-03-16
add hswish activation function

brettkoonce:


brettkoonce:
derivative(hswish(x)) :: [ relu6(x+3)+ x * _Raw.relu6Grad(x+3) ]  / 6
@dan-zheng does this implementation look correct?

Shashi456:
Just a question, should we add lesser known activation functions to the main api base or just maintain them in swift-models for now and think about adding them later? Other than mobilenetv3, I know few models which use this activation function. This case is the same with gelu activation.

t-ae:
@brettkoonce 
The gradient looks wrong.

```swift
import TensorFlow

@inlinable
@differentiable
public func hswish<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
  x * relu6(x+3) / 6.0
}

@inlinable
@derivative(of: hswish)
func _vjpHSwish<T: TensorFlowFloatingPoint>(
  _ x: Tensor<T>
) -> (value: Tensor<T>, pullback: (Tensor<T>) -> Tensor<T>) {
	(
	hswish(x), { v in
		let relu = relu6(v+3)
		let rawRelu = _Raw.relu6Grad(gradients: v, features: x + 3)
		return (relu + rawRelu * v) / 6.0
	})
}

@differentiable
public func hswish2<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
  x * relu6(x+3) / 6.0
}

let tensor = Tensor<Float>([-4, -2, 0, 2, 4])

print(gradient(at: tensor) { hswish($0).sum() }) // [0.6666667, 0.8333333, 0.8333333, 0.8333333, 0.6666667]
print(gradient(at: tensor) { hswish2($0).sum() }) // [        0.0, -0.16666667,         0.5,   1.1666667,         1.0]
```

You should add test.
(Do we need `_vjpHSwish` in this case, BTW?)

brettkoonce:
@Shashi456 thought i would add it here first, but if needed can yeah just let it live in swift-models, nbd

@t-ae you are correct, i was doing the math wrong, fixed and updated the test to work properly now! you are also correct, we do not need a vjp pullback in this case (or rather, the relu6 call handles it for us), so removed all of that!  thanks for the demo code!

sgugger:
LGTM, thanks a lot!

t-ae:
It's a bit late, but I think `hsigmoid` should be renamed to `hardSigmoid` and `hswish` to `hardSwish`.
https://www.tensorflow.org/api_docs/python/tf/keras/activations/hard_sigmoid

t-ae:
> For reference, the hard sigmoid function may be defined differently in different places.
https://stackoverflow.com/a/39625795

`hsigmoid(x) = min(max((x+3) / 6, 0), 1)`
`tf.keras.activations.hard_sigmoid(x) = min(max(0.2 * x + 0.5, 0), 1)`

The slope value is different.
I found japanese article that describes `slope ≈ 0.2` minimizes the mean squared error from `sigmoid`.
https://qiita.com/hsjoihs/items/88d1569aaef01659bbd5

`hsigmoid` should be identical to `tf.keras.activations.hard_sigmoid`.

brettkoonce:
@t-ae that implementation is the one in the paper "Semantic segmentation of satellite images using a modified cnn with hard-swish activation function."

PR #745, opened 2020-03-16
CMake: add a proper option for X10

compnerd:
Rather than implicitly building X10 support if `X10_INCLUDE_DIR` is
specified, have a proper `BUILD_X10` option.


compnerd:
@asuhan that doesn't change here.  Re-using X10's include path to indicate building X10 support is invalid use of the parameter and confusing.  Adding the explicit option to build X10 support just makes it clearer, the header directory specification doesn't break because of this.

asuhan:
Ok, but then we need a counterpart change in https://github.com/apple/swift/tree/tensorflow to set `BUILD_X10` when passing `--x10` to the build script.

compnerd:
Correct you are!  That should go in first.  I realized that I needed that change, but failed to upload it :-(

PR #747, opened 2020-03-17
Group/Instance normalization layer

t-ae:
This PR adds `GroupNorm`, `InstanceNorm`, and their tests.
Related: #685


Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -1935,6 +1935,58 @@ final class LayerTests: XCTestCase {
     XCTAssertEqual(transformerTensor.shape, transformerResult.shape)
   }
 
+  func testGroupNorm() {
```
Shashi456: Could you also add gradient tests for these layers?

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -224,3 +242,165 @@ public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
+  /// The variance epsilon value.
+  @noDerivative public let epsilon: Scalar
+
+  /// Creates a group normalization layer.
+  public init(
+    offset: Tensor<Scalar>,
+    scale: Tensor<Scalar>,
+    groups: Int,
+    axis: Int,
+    epsilon: Scalar
```
Shashi456: I think It's a good idea to support a default value for epsilon, as mentioned in the paper and most implementations as 1e-5.

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -2005,5 +2057,7 @@ final class LayerTests: XCTestCase {
     ("testBatchNormInference", testBatchNormInference),
     ("testLayerNorm", testLayerNorm),
     ("testLayerNormInference", testLayerNormInference),
+    ("testGroupNorm", testGroupNorm),
+    ("testInstanceNorm", testInstanceNorm)
```
Shashi456:
```suggestion
    ("testInstanceNorm", testInstanceNorm),
```

Sources/TensorFlow/Layers/Normalization.swift
t-ae: There's two initializers and one has default values.
I replicated `BatchNorm`'s design.
Default `epsilon` value `1e-3` is from [tfa.layers.GroupNormalization](https://github.com/tensorflow/addons/blob/v0.7.1/tensorflow_addons/layers/normalizations.py#L73-L99).

Tests/TensorFlowTests/LayerTests.swift
t-ae: Added.
The code got awful after `swift-format`😂

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -1935,6 +1935,152 @@ final class LayerTests: XCTestCase {
...
+      ],
+      [
+        [[-0.9999442, -0.9999442, -0.9999447, -0.9999447, -0.9999442, -0.9999442]],
+        [[0.9999447, 0.9999442, 0.9999442, 0.9999442, 0.9999447, 0.9999447]]
+      ]
+    ]
+    XCTAssertEqual(output, expected)
+    let grad = gradient(at: tensor, layer) { $1($0).squared().sum() }
+    let expectedGrad: Tensor<Float> = [
+      [
```
Shashi456: Reformat these brackets like other tests in the file.

Tests/TensorFlowTests/LayerTests.swift
Shashi456: @t-ae I'd still suggest manual refactoring tho xD

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -12,6 +12,20 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+/// Normalizes a tensor by `mean` and `variance`, and applies a `scale` to it, as well as an `offset`.
+@differentiable
+func _batchNorm<Scalar: TensorFlowFloatingPoint>(
```
sgugger: I think this function should be private (and without `_`). Also, wouldn't `normalize` be a better name? It does not apply batch normalization all the time since you use it for group norm (and indirectly layer norm).

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -12,6 +12,20 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+/// Normalizes a tensor by `mean` and `variance`, and applies a `scale` to it, as well as an `offset`.
```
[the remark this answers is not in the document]
dabrahams: Good instinct to factor out the common calculation, for sure.  Sometimes when these functions arise as a consequence of looking at code, it's hard to find a good name.

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -224,3 +244,166 @@ public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
+/// Reference: [Group Normalization](https://arxiv.org/abs/1803.08494).
+@frozen
+public struct GroupNorm<Scalar: TensorFlowFloatingPoint>: Layer {
+  /// The offset value, also known as beta.
+  public var offset: Tensor<Scalar>
+  /// The scale value, also known as gamma.
+  public var scale: Tensor<Scalar>
+  /// The number of groups.
+  @noDerivative public let groups: Int
+  /// The axis.
```
dabrahams: This is a little vague.  Can we say something like “the axis along which the groups are defined” for example? I don't know what is special about this axis, but right now the comment just says, “this represents an axis” without saying what the axis means.

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -224,3 +244,166 @@ public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
+  @noDerivative public let axis: Int
+  /// The variance epsilon value.
+  @noDerivative public let epsilon: Scalar
+
+  /// Creates a group normalization layer.
```
dabrahams: This summary isn't adding any useful information to the code.  We know it creates an instance of `GroupNorm`— that's given by the language—which is documented as a group normalization layer.  Usually we want to say something about the characteristics of the instance created, and how it relates to the arguments passed.  Sometimes that's hard, and we might not be able to do better than “Creates an instance having the given attributes,” but it's good to try.  And if that's all we can say, we should have a `- Parameters:` clause that documents each of the parameters, and what they mean.  I realize that may end up repeating the doc comments you've got on stored properties, but the stored properties don't show up to the person looking at the init in the API documentation or the quick help in an IDE.

dabrahams: Also, the preconditions should be noted in the doc comment as well

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -224,3 +244,166 @@ public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
+    self.epsilon = epsilon
+  }
+
+  /// Creates a group normalization layer.
+  ///
+  /// - Parameters:
+  ///   - featureCount: The number of features.
+  ///   - groups: The number of groups.
+  ///   - axis: The axis that should be normalized.
+  ///   - epsilon: The small scalar added to variance.
```
dabrahams: Good job on the doc comment!

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -224,3 +244,166 @@ public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
+  public init(
+    featureCount: Int,
+    groups: Int,
```
dabrahams: this should probably be named `groupCount` for consistency with `featureCount`.

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -224,3 +244,166 @@ public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
+    axis: Int = -1,
+    epsilon: Scalar = 1e-3
```
dabrahams: The doc comment needs to describe the default values of any parameters.  Also, if -1 is a special value that is not normally a valid axis value (I don't know), the doc comment should describe it, e.g. something like:

  ///   - axis: The axis that should be normalized, or -1 to indicate all axes.

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -224,3 +244,166 @@ public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
+  ) {
+    precondition(
```
dabrahams: Please describe the precondition in the doc comment.

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -224,3 +244,166 @@ public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
+    self.init(
+      offset: Tensor(zeros: [featureCount]),
+      scale: Tensor(ones: [featureCount]),
+      groups: groups,
+      axis: axis,
+      epsilon: epsilon
+    )
+  }
+
+  /// Returns the output obtained from applying the layer to the given input.
```
dabrahams: Nit: kinda wordy; how about ”Returns the result of applying this layer to `input`.”?

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -224,3 +244,166 @@ public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
+  ///
+  /// - Parameter input: The input to the layer.
+  /// - Returns: The output.
```
dabrahams: Those bullets aren't adding anything to what's in the summary; I suggest striking them.

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -224,3 +244,166 @@ public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
+  @differentiable
+  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+    let positiveAxis = (input.rank + axis) % input.rank
+    precondition(positiveAxis != 0, "The axis cannot be batch axis.")
```
dabrahams: Please describe the precondition in the doc comment.

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -224,3 +244,166 @@ public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
+    return normalized.reshaped(to: input.shape)
+  }
+}
+
+/// A layer that applies instance normalization over a mini-batch of inputs.
+///
+/// Reference: [Instance Normalization: The Missing Ingredient for Fast Stylization](https://arxiv.org/abs/1607.08022).
+@frozen
+public struct InstanceNorm<Scalar: TensorFlowFloatingPoint>: Layer {
+  /// Internal group normalization.
```
dabrahams: This doc comment isn't telling me anything that isn't evident from the code.  Can we do better?  What role does this thing play in the `InstanceNorm` calculation?

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -224,3 +244,166 @@ public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
+  var groupNorm: GroupNorm<Scalar>
+
+  /// The offset value, also known as beta.
+  public var offset: Tensor<Scalar> {
+    _read { yield groupNorm.offset }
+    _modify { yield &groupNorm.offset }
```
dabrahams: Have we verified that this code's performance benefits from using the unofficial language `_read`/`_modify`/`yield` language features, and that a simple getter/setter pair generates worse code?

Same applies to `scale`.

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -224,3 +244,166 @@ public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
+  }
+
+  /// The scale value, also known as gamma.
+  public var scale: Tensor<Scalar> {
+    _read { yield groupNorm.scale }
+    _modify { yield &groupNorm.scale }
+  }
+
+  /// The axis.
```
dabrahams: Can we do better (see my other remark about a similar comment)?

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -224,3 +244,166 @@ public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
+  public var axis: Int { groupNorm.axis }
+  /// The variance epsilon value.
+  public var epsilon: Scalar { groupNorm.epsilon }
+
+  /// Creates a instance normalization layer.
```
dabrahams: Please see my notes on `GroupNorm.init`.

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -224,3 +244,166 @@ public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
+  }
+}
+
+/// A layer that applies instance normalization over a mini-batch of inputs.
+///
+/// Reference: [Instance Normalization: The Missing Ingredient for Fast Stylization](https://arxiv.org/abs/1607.08022).
+@frozen
+public struct InstanceNorm<Scalar: TensorFlowFloatingPoint>: Layer {
+  /// Internal group normalization.
+  var groupNorm: GroupNorm<Scalar>
```
dabrahams: When you find yourself naming a variable after its type, you've almost always missed an opportunity to say something more descriptive and valuable.  What role does this `GroupNorm` play in the  `InstanceNorm`?  That question usually leads to a much better name.  In this case it looks like the best we can do might be `implementation` or `delegate`, but that's just my quick impression—you may be able to do better.

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -224,3 +244,166 @@ public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
+  public var epsilon: Scalar { groupNorm.epsilon }
+
+  /// Creates a instance normalization layer.
+  public init(
+    offset: Tensor<Scalar>,
+    scale: Tensor<Scalar>,
+    axis: Int,
+    epsilon: Scalar
+  ) {
+    groupNorm = GroupNorm(
```
dabrahams: The use of `GroupNorm` as an implementation detail implies all the `init`s of`InstaneNorm` inherit certain preconditions that need to be documented… otherwise it's not an implementation detail! :wink:

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -224,3 +244,166 @@ public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
+    epsilon: Scalar = 1e-3
+  ) {
+    groupNorm = GroupNorm(
+      featureCount: featureCount,
+      groups: featureCount,
+      axis: axis,
+      epsilon: epsilon)
+  }
+
+  /// Returns the output obtained from applying the layer to the given input.
```
dabrahams: Please see my notes on the previous `callAsFunction` doc comment.

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -1935,6 +1935,112 @@ final class LayerTests: XCTestCase {
     XCTAssertEqual(transformerTensor.shape, transformerResult.shape)
   }
 
+  func testGroupNorm() {
+    // The expected values were computed using the following Python code:
+    // ```
```
dabrahams: Nit: markdown has no value outside of documentation (`///`) comments.  I suggest removing the backticks and adding 2 spaces after `//` on each line containing Python code; that will make it easier to read.  Actually I might start with `#!python`, just to clue in the reader who tends to see the "shape" of the comment before reading details.

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -1935,6 +1935,112 @@ final class LayerTests: XCTestCase {
...
+    // import tensorflow as tf
+    // import tensorflow_addons as tfa
+    // x = tf.reshape(tf.range(24, dtype=tf.float32), [2, 2, 1, 6])
+    // layer = tfa.layers.GroupNormalization(groups=2)
+    // with tf.GradientTape() as tape:
+    //     tape.watch(x)
+    //     y = layer(x)
+    //     z = tf.math.reduce_sum(tf.math.square(y))
+    // print(y, tape.gradient(z, [x] + layer.trainable_variables))
```
dabrahams: Great job describing how the expected values were derived!  People often leave that information out.

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -1935,6 +1935,112 @@ final class LayerTests: XCTestCase {
...
+        8.275006, 7.4475055, 8.275006
+      ]))
+    XCTAssert(
+      grad.1.offset.isAlmostEqual(to: [
+        -2.572937, 0, 2.572937,
+        -2.572937, 0, 2.572937
+      ]))
+  }
+
+  func testInstanceNorm() {
```
dabrahams: Please see my notes on `testGroupNorm`

Sources/TensorFlow/Layers/Normalization.swift
t-ae: `_batchNorm` corresponds to [`tf.nn.batch_normalization`](https://github.com/tensorflow/tensorflow/blob/v2.1.0/tensorflow/python/ops/nn_impl.py#L1373-L1434).
I took name from it but `normalize` sounds better.

dabrahams: Sounding better is not usually the criterion we're looking for when choosing a name; the first priority is semantic faithfulness.  Is this part of the calculation actually returning a normalized version of its first argument for some common understanding of the meaning of “normalization?”  If so, `normalized` is great.

Sources/TensorFlow/Layers/Normalization.swift
t-ae: `InstanceNorm` is the special case of `GroupNorm` where `featureCounts` == `groupCounts`.
[`tfa.layers.InstanceNormalization`](https://github.com/tensorflow/addons/blob/v0.7.1/tensorflow_addons/layers/normalizations.py#L277-L318) is a subclass of `tfa.layers.GroupNormalization` but our `InstanceNorm` is struct, so I made it has internal `GroupNorm`.
I'm not sure this design is good.

Tests/TensorFlowTests/LayerTests.swift
t-ae: This is the standard style all other tests using.
Please see #499.

Sources/TensorFlow/Layers/Normalization.swift
t-ae: There are many other `Returns the output obtained from applying the layer to the given input.` in this repository.
It seems this is the standard(?) comment for `Layer.callAsFunction`.
I leave it and wait someone will change all of them.

Sources/TensorFlow/Layers/Normalization.swift
t-ae: Please see https://github.com/tensorflow/swift-apis/pull/747#discussion_r394725120

Sources/TensorFlow/Layers/Normalization.swift
t-ae: If it is `get`/`set`, the code `layer.scale[0] += 1` will be equivalent to:

```swift
var scale = layer.scale
scale[0] += 1
layer.scale = scale
```

because `Tensor` is value type.
There will be CoW and `scale` can have many elements.
With `_read`/`_modify` there'll be no CoW.

Sources/TensorFlow/Layers/Normalization.swift
t-ae: `-1` is valid axis.
If the axis is negative, it means target axis is `targetTensor.rank + axis`.

Sources/TensorFlow/Layers/Normalization.swift
t-ae: It depends on the shape of parameter tensors If this is `batchNorm` or not, so `normalize` is better.

Sources/TensorFlow/Layers/Normalization.swift
dabrahams: By golly, you're right.  I thought they had fixed the optimizer for inlinable getters/setters but I just did some experiments that indicate otherwise.  Thanks for the insight!

t-ae:
@dabrahams 
Thank you for review.

Most of my documentation comments were from `BatchNorm`/`LayerNorm`.
There should be another PR to update their documentation comments too.

Sources/TensorFlow/Layers/Normalization.swift
dabrahams: The aggregation is _very_ good!  We just need to figure out how to talk about it.  How about

```
/// The general normalization layer of which `self` is a special case.
```

dabrahams:
@dabrahams to follow up

dabrahams:
@t-ae I think we need your comment "@googlebot I consent" to get past the CLA check.  Thanks!

dabrahams:
@googlebot I fixed it

dabrahams:
@googlebot I fixed it

PR #748, opened 2020-03-17
Make reversed on Tensor public

kongzii:
Hello,

I was excitedly going to use Tensor.reversed() in my project and find out that it is not public in swift-APIs. I looked at https://github.com/tensorflow/swift-apis/pull/709 and it does not seem to be intended. Hope this is helpful PR.


saeta:
LGTM! Thank you!

PR #751, opened 2020-03-17
tests: shuffle tests into the tests directory

compnerd:
X10 tests were previously located in the X10 directory rather than in
the tests directory.  This consolidates the tests into a single location
which prevents unnecessarily building them if tests are disabled.


compnerd:
This will allow us to run tests with `ninja test`

PR #756, opened 2020-03-19
Refactor and improve error messages.

saeta:
This change is a follow-on to #673 based on post-merge feedback and additional
observations of how to improve the quality of error messages, and simplify the
code.

### DIFF
--- a/Sources/TensorFlow/Operators/Basic.swift
+++ b/Sources/TensorFlow/Operators/Basic.swift
@@ -1325,74 +1325,79 @@ extension Tensor.IndexPath {
 //===------------------------------------------------------------------------------------------===//
 
 extension Tensor {
-  /// Returns `true` iff `axis` can refer to an axis of `self`.
+  /// Returns `true` iff `k` denotes an axis of `self`.
   @usableFromInline
-  internal func isValid<T: BinaryInteger>(axis: T) -> Bool {
-    let axis = Int(axis)
+  internal func isValid<T: BinaryInteger>(axis k: T) -> Bool {
+    let axis = Int(k)
     return axis >= -rank && axis < rank
   }
 
-  /// Returns `true` iff all scalars can refer to an axis of `self`.
+  /// Returns `true` iff each element of `dims` denotes an axis of `self`.
   @usableFromInline
-  internal func areValid<T: BinaryInteger>(axes: [T]) -> Bool {
-    return axes.allSatisfy { isValid(axis: $0) }
+  internal func areValid<T: BinaryInteger>(axes dims: [T]) -> Bool {
+    return dims.allSatisfy { isValid(axis: $0) }
   }
 
-  /// Returns `true` iff all scalars can refer to an axis of `self`.
+  /// Returns `true` iff each element of `dims` denotes an axis of `self`.
+  ///
+  /// - Precondition: `dims` is rank 0 or rank 1.
   @usableFromInline
   internal func areValid(
-    axes: Tensor<Int32>,
+    axes dims: Tensor<Int32>,
     file: StaticString = #file,
     line: UInt = #line
   ) -> Bool {
     precondition(
-      axes.rank < 2,
-      "Axes must have rank 0 or rank 1; axes has rank \(axes.rank) with values \(axes.scalars).",
+      dims.rank < 2,
+      "Axes must have rank 0 or rank 1; axes has rank \(dims.rank) with values \(dims.scalars).",
       file: file,
       line: line)
-    return areValid(axes: axes.scalars)
+    return areValid(axes: dims.scalars)
   }
 
-  /// Ensures the provided axes refer to axis of `self` (are in range `-rank..<rank`).
+  /// Checks that each element of `dims` denotes an axis of `self`, and stops the program with a
+  /// diagnostic otherwise.
   @usableFromInline
   func ensureValid(
-    axes: Tensor<Int32>,
+    axes dims: Tensor<Int32>,
     function: StaticString = #function,
     file: StaticString = #file,
     line: UInt = #line
   ) {
     precondition(
-      areValid(axes: axes, file: file, line: line),
-      "All axes must be in `-rank..<rank` when calling \(function) (rank: \(rank), axes: \(axes))",
+      areValid(axes: dims, file: file, line: line),
+      "All axes must be in `-rank..<rank` when calling \(function) (rank: \(rank), axes: \(dims))",
       file: file,
       line: line)
   }
 
-  /// Ensures the provided axes refer to axis of `self` (are in range `-rank..<rank`).
+  /// Checks that each element of `dims` denotes an axis of `self`, and stops the program with a
+  /// diagnostic otherwise.
   @usableFromInline
   func ensureValid(
-    axes: [Int],
+    axes dims: [Int],
     function: StaticString = #function,
     file: StaticString = #file,
     line: UInt = #line
   ) {
     precondition(
-      areValid(axes: axes),
-      "All axes must be in `-rank..<rank` when calling \(function) (rank: \(rank), axes: \(axes))",
+      areValid(axes: dims),
+      "All axes must be in `-rank..<rank` when calling \(function) (rank: \(rank), axes: \(dims))",
       file: file,
       line: line)
   }
 
+  /// Checks that `k` denotes an axis of `self`, and stops the program with a diagnostic otherwise.
   @usableFromInline
   func ensureValid(
-    axis: Int,
+    axis k: Int,
     function: StaticString = #function,
     file: StaticString = #file,
     line: UInt = #line
   ) {
     precondition(
-      isValid(axis: axis),
-      "Axis must be in `-rank..<rank` when calling \(function) (rank: \(rank), axis: \(axis))",
+      isValid(axis: k),
+      "Axis must be in `-rank..<rank` when calling \(function) (rank: \(rank), axis: \(k))",
       file: file,
       line: line)
   }
--- a/Sources/TensorFlow/Operators/Basic.swift
+++ b/Sources/TensorFlow/Operators/Basic.swift
@@ -1332,57 +1332,57 @@ extension Tensor {
     return axis >= -rank && axis < rank
   }
 
-  /// Returns `true` iff each element of `dims` denotes an axis of `self`.
+  /// Returns `true` iff each element of `axes` denotes an axis of `self`.
   @usableFromInline
-  internal func areValid<T: BinaryInteger>(axes dims: [T]) -> Bool {
-    return dims.allSatisfy { isValid(axis: $0) }
+  internal func areValid<T: BinaryInteger>(axes: [T]) -> Bool {
+    return axes.allSatisfy { isValid(axis: $0) }
   }
 
-  /// Returns `true` iff each element of `dims` denotes an axis of `self`.
+  /// Returns `true` iff each element of `axes` denotes an axis of `self`.
   ///
-  /// - Precondition: `dims` is rank 0 or rank 1.
+  /// - Precondition: `axes` has rank 0 or rank 1.
   @usableFromInline
   internal func areValid(
-    axes dims: Tensor<Int32>,
+    axes: Tensor<Int32>,
     file: StaticString = #file,
     line: UInt = #line
   ) -> Bool {
     precondition(
-      dims.rank < 2,
-      "Axes must have rank 0 or rank 1; axes has rank \(dims.rank) with values \(dims.scalars).",
+      axes.rank < 2,
+      "Axes must have rank 0 or rank 1; axes has rank \(axes.rank) with values \(axes.scalars).",
       file: file,
       line: line)
-    return areValid(axes: dims.scalars)
+    return areValid(axes: axes.scalars)
   }
 
-  /// Checks that each element of `dims` denotes an axis of `self`, and stops the program with a
+  /// Checks that each element of `axes` denotes an axis of `self`, and stops the program with a
   /// diagnostic otherwise.
   @usableFromInline
   func ensureValid(
-    axes dims: Tensor<Int32>,
+    axes: Tensor<Int32>,
     function: StaticString = #function,
     file: StaticString = #file,
     line: UInt = #line
   ) {
     precondition(
-      areValid(axes: dims, file: file, line: line),
-      "All axes must be in `-rank..<rank` when calling \(function) (rank: \(rank), axes: \(dims))",
+      areValid(axes: axes, file: file, line: line),
+      "All axes must be in `-rank..<rank` when calling \(function) (rank: \(rank), axes: \(axes))",
       file: file,
       line: line)
   }
 
-  /// Checks that each element of `dims` denotes an axis of `self`, and stops the program with a
+  /// Checks that each element of `axes` denotes an axis of `self`, and stops the program with a
   /// diagnostic otherwise.
   @usableFromInline
   func ensureValid(
-    axes dims: [Int],
+    axes: [Int],
     function: StaticString = #function,
     file: StaticString = #file,
     line: UInt = #line
   ) {
     precondition(
-      areValid(axes: dims),
-      "All axes must be in `-rank..<rank` when calling \(function) (rank: \(rank), axes: \(dims))",
+      areValid(axes: axes),
+      "All axes must be in `-rank..<rank` when calling \(function) (rank: \(rank), axes: \(axes))",
       file: file,
       line: line)
   }
--- a/Sources/TensorFlow/Operators/Basic.swift
+++ b/Sources/TensorFlow/Operators/Basic.swift
@@ -59,7 +59,7 @@ extension Tensor {
   @inlinable
   @differentiable( where Scalar: TensorFlowFloatingPoint)
   public func unstacked(alongAxis axis: Int = 0) -> [Tensor] {
-    precondition(isAxisInRange(axis), "Axis must be in the range `[-rank, rank)`.")
+    ensureValid(axis: axis)
     let posAxis = axis < 0 ? axis + rank : axis
     return _Raw.unpack(value: self, num: Int64(shape[posAxis]), axis: Int64(posAxis))
   }
@@ -89,7 +89,7 @@ extension Tensor {
   @inlinable
   @differentiable( where Scalar: TensorFlowFloatingPoint)
   public func split(count: Int, alongAxis axis: Int = 0) -> [Tensor] {
-    precondition(isAxisInRange(axis), "Axis must be in the range `[-rank, rank)`.")
+    ensureValid(axis: axis)
     precondition(
       shapeTensor[axis].scalarized() % Int32(count) == 0,
       "Number of ways to split should evenly divide the split dimension.")
@@ -122,7 +122,7 @@ extension Tensor {
   @inlinable
   @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
   public func split(sizes: Tensor<Int32>, alongAxis axis: Int = 0) -> [Tensor] {
-    precondition(isAxisInRange(axis), "Axis must be in the range `[-rank, rank)`.")
+    ensureValid(axis: axis)
     precondition(
       shapeTensor[axis] == sizes.sum(),
       "The values in sizes must add up to the size of dimension axis.")
@@ -396,7 +396,7 @@ extension Tensor {
   @inlinable
   @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
   public func reversed(inAxes axes: Tensor<Int32>) -> Tensor {
-    precondition(areAxesInRange(axes), "All axes must be in the range `[-rank, rank)`.")
+    ensureValid(axes: axes)
     return _Raw.reverseV2(self, axis: axes)
   }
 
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1325,34 +1325,37 @@ extension Tensor.IndexPath {
 //===------------------------------------------------------------------------------------------===//
 
 extension Tensor {
-  /// Returns `true` if the given axis is in the range `[-rank, rank)`.
+  /// Returns `true` if `axis` is in `-rank..<rank`, indicating it refers to an axis of `self`.
```
dabrahams: IMO exposing the implementation in the summary has negative value; it merely obscures the high-level semantics and breaks the abstraction provided by the function.  When I'm reading code that uses it, I want to know what calling this method _means,_ not how it does its work.  I can understand needing to have the information *somewhere*, but I'd expect to find the information about valid axis in documentation on `Tensor`, which surely has `public` APIs that depend on understanding thisrelationship.  If this *were* really the only place to describe it, I'd get it out of the summary and put it in a separate "paragraph."

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1325,34 +1325,37 @@ extension Tensor.IndexPath {
...
   @usableFromInline
-  internal func isAxisInRange<T: BinaryInteger>(_ axis: T) -> Bool {
+  internal func isInRange<T: BinaryInteger>(axis: T) -> Bool {
     let axis = Int(axis)
     return axis >= -rank && axis < rank
   }
 
-  /// Returns `true` if the given scalar tensor is in the range `[-rank, rank)`.
+  /// Returns `true` iff all elements in `axis` are in `-rank..<rank`, indicating they refer to an
+  /// axis of `self`.
```
dabrahams: See remark on the scalar version's comment.

Also, this appears to say this works on arbitrary tensors of axes, when it requires a “scalar tensor” according to the `precondition` call in the body.  If the doc comment were right, I'd expect this to be `areInRange(axes:)`, so I am guessing you need to fix the comment to reflect the precondition?

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1361,6 +1364,50 @@ extension Tensor {
...
         file: file,
         line: line)
-    return areAxesInRange(axes.scalars)
+    return isInRange(axes: axes.scalars)
+  }
+
+  /// Ensures the provided axes refer to axis of `self` (are in range `-rank..<rank`).
+  @usableFromInline
+  func ensureValid(
+      axes: Tensor<Int32>,
```
dan-zheng: Please consistently use 2-space indentation.

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1325,34 +1325,37 @@ extension Tensor.IndexPath {
...
-  internal func isAxisInRange(
-    _ axis: Tensor<Int32>,
+  internal func isInRange(
+    axis: Tensor<Int32>,
     file: StaticString = #file,
     line: UInt = #line
   ) -> Bool {
     precondition(axis.rank == 0, "Axis must have rank 0.", file: file, line: line)
-    return areAxesInRange(axis.scalars)
+    return isInRange(axes: axis.scalars)
```
dabrahams: The plurality of `is` and `axes` definitely clash here!  `areInRange(axes:)` or `isInRange(axis:)` please.

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1325,34 +1325,37 @@ extension Tensor.IndexPath {
...
   }
 
-  /// Returns `true` if all given axes are in the range `[-rank, rank)`.
+  /// Returns `true` iff all elements in `axis` are in `-rank..<rank`, indicating they refer to an
+  /// axis of `self`.
   @usableFromInline
-  internal func areAxesInRange<T: BinaryInteger>(_ axes: [T]) -> Bool {
-    return !axes.contains(where: { !isAxisInRange($0) })
+  internal func isInRange<T: BinaryInteger>(axes: [T]) -> Bool {
+    return !axes.contains { !isInRange(axis: $0) }
```
dabrahams:
```suggestion
    return axes.allSatisfy { isInRange(axis: $0) }
```

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1325,34 +1325,37 @@ extension Tensor.IndexPath {
...
+    return isInRange(axes: axis.scalars)
   }
 
-  /// Returns `true` if all given axes are in the range `[-rank, rank)`.
+  /// Returns `true` iff all elements in `axis` are in `-rank..<rank`, indicating they refer to an
+  /// axis of `self`.
   @usableFromInline
-  internal func areAxesInRange<T: BinaryInteger>(_ axes: [T]) -> Bool {
-    return !axes.contains(where: { !isAxisInRange($0) })
+  internal func isInRange<T: BinaryInteger>(axes: [T]) -> Bool {
```
dabrahams: Since you're taking the trouble to thread `#file` and `#line` through `isInRange(axis:)`, don't you want them here as well, forwarding on to the call in the body?

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1325,34 +1325,37 @@ extension Tensor.IndexPath {
...
   }
 
-  /// Returns `true` if all scalars of the given 1-D tensor are in the range `[-rank, rank)`.
+  /// Returns `true` iff all scalars of the 1-D tensor `axis` are in `-rank..<rank`, indicating they
+  /// refer to an axis of `self`.
   @usableFromInline
-  internal func areAxesInRange(
-    _ axes: Tensor<Int32>,
+  internal func isInRange(
+    axes: Tensor<Int32>,
```
dabrahams: another plural/singular clash

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1361,6 +1364,50 @@ extension Tensor {
         "Axes must have rank 0 or rank 1; axes has rank \(axes.rank) with values \(axes.scalars).",
         file: file,
         line: line)
-    return areAxesInRange(axes.scalars)
+    return isInRange(axes: axes.scalars)
+  }
+
+  /// Ensures the provided axes refer to axis of `self` (are in range `-rank..<rank`).
+  @usableFromInline
+  func ensureValid(
```
dabrahams: OK, so we should choose either “valid” or “in range” to denote whether the something refers to a valid axis.  I vote for the former.

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1361,6 +1364,50 @@ extension Tensor {
         "Axes must have rank 0 or rank 1; axes has rank \(axes.rank) with values \(axes.scalars).",
         file: file,
         line: line)
-    return areAxesInRange(axes.scalars)
+    return isInRange(axes: axes.scalars)
```
dabrahams: You didn't address the clash between the precondition check and the comment that I noted in https://github.com/tensorflow/swift-apis/pull/673#discussion_r394651649

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1361,6 +1364,50 @@ extension Tensor {
...
+  @usableFromInline
+  func ensureValid(
+      axes: Tensor<Int32>,
+      function: StaticString = #function,
+      file: StaticString = #file,
+      line: UInt = #line
+  ) {
+      precondition(
+          isInRange(axes: axes, file: file, line: line),
+          "All axes must be in `-rank..<rank` when calling \(function) (rank: \(rank), axes: \(axes))",
```
dabrahams: C.f. https://github.com/tensorflow/swift-apis/pull/673#discussion_r394652982, What purpose do the backticks serve here?  If we want to set a precedent that all code embedded in error messages should be markdown-ified, I guess that's fine with me (and we should write it down somewhere), but there's no precedent for it.  I note you don't have backticks around the function name.

Sources/TensorFlow/Operators/Basic.swift
saeta: Sorry! I appear to have missed that note. I believe I've fixed this by removing the extra "is..." overload. This should take either a rank 0 (scalar) or rank 1 (vector) tensor, which was the cause for the change in behavior from before (as it was ambiguous).

Sources/TensorFlow/Operators/Basic.swift
saeta: Good catch; thanks!

Sources/TensorFlow/Operators/Basic.swift
saeta: Renamed all of the above to `{is|are}Valid` to work with `ensureValid`'s below.

Sources/TensorFlow/Operators/Basic.swift
saeta: I thought that it was more readable when it is clear that we've switched from English text to Swift-equivalent code, as the syntax `-rank..<rank` is not normal written English, whereas everything else (including the function name) is reasonable English. Does that make sense? Do you think it reads better with the backticks removed?

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1325,76 +1325,62 @@ extension Tensor.IndexPath {
 //===------------------------------------------------------------------------------------------===//
 
 extension Tensor {
-  /// Returns `true` if `axis` is in `-rank..<rank`, indicating it refers to an axis of `self`.
+  /// Returns `true` iff `axis` can refer to an axis of `self`.
```
dabrahams: Nit 1,

```suggestion
  /// Returns `true` iff `axis` denotes an axis of `self`.
```

Sources/TensorFlow/Operators/Basic.swift
dabrahams: Very much improved, thanks!  I left some notes in case you care.

Sources/TensorFlow/Operators/Basic.swift
dabrahams: Sorry for the noise, I obviously got lost in the tangle.

Sources/TensorFlow/Operators/Basic.swift
dabrahams: W00t!

Sources/TensorFlow/Operators/Basic.swift
dabrahams: Not necessarily; it's just that there's no precedent for using backticks this way when printing to the console AFAIK.  The swift compiler, for example, encloses type names in single quotes when printing diagnostics.  Backticks can work too I suppose, but we should choose consciously and record what we choose.  One argument in favor of backticks is that I'd guess people who are used to writing markdown may gravitate toward them naturally.  You might even claim that Markdown is now lingua franca for text with embedded code, which makes backticks “the right” answer.

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1325,42 +1325,75 @@ extension Tensor.IndexPath {
...
   @usableFromInline
-  internal func isAxisInRange<T: BinaryInteger>(_ axis: T) -> Bool {
+  internal func isValid<T: BinaryInteger>(axis: T) -> Bool {
     let axis = Int(axis)
     return axis >= -rank && axis < rank
   }
 
-  /// Returns `true` if the given scalar tensor is in the range `[-rank, rank)`.
+  /// Returns `true` iff all scalars can refer to an axis of `self`.
```
dabrahams: Hmm, that ain't right; it would have to return `false` unconditionally!  Maybe 
```suggestion
  /// Returns `true` iff all elements of axes denote an axis of `self`.
```
Again, you might consider a name for the argument separate from its argument label.

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1325,42 +1325,75 @@ extension Tensor.IndexPath {
...
   @usableFromInline
-  internal func isAxisInRange(
-    _ axis: Tensor<Int32>,
+  internal func areValid<T: BinaryInteger>(axes: [T]) -> Bool {
+    return axes.allSatisfy { isValid(axis: $0) }
+  }
+
+  /// Returns `true` iff all scalars can refer to an axis of `self`.
```
dabrahams: Has a precondition that needs documenting.

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1325,42 +1325,75 @@ extension Tensor.IndexPath {
 //===------------------------------------------------------------------------------------------===//
 
 extension Tensor {
-  /// Returns `true` if the given axis is in the range `[-rank, rank)`.
+  /// Returns `true` iff `axis` can refer to an axis of `self`.
   @usableFromInline
-  internal func isAxisInRange<T: BinaryInteger>(_ axis: T) -> Bool {
+  internal func isValid<T: BinaryInteger>(axis: T) -> Bool {
     let axis = Int(axis)
```
dabrahams: I just noticed this.  You shouldn't need to convert the type of axis (and theoretically risk an out-of-range trap) if `rank` is an integer type.  Swift has heterogeneous numeric comparisons that actually work :-)

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1325,42 +1325,75 @@ extension Tensor.IndexPath {
...
+    precondition(
+      axes.rank < 2,
+      "Axes must have rank 0 or rank 1; axes has rank \(axes.rank) with values \(axes.scalars).",
+      file: file,
+      line: line)
+    return areValid(axes: axes.scalars)
   }
 
-  /// Returns `true` if all given axes are in the range `[-rank, rank)`.
+  /// Ensures the provided axes refer to axis of `self` (are in range `-rank..<rank`).
```
dabrahams: “Ensures” doesn't have a very crisp meaning.  To me it sounds like it's maybe going to modify `axes` so its elements are in range.  
```suggestion
  /// Checks that each element of `axes` denotes an element of
  /// `self`, and stops the program with a diagnostic otherwise. 
```

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1325,42 +1325,75 @@ extension Tensor.IndexPath {
...
+  ) {
+    precondition(
+      areValid(axes: axes, file: file, line: line),
+      "All axes must be in `-rank..<rank` when calling \(function) (rank: \(rank), axes: \(axes))",
+      file: file,
+      line: line)
   }
 
-  /// Returns `true` if all scalars of the given 1-D tensor are in the range `[-rank, rank)`.
+  /// Ensures the provided axes refer to axis of `self` (are in range `-rank..<rank`).
```
dabrahams: See my notes on the tensor overload.

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1325,42 +1325,75 @@ extension Tensor.IndexPath {
...
-  ) -> Bool {
+  ) {
+    precondition(
+      areValid(axes: axes),
+      "All axes must be in `-rank..<rank` when calling \(function) (rank: \(rank), axes: \(axes))",
+      file: file,
+      line: line)
+  }
+
+  @usableFromInline
```
dabrahams: Sorry to mention it, but this should have a comment.

Sources/TensorFlow/Operators/Basic.swift
dabrahams: Also see the suggestion on the overload above.

Sources/TensorFlow/Operators/Basic.swift
saeta: Works for me! Took both. Thanks for the suggestions.

Sources/TensorFlow/Operators/Basic.swift
saeta: I think it reads a bit better with the separate argument name vs label. Picking a name for this one is tricky. I've come up with `dims` but not fully sure I like it yet.

Sources/TensorFlow/Operators/Basic.swift
saeta: Documented precondition; also which suggestion are you referring to?

Sources/TensorFlow/Operators/Basic.swift
saeta: Sorry, what do you mean here? I'm confused.

Sources/TensorFlow/Operators/Basic.swift
saeta: Good catch; thank you!

Sources/TensorFlow/Operators/Basic.swift
saeta: Should be fixed. Thanks!

saeta:
Thank you @dabrahams for the super speedy reviews! I appreciate it.

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1325,42 +1325,80 @@ extension Tensor.IndexPath {
...
   @usableFromInline
-  internal func isAxisInRange<T: BinaryInteger>(_ axis: T) -> Bool {
-    let axis = Int(axis)
+  internal func isValid<T: BinaryInteger>(axis k: T) -> Bool {
+    let axis = Int(k)
     return axis >= -rank && axis < rank
   }
 
-  /// Returns `true` if the given scalar tensor is in the range `[-rank, rank)`.
+  /// Returns `true` iff each element of `dims` denotes an axis of `self`.
```
dabrahams: IMO `dims` is a bit worse than something meaningless like `r`

1. It is an abbrev (strike one)
2. it has a meaning when not read as an abbrev., which makes it confusing.  This has nothing to do with dimming
3. When you do interpret it correctly,  it doesn't seem apt.  This isn't describing the dimensions of something, which is what you would expect from a parameter called “dimensions”

Actually I'm only assuming you mean “dimensions”, because it's an abbrev.  ;-)

dabrahams: I know you're going to chafe every time I refer to [this](https://swift.org/documentation/api-design-guidelines/#choose-parameter-names-to-serve-doc), but it doesn't matter whether the name helps readability in the body of your function; you can always create a new binding there.

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1325,42 +1325,80 @@ extension Tensor.IndexPath {
...
   @usableFromInline
-  internal func isAxisInRange(
-    _ axis: Tensor<Int32>,
+  internal func areValid<T: BinaryInteger>(axes dims: [T]) -> Bool {
+    return dims.allSatisfy { isValid(axis: $0) }
+  }
+
+  /// Returns `true` iff each element of `dims` denotes an axis of `self`.
+  ///
+  /// - Precondition: `dims` is rank 0 or rank 1.
```
dabrahams:
```suggestion
  /// - Precondition: `dims` has rank 0 or rank 1.
```

or

```suggestion
  /// - Precondition: `(0...1).contains(dims.rank`
```

The first one is probably more natural.

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1325,42 +1325,80 @@ extension Tensor.IndexPath {
...
+    precondition(
+      dims.rank < 2,
+      "Axes must have rank 0 or rank 1; axes has rank \(dims.rank) with values \(dims.scalars).",
+      file: file,
+      line: line)
+    return areValid(axes: dims.scalars)
+  }
+
+  /// Checks that each element of `dims` denotes an axis of `self`, and stops the program with a
+  /// diagnostic otherwise.
```
dabrahams: Just so you know I'm not being arbitrary and thoughtless in my suggestions, there are reasons I thought my suggested phrasing with the semicolon was better, FWIW.  I certainly respect your perogative to make a different choice so I'm not going to get into it unless you really want to discuss it.

dabrahams:
> Thank you @dabrahams for the super speedy reviews! I appreciate it.

n/p I'm as eager to close this one out as I'm sure you are!

Sources/TensorFlow/Operators/Basic.swift
saeta: Oh, I have no objections to [that](https://swift.org/documentation/api-design-guidelines/#choose-parameter-names-to-serve-doc); I've reverted to just `axes`.

Sources/TensorFlow/Operators/Basic.swift
saeta: Definitely strongly prefer the first one. Thanks!

saeta:
PTAL?

saeta:
Note: I tested locally and all tests passed. Looking at the [error logs](https://source.cloud.google.com/results/invocations/f307ded2-ccc3-4741-a41e-5465dd862d02/log), it looks like the issue is:

```
Step 4/37 : RUN curl -fSsL $swift_tf_url -o swift.tar.gz     && mkdir usr     && tar -xzf swift.tar.gz --directory=usr --strip-components=1     && rm swift.tar.gz
 ---> Running in 8194eccf8bd0
curl: (22) The requested URL returned error: 404
The command '/bin/sh -c curl -fSsL $swift_tf_url -o swift.tar.gz     && mkdir usr     && tar -xzf swift.tar.gz --directory=usr --strip-components=1     && rm swift.tar.gz' returned a non-zero code: 22
```

which is hopefully a transient error. I'm going to re-kick off kokoro now.

dabrahams:
LGTM

PR #767, opened 2020-03-21
Fix bug in `batchGathering`

saeta:
Previously, if `batchGathering` was operating upon a batch of data, it would get
stuck during the final reshape, due to using the wrong input tensor shape. This
change fixes that and includes a test case that caught the bug before the fix was
applied.

```
Test Case '-[TensorFlowTests.BasicOperatorTests testBatchGatheringAcrossBatch]' started.
Fatal error: Input to reshape is a tensor with 6 values, but the requested shape has 3: file /Users/saeta/src/swift-apis/Sources/TensorFlow/Bindings/EagerExecution.swift, line 301
Exited with signal code 4
```


saeta:
PTAL @eaplatanios

saeta:
Note: this unblocks https://github.com/tensorflow/swift-models/pull/344

eaplatanios:
Sorry @saeta didn't manage to take a look into this earlier because these days are kind of crazy. The fix looks good to me!

saeta:
hehe, no worries, thanks @eaplatanios !!

PR #772, opened 2020-03-22
Fix out-of-tree builds

asuhan:
Fix issues with building outside swift-apis repository. Add the relevant headers to the top-level module.modulemap, they'll be distributed as part of the toolchain.


asuhan:
Depends on https://github.com/apple/swift/pull/30568.

compnerd:
I really think that this needs to be adjusted - the X10 modules should not be part of `CTensorFlow`.

PR #774, opened 2020-03-23
Add runtime dispatching.

pschuh:


asuhan:
1. Does this still allow using x10 tensors directly? I assume so, given that you haven't touched the tests.
2. Does this add new functionality (going to either x10 or eager) or not yet?

pschuh:
All this added is that now you can construct: x10_tensor.Tensor(3, on: Device.defaultTFEager) and get an eager tensor and then actually use that for running ops on the eager device. I'm contemplating making that be the default device so that you have to opt into xla for running on accelerators.

Sources/TensorFlow/Bindings/generate_wrappers.py
```diff
@@ -17,8 +17,10 @@
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
+from builtins import bytes
 
 import os
+import json
```
asuhan: Should we alphabetize here (`json` before `os`)?

asuhan:
Could you add a basic test? No need to cover all operators or anything, just a simple add would suffice.

Sources/TensorFlow/Operators/NN.swift
```diff
@@ -64,7 +64,7 @@ extension Padding {
   }
 
   @inlinable
-  internal var raw2: _Raw.Padding2 {
+  internal var raw2: _Raw.Padding1 {
```
asuhan: What's the story with `Padding2` vs `Padding1`, did these change because of code generation changes?

pschuh: The code generator would use a single number for deduping all enum types. This means that if you had two Padding and two Alignment or something it could show up like so:
Alignment
Alignment1
Padding
Padding2
This means that if another duplicate showed up somewhere, it could bump all the names. I changed it so that it just has a counter per base name instead of this.

PR #781, opened 2020-03-24
build: use `X10_LIBRARY` to reference the x10 library

compnerd:
This allows explicitly specifying the x10 library rather than a single
tensorflow library which is assumed to be x10.


dabrahams:
@asuhan plans to test this

asuhan:
@compnerd Still getting the same https://pastebin.com/KXfNRrcE as before. The ops_test executable links to both libtensorflow.so.2 and libx10.so, so the duplication problems persist.

pschuh:
@asuhan Do we have to do the change where tensorflow.so is just a symlink to x10.so in the toolchain first?

asuhan:
@pschuh We do that already without this patch. Not sure what the difference would be in that case.

PR #784, opened 2020-03-27
Add AlphaDropout Layer

Shashi456:
Adding another layer supported in keras. Build and Test pass locally. Look at the keras implementation [here](https://github.com/tensorflow/tensorflow/blob/e5bf8de410005de06a7ff5393fafdf832ef1d4ad/tensorflow/python/keras/layers/noise.py#L140).

The calculation of x is broken into two lines because the compiler can't type check in time, I can add a comment preceding the statements if need be.


Sources/TensorFlow/Layers/Dropout.swift
```diff
@@ -73,3 +73,52 @@ public struct Dropout<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
...
+/// Alpha Dropout is a `Dropout` that keeps mean and variance of inputs to their 
+/// original values, in order to ensure the self-normalizing property even after this
+/// dropout. Alpha Dropout fits well to Scaled Exponential Linear Units by randomly
+/// setting activations to the negative saturation value.
+///
+/// Source : Self-Normalizing Neural Networks: https://arxiv.org/abs/1706.02515
+@frozen
+public struct AlphaDropout<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
+  @noDerivative public let probability: Double
+  //@noDerivative public let rate: Double
```
saeta, line 88: Tinkercruft that could be removed?

Sources/TensorFlow/Layers/Dropout.swift
```diff
@@ -73,3 +73,52 @@ public struct Dropout<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
...
+    precondition(
+      0...1 ~= probability,
+      "Probability must be a value between 0 and 1 (inclusive) but is \(probability)")
+    self.probability = probability
+  }
+
+  /// Returns the output obtained from applying the layer to the given input.
+  ///
+  /// - Parameter input: The input to the layer.
+  /// - Returns: The output.
```
saeta, line 105: I think the documentation as written is perhaps less helpful than possible. What do you think about the following suggestion:

```suggestion
  /// Adds noise to `input` during training, and is a no-op during inference.
```

Sources/TensorFlow/Layers/Dropout.swift
```diff
@@ -73,3 +73,52 @@ public struct Dropout<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
...
+/// setting activations to the negative saturation value.
+///
+/// Source : Self-Normalizing Neural Networks: https://arxiv.org/abs/1706.02515
+@frozen
+public struct AlphaDropout<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
+  @noDerivative public let probability: Double
+  //@noDerivative public let rate: Double
+
+
+  /// Creates a dropout layer.
```
saeta, line 91:
```suggestion
  /// Initializes an `AlphaDropout` layer with a configurable `probability`.
```

Sources/TensorFlow/Layers/Dropout.swift
```diff
@@ -73,3 +73,52 @@ public struct Dropout<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
...
+    case .training:
+      let alpha = 1.6732632423543772848170429916717
+      let scale = 1.0507009873554804934193349852946
+      let alpha_p = -alpha * scale
+      let uniform = Tensor<Scalar>(randomUniform: input.shape)
+      let noise = uniform .>= Scalar(probability)
+      let a = pow(((1 - probability) * (1 + probability * pow(alpha_p, 2))), -0.5)
+      let b = -a * alpha_p * probability
+      var x = input * Tensor(noise) 
+      x = x + Scalar(alpha_p) * (1 - Tensor(noise))
```
saeta, line 118: The [Keras implementation](https://github.com/tensorflow/tensorflow/blob/f9e899854cc96db28564fa65f22d32a647268fc1/tensorflow/python/keras/layers/noise.py#L183) has a nice description of what's going on at each phase. Do you think you could add similar comments for the reader to make it easier to maintain the code over time? Thanks!

Tests/TensorFlowTests/ContextTests.swift
```diff
@@ -50,8 +50,24 @@ final class ContextTests: XCTestCase {
...
+    XCTAssertEqual(dropout(x), x)
+    withLearningPhase(LearningPhase.inference) {
+      XCTAssertEqual(dropout(x), x)
+      withLearningPhase(LearningPhase.training) {
+        XCTAssertNotEqual(dropout(x), x)
+      }
+      XCTAssertEqual(dropout(x), x)
+    }
+    XCTAssertEqual(dropout(x), x)
+  }
```
t-ae, line 66: The test for `AlphaDropout` should be in LayerTests.swift.
Please see [my similar comment](https://github.com/tensorflow/swift-apis/pull/765/files#r403780571).

saeta: +1 thank you @t-ae for noticing that!

saeta:
Hi @Shashi456 ! Just wanted to follow-up; any update here? Thanks!

Shashi456:
@saeta done. Extremely sorry for the delay

PR #793, opened 2020-03-31
Build libx10.so from source at the beginning before starting docker.

pschuh:


asuhan:
How will this interact with https://github.com/tensorflow/swift-apis/pull/735, could that one be simplified - maybe always call this script thanks to caching?

Also, I think we should build with CUDA support.

pschuh:
This script just always runs at the beginning outside of the docker file. We just need to copy the libx10.so into the docker container set up at the end of this script. Cuda might require some more work. It may be possible to move all of this into the docker container later on, but I'm doing it here so I have access to the gcloud apis. (I think it is just a matter of exporting the right things into the docker container)

Sources/x10/xla_tensor/ops/as_strided_view_update.cpp
```diff
@@ -47,7 +47,7 @@ xla::XlaOp LowerAsStridedViewUpdate(xla::XlaOp target, xla::XlaOp input,
     xla::XlaOp r1_target = XlaHelpers::Flatten(target);
     transposed_input = xla::DynamicUpdateSlice(
         r1_target, r1_input,
-        {XlaHelpers::ScalarValue<xla::int64>(storage_offset, input.builder())});
+        absl::Span<const xla::XlaOp>({XlaHelpers::ScalarValue<xla::int64>(storage_offset, input.builder())}));
```
asuhan, line 50: Out of curiosity, why do we need this change (and the other one to the header), isn't this going to use the same compiler we use for Kokoro?

pschuh: This is because it is using a different compiler in the outside VM. I think I'm going to try and add my caching to the other bazel build command.

PR #795, opened 2020-04-01
Deprecate Datasets.

saeta:
Datasets will be removed due to the tracing infrastructure they rely upon
being deprecated in an upcoming TensorFlow core version. Instead, we will
be moving S4TF to the new batcher API instead.


saeta:
Note: I'll try and work on a follow-up PR that deprecates the underlying `_Raw` infrastructure.

PR #800, opened 2020-04-03
Clean up some #ifs and add an option to allow different default backends

pschuh:


Sources/x10/swift_bindings/Device.swift
```diff
@@ -114,8 +114,12 @@ public struct Device {
 
   /// The default `Device`.
   public static var `default`: Device {
+  #if DEFAULT_BACKEND_EAGER
```
asuhan, line 117: When would we set `DEFAULT_BACKEND_EAGER`?

pschuh: We should set this flag while compiling TensorFlow in order to get an x10 enabled TensorFlow module.

PR #801, opened 2020-04-03
build: build tensorflow as a sub-build

compnerd:
Build tensorflow as part of the swift-apis build rather than injecting
it by configuration parameters.


CMakeLists.txt
```diff
@@ -26,7 +26,32 @@ option(BUILD_X10 "enable the x10 tensor library" OFF)
...
+  ExternalProject_Add(tensorflow
+    GIT_REPOSITORY
+      git://github.com/tensorflow/tensorflow
+    GIT_TAG
+      r2.2
+    CONFIGURE_COMMAND
+      ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/Sources/x10 <SOURCE_DIR>/tensorflow/compiler/tf2xla
+      yes "" | <SOURCE_DIR>/configure.py
+    BUILD_COMMAND
+      bazel build -c opt --define framework_shared_object=false //tensorflow/compiler/tf2xla:libx10.so
```
pschuh, line 41: Can you add a flag for the bazel caching flags?

compnerd: Yeah, we can add flags, but they should be generic so that anyone can run with those flags without issues

PR #805, opened 2020-04-08
Remove Tensorflow dep from ops_test.

pschuh:
Now that runtime dispatching is supported, we can import Tensor once and use runtime dispatching to test both backends against each other.


Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1229,7 +1229,7 @@ public func softmax<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
 @differentiable
 public func softmax<T: TensorFlowFloatingPoint>(_ x: Tensor<T>, alongAxis axis: Int) -> Tensor<T> {
   let xExp = exp(x)
-  return xExp / xExp.sum(alongAxes: Tensor<Int32>(Int32(axis)))
+  return xExp / xExp.sum(alongAxes: Tensor<Int32>(Int32(axis), on: xExp.device))
```
asuhan, line 1232: I assume you're confident you caught all places where this is needed thanks to the fact that there are no implicit device transfers.

pschuh: Only mostly so.

PR #806, opened 2020-04-08
only copy gcloud if it exists.

pschuh:


Dockerfile
```diff
@@ -11,9 +11,9 @@ RUN echo "deb [arch=amd64] https://storage.googleapis.com/bazel-apt stable jdk1.
 RUN apt-get update
 
 WORKDIR /opt
-COPY google-cloud-sdk /opt/google-cloud-sdk
+COPY google-cloud-sd[k] /opt/google-cloud-sdk
```
compnerd, line 14: Why the `[k]` here?  I don't see `google-cloud-sdk` in the tree still, so I don't think that this works.

pschuh: I was trying to work around not having conditional copies.

Dockerfile
```diff
@@ -10,15 +10,13 @@ RUN curl https://bazel.build/bazel-release.pub.gpg | apt-key add -
...
-COPY google-cloud-sdk /opt/google-cloud-sdk
-
-RUN /opt/google-cloud-sdk/bin/gcloud auth list
-
 # Copy the kernel into the container
 WORKDIR /swift-apis
 COPY . .
 
+RUN if test -f google-cloud-sdk; then mv google-cloud-sdk /opt/google-cloud-sdk; fi
+RUN if test -f /opt/google-cloud-sdk/bin/gcloud; then /opt/google-cloud-sdk/bin/gcloud auth list; fi
```
compnerd, line 18: Can you make this a single `RUN` please, this is creating an unnecessary intermittent layer that is unnecessary.

Dockerfile
```diff
@@ -10,15 +10,12 @@ RUN curl https://bazel.build/bazel-release.pub.gpg | apt-key add -
...
-WORKDIR /opt
-COPY google-cloud-sdk /opt/google-cloud-sdk
-
-RUN /opt/google-cloud-sdk/bin/gcloud auth list
-
 # Copy the kernel into the container
 WORKDIR /swift-apis
 COPY . .
 
+RUN if test -f google-cloud-sdk; then mv google-cloud-sdk /opt/google-cloud-sdk; opt/google-cloud-sdk/bin/gcloud auth list; fi
```
saeta, line 17: Do you need `opt/google-cloud-sdk/bin/gcloud auth list` also in the command, or can it be skipped?

PR #807, opened 2020-04-08
Fix ops_test when default is switched to defaultTFEager.

pschuh:


pschuh:
I'll be easier to add that version once TensorFlow module builds against libx10.so.

PR #809, opened 2020-04-09
rm minor warnings

brettkoonce:


Tests/TensorFlowTests/XCTestManifests.swift
```diff
@@ -21,7 +21,6 @@ import XCTest
       testCase(BasicOperatorTests.allTests),
       testCase(ComparisonOperatorTests.allTests),
       testCase(ContextTests.allTests),
-      testCase(DatasetTests.allTests),
```
saeta, line 24: I think we still want to be running these tests, even though Datasets are deprecated. Alas, I don't know of a way to suppress the warning without a lot of contortions. :-( Once we cut the v0.9 release, then we can remove this (and the rest of the datasets implementation).

brettkoonce: kk, tweaked!

PR #812, opened 2020-04-10
Move bazel flags to bazelrc.

pschuh:


pschuh:
~ +2 minutes.
I think Saleem is going to fold it into: https://github.com/tensorflow/swift-apis/pull/801

PR #816, opened 2020-04-13
Add FractionalMaxPool

jarnefeltoliver:
Implementation of `FractionalMaxPool` (2D) layer, `fractionalMaxPool` operation. X10 operation not added. 

One known defect is that the gradient function of `fractionalMaxPool` uses directly `_Raw.fractionalMaxPoolGrad` and thus is not higher order differentiable.

 Also since as of now tuples are not differentiable, the `fractionalMaxPool` returns only the first value (`output`) of `_Raw.fractionalMaxPool` in the forward pass, which is called again when calculating the gradient where the second (`rowPoolingSequence`) and third (`colPoolingSequence`) values are also needed. 

This feature was requested in [TF-1236](https://bugs.swift.org/browse/TF-1236?filter=-1).


Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -1079,7 +1079,7 @@ final class LayerTests: XCTestCase {
             [[0.0, 0.0], [0.0, 0.0]],
             [[0.0, 0.0], [0.0, 0.0]],
           ],
-          [
+          [
```
saeta, line 1082:
```suggestion
          [
```

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -1097,6 +1097,35 @@ final class LayerTests: XCTestCase {
...
+    let x = Tensor(shape: [1, 2, 2, 1], scalars: (0..<4).map(Float.init))
+    let computedGradient = gradient(at: x, layer) { $1($0).sum() }
+    // The expected value of the gradient was computed using the following Python code:
+    // ```
+    // import tensorflow as tf
+    // x = tf.reshape(tf.range(4, dtype=tf.float32), [1, 2, 2, 1])
+    // with tf.GradientTape() as tape:
+    //     tape.watch(x)
+    //     y = tf.math.reduce_sum(tf.nn.fractional_max_pool(x, pooling_ratio=[1, 1.4, 1.5, 1], pseudo_random=False)[0])
+    // print(tape.gradient(y, x))
```
saeta, line 1119:
```suggestion
    // ```
    // import tensorflow as tf
    // x = tf.reshape(tf.range(4, dtype=tf.float32), [1, 2, 2, 1])
    // with tf.GradientTape() as tape:
    //     tape.watch(x)
    //     y = tf.math.reduce_sum(tf.nn.fractional_max_pool(x, pooling_ratio=[1, 1.4, 1.5, 1], pseudo_random=False)[0])
    // print(tape.gradient(y, x))
    // ```
```

Sources/TensorFlow/Layers/Pooling.swift
```diff
@@ -424,3 +424,49 @@ public struct GlobalMaxPool3D<Scalar: TensorFlowFloatingPoint>: ParameterlessLay
     return input.max(squeezingAxes: [1, 2, 3])
   }
 }
+
+/// A fractional max pooling layer.
```
saeta, line 428:
```suggestion
/// A fractional max pooling layer.
///
/// Note: `FractionalMaxPool` does not have an XLA implementation, and thus may have performance implications.
```

Sources/TensorFlow/Operators/NN.swift
```diff
@@ -821,6 +821,84 @@ func _vjpAvgPool3D<Scalar: TensorFlowFloatingPoint>(
   )
 }
 
+/// Returns a 2-D fractional max pooling, with the specified pooling ratios.
+///
```
saeta, line 825:
```suggestion
/// Returns a 2-D fractional max pooling, with the specified pooling ratios.
///
/// Note: `fractionalMaxPool` does not have an XLA implementation, and thus may have performance implications.
///
```

Sources/TensorFlow/Layers/Pooling.swift
```diff
@@ -424,3 +424,51 @@ public struct GlobalMaxPool3D<Scalar: TensorFlowFloatingPoint>: ParameterlessLay
...
+///
+/// Note: `FractionalMaxPool` does not have an XLA implementation, and thus may have performance implications.
+@frozen
+public struct FractionalMaxPool<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
```
t-ae, line 432: It seems `FractionalMaxPool` supports only 2D inputs.
Shouldn't it be `FractionalMaxPool2D` like `MaxPool2D`?

Sources/TensorFlow/Operators/NN.swift
```diff
@@ -821,6 +821,86 @@ func _vjpAvgPool3D<Scalar: TensorFlowFloatingPoint>(
...
+///     generates the pooling sequence in a pseudorandom fashion, otherwise, in a random fashion.
+///   - overlapping: An optional `Bool`. Defaults to `false`. When set to `true`, it means
+///     when pooling, the values at the boundary of adjacent pooling cells are used by both cells.
+///   - deterministic: An Optional `Bool`. When set to `true`, a fixed pooling region will be
+///     used when iterating over a fractionalMaxPool node in the computation graph.
+///   - seed: An optional `Int64`. Defaults to `0`. If set to be non-zero, the random number
+///     generator is seeded by the given seed.
+///   - seed2: An optional `Int64`. Defaults to `0`. A second seed to avoid seed collision.
+@differentiable(wrt: input)
+public func fractionalMaxPool<Scalar: TensorFlowFloatingPoint>(
```
t-ae, line 842: `fractionalMaxPool` supports only 2D input.
It should be `fractionalMaxPool2D` like `maxPool2D`.

Sources/TensorFlow/Layers/Pooling.swift
```diff
@@ -424,3 +424,51 @@ public struct GlobalMaxPool3D<Scalar: TensorFlowFloatingPoint>: ParameterlessLay
...
+  /// Determines whether pooling sequence is generated by pseudorandom fashion.
+  @noDerivative public let pseudoRandom: Bool
+  /// Determines whether values at the boundary of adjacent pooling cells are used by both cells
+  @noDerivative public let overlapping: Bool
+  /// Seed for the random number generator
+  @noDerivative public let seed: Int64
+  /// A second seed to avoid seed collision
+  @noDerivative public let seed2: Int64
+  /// Initializes a `FractionalMaxPool` layer with configurable `poolingRatio`.
+  public init(poolingRatio: (Double, Double, Double, Double), pseudoRandom: Bool = false,
```
t-ae, line 445: `poolingRatio.0` and `poolingRatio.3` always must be 1.0.
It would be good to have simplified initializer like `MaxPool2D` (at least from the point of view of consistency).
https://github.com/tensorflow/swift-apis/blob/010a390dc3a96836f73f801a1704dd2a9114e8a6/Sources/TensorFlow/Layers/Pooling.swift#L88-L101

jarnefeltoliver: Thanks this was a a good idea!

Sources/TensorFlow/Operators/NN.swift
jarnefeltoliver: Done

Sources/TensorFlow/Layers/Pooling.swift
jarnefeltoliver: Done

PR #818, opened 2020-04-13
Remove some x10_device imports.

pschuh:


Sources/x10/swift_bindings/training_loop.swift
```diff
@@ -12,9 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import x10_device
 import x10_tensor
-import x10_xla_tensor_wrapper
+@_implementationOnly import x10_xla_tensor_wrapper
```
compnerd, line 16: Would be nice if this were a separate commit.

PR #821, opened 2020-04-14
rm -> remove_directory

pschuh:


compnerd:
Hmm, I thought that the remove_directory didn’t take arguments, but if CMake is happy, LGTM

PR #826, opened 2020-04-15
Allow reusing x10 from the toolchain.

pschuh:


cmake/modules/FindX10.cmake
```diff
@@ -0,0 +1,11 @@
+file(TO_CMAKE_PATH "${X10_TOOLCHAIN_PATH}" X10_CMAKE_TOOLCHAIN_PATH)
```
compnerd, line 1: Please don't do this.  You can use `get_filename_component` on `${CMAKE_C_COMPILER}` and use that to compute the path you want and use that as a hint, but please don't base this behaviour on this random variable.  There is a standard behaviour expected of the file package (https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html).

cmake/modules/FindX10.cmake
```diff
@@ -0,0 +1,11 @@
...
+
+find_file(X10_LIBRARY libx10.so ${X10_TOOLCHAIN_PATH} PATH_SUFFIXES usr/lib/swift/linux)
```
compnerd, line 3: Please don't look for `libx10.so`.  Additionally, I think that `find_library` is a much better approach than `find_file` for this case.

cmake/modules/FindX10.cmake
```diff
@@ -0,0 +1,11 @@
...
+
+add_library(x10 IMPORTED UNKNOWN)
+set_target_properties(x10 PROPERTIES IMPORTED_LOCATION ${X10_LIBRARY}
```
compnerd, line 6: Please break after `PROPERTIES`.

cmake/modules/FindX10.cmake
```diff
@@ -0,0 +1,11 @@
...
+
+find_file(X10_LIBRARY libx10.so ${X10_TOOLCHAIN_PATH} PATH_SUFFIXES usr/lib/swift/linux)
+
+add_library(x10 IMPORTED UNKNOWN)
+set_target_properties(x10 PROPERTIES IMPORTED_LOCATION ${X10_LIBRARY} 
+	INTERFACE_INCLUDE_DIRECTORIES ${X10_CMAKE_TOOLCHAIN_PATH}/usr/lib/swift/x10/include)
+
+add_library(tensorflow UNKNOWN IMPORTED)
+set_target_properties(tensorflow PROPERTIES IMPORTED_LOCATION ${X10_LIBRARY} 
+	INTERFACE_INCLUDE_DIRECTORIES ${X10_CMAKE_TOOLCHAIN_PATH}/usr/lib/swift/x10/include)
```
compnerd, line 11: Why does `FindX10` define a `tensorflow` library?

PR #828, opened 2020-04-15
x10: port for Darwin

compnerd:
The Xcode 11.3 compiler does objects to the use of the extension to
provide the overload for a std::operator<< in a different namespace.
Move the definition into the `std::` namespace.  Although this is
placing something into a restricted namespace, it is adding an overload
and is permissible.  This allows the streaming of the `grpc::Status`
type to the logging.

Adjust the allocation to use `posix_malloc` for the alignment rather
than the C11 `aligned_alloc` as the latter is unavailable in C++11 mode
and C11 needs to be explicitly opted into.  It should be available as
part of C++17 on macOS, however, tensorflow is still adhering C++11.

### DIFF
--- a/Sources/x10/xla_client/xrt_computation_client.cc
+++ b/Sources/x10/xla_client/xrt_computation_client.cc
@@ -136,6 +136,10 @@ class TensorAllocator : public tensorflow::Allocator {
     void* ptr = _aligned_malloc(
         alloc_blocks->alloc_key.alignment,
         alloc_blocks->alloc_key.alignment + alloc_blocks->alloc_key.num_bytes);
+#elif defined(__APPLE__)
+    void* ptr = nullptr;
+    posix_memalign(&ptr, alloc_blocks->alloc_key.alignment,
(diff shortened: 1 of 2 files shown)


Sources/x10/xla_client/xrt_computation_client.cc
```diff
@@ -136,6 +136,10 @@ class TensorAllocator : public tensorflow::Allocator {
     void* ptr = _aligned_malloc(
         alloc_blocks->alloc_key.alignment,
         alloc_blocks->alloc_key.alignment + alloc_blocks->alloc_key.num_bytes);
+#elif defined(__APPLE__)
```
asuhan, line 139: It'd be better to create a helper for platform-dependent aligned allocation, I think. It becomes hard to read with all these preprocessor noise inlined.

compnerd: Sure, I can try to do that.  Would you mind if I did that in a follow up?

asuhan: No, sounds good.

PR #830, opened 2020-04-16
XLA: extract the allocation traits for Tensors

compnerd:
Extract the operations into a helper traits for the allocator.  This
allows the main path of the allocator be agnostic to the aligned
allocation implementation.


compnerd:
@asuhan not sure if this is what you had in mind.  Did you want this to be more split apart with the allocator traits being specialized per platform instead as is normal and expected?  This avoids ensuring that the signature matches and reduces overall code, but, Im happy to switch to the traditional model if you so prefer.

PR #839, opened 2020-04-17
X10: use `CTensorFlow`

compnerd:
Add the `CTensorFlow` dependency as the X10 module map is merged into
`CTensorFlow`.  Until they are split up, this allows building
`x10_device` again with non-bundled toolchains.


asuhan:
Is this meant to solve the `import x10_device_wrapper` issue?

compnerd:
@asuhan yes, that is precisely the issue this is meant to address :)

PR #847, opened 2020-04-20
Make tensors be on consistent device for bandpart ops

xihui-wu:
This is a bug detected while training GPT2 on TPU with x10. We saw an error complaining non-x10 tensor mixed in the model. It turns out bandPart is the cause which is used by GPT2 [here](https://github.com/tensorflow/swift-models/blob/fe419a9d4b9eda9dffdb384567b2121d83107d3e/Models/Text/GPT2/TransformerLM.swift#L120). 
The PR is to fix it.


pschuh:
We'll need a proper fix for the generated code, but this looks fine.

PR #852, opened 2020-04-20
x10: install the x10 headers to repair CI

compnerd:
The x10 headers should not be distributed.  However, the conflated
tensorflow and x10 modules results in a mixup where the headers are
needed.


compnerd:
Going to abandon in favour of #854

PR #860, opened 2020-04-22
Deprecate `Layer.call(_:)`.

dan-zheng:
`Layer.call(_:)` was added as a temporary shim while [SE-0253](https://github.com/apple/swift-evolution/blob/master/proposals/0253-callable.md) was under review.
`Layer.callAsFunction(_:)` is the permanent name.


saeta:
👍

PR #867, opened 2020-04-23
Honor device for initializer with repeating value

asuhan:


asuhan:
Note that this isn't easy to test since we don't always have any other device than CPU.

Sources/TensorFlow/Initializers.swift
```diff
@@ -39,9 +39,13 @@ extension Tensor {
     repeating repeatedValue: Scalar, shape: TensorShape,
     on device: Device = .default
   ) {
-    self = _Raw.fill(
-      dims: Tensor<Int32>(shape.dimensions.map(Int32.init), on: device),
-      value: Tensor(repeatedValue, on: device))
+    let dims = Tensor<Int32>(shape.dimensions.map(Int32.init), on: device)
+    let value = Tensor(repeatedValue, on: device)
+    #if USING_X10_BACKEND
+      self = _RawXLA.fill(dims: dims, value: value, device: device)
```
pschuh, line 45: This one should switch on device.backend. like: https://github.com/tensorflow/swift-apis/blob/master/Sources/TensorFlow/Core/Tensor.swift#L211

asuhan: Fixed the `_RawXLA.fill` implementation instead.

PR #870, opened 2020-04-23
Fix formatting

asuhan:


Sources/TensorFlow/Core/MixedPrecision.swift
```diff
@@ -206,12 +206,12 @@ extension Tensor {
 #endif
 
 extension Tensor where Scalar: TensorFlowFloatingPoint {
-  @derivative(of: toReducedPrecision)
+  @derivative(of:toReducedPrecision)
```
dan-zheng, line 209: The new formatting for `@derivative(of:)` attributes incorrectly strips whitespace after the colon.

For me, `swift-format` doesn't strip the trailing space:
```console
$ cat derivative.swift
@derivative(of: foo)
func derivative(_ x: Float) -> (value: Float, differential: (Float) -> Float)

$ swift-format derivative.swift
@derivative(of: foo)
func derivative(_ x: Float) -> (value: Float, differential: (Float) -> Float

$ swift-format -v
0.0.1
```

It'd be nice to keep the existing correct formatting if possible.

I wonder if we're using different versions of `swift-format` - it's hard to tell from the `0.0.1` version number. I just upgraded to `2020.04.20RC00`, which seems up-to-date.

asuhan: Why is stripping the whitespace after the colon incorrect?

Regarding `swift-format`, I'm just using the one built from master, which should be the one we're using internally as well.

dan-zheng: > Why is stripping the whitespace after the colon incorrect?

The `of:` in `@derivative(of: foo)` is like an argument label, which should have a trailing space. ([Google Swift Style Guide](https://google.github.io/swift/#horizontal-whitespace))

> Regarding `swift-format`, I'm just using the one built from master, which should be the one we're using internally as well.

Hmm, `swift-format` behaves differently for us, perhaps because we're using different versions.

I'll push a commit fixing `@derivative` attribute formatting (via `swift-format -r -i .`), if you don't mind.

asuhan: That bug looks unrelated, that formatting bug made the code invalid, this one doesn't - just looks different from what you want.

I'm using `swift-format` built from https://github.com/apple/swift-format today, but I now see it behaves differently from whatever Cider is using. Your commit is ok in that case, but it means we still can't tell people to just use `swift-format` until it's consistent with our internal version.

dan-zheng: SR-12002 does actually mention `@derivative(of: foo, wrt: x) -> @derivative(of:foo,wrt:x)` incorrect formatting.

It's a bit surprising that internal and GitHub versions of `swift-format` behave differently. I didn't realize I was locally using an internal version, sorry - I'll try testing the GitHub version.

PR #876, opened 2020-04-23
Test against notf toolchain.

pschuh:


Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM gcr.io/swift-tensorflow/base-deps-cuda10.1-cudnn7-ubuntu18.04
 
 # Allows the caller to specify the toolchain to use.
-ARG swift_tf_url=https://storage.googleapis.com/swift-tensorflow-artifacts/nightlies/latest/swift-tensorflow-DEVELOPMENT-cuda10.1-cudnn7-ubuntu18.04.tar.gz
+ARG swift_tf_url=https://storage.googleapis.com/swift-tensorflow-artifacts/custom-build/latest/swift-tensorflow-DEVELOPMENT-notf-ubuntu18.04.tar.gz
```
marcrasi, line 4: You should use the "notf" toolchain at `https://storage.googleapis.com/swift-tensorflow-artifacts/nightlies/latest/swift-tensorflow-DEVELOPMENT-notf-ubuntu18.04.tar.gz`. It's not there yet, but the first automatically triggered build is almost finished and it should appear there any minute now...

Dockerfile
```diff
@@ -95,3 +80,16 @@ RUN /swift-tensorflow-toolchain/usr/bin/swift build -c release
 
 WORKDIR /open_spiel
 RUN /swift-tensorflow-toolchain/usr/bin/swift test
+
+WORKDIR /swift-apis
+# TODO: move into bash scripts...
+RUN rm -f /swift-tensorflow-toolchain/usr/lib/swift/linux/x86_64/TensorFlow.swiftinterface
+RUN rm -f /swift-tensorflow-toolchain/usr/lib/swift/linux/x86_64/TensorFlow.swiftdoc
+RUN rm -f /swift-tensorflow-toolchain/usr/lib/swift/linux/x86_64/TensorFlow.swiftmodule
+RUN rm -f /swift-tensorflow-toolchain/usr/lib/swift/linux/libswiftTensorFlow.so
```
marcrasi, line 89: Are these rm's still necessary?

pschuh: The command above does an install.

marcrasi: Oh I see, `-D CMAKE_INSTALL_PREFIX=/swift-tensorflow-toolchain/usr` tells it to put the stuff here.

Dockerfile
marcrasi: Ok, it's there now!

Dockerfile
compnerd: @marcrasi yeah, that caught me too.  Perhaps this should get commented?’

pschuh: This is testing that swift-pm works against the installed cmake build. They can go if the swift-pm goes.

PR #877, opened 2020-04-23
Dropout fix: tensors involed in mul should be on same device

xihui-wu:
It complains:
Fatal error: Op must have the same backend type: XLA vs TF_EAGER: file third_party/swift/tensorflow_apis/Sources/TensorFlow/Bindings/RawOpsDispatching.swift, line 18758
otherwise.


texasmichelle:
It looks like this was resolved in #881 👍

xihui-wu:
> It looks like this was resolved in #881 👍

Yep. Closing.

PR #878, opened 2020-04-24
Pick the best device automatically

asuhan:


Sources/x10/swift_bindings/Device.swift
```diff
@@ -132,7 +132,8 @@ public struct Device {
 
   /// The default XLA device.
   public static var defaultXLA: Device {
-    return Device(kind: .CPU, ordinal: 0, backend: .XLA)
+    let defaultDeviceType = getDefaultDeviceType()
+    return Device(kind: defaultDeviceType.kind, ordinal: 0, backend: .XLA)
```
pschuh, line 136: Is there a reason why we're hard-coding ordinal: 0 rather than getting this information fully from c like we had before with :
```
    let cdevice = DefaultDevice()
    return cdevice.device
```

asuhan: I'd have to either duplicate `CDevice` or depend on it from the XRT client, which is a bit tedious. I valued simplicity a bit more in this case, but we can revisit if it ever becomes a problem.

PR #881, opened 2020-04-24
Make Device more explicit.

pschuh:
This should help construct the tensor on the right device for better xla-tensor support.


Sources/TensorFlow/Layers/Core.swift
```diff
@@ -36,7 +36,7 @@ public struct Flatten<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
 @frozen
 public struct Reshape<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
   /// The target shape.
-  @noDerivative public let shape: Tensor<Int32>
+  @noDerivative public var shape: Tensor<Int32>
```
asuhan, line 39: Why was this required?

pschuh: CopyToDevice requires mutation.

PR #882, opened 2020-04-24
Avoid unnecessary tensor materialization in tensor-scalar op derivatives.

dan-zheng:
Add derivatives for `(Tensor, Scalar) -> Tensor` operations differentiable with respect to only the `Tensor` argument, to avoid unnecessary tensor materialization in pullbacks.

## Background

Currently, `(Tensor, Scalar) -> Tensor` operations like `Tensor * Scalar` have one derivative defined with respect to both both `lhs` and `rhs`:

```swift
  @inlinable
  @derivative(of: * wrt: (lhs, rhs))
  static func _vjpMultiply(lhs: Tensor, rhs: Scalar) -> (
    value: Tensor, pullback: (Tensor) -> (Tensor, Scalar)
  ) {
    return (lhs * rhs, { v in (v * rhs, (v * lhs).sum().scalarized()) })
  }
}
```

What if we only care about the derivative with respect to the `lhs` `Tensor`, like in `x * 3`?

Currently, differentiation generates "subset parameter derivative thunks" which discard unused pullback results. The thunk for `_vjpMultiply` wrt only `lhs` (the `Tensor`) calls the user-defined derivative wrt both `lhs` and `rhs`, so it's not avoiding the work of computing the derivative wrt `rhs` (the `Scalar`).

## Idea

This patch adds extra derivatives for `(Tensor, Scalar) -> Tensor` operations differentiable with respect to only the `Tensor` argument:

```swift
  // This derivative for `(lhs: Tensor) * (rhs: Scalar)` copies the one above, but is
  // only with respect to only `lhs`.
  @inlinable
  @derivative(of: *, wrt: lhs)
  static func _vjpMultiply(lhs: Tensor, rhs: Scalar) -> (
    value: Tensor, pullback: (Tensor) -> Tensor
  ) {
    return (lhs * rhs, { v in v * rhs })
  }
```

This avoids unnecessary work to compute the derivative with respect to `Scalar` arguments, which involves calling `Tensor.scalarized()`, thereby triggering X10 tensor materialization.

## Future directions

It would be cool if differentiation could generate "efficient subset parameter derivative thunks" from user-defined derivatives that *do* actually avoid extra work, without affecting usability.

---

@pschuh confirmed that this patch indeed avoids unnecessary materialization.

### DIFF
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -693,6 +693,80 @@ extension Tensor where Scalar: TensorFlowFloatingPoint {
   }
 }
 
+// MARK: Optimized derivatives for tensor-scalar operations
+
+// Note: these derivatives for `(Tensor, Scalar) -> Tensor` binary operations copy ones from above,
+// but are differentiable only with respect to the `Tensor` argument.
+//
+// This avoids unnecessary work to compute the derivative with respect to `Scalar` arguments, which
+// involves calling `Tensor.scalarized()` and thereby triggering X10 tensor materialization.
+
+extension Tensor where Scalar: TensorFlowFloatingPoint {
+  @inlinable
+  @derivative(of: +, wrt: lhs)
+  static func _vjpAdd(lhs: Tensor, rhs: Scalar) -> (
+    value: Tensor, pullback: (Tensor) -> Tensor
+  ) {
+    return (lhs + rhs, { v in v })
+  }
+
+  @inlinable
+  @derivative(of: +, wrt: rhs)
+  static func _vjpAdd(lhs: Scalar, rhs: Tensor) -> (
+    value: Tensor, pullback: (Tensor) -> Tensor
+  ) {
+    return (lhs + rhs, { v in v })
+  }
+
+  @inlinable
+  @derivative(of: -, wrt: lhs)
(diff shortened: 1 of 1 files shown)


Sources/TensorFlow/Operators/Math.swift
```diff
@@ -647,6 +647,18 @@ extension Tensor where Scalar: TensorFlowFloatingPoint {
     return (lhs * rhs, { v in (v * rhs, (v * lhs).sum().scalarized()) })
   }
 
+  /// - Note: this derivative for `(lhs: Tensor) * (rhs: Scalar)` with respect to only `lhs` exists
+  ///   to prevent unnecessary tensor materialization in the pullback. Computing the derivative with
+  ///   respect to `rhs` requires a call to `Tensor.scalarized()`, but that value is discarded when
+  ///   differentiating only with respect to `rhs`.
```
texasmichelle, line 653: `lhs`?

dan-zheng: Fixed, thanks!

dan-zheng:
@pschuh pointed out that this optimization is relevant for all `(Tensor, Scalar) -> Tensor` operations, not just `Tensor * Scalar`. Updated.

texasmichelle:
Many thanks! This unblocks BERT on XLA 🎉

PR #887, opened 2020-04-25
TensorFlow: attempt to repair the Windows build

compnerd:
The type generic math needs to be imported through the use of
`MSVCRT` as that is the equivalent to the `Darwin` or `Glibc` modules.
`ucrt` is the pure C library without any additions.

### DIFF
--- a/Sources/TensorFlow/Layers/Dropout.swift
+++ b/Sources/TensorFlow/Layers/Dropout.swift
@@ -12,6 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#if os(Windows)
+import func MSVCRT.sqrt
+#endif
+
 extension Tensor where Scalar: TensorFlowFloatingPoint {
   /// Computes dropout given a probability.
   // TODO: Remove the underscore once `droppingOut(probability:)` has been removed.


compnerd:
CC: @saeta @vballoli

compnerd:
Windows Run: https://dev.azure.com/compnerd/swift-build/_build/results?buildId=30744&view=results

compnerd:
The Windows run was clean.

saeta:
Thanks for the speedy fix @compnerd!  I think I'm the medium term, it's going to be important to ensure we don't need conditional imports for Windows or Android for these sorts of common operations. Do you think you can help put together a list of these issues as they come up so we can get a decent survey of them over time?

compnerd:
@saeta that’s going to be required.  Anytime you use the underlying system’s libc there’s going to be conditional imports.  Furthermore, sqrt is definitely in different modules per platform in the swift implementation (AFIACT intentionally to prevent circular dependencies).  This isn’t a special thing about Windows, it’s how swift chose to implement the core libraries (Darwin, MSVCRT, Glibc depend on swiftCore, which cannot depend on them).

One option would be to implement a new module System that exposes the entire C library interface (import C) and add a second module that exposes the POSIX interfaces (import POSIX).  A third module that should be implemented would implement the support for thread local storage and threading (import Threading).  That’s been something which has been brought up a few times, but no one has spent the time to actually present what such a surface would look like.

PR #888, opened 2020-04-26
NFC: clean up `Sequential` gyb.

dan-zheng:
Refactor `Sequential` gyb logic into helper functions for clarity.

---

This was a step towards gyb'ing `Sequential{n}` typealiases, which actually doesn't work due to [generic typealias limitations](https://github.com/apple/swift-evolution/blob/master/proposals/0048-generic-typealias.md#detail-design).

### DIFF
--- a/Sources/TensorFlow/Layers/Sequential.swift
+++ b/Sources/TensorFlow/Layers/Sequential.swift
@@ -201,8 +201,7 @@ public struct LayerBuilder {
     -> Sequential<
       L1,
       Sequential<
-        L2,
-        Sequential<L3, Sequential<L4, Sequential<L5, Sequential<L6, Sequential<L7, L8>>>>>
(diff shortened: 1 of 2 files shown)


compnerd:
I think that we should be working to ween ourselves off of gyb.  Although this change doesn't look like it shouldn't break anything, but Id like to understand the broader plans for how this would enable something like that.

dan-zheng:
That's a fair question. I think `gyb` is not the nicest, but ultimately useful for avoiding boilerplate.

There are other `gyb` files too - whether or not to de-`gyb` in general seems like a team decision. In the meantime, I think cleaning up `gyb` files is acceptable.

dan-zheng:
I'll land this now to unblock a draft PR.

We can revisit whether to de-`gyb` later - for `Sequential.swift`, it would be as easy as deleting `Sequential.swift.gyb`. I currently prefer "`gyb` -> `.swift` without `#sourceLocation` line directives" over maintaining boilerplate code.

PR #905, opened 2020-04-28
Revert SR-10697 workaround for  in swift-apis TF-507

ematejska:
This implements https://bugs.swift.org/browse/TF-507

### DIFF
--- a/Tests/TensorFlowTests/LayerTests.swift
+++ b/Tests/TensorFlowTests/LayerTests.swift
@@ -1373,14 +1373,10 @@ final class LayerTests: XCTestCase {
     var cell = BasicRNNCell<Float>(inputSize: 2, hiddenSize: 5)
     cell.weight = weight
     cell.bias = bias
-    let state = BasicRNNCell.State(
-      Tensor<Float>(ones: [1, 5]) * Tensor<Float>([1, 0.2, 0.5, 2, 0.6])
-    )
+    let state = Tensor<Float>(ones: [1, 5]) * Tensor<Float>([1, 0.2, 0.5, 2, 0.6])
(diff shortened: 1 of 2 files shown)


Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -1373,14 +1373,10 @@ final class LayerTests: XCTestCase {
...
-    let state = BasicRNNCell.State(
-      Tensor<Float>(ones: [1, 5]) * Tensor<Float>([1, 0.2, 0.5, 2, 0.6])
-    )
+    let state = Tensor<Float>(ones: [1, 5]) * Tensor<Float>([1, 0.2, 0.5, 2, 0.6])
     let input = Tensor<Float>(ones: [1, 2]) * Tensor<Float>([0.3, 0.7])
     let output = cell(input: input, state: state).state
-    let expected = BasicRNNCell.State(
-      Tensor<Float>([[0.9921227, 0.9999934, 0.9921227, 0.9999934, 0.9921227]])
-    )
+    let expected = Tensor<Float>([[0.9921227, 0.9999934, 0.9921227, 0.9999934, 0.9921227]])
```
dan-zheng, line 1379: It seems like these expected values stayed the same, which is nice as a mini correctness test.

dan-zheng:
This is a source-breaking change. I don't think it's possible to be backwards compatible, but that's okay since this is an obvious improvement that any users would be happy about.

Let's merge if this makes sense to others, too!

ematejska:
@saeta What do you think?

dan-zheng:
I can help merge with an informative commit description now!
I'll explain why reverting SR-10697 is now possible.

PR #906, opened 2020-04-29
Revert "Revert SR-10697 workaround (#905)"

asuhan:
This reverts commit 5b24a71b3978717a95d07eeb597d2e63ffeb30e1.


asuhan:
This breaks in Piper, revert for now. We can revisit this after the release.

asuhan:
@dan-zheng It's not just that test, I only posted one failure. More tests fail for similar reasons.

dan-zheng:
> @dan-zheng It's not just that test, I only posted one failure. More tests fail for similar reasons.

Thanks, updated my comment. Do you know why the failures weren't caught by Kokoro?

asuhan:
@dan-zheng Because Kokoro doesn't test anything in Piper.

ematejska:
Thanks for reverting?  I'll sync up offline about how to test/fix in Piper.

PR #907, opened 2020-04-29
Log new graphs and shape changes

asuhan:
Add ability to log new graphs and shape changes, with stack trace information. This should help a lot with tracing the cause of dynamic shapes.

### DIFF
--- a/Sources/x10/xla_tensor/ir_dump_util.cpp
+++ b/Sources/x10/xla_tensor/ir_dump_util.cpp
@@ -20,10 +20,11 @@
 
 #include "absl/container/node_hash_map.h"
 #include "absl/types/optional.h"
-#include "tensorflow/compiler/xla/xla_client/debug_macros.h"
-#include "tensorflow/compiler/xla/xla_client/xla_util.h"
+#include "tensorflow/compiler/tf2xla/xla_tensor/debug_util.h"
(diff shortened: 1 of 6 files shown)


Sources/x10/xla_tensor/ir_dump_util.cpp
```diff
@@ -262,5 +279,48 @@ std::string DumpUtil::ToHlo(absl::Span<const Value> values) {
   return ConsumeValue(xla::util::GetComputationHloText(computation));
 }
 
+std::string DumpUtil::GetGraphChangeLog(absl::Span<const Node* const> roots) {
+  auto post_order = Util::ComputePostOrder(roots);
+  std::unordered_map<const Node*, size_t> roots_ids = GetRootsIds(roots);
+  NodeIdMap id_map = GenerateIdMap(post_order);
+  std::vector<ChangeLogNode> change_log;
+  xla::hash_t h = 0x85ebca77c2b2ae63;
```
pschuh, line 287: Does this need to stay the same as in: x10/xla_client/util.h ?

asuhan: Not really, can be anything, which is why I didn't make an effort to make a header for it.

PR #910, opened 2020-04-30
NonuniformInferenceBatches, as a type.

dabrahams:
The alternative to the typealias and the protocol hack would be to write a
wrapper around the type we've defined and make it conform to Collection, etc.


sgugger:
@googlebot I fixed it (let's try if this works by any chance.)

PR #911, opened 2020-04-30
Replace std::unordered_map in OpByOpExecutor

asuhan:
This is required by the internal style guide.


compnerd:
Could you please add some context for the change to the commit message?

PR #912, opened 2020-04-30
Create swift.yml

compnerd:
Add an initial GitHub workflow to enable a `swift-format` linting action.

### DIFF
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -31,9 +31,10 @@ jobs:
           cd ..
 
       - name: lint
-        uses: compnerd/lint-action@swift-format
+        uses: compnerd/lint-action@s4tf
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
+          auto_fix: true
           appleswiftformat: true
 
   # build:
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -1,0 +1,49 @@
+name: Swift
+
+on:
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  lint:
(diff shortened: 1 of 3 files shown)


compnerd:
@googlebot I fixed it.

Sources/TensorFlow/Core/MixedPrecision.swift
```diff
@@ -157,7 +157,7 @@ extension Tensor {
   }
 
   /// Promotes a scalar to a tensor with the same device and precision as the given tensor.
-  @differentiable(where Scalar: TensorFlowFloatingPoint)
+  @differentiable( where Scalar: TensorFlowFloatingPoint)
```
saeta, line 160: Note: seems like it's mis-formatting this at the moment. (I think you're already aware, but wanted to make sure to flag them here.)

Sources/TensorFlow/Core/MixedPrecision.swift
```diff
@@ -206,12 +206,12 @@ extension Tensor {
 #endif
 
 extension Tensor where Scalar: TensorFlowFloatingPoint {
-  @derivative(of: toReducedPrecision)
+  @derivative(of:toReducedPrecision)
```
saeta, line 209: And here.

Sources/TensorFlow/Core/MixedPrecision.swift
compnerd: Yeah, @allevato is going to be looking into backporting the fixes for that to the 5.2 release so we can get these correct.

compnerd:
Superseded by #931

PR #915, opened 2020-04-30
swift-format: update configuration to current cnfiguration for swift-…

compnerd:
…format

The file is the output of `swift-format -m dump-configuration` from the
swift-5.2-branch of swift-format.

### DIFF
--- a/.swift-format
+++ b/.swift-format
@@ -1,14 +1,53 @@
 {
-    "version": 1,
-    "lineLength": 100,
-    "indentation": {
-        "spaces": 2
-    },
-    "maximumBlankLines": 1,
-    "respectsExistingLineBreaks": true,
-    "blankLineBetweenMembers": {
-        "ignoreSingleLineProperties": true
-    },
-    "lineBreakBeforeControlFlowKeywords": false,
-    "lineBreakBeforeEachArgument": false
+  "fileScopedDeclarationPrivacy" : {
+    "accessLevel" : "private"
+  },
+  "indentation" : {
+    "spaces" : 2
+  },
+  "indentConditionalCompilationBlocks" : true,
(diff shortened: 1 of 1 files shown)


.swift-format
```diff
@@ -1,14 +1,53 @@
...
+  "indentConditionalCompilationBlocks" : true,
+  "lineBreakAroundMultilineExpressionChainComponents" : false,
+  "lineBreakBeforeControlFlowKeywords" : false,
+  "lineBreakBeforeEachArgument" : false,
+  "lineBreakBeforeEachGenericRequirement" : false,
+  "lineLength" : 100,
+  "maximumBlankLines" : 1,
+  "prioritizeKeepingFunctionOutputTogether" : false,
+  "respectsExistingLineBreaks" : true,
+  "rules" : {
```
saeta, line 17: Did we want to start by setting these all to `false`, and then selectively enabling them once there are no warnings? Or am I missing something?

compnerd: I was going to go in the opposite direction: start from the default configuration and then switch them to what we want so that we can at least have some explanation of why we chose to change the configuration (since we cannot have comments in the configuration itself).

dan-zheng: @compnerd: do you want to be explicit about all configuration options, instead of inheriting defaults and defining just the "customized" options that we want?

I think there are points in favor of the latter (the file is smaller, there's no information noise because only the necessary options are defined). I also think you tend to prefer explicitness 🙂

compnerd: Quite true, I do prefer the explicitness.  The benefit here would be although this is a longer file (which honestly doesn't matter too much), it makes it easy to identify the setting and the value we are using and ensures that future changes to the tool's defaults to do not alter the behaviour.

PR #916, opened 2020-04-30
correct whitespace in `Epochs` implementation

compnerd:
While trying to enable `swift-format` based linting, these files were
noted to have whitespace errors.  Correct the errors to allow enabling
the linter.


compnerd:
Manually had to do it for now ... trying to split off the changes so that enabling the auto-application of the lint/formatting is done once 5.2 has the needed change backported.

PR #917, opened 2020-04-30
Make PrintX10Metrics available

asuhan:


Sources/x10/swift_bindings/XLATensor.swift
```diff
@@ -1069,3 +1069,7 @@ extension XLATensor {
     }
   }
 }
+
+public func PrintX10Metrics() {
+  x10_xla_tensor_wrapper.PrintMetrics()
```
compnerd, line 1074: I think that you can get away with `PrintMetrics()` but I may be mistaken.

compnerd:
Oh, and did you ensure that `PrintMetrics` is properly annotated for export on Windows from the C++ side?

Sources/x10/swift_bindings/XLATensor.swift
asuhan: Done.

asuhan:
@compnerd It has `XLA_API` already, do we need anything else?
@saeta I don't know if we can test this effectively from this repository, the main concern is visibility. The ultimate test would be to call it from swift-models, I think.

compnerd:
`XLA_API` is exactly what is needed, if it has it already, then nothing to do here, move along :)

PR #924, opened 2020-05-01
TensorFlow: coddle some colons (NFC)

compnerd:
Adjust style as per swift-format.


compnerd:
Yeah, I suppose I can batch them up, though there are a different set of issues and we are nearly through them all :) (these are caught by swift-format)

PR #929, opened 2020-05-04
Make derivatives of public functions be `@usableFromInline`.

dan-zheng:
Friend PR: https://github.com/apple/swift/pull/31527

> Require `@derivative` functions and their original functions to have the same access level.
> Public original functions may also have internal `@usableFromInline` derivatives, as a special case. Diagnose access level mismatches.
> 
> This simplifies derivative registration rules, and may enable simplification of AutoDiff symbol linkage rules.

### DIFF
--- a/Sources/TensorFlow/Core/MixedPrecision.swift
+++ b/Sources/TensorFlow/Core/MixedPrecision.swift
@@ -206,11 +206,13 @@ extension Tensor {
 #endif
 
 extension Tensor where Scalar: TensorFlowFloatingPoint {
+  @usableFromInline
   @derivative(of: toReducedPrecision)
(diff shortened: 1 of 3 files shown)


dan-zheng:
I built a toolchain from https://github.com/apple/swift/pull/31527 (requiring same original/derivative access levels), and confirmed that [tensorflow/swift-apis](https://github.com/tensorflow/swift-apis) builds after this patch.

[tensorflow/swift-models](https://github.com/tensorflow/swift-models) builds without any patches.

dan-zheng:
It's fine to merge this patch now before https://github.com/apple/swift/pull/31527 lands.

PR #937, opened 2020-05-06
Version of #747 that can be swift-format'ed

dabrahams:
See #747


dabrahams:
@googlebot I fixed it

dabrahams:
@googlebot I fixed it

dabrahams:
Landed #747 so this is unneeded.

Issue #943, opened 2020-05-08
Bidirectional RNN


kongzii:
This is follow up to discussion https://groups.google.com/a/tensorflow.org/forum/#!topic/swift/xuspcIrHLoQ with @dan-zheng .

My current (naive) implementation looks like this:

```swift
public struct BiLSTM: Layer {
    @noDerivative 
    public let activation: Activation

    var forward, backward: RNN<LSTMCell<Float>>

    public typealias Activation = @differentiable(Tensor<Float>) -> Tensor<Float>

    public init(inputSize: Int, hiddenSize: Int, activation: @escaping Activation = identity) {
        precondition(hiddenSize % 2 == 0)
        let hiddenSizePerLSTM = hiddenSize / 2

        self.activation = activation
        forward = RNN(LSTMCell<Float>(inputSize: inputSize, hiddenSize: hiddenSizePerLSTM))
        backward = RNN(LSTMCell<Float>(inputSize: inputSize, hiddenSize: hiddenSizePerLSTM))
    }

    @differentiable
    public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
        let forwardSequence = input.unstacked(alongAxis: 1)
        let backwardSequence = input.reversed(inAxes: 1).unstacked(alongAxis: 1)

        let forwardCells = forward(forwardSequence)
        let backwardCells = backward(backwardSequence)

        return activation(
            forwardCells[input.shape[1] - 1].hidden
                .concatenated(with: backwardCells[input.shape[1] - 1].hidden, alongAxis: 1)
        )
    }
}
```

So basically like dan said, layer is storing two LSTMs and forwarding input through both of them. I believe this is the way how bidirectional LSTMS are done, but please correct me if there is a better alternative.

I read many ways about what to do with outputs from two LSTM: concatenate like above, sum, multiplication, difference. I wonder if it is better to provide only one or allows users to choose between modes at layer init?

Also, it would be probably best to do this in a more universal way, so all RNN layers will be able to be bidirectional?

Could somebody please give me some hints to kick-start my implementation in swift-apis way?

Thanks.

dan-zheng:
Thank you Peter for creating this issue!



> Also, it would be probably best to do this in a more universal way, so all RNN layers will be able to be bidirectional?

kongzii:
@dan-zheng Thanks! Could you please assign me the issue? It seems that I can not do it as a Contributor.

kongzii:
[...]

PR #946, opened 2020-05-09
Bidirectional RNN

kongzii:
As discussed in https://github.com/tensorflow/swift-apis/issues/943 this is a draft for `BidirectionalRecurrentLayer`. I will add other merge modes and support for LSTMCell and GRUCell once we agree on the implementation.

Because `State` of various `RecurrentLayerCell` are implemented differently, I could not create one universal merge method. Initially, I tried to follow @dan-zheng advice. 

> initializer taking a "merge mode" function

and I tried to implement it like this:

```swift
// Merge functions
@differentiable
public func sum<Scalar: TensorFlowFloatingPoint>(
  _ a: BasicRNNCell<Scalar>.TimeStepOutput, 
  _ b: BasicRNNCell<Scalar>.TimeStepOutput
) -> BasicRNNCell<Scalar>.TimeStepOutput {
  return BasicRNNCell<Scalar>.TimeStepOutput(a.value + b.value)
}

@differentiable
public func sum<Scalar: TensorFlowFloatingPoint>(
  _ a: LSTMCell<Scalar>.TimeStepOutput,
  _ b: LSTMCell<Scalar>.TimeStepOutput
) -> LSTMCell<Scalar>.TimeStepOutput {
  return LSTMCell<Scalar>.TimeStepOutput(cell: a.cell + b.cell, hidden: a.hidden + b.hidden)
}

@differentiable
public func sum<Scalar: TensorFlowFloatingPoint>(
  _ a: GRUCell<Scalar>.TimeStepOutput,
  _ b: GRUCell<Scalar>.TimeStepOutput
) -> GRUCell<Scalar>.TimeStepOutput {
  return GRUCell<Scalar>.TimeStepOutput(hidden: a.hidden + b.hidden)
}

public struct BidirectionalRecurrentLayer<Cell: RecurrentLayerCell>: Layer {
  ...
  public typealias Merge = @differentiable (Cell.TimeStepOutput, Cell.TimeStepOutput) -> Cell.TimeStepOutput
  @noDerivative public let merging: Merge
  public init(_ cell: @autoclosure () -> Cell, merge: @escaping Merge) {
    forward = RecurrentLayer(cell())
    backward = RecurrentLayer(cell())

    // Exited with signal code 11 due to this
    merging = merge
  }
  ...
}
```

the build was successful but when I tried to initialize it in tests:

```swift
BiBasicRNN<Float>(BasicRNNCell(inputSize: 4, hiddenSize: 4, seed: (0xFeed, 0xBeef)), merge: sum)
```

It crashed at runtime with `Exited with signal code 11`.  Perhaps there is just some bug in the code?

In this implementation, I created `Mergable` protocol, now if `State` of  `RecurrentLayerCell` comforts to this protocol, it can be used in `BidirectionalRecurrentLayer` and merge mode is distinguished by an enum.

I am open to any advice and will happily edit it, thanks for reviews :)


Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -444,10 +451,61 @@ public struct RecurrentLayer<Cell: RecurrentLayerCell>: Layer {
...
+public struct BidirectionalRecurrentLayer<Cell: RecurrentLayerCell>: Layer 
+where Cell.TimeStepOutput: Mergable {
+  public typealias Input = [Cell.TimeStepInput]
+  public typealias Output = [Cell.TimeStepOutput]
+  public typealias Merge = @differentiable (Cell.TimeStepOutput, Cell.TimeStepOutput) -> Cell.TimeStepOutput
+
+  @noDerivative public let mergeMode: MergeMode
+  public var forward, backward: RecurrentLayer<Cell>
+
+  public init(_ cell: @autoclosure () -> Cell, mergeMode: MergeMode = .sum) {
```
dan-zheng, line 473: Rather than using a `MergeMode` enum, how about directly taking merge mode as a `@differentiable` closure instead? Closures are more flexible than enums, which are not extensible.
```suggestion
  public init(_ cell: @autoclosure () -> Cell, mergeMode: @differentiable (Cell.TimeStepOutput, Cell.TimeStepOutput) -> Cell.TimeStepOutput = sum) {
```
This should also allow you to remove `protocol Mergable`.

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -1520,6 +1520,24 @@ final class LayerTests: XCTestCase {
...
+        Tensor<Float>(shape: TensorShape([4, 4]), scalars: [
+          0.20775771, 0.20080023, -0.13768704, -0.18534681,
+          0.22666009, 0.30019346, -0.19720285, -0.14683801,
+          0.23758979, 0.32101023, -0.20359215, -0.1787096,
+          0.24337786, 0.3389194, -0.21143384, -0.1675081,
+        ].map { $0 + $0 }),
+        accuracy: 1e-6)
+    }
+  }
+
```
dan-zheng, line 1540: I wonder if you've tested bidirectional recurrent layer correctness using a Python reference implementation?

We have an [existing `lstm.py` script](https://github.com/tensorflow/swift-apis/blob/master/Utilities/ReferenceImplementations/lstm.py) for testing LSTM correctness versus Keras. Could you please copy it to `bidirectional_rnn.py` and adapt it to implement the same `Bidirectional` program as the one here in Swift?

Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -444,10 +451,61 @@ public struct RecurrentLayer<Cell: RecurrentLayerCell>: Layer {
...
+  public func lastOutput(_ inputs: Input) -> Cell.TimeStepOutput {
+    self(inputs)[withoutDerivative(at: inputs.count - 1)]
+  }
+}
+
 extension RecurrentLayer: Equatable where Cell: Equatable {}
 extension RecurrentLayer: AdditiveArithmetic where Cell: AdditiveArithmetic {}
 
 public typealias BasicRNN<Scalar: TensorFlowFloatingPoint> = RecurrentLayer<BasicRNNCell<Scalar>>
+public typealias BiBasicRNN<Scalar: TensorFlowFloatingPoint> = BidirectionalRecurrentLayer<BasicRNNCell<Scalar>>
```
dan-zheng, line 508: The Swift API Design Guidelines emphasize clear names rather than short ones: [clarity over brevity](https://swift.org/documentation/api-design-guidelines/#clarity-over-brevity).

How about naming typealiases as `BidirectionalXXX`? I think it's very clear and reasonably short.
```swift
public typealias BidirectionalBasicRNN<Scalar: TensorFlowFloatingPoint> = BidirectionalRecurrentLayer<BasicRNNCell<Scalar>>
public typealias BidirectionalLSTM<Scalar: TensorFlowFloatingPoint> = BidirectionalRecurrentLayer<LSTMCell<Scalar>>
public typealias BidirectionalGRU<Scalar: TensorFlowFloatingPoint> = BidirectionalRecurrentLayer<GRUCell <Scalar>>
```

Sources/TensorFlow/Layers/Recurrent.swift
kongzii: Yes, I wanted to, but I hit the issue described above, and I wanted to do a PR with working tests. I will try it again, but if I will not be able to fix it, can I do a PR with a runtime error and perhaps you will be able to correct me?

dan-zheng: Ah, I see. Yes, I can help take a look if you share a branch reproducing the runtime error.

Tests/TensorFlowTests/LayerTests.swift
kongzii: I did not, but this code is from `testRNN`, only difference is `.map { $0 + $0 }` because merge mode is `sum`. 
I will check the script and adapt it 👍

Sources/TensorFlow/Layers/Recurrent.swift
kongzii: Yes, renamed.

Sources/TensorFlow/Layers/Recurrent.swift
kongzii: I updated PR. I think the issue is that Swift can not take right `sum` function? For example if I put default function to the init: `merge: @escaping MergeFunction = sum` it crashes on 

```swift
/Users/peter.jung/Projects/swift-apis/Sources/TensorFlow/Layers/Recurrent.swift:484:81: error: no exact matches in reference to global function 'sum'
  public init(_ cell: @autoclosure () -> Cell, merge: @escaping MergeFunction = sum) {
                                                                                ^
/Users/peter.jung/Projects/swift-apis/Sources/TensorFlow/Layers/Recurrent.swift:450:13: note: found this candidate
public func sum<Scalar: TensorFlowFloatingPoint>(
            ^
/Users/peter.jung/Projects/swift-apis/Sources/TensorFlow/Layers/Recurrent.swift:458:13: note: found this candidate
public func sum<Scalar: TensorFlowFloatingPoint>(
            ^
/Users/peter.jung/Projects/swift-apis/Sources/TensorFlow/Layers/Recurrent.swift:469:13: note: found this candidate
public func sum<Scalar: TensorFlowFloatingPoint>(
            ^
[1/3] Compiling TensorFlow Recurrent.swift
```

dan-zheng: Ah, I see. You've added three concrete overloads of `sum` for various `RecurrentLayerCell.TimeStepOutput` types:

<details>
<p>

```swift
@differentiable
public func sum<Scalar: TensorFlowFloatingPoint>(
  _ firstOutput: BasicRNNCell<Scalar>.TimeStepOutput, 
  _ secondOutput: BasicRNNCell<Scalar>.TimeStepOutput
) -> BasicRNNCell<Scalar>.TimeStepOutput {
  return BasicRNNCell<Scalar>.TimeStepOutput(firstOutput.value + secondOutput.value)
}

@differentiable
public func sum<Scalar: TensorFlowFloatingPoint>(
  _ firstOutput: LSTMCell<Scalar>.TimeStepOutput,
  _ secondOutput: LSTMCell<Scalar>.TimeStepOutput
) -> LSTMCell<Scalar>.TimeStepOutput {
  return LSTMCell<Scalar>.TimeStepOutput(
    cell: firstOutput.cell + secondOutput.cell, 
    hidden: firstOutput.hidden + secondOutput.hidden
  )
}

@differentiable
public func sum<Scalar: TensorFlowFloatingPoint>(
  _ firstOutput: GRUCell<Scalar>.TimeStepOutput,
  _ secondOutput: GRUCell<Scalar>.TimeStepOutput
) -> GRUCell<Scalar>.TimeStepOutput {
  return GRUCell<Scalar>.TimeStepOutput(hidden: firstOutput.hidden + secondOutput.hidden)
}
```

</p>
</details>

However, you need a generic "sum" function in order pass type-checking - `RecurrentLayerCell.TimeStepOutput` must be constrained to a protocol that defines a "sum" function.

Previously, you had a `protocol Mergable` representing this. However, looking at the implementations of `TimeStepOutput.sum`, the function is really just performing memberwise  `AdditiveArithmetic.+`.

---

Can you try adding a constraint:
```swift
public struct BidirectionalRecurrentLayer<Cell: RecurrentLayerCell>: Layer
where Cell.TimeStepOutput: AdditiveArithmetic {
```

And using `+` as the default parameter value for `mergeMode`:
```swift
  public init(_ cell: @autoclosure () -> Cell, mergeMode: @differentiable (Cell.TimeStepOutput, Cell.TimeStepOutput) -> Cell.TimeStepOutput = +) {
```

dan-zheng: If we want to extend "merge mode" to support other functions like concatenation, which aren't defined as requirements of any pre-existing protocol, we will need to add a new protocol defining all "merge functions" as requirements.

To make incremental progress, we can look into extending "merge mode" in a follow-up PR! Using and testing just `+` for now is totally reasonable.

Tests/TensorFlowTests/LayerTests.swift
kongzii: I am making progress on tests, all seems to be ok except that the last row of `gradInputs` does not correspond to the last row of `expectedGradX`.  I am trying to figure out why now.

Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -444,12 +461,99 @@ public struct RecurrentLayer<Cell: RecurrentLayerCell>: Layer {
...
+  static func +(lhs: Self, rhs: Self) -> Self
+}
+
+@differentiable
+public func sum<TimeStepOutput: Mergeable>(
+  _ first: TimeStepOutput, 
+  _ second: TimeStepOutput
+) -> TimeStepOutput {
+  return first + second
+}
```
dan-zheng, line 476: Rather than adding a `Mergeable` protocol in this PR, how about conforming `RNNCell.State` types to `AdditiveArithmetic` instead, and using `AdditiveArithmetic.+` instead of a new `func sum`?

Context here: https://github.com/tensorflow/swift-apis/pull/946#discussion_r422509815

My main concern is that a new public protocol `Mergeable` deserves more motivation and review, and it's not quite justified yet in this PR since `sum` is the only merge function. I would prefer if you investigated `Mergeable` and more merge functions in a follow-up PR, to make changes incremental!

dan-zheng: Oh, I think I see an issue: protocol `Mergeable` is necessary because it requires that `static func +` has a `@differentiable` attribute, unlike `AdditiveArithmetic`.

If that is the case, keeping protocol `Mergeable` in this PR is fine (because it's unavoidable). However, could you try deleting `public func sum` and replacing it with `+` (from `Mergeable`)?

```swift
  public init(_ cell: @autoclosure () -> Cell, merge: @escaping MergeFunction = (+)) {
```

Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -444,12 +461,99 @@ public struct RecurrentLayer<Cell: RecurrentLayerCell>: Layer {
   }
 }
 
+public protocol Mergeable 
+where Self: Differentiable {
```
dan-zheng, line 465:
```suggestion
public protocol Mergeable: Differentiable, AdditiveArithmetic {
```

Sources/TensorFlow/Layers/Recurrent.swift
kongzii: Yes, I agree. Sorry for the misunderstanding. I was trying to fix (https://github.com/tensorflow/swift-apis/pull/946/files#diff-85b8b5016877b6312b0c89c0862c5b56R484)

```
  // TODO: Runtime crash when I try to set `sum` function to property
  // @noDerivative public let merge: MergeFunction
```

so I created a new protocol to test if it helps but it didn't. I was also not able to do

```
public init(_ cell: @autoclosure () -> Cell, mergeMode: @differentiable (Cell.TimeStepOutput, Cell.TimeStepOutput) -> Cell.TimeStepOutput = +
```

because of 

```
error: expected initial value after '='
```

I will fix it after I figure out tests.
Just wanted to share progress here, so that PR does not look abandoned.

dan-zheng: Thanks for the context! No worries about "looking abandoned", feel free to make progress at your own pace.

kongzii: Oh, so you figured out the error before I sent a reply. Thanks!

kongzii:
```
note: differentiated functions in '@inlinable' functions must be marked '@differentiable' or have a public '@derivative'; this is not possible with a closure, make a top-level function instead
  public init(_ cell: @autoclosure () -> Cell, merge: @escaping MergeFunction = (+))
```

Perhaps I am just doing something wrong, but are you sure  that this should work? 🤔

dan-zheng: Ah, I understand this error: the expression `(+)` creates a thunk, triggering the error.
In that case, please continue to use the top-level `func sum<T: Mergeable>` function.

I'll discuss this issue with @marcrasi, who added the error, and see if it's still relevant in light of recent differentiation-related changes.

Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -444,12 +461,98 @@ public struct RecurrentLayer<Cell: RecurrentLayerCell>: Layer {
   }
 }
 
+public protocol Mergeable: Differentiable, AdditiveArithmetic {
+  @differentiable
+  static func +(lhs: Self, rhs: Self) -> Self
+}
+
+@differentiable
+public func sum<TimeStepOutput: Mergeable>(
```
dan-zheng, line 470: This `sum` function is generic over `Mergeable`, but nothing should be specifically named `TimeStepOutput`.
```suggestion
public func sum<T: Mergeable>(
```

Tests/TensorFlowTests/LayerTests.swift
kongzii: @dan-zheng I finally get the tests working 🎉 But I needed to do some changes (I got inspiration from Keras implementation), could you please check last commit and give me feedback? Thank you.

kongzii:
Could someone help me, please? Implementation seems to be OK, all tests against Keras are passing (reversed rnn/lstm/gru, bidirectional rnn/lstm/gru).
But I have a problem with the following code:

```
// TODO: Runtime crash when I try to set `sum` function to property
  @noDerivative public let merge: MergeFunction
  public var forward, backward: RecurrentLayer<Cell>

  public init(_ cell: @autoclosure () -> Cell, merge: @escaping MergeFunction = sum) {
    forward = RecurrentLayer(cell())
    backward = RecurrentLayer(cell(), backwardDirection: true)
    // TODO: Runtime crash when I try to set `sum` function to property
    self.merge = merge
  }
```

It compiles OK, but it will crash when executed with `swift test` with the error:

```
Exited with signal code 11
```

If I comment it out:

```
// TODO: Runtime crash when I try to set `sum` function to property
  // @noDerivative public let merge: MergeFunction
  public var forward, backward: RecurrentLayer<Cell>

  public init(_ cell: @autoclosure () -> Cell, merge: @escaping MergeFunction = sum) {
    forward = RecurrentLayer(cell())
    backward = RecurrentLayer(cell(), backwardDirection: true)
    // TODO: Runtime crash when I try to set `sum` function to property
    // self.merge = merge
  }
```

it works. 

MergeFunction is type alias for:

```
public typealias MergeFunction = @differentiable (Cell.TimeStepOutput, Cell.TimeStepOutput) -> Cell.TimeStepOutput
```

Thanks for any hints.

dan-zheng:
> It compiles OK, but it will crash when executed with `swift test` with the error:
> 
> ```
> Exited with signal code 11
> ```

I reproduced the runtime crash. It seems similar to TF-1122: crash regarding `@differentiable` function-typed stored properties.

I'll try to look into the issue soon. In the meantime, I believe we can workaround it by using a private class wrapping the `@differentiable` function stored property ([done here](https://github.com/sgugger/swift-training/blob/8c596fe128eb9837e7777ddfb5c48c6a595cb859/DifferentiableStep/TrainingLoop.swift#L4-L10)). I'll look into that now.

kongzii:
@dan-zheng Thank you for your help! :) However, there are still some errors I do not know how to deal with :-(.

This results in the compile error:

```swift
var rnn = BidirectionalBasicRNN<Float>(
  BasicRNNCell(inputSize: 4, hiddenSize: 4),
  mergeFunction: sum
)
```

The error is:

```
Undefined symbols for architecture x86_64:
  "_$s15TensorFlowTests09_AD__$s10A57Flow0A0VySfGA2DIegnnr_A3DIegggo_TR_bb0__PB__src_0_wrt_0_1VMn", referenced from:
      _symbolic _____ 15TensorFlowTests09_AD__$s10A57Flow0A0VySfGA2DIegnnr_A3DIegggo_TR_bb0__PB__src_0_wrt_0_1V in LayerTests.swift.o
ld: symbol(s) not found for architecture x86_64
[2/3] Linking TensorFlowPackageTests
```

But this works OK:

```swift
var rnn = BidirectionalBasicRNN<Float>(
  BasicRNNCell(inputSize: 4, hiddenSize: 4),
  mergeFunction: BasicRNNCell<Float>.State.sum
)
```

The second problem is, that after merge with the latest master if I do this:

```swift
extension Tensor: Mergeable where Scalar: TensorFlowFloatingPoint {}
```

It results in a compile-time error. I saved stack dump at https://pastebin.com/PxZTir6F. For now, I solved it by changing Mergeable `+` into `sum`:

```swift
public protocol Mergeable: Differentiable, AdditiveArithmetic {
  /// - Note: Trying to use `+` instead of `sum` results in compile error,
  ///   when Tensor is extended for Mergeable comforting.
  ///   Stack dump is stored at https://pastebin.com/PxZTir6F.
  @differentiable
  static func sum(_ lhs: Self, _ rhs: Self) -> Self
}
```

But I am not sure that it is an acceptable solution.

kongzii:
@dan-zheng  Any update, please? :)

dan-zheng:
Sorry for the late reply. Thanks for sharing the compiler crash details, I'll look into them this week.

ematejska:
@dan-zheng is still working on the compiler crashers.

kongzii:
> @dan-zheng is still working on the compiler crashers.

Thanks for the update, I can imagine that it's not simple.

Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -415,17 +426,146 @@ public struct RecurrentLayer<Cell: RecurrentLayerCell>: Layer {
...
+  ///   stored properties (TF-1122).
+  private class _MergeFunction {
+    var function: MergeFunction
+    init(_ function: @escaping MergeFunction) {
+      self.function = function
+    }
+  }
+
+  @noDerivative private let _mergeFunction: _MergeFunction
+
```
dan-zheng, line 475: Please add documentation comments to public stored properties.
```suggestion
  /// The forward recurrent layer.
```

Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -415,17 +426,146 @@ public struct RecurrentLayer<Cell: RecurrentLayerCell>: Layer {
...
+  public var forward: RecurrentLayer<Cell>
+
```
dan-zheng, line 477:
```suggestion
  /// The backward recurrent layer.
```

Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -415,17 +426,146 @@ public struct RecurrentLayer<Cell: RecurrentLayerCell>: Layer {
...
+  public var backward: RecurrentLayer<Cell>
+
```
dan-zheng, line 479:
```suggestion
  /// The differentiable function used for merging forward and backward recurrent layer outputs.
```

Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -324,9 +329,11 @@ public struct RecurrentLayer<Cell: RecurrentLayerCell>: Layer {
   public typealias Output = [Cell.TimeStepOutput]
 
   public var cell: Cell
+  @noDerivative public let backwardDirection: Bool
 
-  public init(_ cell: @autoclosure () -> Cell) {
+  public init(_ cell: @autoclosure () -> Cell, backwardDirection: Bool = false) {
```
dan-zheng, line 334: It seems strange to add a `backwardDirection` parameter to `RecurrentLayer`'s initializer. It seems like this parameter only exists to support `BidirectionalRecurrentLayer` and doesn't have precedent in other neural network libraries, so it's not really a public API that we want to expose to users.

Suggestion: rather than adding the `backwardDirection` stored property and parameter, how about changing `BidirectionalRecurrentLayer.callAsFunction` to apply the backward recurrent layer to `inputs.reversed()` instead?

```swift
    let backwardOutputs = backward(
      inputs.reversed(), initialState: initialStateBackward)
```

This is less efficient in the short-term but allows us to keep the public APIs that we want. Eventually, we can restore efficiency by making `RecurrentLayer.callAsFunction` take an arbitrary `Collection` of `Cell.TimeStepInput` (e.g. [`ReversedCollection<LazySequence<[Cell.TimeStepInput]>>`](https://developer.apple.com/documentation/swift/reversedcollection)) of instead `[Cell.TimeStepInput]`.

Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -415,17 +426,146 @@ public struct RecurrentLayer<Cell: RecurrentLayerCell>: Layer {
...
+  public var forward: RecurrentLayer<Cell>
+
+  public var backward: RecurrentLayer<Cell>
+
+  @noDerivative public var mergeFunction: MergeFunction {
+    _mergeFunction.function
+  }
+
+  /// - Note: Default merge mode in Keras is `concat`, which is not supported right now.
+  ///   When `concat` mode is added, it can be used as default value to not confuse users.
```
dan-zheng, line 485:
```suggestion
  /// Creates an instance from the given recurrent layer cell and merge function.
  ///
  /// - Note: The default merge mode in Keras is `concat`, which is not currently supported.
  ///   When the `concat` mode is added, it can be used as the default value.
```

Sources/TensorFlow/Layers/Recurrent.swift
kongzii: I followed Keras, and they do have a `go_backwards` option publicly available, look at https://keras.io/api/layers/recurrent_layers/rnn/. But I am not sure there will be another usage except for `BidirectionalRecurrentLayer`.

But of course, I can change it. Do you think it will make more sense?

Utilities/ReferenceImplementations/lstm.py
```diff
@@ -1,9 +1,13 @@
-# Computes expected results for `testLSTM()` in `Tests/TensorFlowTests/LayerTests.swift`.
+# Computes expected results for `testLSTM()` or `testReversedLSTM()` in `Tests/TensorFlowTests/LayerTests.swift`.
 # Requires 'tensorflow>=2.0.0a0' (e.g. "pip install tensorflow==2.0.0b1").
 
+import sys
 import numpy
 import tensorflow as tf
 
+# Set random seed for repetable results
```
dan-zheng, line 8:
```suggestion
# Set random seed for repeatable results.
```

Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -415,17 +426,148 @@ public struct RecurrentLayer<Cell: RecurrentLayerCell>: Layer {
...
+    forward = RecurrentLayer(cell())
+    backward = RecurrentLayer(cell(), backwardDirection: true)
+    _mergeFunction = .init(mergeFunction)
+  }
+
+  @differentiable
+  public func callAsFunction(
+    _ inputs: Input,
+    initialStateForward: Cell.State,
+    initialStateBackward: Cell.State
```
dan-zheng, line 498:
```suggestion
    initialForwardLayerState: Cell.State,
    initialBackwardLayerState: Cell.State
```

Sources/TensorFlow/Layers/Recurrent.swift
dan-zheng: Aha, thanks for sharing the precedent from Keras, that's good to know.

Unless others feel differently, I still think it's better to avoid adding a `backwardDirection` parameter since it's equivalent to reversing the input and seems ad-hoc.

Could you please pursue the design above instead, which avoids expanding the API surface? If there are strong use cases for a `backwardDirection` parameter outside of `BidirectionalRecurrentLayer`, it can always be added later.

Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -415,17 +426,148 @@ public struct RecurrentLayer<Cell: RecurrentLayerCell>: Layer {
...
+  ) -> Output {
+    let forwardOutputs = forward(
+      inputs, initialState: initialStateForward)
+    let backwardOutputs = backward(
+      inputs, initialState: initialStateBackward)
```
dan-zheng, line 503:
```suggestion
    let forwardOutputs = forward(
      inputs, initialState: initialForwardLayerState)
    let backwardOutputs = backward(
      inputs, initialState: initialBackwardLayerState)
```

Sources/TensorFlow/Layers/Recurrent.swift
kongzii: Of course, I will look into it later today or tomorrow and change it as you suggested.

kongzii: I have implemented the changes, unfortunately,`.reversed()` is not yet differentiable, so I made a workaround. In the next PRs, I could implement differentiation for it, and they fix it here, what do you think?

Sources/TensorFlow/Layers/Recurrent.swift
kongzii: Of course, fixed everywhere.

Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -420,12 +425,150 @@ public struct RecurrentLayer<Cell: RecurrentLayerCell>: Layer {
   }
 }
 
+public protocol Mergeable: Differentiable, AdditiveArithmetic {
```
dan-zheng, line 428: Please add a doc comment:
```swift
/// A type with values that support differentiable binary operations.
///
/// Used by `BidirectionalRecurrentLayer` as a generic requirement for merge functions.
```

Tests/TensorFlowTests/LayerTests.swift
```diff
@@ -1600,6 +1600,189 @@ final class LayerTests: XCTestCase {
...
+    let expectedSum = Tensor<Float>(
+      -1.3748394
+    )
+    let expectedStates = Tensor<Float>(
+        [[[0.039731443,0.30118462,-1.1362883,-0.6404417],
+          [-0.36877188,0.23013103,0.053439736,0.29040766],
+          [-0.7231241,0.4473409,-0.46913218,-0.64710104],
+          [0.41570857,-1.500107,-0.20221666,-0.08822429]]]
+    )
+    // TODO: Modify BidirectionalRecurrentLayer to return those two and test it
```
dan-zheng, line 1657: The newly added tests have some unused value warnings:
```
tensorflow-swift-apis/Tests/TensorFlowTests/LayerTests.swift:1658:9: warning: initialization of immutable value 'expectedFinalStateForward' was never used; consider replacing with assignment to '_' or removing it
    let expectedFinalStateForward = Tensor<Float>(
    ~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
    _
tensorflow-swift-apis/Tests/TensorFlowTests/LayerTests.swift:1661:9: warning: initialization of immutable value 'expectedFinalStateBackward' was never used; consider replacing with assignment to '_' or removing it
    let expectedFinalStateBackward = Tensor<Float>(
    ~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
    _
tensorflow-swift-apis/Tests/TensorFlowTests/LayerTests.swift:2091:9: warning: initialization of immutable value 'expectedFinalStateHiddenForward' was never used; consider replacing with assignment to '_' or removing it
    let expectedFinalStateHiddenForward = Tensor<Float>(
    ~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    _
tensorflow-swift-apis/Tests/TensorFlowTests/LayerTests.swift:2094:9: warning: initialization of immutable value 'expectedFinalStateCellForward' was never used; consider replacing with assignment to '_' or removing it
    let expectedFinalStateCellForward = Tensor<Float>(
    ~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    _
tensorflow-swift-apis/Tests/TensorFlowTests/LayerTests.swift:2097:9: warning: initialization of immutable value 'expectedFinalStateHiddenBackward' was never used; consider replacing with assignment to '_' or removing it
    let expectedFinalStateHiddenBackward = Tensor<Float>(
    ~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    _
tensorflow-swift-apis/Tests/TensorFlowTests/LayerTests.swift:2100:9: warning: initialization of immutable value 'expectedFinalStateCellBackward' was never used; consider replacing with assignment to '_' or removing it
    let expectedFinalStateCellBackward = Tensor<Float>(
    ~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    _
tensorflow-swift-apis/Tests/TensorFlowTests/LayerTests.swift:2672:9: warning: initialization of immutable value 'expectedFinalStateForward' was never used; consider replacing with assignment to '_' or removing it
    let expectedFinalStateForward = Tensor<Float>(
    ~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
    _
tensorflow-swift-apis/Tests/TensorFlowTests/LayerTests.swift:2675:9: warning: initialization of immutable value 'expectedFinalStateBackward' was never used; consider replacing with assignment to '_' or removing it
    let expectedFinalStateBackward = Tensor<Float>(
    ~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
    _
```

Is it easy to test these values? If not, could you please comment them to silence the warnings?

Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -420,12 +425,150 @@ public struct RecurrentLayer<Cell: RecurrentLayerCell>: Layer {
...
+  /// - Note: Trying to use `+` instead of `sum` results in compile error,
+  ///   when Tensor is extended for Mergeable comforting.
+  ///   Stack dump is stored at https://pastebin.com/PxZTir6F.
```
dan-zheng, line 431: Minor rewording, with a link to the issue:
```suggestion
  /// - Note: renaming `sum` to `+` results in a compiler crash when conforming `Tensor` to
  /// `Mergeable` (SR-13229).
```

Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -420,12 +425,150 @@ public struct RecurrentLayer<Cell: RecurrentLayerCell>: Layer {
...
+    for forwardIndex in  0 ..< withoutDerivative(at: inputs.count) {
+        let backwardIndex = withoutDerivative(at: inputs.count - 1 - forwardIndex)
+        outputs.append(mergeFunction(forwardOutputs[forwardIndex], backwardOutputs[backwardIndex]))
+    }
+
+    return outputs
+  }
+
+  @differentiable
+  public func callAsFunction(_ inputs: Input) -> Output {
```
dan-zheng, line 519: Please add a precondition:
```swift
    precondition(!inputs.isEmpty, "'inputs' must be non-empty.")
```

Same for the `lastOutput` methods below.

Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -420,12 +425,150 @@ public struct RecurrentLayer<Cell: RecurrentLayerCell>: Layer {
   }
 }
 
+public protocol Mergeable: Differentiable, AdditiveArithmetic {
+  /// - Note: Trying to use `+` instead of `sum` results in compile error,
+  ///   when Tensor is extended for Mergeable comforting.
+  ///   Stack dump is stored at https://pastebin.com/PxZTir6F.
+  @differentiable
+  static func sum(_ lhs: Self, _ rhs: Self) -> Self
```
dan-zheng, line 433: Please add a doc comment:
```swift
  /// Adds two values and produces their sum.
```

Tests/TensorFlowTests/LayerTests.swift
kongzii: It is easy to return them but it seems kind of weird to me: I could modify `BidirectionalRecurrentLayer.callAsFunction` to return something like `(finalStateForward: Cell.TimeStepOutput, ..., output: Output)`. In Keras, it is done in a similar way if you check tests: https://github.com/tensorflow/swift-apis/pull/946/files#diff-5a1997b5e5dbf8509079fa2f4d5ae9d6R129. 

If you have an idea of how they should be returned I will implement it. For now, I commented out the variables.

Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -420,12 +425,150 @@ public struct RecurrentLayer<Cell: RecurrentLayerCell>: Layer {
...
+  }
+
+  public static func +(lhs: Self, rhs: Self) -> Self {
+    return Self(lhs.cell + rhs.cell)
+  }
+
+  public static func -(lhs: Self, rhs: Self) -> Self {
+    return Self(lhs.cell - rhs.cell)
+  }
+}
```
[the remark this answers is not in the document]
kongzii: Oh sorry, I forgot about it, yes it can.

Sources/TensorFlow/Layers/Recurrent.swift
kongzii: Added.

Sources/TensorFlow/Layers/Recurrent.swift
kongzii: Added. :)

Tests/TensorFlowTests/LayerTests.swift
dan-zheng: I see: commenting the tests sounds for now, thanks! I think the current versions of `callAsFunction` make sense for now.

Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -420,12 +426,146 @@ public struct RecurrentLayer<Cell: RecurrentLayerCell>: Layer {
...
+  public var backward: RecurrentLayer<Cell>
+  /// The differentiable function used for merging forward and backward recurrent layer outputs.
+  @noDerivative public var mergeFunction: MergeFunction {
+    _mergeFunction.function
+  }
+
+  /// Creates an instance from the given recurrent layer cell and merge function.
+  ///
+  /// - Note: The default merge mode in Keras is `concat`, which is not currently supported.
+  ///   When the `concat` mode is added, it can be used as the default value.
```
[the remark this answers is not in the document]
kongzii: Yeah, I will look into it and add tests too 👍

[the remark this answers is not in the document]
kongzii: Sorry for the delay, but I encouraged a weird bug. I have added `concatenate` merge function and tests for it, but at runtime I got this error:

```
Precondition failed: Axis must be in `-rank..<rank` when calling split(sizes:alongAxis:) (rank: 0, axis: 1): file swift-apis/Sources/TensorFlow/Operators/Basic.swift, line 142
2020-09-02 19:57:25.801775+0200 xctest[93460:1943115] Precondition failed: Axis must be in `-rank..<rank` when calling split(sizes:alongAxis:) (rank: 0, axis: 1): file swift-apis/Sources/TensorFlow/Operators/Basic.swift, line 142
```

It is caused by using `lhs.concatenated(with: rhs, alongAxis: 1)` instead of `lhs + rhs`, but I did not find the error yet.

Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -439,13 +449,28 @@ public protocol Mergeable: Differentiable, AdditiveArithmetic {
 }
 
 extension Tensor: Mergeable where Scalar: TensorFlowFloatingPoint {
+  /// Adds two values and produces their concatenation.
+  @differentiable
+  public static func concatenate(_ lhs: Tensor, _ rhs: Tensor) -> Tensor {
+    lhs.concatenated(with: rhs, alongAxis: 1)
```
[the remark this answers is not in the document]
kongzii: not really and I have just checked out Keras and they are concatenating with `axis=-1`, so I will edit it accordingly.

Sources/TensorFlow/Layers/Recurrent.swift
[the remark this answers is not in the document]
dan-zheng: Is this issue fixed now that concatenation happens along axis `-1` instead of axis `1`?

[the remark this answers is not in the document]
kongzii: Unfortunately not. I was trying to create some minimal example that would reproduce the error but without any luck. It is weird because if I just replace `lhs.concatenated(with: rhs, alongAxis: X)` with `lhs + rhs`, it just works. And all tests for sum merge mode against Keras are passing. And I did not find any special code for concatenated method in Keras.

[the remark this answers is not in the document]
dan-zheng: I see, thanks for the prompt response.

A minimal example would be really helpful. When you call `lhs.concatenated(with: rhs, alongAxis: -1)` and it triggers the precondition failure, what are the shapes of `lhs` and `rhs`?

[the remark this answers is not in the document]
kongzii: I found out that it fails at `_vjpConcatenated`. I added the following prints into the function in Basic.swift:

```
@inlinable
  @derivative(of: concatenated)
  func _vjpConcatenated(
    with other: Tensor,
    alongAxis axis: Int
  ) -> (value: Tensor, pullback: (Tensor) -> (Tensor, Tensor)) {
    let posAxis = axis < 0 ? axis + rank : axis
    let splits = [shape[posAxis], other.shape[posAxis]]
    return (
      concatenated(with: other, alongAxis: axis),
      { result in
        print("self", self)
        print("other", other)
        print("result", result)
        print("posAxis", posAxis)
        print("splits", splits)
        let gradients = result.split(sizes: splits, alongAxis: axis)
        print("gradients", gradients)
        return (gradients[0], gradients[1])
      }
    )
  }
```

the output after running `testBidirectionalBasicRNNConcatenateMerge` is:

```
self [[0.81258523, -0.5553411, 0.27422628, 0.73577887]]
other [[-0.39687672, -0.94476587, -0.47644293, -0.82400316]]
result [[1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0]]
posAxis 1
splits [4, 4]
gradients [[[1.0, 1.0, 1.0, 1.0]], [[1.0, 1.0, 1.0, 1.0]]]
self [[-0.19038992,  0.44098023, 0.062064234,  -0.9267066]]
other [[  -0.5327341, 0.0063607423,   -0.5311964,    0.2796056]]
result 0.0
posAxis 1
splits [4, 4]
Precondition failed: Axis must be in `-rank..<rank` when calling split(sizes:alongAxis:) (rank: 0, axis: -1): file /Users/peter.jung/Projects/swift-apis/Sources/TensorFlow/Operators/Basic.swift, line 142
2020-09-09 20:08:04.530498+0200 xctest[69563:4338687] Precondition failed: Axis must be in `-rank..<rank` when calling split(sizes:alongAxis:) (rank: 0, axis: -1): file /Users/peter.jung/Projects/swift-apis/Sources/TensorFlow/Operators/Basic.swift, line 142
```

So, for some reason, at some point, `result` is `0` and then `result.split(...)` fails.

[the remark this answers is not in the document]
kongzii: @dan-zheng I have finally some free time and I was able to create a script to reproduce the error along with equivalent (working) python script. Please look at https://github.com/tensorflow/swift-apis/issues/1087.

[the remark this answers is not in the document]
kongzii: @dan-zheng Thanks to the @xihui-wu, I implemented a workaround. Concat-merge mode is now fully functional and set as a default merge function. Please check out the PR, everything seems to be working now :) :tada:

PR #951, opened 2020-05-13
Update `@_Freezable` after wrapped property differentiation changes.

dan-zheng:
Wrapped property differentiation was recently simplified:

<details>

- `Differentiable` conformance derivation now "peers through" property
  wrappers.
- Synthesized `TangentVector` structs contain wrapped properties'
  `TangentVector`s as stored properties, not wrappers' `TangentVector`s.
- It is irrelevant whether property wrapper types conform to `Differentiable`.

</details>

In light of these changes:
- `@_Freezable` no longer conforms to `Differentiable`.
- Rather than providing zero derivatives, `@_Freezable` now disables updates to
  `var wrappedValue`. This better fits the definition of "frozen".

---

This change requires and is required by https://github.com/apple/swift/pull/31731: the 2020-05-12 apple/swift `master -> tensorflow` merge.

Alternatively, we could consider deleting `@_Freezable` since it's not clear that it's useful.

### DIFF
--- a/Sources/TensorFlow/Freezable.swift
+++ b/Sources/TensorFlow/Freezable.swift
@@ -12,64 +12,49 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-/// A wrapper around a differentiable value with "freezable" derivatives.
+/// A wrapper around a value whose value can be frozen.
 ///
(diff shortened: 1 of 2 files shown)


dan-zheng:
Confirmed that extended test suite passes. Merging to unblock progress.

PR #957, opened 2020-05-13
Hides the symbols except for those that are marked as visible

ematejska:
This fixes the immediate mode issue when the TensoFlow modules was used with swift because the symbols are not being loaded twice for llvm.

https://bugs.swift.org/projects/TF/issues/TF-1273

### DIFF
--- a/Sources/CX10/xla_tensor_wrapper.cc
+++ b/Sources/CX10/xla_tensor_wrapper.cc
@@ -15,7 +15,7 @@
 #if defined(_WIN32)
 #define XLA_API __declspec(dllexport)
 #else
-#define XLA_API __attribute__((visibility("default")))
+#define XLA_API __attribute__((__visibility__("default")))
 #endif
 
 #include "xla_tensor_wrapper.h"
--- a/Sources/CX10/xla_tensor_wrapper.cc
+++ b/Sources/CX10/xla_tensor_wrapper.cc
@@ -15,7 +15,7 @@
 #if defined(_WIN32)
 #define XLA_API __declspec(dllexport)
 #else
-#define XLA_API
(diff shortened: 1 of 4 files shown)


ematejska:
This is not to be merged yet because the TFE_GetServerDef symbol in Sources/TensorFlow/Core/Runtime.swift ends up undefined. (It's an experimental symbol in tensorflow/c/c_api_experimental.h whose visibility is not public.

compnerd:
`TFE_GetServerDef` is in TensorFlow and needs to be annotated with default visibility.

Sources/CX10/xla_tensor_wrapper.cc
```diff
@@ -15,7 +15,7 @@
 #if defined(_WIN32)
 #define XLA_API __declspec(dllexport)
 #else
-#define XLA_API
+#define XLA_API __attribute__((visibility("default")))
```
compnerd, line 18: Could you please use the reserved spelling?  `__attribute__((__visibility__("default")))`

ematejska:
Yeah, I wonder if TensorFlow will accept that change or if there is anyther API that can be used.

ematejska:
I got the PR for tensorflow here: https://github.com/tensorflow/tensorflow/pull/39551 but if it's accepted, we have to wait for S4TF to pull that version of Tensorflow in before merging this.

ematejska:
Removed reliance of TFE_GetServerDef here: https://github.com/tensorflow/swift-apis/pull/976. If testing passes, we're good to merge.

Sources/CX10/xla_tensor_wrapper.cc
ematejska: Done

Issue #936, opened 2020-05-06
GRUCell internal tensors undersized for inputSize != hiddenSize


lindyhoptrader:
The internal weight and hidden state matrices for GRUCell are undersized and fail at matmul if inputSize != hiddenSize.

They should be initialized more like this:

@noDerivative public var stateShape: TensorShape {
    //[1, updateWeight1.shape[0]]
     updateWeight1.shape
  }

/// Creates a `GRUCell` with the specified input size and hidden state size.
  ///
  /// - Parameters:
  ///   - inputSize: The number of features in 2-D input tensors.
  ///   - hiddenSize: The number of features in 2-D hidden states.
  public init(
    inputSize: Int,
    hiddenSize: Int,
    weightInitializer: ParameterInitializer<Scalar> = glorotUniform(),
    biasInitializer: ParameterInitializer<Scalar> = zeros()
  ) {
    let inputWeightShape = TensorShape([inputSize, hiddenSize])
    let hiddenWeightShape = TensorShape([hiddenSize, hiddenSize])
    let gateBiasShape = TensorShape([hiddenSize])
    self.updateWeight1 = weightInitializer(inputWeightShape)
    self.updateWeight2 = weightInitializer(hiddenWeightShape)
    self.updateBias = biasInitializer(gateBiasShape)
    self.resetWeight1 = weightInitializer(inputWeightShape)
    self.resetWeight2 = weightInitializer(hiddenWeightShape)
    self.resetBias = biasInitializer(gateBiasShape)
    self.outputWeight1 = weightInitializer(inputWeightShape)
    self.outputWeight2 = weightInitializer(hiddenWeightShape)
    self.outputBias = biasInitializer(gateBiasShape)
  }

Here's another detailed python GRUCell implementation for reference, showing hiddenWeightShapes as square 2D hidden X hidden matrices:

https://github.com/dennybritz/rnn-tutorial-gru-lstm/blob/master/gru_theano.py

testGRU() in LayerTests.swift should be supplemented with an example where inputSize != hiddenSize

Also should check LSTMCell for possibility of similar tensor sizing errors.

dan-zheng:
Hi @lindyhoptrader,

It's a bit hard to understand your issue without more context. If you feel `GRUCell` initialization is in need of a fix, could you please turn your changes above into a PR and update tests?

kongzii:
I noticed this too. Also, If I test outputs against reference implementation in Python, it does not match.

[...]

PR #960, opened 2020-05-16
Fix GRU and add tests

kongzii:
Hi, this PR does several things:

- fixes https://github.com/tensorflow/swift-apis/issues/936
- fixes incorrect equations in GRU
- separates `bias` and `recurrentBias`, as it is in Keras
- renames tensors in GRU (I think that `updateKernel, updateRecurrentKernel` is more clear and also swiftier than `updateWeight1, updateWeight2`)
- added test against reference implementation in Keras (with different input and hidden size and it is asserting shapes of tensors before replacing them with generated code, I think this would be helpful in other tests too)

I hope all is OK with GRUCell now.


kongzii:
Could I get some feedback on this, please? :) When GRU will be fixed, I can continue with https://github.com/tensorflow/swift-apis/pull/946.

PR #962, opened 2020-05-18
Gelu reduced precision

mikowals:
`gelu(x)` on X10 fails when x is reduced precision.  I added a check on x and reduce precision of all the tensor constants it uses if necessary.

I also implemented tests (based off leakyRelu) in X10/ops_tests.swift.  **I have not run the tests I added yet** as I don't know how to trigger them locally (Xcode or command line in MacOS).  I think this is the correct place for them and they were previously omitted maybe because `gelu()` is not reimplemented in X10's rawOpsManual.swift.

### DIFF
--- a/Sources/TensorFlow/Operators/Math.swift
+++ b/Sources/TensorFlow/Operators/Math.swift
@@ -1365,19 +1365,16 @@ func _vjpElu<T: TensorFlowFloatingPoint>(
 @inlinable
 @differentiable
 public func gelu<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
-  let useReducedPrecision = withoutDerivative(at: x) { $0.isReducedPrecision }
-  func _maybeReducedPrecision(_ x: Tensor<T>) -> Tensor<T> {
-    useReducedPrecision ? x.toReducedPrecision : x
-  }
+  let xWithoutDerivative = withoutDerivative(at: x) { $0 }
   // An approximation of √(2/π).
-  let ratio1 = _maybeReducedPrecision(Tensor<T>(0.7978845608, on: x.device))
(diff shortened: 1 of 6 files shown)


asuhan:
Cool! I'll defer to @dan-zheng on matters of style, the logic looks correct to me.

mikowals:
I found the `Tensor(... , deviceAndPrecisonLike: )` so I used it to clean up the code.

I added a gradient test for `gelu()` because I initially didn't realise `withoutDerivative` was needed and all tests passed still passed.  But backprop failed in a network locally.  So I added a test that would have caught it.

Sources/TensorFlow/Operators/Math.swift
```diff
@@ -1365,14 +1365,16 @@ func _vjpElu<T: TensorFlowFloatingPoint>(
 @inlinable
 @differentiable
 public func gelu<T: TensorFlowFloatingPoint>(_ x: Tensor<T>) -> Tensor<T> {
-  let ratio1 = Tensor<T>(0.7978845608, on: x.device)  // An approximation of √(2/π).
+  let xWithoutDerivative = withoutDerivative(at: x) { $0 }
```
dan-zheng, line 1368: Is it necessary to use `withoutDerivative(at: x)` instead of `x` directly here?

dan-zheng, line 1368: If `withoutDerivative(at: x)` is necessary, you can simplify it to this version:
```suggestion
  let xWithoutDerivative = withoutDerivative(at: x)
```

mikowals: Thanks. I did not realise I could write it that way.  I will be able to use `x` directly once #963 is merged.

mikowals: done now.

mikowals:
I have messed up git.  I will close this and make another pull request.

PR #963, opened 2020-05-20
add @noDerivative to isReducedPrecision declaration

mikowals:
I don't think differentiating `.isReducedPrecision` has a use so is it okay to flag it with `@noDerivative` to avoid using `withoutDerivative` frequently in differentiable functions? 

I also removed the `withoutDerivative(at: input)` for both `.isReducedPrecision` and `Tensor(... , deviceAndPrecisionLike: input)` calls in layers/normalizing.swift since they are no longer needed.

I can also remove `withoutDerivative` use in #962.


ematejska:
Looks good, just waiting for CI results.

PR #965, opened 2020-05-21
match precision in gelu 

mikowals:
This replaces #962 

The tensors created in gelu are matched to the precision of the input.

I added X10 tests that would have caught the mismatch.  I have not run them since I don't know how to trigger the x10 tests.  The added tests are based off leakyRelu X10 tests.

I added a test of gelu gradient to pick up an error that was missed by tests at an intermediate step in these changes.


mikowals:
I saw that tests failed.  I updated the grad values in testGeluGrad and increased the error tolerance in the testGelu for reduced precision.  

I have made a reproduction of the tests that I can run and these changes were sufficient.  So I am not completely shooting in the dark but still not running the actual tests locally.

mikowals:
I removed the changed scalars in the gradient test as they are irrelevant and don't need to look so precise.  My mock test is passing because I do not have a compiled version with #963 and had resorted to `withoutDerivative()` for `isReducedPrecision`.

The error message in the test log is:

    Fatal error: Op must have the same backend type: TF_EAGER vs XLA: file /swift-apis/Sources/TensorFlow/Bindings/RawOpsDispatching.swift, line 370.

I am not sure what causes the failure and how `withoutDerivative()` is able to resolve it.

asuhan:
@mikowals What's the version of the code which reported the backend type mismatch? The error simply says we don't accept operations between different backends, which was a conscious design choice to avoid implicit conversions with dramatic performance implications.

mikowals:
That is the error reported in the kokoro test logs from commit 6201e0d above.  

I can get the same error with s4tf 0.9 release if I backprop through gelu as its written in this pull.  The error will go away if I add `withoutDerivative(at: x)` for the device and precision checks.  That should no longer be necessary though after #963.

So I could add `withoutDerivative(at: x)` back to this pull request and I expect it would make tests pass but it should not be necessary.

mikowals:
I added back the use of `withoutDerivative(at: x)`.  Tests should pass and this can be merged.

asuhan:
Thanks @mikowals . Right now I'm looking into performance issues (broken traces) which appear to have been triggered by https://github.com/tensorflow/swift-apis/pull/963. Once that's sorted out, we can come back to this patch - shouldn't take more than a few days.

compnerd:
@asuhan is going to tentatively going to look into this this week

Tests/TensorFlowTests/TensorAutoDiffTests.swift
```diff
@@ -1104,6 +1104,11 @@ final class TensorAutoDiffTests: XCTestCase {
       ])
   }
 
+  func testGelu() {
+    func f(a: Tensor<Float>) -> Tensor<Float> { gelu(a).sum() }
+    XCTAssertEqual(gradient(at: [5, -5, 0], in: f), [1, 0, 0.5])
```
asuhan, line 1109: Use `XCTAssert(gradient(at: [5, -5, 0], in: f).isAlmostEqual(to: [1, 0, 0.5]))` instead, this doesn't pass on my machine.

Tests/x10/ops_test.swift
```diff
@@ -1273,6 +1273,47 @@ final class TensorTests: XCTestCase {
       }
     }
   }
+
```
asuhan, line 1276: Remove the space at the beginning of the line.

Tests/TensorFlowTests/TensorAutoDiffTests.swift
mikowals: Done.  It matched exactly on my machine so I had to guess a tolerance level.

Tests/x10/ops_test.swift
mikowals: Done.

PR #967, opened 2020-05-24
Fix bugs and incorrect comments in Sampling.swift

dabrahams:
@sgugger I was a little shocked at the brokenness of `index(_:offsetBy:)` and not pointing any fingers 'cause I might have written it that way, but do you think you could come up with tests for all these APIs that verify them?

### DIFF
--- a/Sources/TensorFlow/Epochs/Sampling.swift
+++ b/Sources/TensorFlow/Epochs/Sampling.swift
@@ -48,15 +48,15 @@ extension Sampling: Collection {
 
 extension Sampling: BidirectionalCollection
 where Selection: BidirectionalCollection {
-  /// Returns the position after `i`.
+  /// Returns the position before `i`.
   public func index(before i: Index) -> Index { selection.index(before: i) }
 }
 
 extension Sampling: RandomAccessCollection
-where Selection: BidirectionalCollection {
+where Selection: RandomAccessCollection {
   /// Returns the position `n` places from `i`.
   public func index(_ i: Index, offsetBy n: Int) -> Index {
-    selection.index(before: i)
+    selection.index(i, offsetBy: n)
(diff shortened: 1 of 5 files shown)


Sources/TensorFlow/Epochs/Sampling.swift
```diff
@@ -48,15 +48,15 @@ extension Sampling: Collection {
...
 extension Sampling: BidirectionalCollection
 where Selection: BidirectionalCollection {
-  /// Returns the position after `i`.
+  /// Returns the position before `i`.
   public func index(before i: Index) -> Index { selection.index(before: i) }
 }
 
 extension Sampling: RandomAccessCollection
-where Selection: BidirectionalCollection {
+where Selection: RandomAccessCollection {
```
dabrahams, line 56: Oops!

sgugger:
Mmm, not sure if it came from messy copy-paste on my side or not. I can look at adding tests, but will be fairly busy with the generic training loop next week.

Sources/TensorFlow/Epochs/Sampling.swift
```diff
@@ -66,15 +66,12 @@ where Selection: BidirectionalCollection {
   public func index(
     _ i: Index, offsetBy distance: Int, limitedBy limit: Index
   ) -> Index? {
-    // FIXME: swift-3-indexing-model: tests.
-    let l = self.distance(from: i, to: limit)
-    if distance > 0 ? l >= 0 && l < distance : l <= 0 && distance < l {
-      return nil
-    }
-    return index(i, offsetBy: distance)
+    selection.index(i, offsetBy: distance, limitedBy: limit)
```
sgugger, line 69: This makes the tests break for some reason.

PR #971, opened 2020-05-27
CMake fix: look for the right directories on MacOSX

dabrahams:


cmake/modules/FindX10.cmake
```diff
@@ -3,7 +3,11 @@ if(DEFINED CMAKE_Swift_COMPILER)
   get_filename_component(_Swift_TOOLCHAIN ${CMAKE_Swift_COMPILER} DIRECTORY)
   get_filename_component(_Swift_TOOLCHAIN ${_Swift_TOOLCHAIN} DIRECTORY)
   get_filename_component(_Swift_TOOLCHAIN ${_Swift_TOOLCHAIN} DIRECTORY)
-  string(TOLOWER ${CMAKE_SYSTEM_NAME} system_lc)
+  if(CMAKE_SYSTEM_NAME STREQUAL Darwin) 
+    set(system_lc macosx)
```
compnerd, line 7: There is some bleeding whitespace at the end of the lines.

PR #974, opened 2020-05-28
Revert "add @noDerivative to isReducedPrecison declaration (#963)"

asuhan:
This reverts commit dc1ad3b53d620e3549bb6a3300af887d7c38eba4.


asuhan:
Unfortunately, AD calls `scalarized` and breaks traces with the change we're reverting. We should understand this better to find a better workaround. For now, this should fix the dramatic benchmark drop.

PR #985, opened 2020-06-05
Optimize split, _vjpConcatenated, tiled and _vjpTiled

asuhan:
Avoid more temporary tensors.


asuhan:
@pschuh Please take a look.

PR #999, opened 2020-06-11
add vjpInitDeviceAndPrecisionLike (fix #973)

### LINKED ISSUE #973: device mismatch in pullback of Tensor(_, deviceAndPrecisionLike:)
This relates to behaviour seen in #965.

I made a small reproducer of the device mismatch.

    import TensorFlow

    @differentiable(wrt: t)
    func scalarTensor(t: Tensor<Float>) -> Tensor<Float> {
        Tensor<Float>(1.0, deviceAndPrecisionLike: t)
    }
    let device = Device.defaultXLA
    let x = Tensor<Float>(0, on: device)
    let (val, pullback) = valueWithPullback(at: x, in: scalarTensor)
    pullback(Tensor<Float>(1, on: device))
    // Fatal error: Op must have the same backend type: TF_EAGER vs XLA:

I can work around by either using `withoutDerivative(at: t)` or I can add a custom derivative like this:

        @derivative(of: init(_:deviceAndPrecisionLike:)) 
        static func vjpDeviceAndPrecisionLike(_ value: Scalar, deviceAndPrecisionLike tensor: Tensor
            ) -> (value: Tensor, pullback: (Tensor) -> (Scalar, Tensor)
        ) {
            return (value: Tensor(value, deviceAndPrecisionLike: tensor), pullback: { 
                ($0.scalarized(), Tensor(0, deviceAndPrecisionLike: tensor))
            })
        }
    }

Declaring `init(_:deviceAndPrecisionLike:)` differentiable only with respect to the first argument also seems sensible but that then just forces the use of `withoutDerivative()` in the second argument to avoid having an undefined pullback.  

I am happy to make a pull request but I am not sure what is a preferred solution.  Maybe there is also another obvious way to fix this that I have overlooked.

mikowals:
Custom pullback returns a zero tensor with device and precision matched to the example `tensor` as per discussion in #973.  

I avoided capturing `tensor` in the pullback closure.    Maybe this is not necessary or I could reduce the redundancy by creating `init(_:on:withReducedPrecision:)`.  I initially did this but it seemed overkill if the real solution of `@noDerivative` in the init makes this all unnecessary. 

When this is merged the `withoutDerivative` attribute could be removed in several places (for example in normalisation layers [here](https://github.com/tensorflow/swift-apis/blob/master/Sources/TensorFlow/Layers/Normalization.swift#L139) and [here](https://github.com/tensorflow/swift-apis/blob/master/Sources/TensorFlow/Layers/Normalization.swift#L155)).  A previous removal of this got [reverted](https://github.com/tensorflow/swift-apis/pull/974)  so I thought doing it in a separate request would make discovering any problem easier.

### DIFF
--- a/Sources/TensorFlow/Core/MixedPrecision.swift
+++ b/Sources/TensorFlow/Core/MixedPrecision.swift
@@ -159,11 +159,32 @@ extension Tensor {
   }
 
   /// Promotes a scalar to a tensor with the same device and precision as the given tensor.
+  // TODO (SR-12968): Mark `tensor` with `@noDerivative` and remove custom vjp below.
   @differentiable(where Scalar: TensorFlowFloatingPoint)
   public init(_ value: Scalar, deviceAndPrecisionLike tensor: Tensor) {
     let device = tensor.device
     let tmp = Tensor(value, on: device)
     self = tensor.isReducedPrecision ? tmp.toReducedPrecision : tmp
+  }
+}
+
+extension Tensor where Scalar: TensorFlowFloatingPoint {
+  // TODO (SR-12968): Remove when `tensor` can be marked `@noDerivative` in `init`.
+  // This currently places the pullback results of `tensor` on the correct device.
+  @usableFromInline
+  @derivative(of: init(_:deviceAndPrecisionLike:))
+  static func vjpInitDeviceAndPrecisionLike(
+    _ value: Scalar,
+    deviceAndPrecisionLike tensor: Tensor
+  ) -> (value: Tensor, pullback: (Tensor) -> (Scalar, Tensor)) {
+    // Get device and precision in forward pass to avoid capturing `tensor` in pullback.
+    let device = tensor.device
+    let useReducedPrecision = tensor.isReducedPrecision
+    let result = Tensor(value, on: device)
+    return (useReducedPrecision ? result.toReducedPrecision : result, {
+      let tmp = Tensor(0, on: device)
+      return ($0.scalarized(), useReducedPrecision ? tmp.toReducedPrecision : tmp)
+    })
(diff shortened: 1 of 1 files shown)


asuhan:
By itself, this patch is fine. The follow-up removal of `withoutDerivative` in normalization layers is the issue. Doing it in `BatchNorm` will break our ResNet x10 traces, because we'd call `scalarized` in `vjpInitDeviceAndPrecisionLike`, one such call site being: https://github.com/tensorflow/swift-apis/blob/5f5f09e9c39e006005e4d2d0abf5f8be6f1fa306/Sources/TensorFlow/Layers/Normalization.swift#L139. 

We could land this for convenience in situations where breaking the trace isn't critical, then do the follow-up removal of `withoutDerivative` once https://bugs.swift.org/browse/SR-12968 is addressed.

mikowals:
IIUC, it is the derivative of `epsilon` in batch norm that is calling `scalarized()`.  I don't think the derivative of epsilon is used so maybe the correct thing to do is:

    withoutDerivative(at: epsilon) { Tensor($0, deviceAndPrecisionLike: input) }

 I tried this locally and can confirm that non-x10 tests pass so the gradient of `epsilon` is not needed.  Even removing `withoutDerivative` altogether though a print statement inside `vjpInitDeviceAndPrecisionLike` is never executed so I think the autodiff (not using x10) figures out `epsilon` does not need a gradient.  I am not sure this gives any insight into how the x10 trace would handle my suggestion.

Removing `withoutDerivative` is not that pressing.  More important might be emitting a warning when using x10 in all the Scalar -> Tensor in forward pass functions that then lead to `scalaraized()` in the backward pass .

PR #1004, opened 2020-06-12
Add curl retry for toolchain download to Dockerfile

BradLarson:
In swift-models PRs [#590](https://github.com/tensorflow/swift-models/pull/590) and [#596](https://github.com/tensorflow/swift-models/pull/596), we found that adding a delayed retry to the curl download of the nightly toolchain helped with the intermittent OpenSSL errors we have been hitting. It doesn't completely solve all cases, but for our CI benchmark runs we went from ~25% failing two days ago to none failing due to network errors last night with this workaround in place.

I noticed that this repository was also experiencing these CI failures, so hopefully this helps.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,8 +18,11 @@ fi
 
 # Download and extract S4TF
 WORKDIR /swift-tensorflow-toolchain
-RUN curl -fSsL $swift_tf_url -o swift.tar.gz \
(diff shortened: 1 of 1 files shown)


dabrahams:
@BradLarson Please squash into single RUN in follow-up.

PR #1008, opened 2020-06-12
Fix erosion2d gradients

akshaan:
This change fixes incorrectly computed gradients for the `erosion2d` op


saeta:
Thanks! Merging in...

PR #1022, opened 2020-06-24
Tagging scalar zero tensors

BradLarson:
During profiling, we noticed that a large number of scalar zero tensors were being generated in the backward pass of certain models. These zero tensors were often involved in simple arithmetic where they did not change the result. As a quick optimization, this allows for a scalar zero Tensor to be tagged as such and to avoid dispatching calculations for which it would not impact the result.

In the case of the WordSeg model, this optimization leads to almost a tripling of throughput when calculating the score and gradient in the eager mode. When using X10, the benefit is reduced due to XLA optimizations, but is still present.

### DIFF
--- a/Sources/TensorFlow/Core/Tensor.swift
+++ b/Sources/TensorFlow/Core/Tensor.swift
@@ -34,7 +34,7 @@ public struct Tensor<Scalar: TensorFlowScalar> {
   /// The underlying `TensorHandle`.
   /// - Note: `handle` is public to allow user defined ops, but should not normally be used.
   public let handle: TensorHandle<Scalar>
-  
+
   @inlinable
   public init(handle: TensorHandle<Scalar>) {
     self.handle = handle
--- a/Sources/TensorFlow/Core/Tensor.swift
+++ b/Sources/TensorFlow/Core/Tensor.swift
@@ -35,6 +35,10 @@ public struct Tensor<Scalar: TensorFlowScalar> {
   /// - Note: `handle` is public to allow user defined ops, but should not normally be used.
   public let handle: TensorHandle<Scalar>
 
+  // The following is a spacer intended to temporarily work around a compiler crash induced
+  // by the zero-tagging optimization residing in TensorHandle.
+  var _spacer: UInt8 = 0
+
   @inlinable
   public init(handle: TensorHandle<Scalar>) {
     self.handle = handle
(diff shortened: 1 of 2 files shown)


Sources/TensorFlow/Core/Tensor.swift
```diff
@@ -35,6 +35,9 @@ public struct Tensor<Scalar: TensorFlowScalar> {
   /// - Note: `handle` is public to allow user defined ops, but should not normally be used.
   public let handle: TensorHandle<Scalar>
 
+  /// An internal marker to identify scalar zero tensors, for use in optimizations.
+  public var _isScalarZero = false
```
dan-zheng, line 39: I believe you can avoid publicly exposing `_isScalarZero` by marking it as `@usableFromInline`:

```suggestion
  @usableFromInline
  internal var _isScalarZero = false
```

BradLarson: Great, thanks. That's definitely preferred over having this implementation detail exposed publicly.

Sources/TensorFlow/Core/Tensor.swift
```diff
@@ -35,6 +35,10 @@ public struct Tensor<Scalar: TensorFlowScalar> {
   /// - Note: `handle` is public to allow user defined ops, but should not normally be used.
   public let handle: TensorHandle<Scalar>
 
+  /// An internal marker to identify scalar zero tensors, for use in optimizations.
```
pschuh, line 38: Could this live on the TensorHandle?

BradLarson: Sorry, kept meaning to check on this and got distracted. Would there be an advantage to having it on TensorHandle? I located it on Tensor because that's the point at which it impacts calculations. Only Tensor addition and subtraction are selectively converted to no-ops, so I didn't see a need to place this deeper within TensorHandle. The logic for marking zeroes would also need to be placed in Tensor.zero, so it made sense to me to have this all at the same level.

pschuh: This might fix your crasher as well. Also, another tensor type might handle zeros itself internally. In addition it makes the Tensor type bigger which makes it slower to copy around. We could also make this a completely different TensorHandle subclass.

BradLarson:
Closing this in favor of PR #1081, which works around the compiler crash.

Issue #1031, opened 2020-06-28
GroupNorm crashes on x10


s1ddok:
I've [tried to fix it](https://github.com/tensorflow/swift-apis/pull/1030), but it didn't work. I believe this is because eps is initialized on TFEager instead of x10 by default. Should tensor's scalar inits accept device as well?

BradLarson:
I believe this should be fixed in PR #1042, which I forgot to pull in yesterday but should show up in tonight's nightly builds. I tested it locally, and it seemed to allow GroupNorm and LayerNorm to work well with X10 in simple models.

Your original fix may not have worked due to the creation of the eager-mode tensor, followed by the copy. `Tensor(self.epsilon, deviceAndPrecisionLike: $0)` created a scalar Tensor directly on device, which is what I think you were looking for. Wrapping that in `withoutDerivative(at:)` further prevents scalarizing the epsilon value in the backward pass, which we know from unfortunate experience will really slow down X10, so I think this approach should be compatible and fast.

If this doesn't fix the issue once it's in the toolchain, let me know and I can take a deeper look into it.

BradLarson:
I'm assuming that this has been fixed, but if it still occurs in your tests, feel free to reopen this issue.

PR #1042, opened 2020-07-08
Making GroupNorm and LayerNorm more compatible with X10

BradLarson:
GroupNorm's local `epsilon` Tensor was being created on the default TF eager device, which led to a crash on X10 devices. This creates an appropriately placed Tensor for that value.

Additionally, wrapping the Tensor creation in `withoutDerivative(at:)` within both GroupNorm and LayerNorm prevents that value from being scalarized in the backwards pass, which would cut the X10 trace at that point. We've found that this has a significant performance impact within BatchNorm, so I've prevented that here.

This should fix issue #1031.

### DIFF
--- a/Sources/TensorFlow/Layers/Normalization.swift
+++ b/Sources/TensorFlow/Layers/Normalization.swift
@@ -136,6 +136,8 @@ public struct BatchNorm<Scalar: TensorFlowFloatingPoint>: Layer {
     }
     runningMean.value += (momentsMean - runningMean.value) * decayMomentum
     runningVariance.value += (momentsVariance - runningVariance.value) * decayMomentum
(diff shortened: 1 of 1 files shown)


mikowals:
Could there be comments added to explain why this is necessary?  

I tripped up earlier removing the code like this from batch norm and think it takes pretty thorough understanding of X10 and autodiff to reason why these are needed.

BradLarson:
@mikowals - Good point, we didn't even know this was an issue until we saw the traces being cut here. I've added notes around the three places that I know will require this right now. There are other locations using `withoutDerivative(at:)` here where I believe it should be safe to remove that without impacting traces, and I've left those alone. They will need to have traces examined before and after to make sure, but if there's no change in the backwards pass we should be good to remove those.

PR #1045, opened 2020-07-22
build: build PythonKit rather than expect it to be provided

compnerd:
This removes the external dependency on PythonKit when building with
CMake.  We now build and vend PythonKit as a dependency when Python
support is enabled.  This will enable extracting PythonKit from the
toolchain and makes tensorflow-swift-apis more self-contained.


ematejska:
When the toolchain is built, we specify the PythonKit build here: https://github.com/apple/swift/blob/0b34b4352646a6337d711d9c9a3930ec1374c276/utils/update_checkout/update-checkout-config.json#L374.  When we branch, sometimes we have to pin the versions.  How will this affect toolchain builds, especially the branched ones?  Anything extra to do?

compnerd:
@ematejska - there is more to do - we can pin versions now in swift-apis and I will be following up to remove the PythonKit checkout from there.

PR #1051, opened 2020-07-30
build: bump TensorFlow to 2.3.0

compnerd:
The TensorFlow 2.3.0 release is GA'ed.  Update the version that we use.

### DIFF
--- a/Tests/TensorFlowTests/OperatorTests/DatasetTests.swift
+++ b/Tests/TensorFlowTests/OperatorTests/DatasetTests.swift
@@ -113,7 +113,7 @@ final class DatasetTests: XCTestCase {
     let dataset = Dataset(elements: scalars)
     let shuffled = dataset.shuffled(sampleCount: 5, randomSeed: 42,
                                     reshuffleForEachIterator: false)
-    XCTAssertEqual([2, 1, 3, 4, 2], shuffled.map { $0.scalar! })
+    XCTAssertEqual([2, 1, 3, 4, 0], shuffled.map { $0.scalar! })
     let evens = shuffled.map { Tensor($0 % 2) .== Tensor(0) }
     XCTAssertEqual(evens.map { $0.scalar! }, [true, false, false, true, true])
   }
--- a/Tests/TensorFlowTests/OperatorTests/DatasetTests.swift
+++ b/Tests/TensorFlowTests/OperatorTests/DatasetTests.swift
@@ -111,10 +111,11 @@ final class DatasetTests: XCTestCase {
   func testMapToDifferentType() {
     let scalars = Tensor<Float>(rangeFrom: 0, to: 5, stride: 1)
     let dataset = Dataset(elements: scalars)
-    let shuffled = dataset.shuffled(sampleCount: 5, randomSeed: 42)
-    XCTAssertEqual([0, 4, 1, 3, 2], shuffled.map { $0.scalar! })
+    let shuffled = dataset.shuffled(sampleCount: 5, randomSeed: 42,
+                                    reshuffleForEachIterator: false)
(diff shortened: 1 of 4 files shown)


texasmichelle:
I believe that setting [`reshuffleForEachIterator`](https://github.com/tensorflow/swift-apis/blob/18b57ca0f660958d4f234a59dd9fb19743e7f5b9/Tests/TensorFlowTests/OperatorTests/DatasetTests.swift#L114) explicitly to false will resolve the test failure:
```
let shuffled = dataset.shuffled(sampleCount: 5, randomSeed: 42, reshuffleForEachIterator: false)
```

texasmichelle:
The failure is caused by shuffle behaving differently, which I suspect is caused by an upstream bugfix. `Datasets.shuffled()` [defaults to `reshuffleForEachIterator:  true`](https://github.com/tensorflow/swift-apis/blob/18b57ca0f660958d4f234a59dd9fb19743e7f5b9/Sources/TensorFlow/Operators/Dataset.swift#L167), which did not have any effect before.

Here's a reduction:
```
import TensorFlow

let scalars = Tensor<Float>(rangeFrom: 0, to: 5, stride: 1)
let dataset = Dataset(elements: scalars)
print("dataset: \(dataset.flatMap { $0.scalars })")
let shuffled = dataset.shuffled(sampleCount: 5, randomSeed: 42)
print("shuffled: \(shuffled.flatMap { $0.scalars })")
print("shuffled: \(shuffled.flatMap { $0.scalars })")
print("shuffled: \(shuffled.flatMap { $0.scalars })")
```
Nightlies:
```
dataset: [0.0, 1.0, 2.0, 3.0, 4.0]
shuffled: [0.0, 4.0, 1.0, 3.0, 2.0]
shuffled: [0.0, 4.0, 1.0, 3.0, 2.0]
shuffled: [0.0, 4.0, 1.0, 3.0, 2.0]
```
TF 2.3:
```
dataset: [0.0, 1.0, 2.0, 3.0, 4.0]
shuffled: [0.0, 4.0, 1.0, 3.0, 2.0]
shuffled: [3.0, 4.0, 1.0, 0.0, 2.0]
shuffled: [2.0, 4.0, 3.0, 0.0, 1.0]
```

Passing in `reshuffleForEachIterator:  false` gives us behavior that matches TF 2.2:

```
dataset: [0.0, 1.0, 2.0, 3.0, 4.0]
shuffled: [2.0, 1.0, 3.0, 4.0, 0.0]
shuffled: [2.0, 1.0, 3.0, 4.0, 0.0]
shuffled: [2.0, 1.0, 3.0, 4.0, 0.0]
```

texasmichelle:
I wonder why the shuffled values are different, given the same seed? It's probably fine just to update the expected values in the test.

Tests/TensorFlowTests/OperatorTests/DatasetTests.swift
```diff
@@ -111,10 +111,11 @@ final class DatasetTests: XCTestCase {
   func testMapToDifferentType() {
     let scalars = Tensor<Float>(rangeFrom: 0, to: 5, stride: 1)
     let dataset = Dataset(elements: scalars)
-    let shuffled = dataset.shuffled(sampleCount: 5, randomSeed: 42)
-    XCTAssertEqual([0, 4, 1, 3, 2], shuffled.map { $0.scalar! })
+    let shuffled = dataset.shuffled(sampleCount: 5, randomSeed: 42,
+                                    reshuffleForEachIterator: false)
+    XCTAssertEqual([2, 1, 3, 4, 2], shuffled.map { $0.scalar! })
```
texasmichelle, line 116:
```suggestion
    XCTAssertEqual([2, 1, 3, 4, 0], shuffled.map { $0.scalar! })
```

PR #1054, opened 2020-08-05
build: add swift-numerics to the build

compnerd:
This allows building swift-numerics as part of tensorflow-swift-apis.
The option `-D ENABLE_SWIFT_NUMERICS` allows user control over this
additional dependency.


compnerd:
Yeah, we will want it in the s-p-m based builds as well, but this gets some of the build work going.  The idea is to actually have the elementary functions come from swift-numerics, so this just sets up the infrastructure for that.

PR #1062, opened 2020-08-18
Change `Dense.bias` type to `Optional`

efremale:
`TrivialModelTests.swift` test is currently failing.

### DIFF
--- a/Sources/TensorFlow/Layers/Dense.swift
+++ b/Sources/TensorFlow/Layers/Dense.swift
@@ -28,17 +28,24 @@ import _Differentiation
 public struct Dense<Scalar: TensorFlowFloatingPoint>: Layer {
   /// The weight matrix.
   public var weight: Tensor<Scalar>
-  /// The bias vector.
-  public var bias: Tensor<Scalar>
+  /// The optional bias vector.
+  public var optionalBias: Tensor<Scalar>?
   /// The element-wise activation function.
   @noDerivative public let activation: Activation
   /// Indicates whether this is a batched dense layer.
   @noDerivative internal let batched: Bool
-  /// Workaround optionals not being handled by AD
-  @noDerivative private let useBias: Bool
 
   /// The element-wise activation function type.
   public typealias Activation = @differentiable (Tensor<Scalar>) -> Tensor<Scalar>
+
+  /// The bias vector.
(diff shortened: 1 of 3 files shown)


Sources/TensorFlow/Layers/Dense.swift
```diff
@@ -111,4 +112,56 @@ extension Dense {
...
+
+extension Optional: KeyPathIterable {
+  public var allKeyPaths: [PartialKeyPath<Self>] {
+    if self != nil {
+      return [\.!]
+    }
+    return []
+  }
+
+  public typealias AllKeyPaths = [PartialKeyPath<Self>]
```
dan-zheng, line 166: Please remove this extension, since it's added in https://github.com/apple/swift/pull/33992.

Sources/TensorFlow/Layers/Dense.swift
```diff
@@ -111,4 +112,56 @@ extension Dense {
...
+    weight: Tensor<Scalar>,
+    bias: Tensor<Scalar>
+  ) {
+    self.init(weight: weight, optionalBias: .init(bias))
+  }
+  
+  /// The bias vector.
+  ///
+  /// - Note: returns `Tensor.zero` if the underlying `optionalBias`  does not exist.
+  //@differentiable
```
dan-zheng, line 151: Removing this `@differentiable` attribute should be fine, since `Dense.TangentVector` isn't differentiated in practice.
```suggestion
```

saeta:
Could you add a test to verify that this will work with `CopyableToDevice`? Thanks!

dan-zheng:
CI fails for a test that seems related to key path iteration:

<details>
<p>

```
```

</p>
</details>

---

Closing for now because optional differentiation isn't fully supported (SR-13700), so usability wouldn't be great (if users can't differentiate through `dense.bias ?? default` nil coalescing).

Investigating the test failure on this branch would be a great step towards a smooth migration to using `Optional` differentiation in deep learning APIs and models (e.g. [ResNet](https://github.com/tensorflow/swift-models/blob/c6366b70fc1392c4f1d6ba1194047f6589724eee/Models/ImageClassification/ResNet.swift#L57-L62)) instead of the current, less efficient workarounds!

PR #1063, opened 2020-08-19
build: bump to CMake 3.16, centralise artifacts

compnerd:
Put all the libraries into lib, all the binaries into bin.  This bumps
up the CMake requirement to 3.16.0 to enable the co-located artifacts.
It should make it easier for people to find the artifacts in the build
tree.


compnerd:
Seems that we didn't document the requirements (I suppose because the intent was that people would be using s-p-m).  I don't think that anything should be needed for that, the layout of the tree should be internal to the project.

PR #1067, opened 2020-08-28
Add model summary

texasmichelle:
Supports model summaries by adding an additional protocol requirement to `Module` and `Layer`. This enables the addition of tensor annotations within `callAsFunction()` and lays the groundwork for additional hooks.

Full explanation: [design doc](https://docs.google.com/document/d/1hEhMiwLtuzsN3RvIC3FAh6NvtTimU8o_qdzMkGvntVg).

- [x] Add `forward()` protocol requirement to `Module` and `Layer`
- [x] Add default implementations of `forward()` and `callAsFunction()` that add annotations to output tensors
- [x] Add a link to the design doc
- [x] Update all native layer types except `Recurrent`
- [x] Add summary formatting


pschuh (approved):

Sources/TensorFlow/Layers/Dropout.swift
```diff
@@ -68,24 +73,28 @@ public struct Dropout<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
...
 /// The noise added always has mean zero, but has a configurable standard deviation.
 public struct GaussianNoise<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
+  /// The input type of the layer.
+  public typealias Input = Tensor<Scalar>
+  /// The output type of the layer.
+  public typealias Output = Tensor<Scalar>
   public typealias TangentVector = EmptyTangentVector
 
-  @noDerivative public let standardDeviation: Tensor<Scalar>
+  @noDerivative public let standardDeviation: Input
```
dan-zheng, line 82: With changes to the `Module` and `Layer` protocols in this PR, it seems that conforming types now need to explicitly specify `Input` and `Output` typealiases. (Previously, the compiler could infer `Input` and `Output` associated types for these conforming types, but that doesn't seem to work anymore.)

Adding `Input` and `Output` typealiases is fine, but how about keeping `Tensor<Scalar>` occurrences explicit instead of renaming them to `Input` and `Output`?
* `Input` and `Output` are closely tied to the `func callAsFunction(_: Input) -> Output` protocol requirement, so it's a bit strange to use them outside of `func callAsFunction`, like as the type of stored properties (which are neither layer inputs nor outputs).
* `Tensor<Scalar>` (a specific concrete type) also conveys more information than `Input` and `Output` (associated types), so it's easier to understand usages like `Tensor<Scalar>(randomNormal: ...)` than `Input(randomNormal: ...)`.

texasmichelle: That makes sense, I'll revert those changes everywhere outside `callAsFunction()`.

dan-zheng: Thank you!

dan-zheng (approved):

Sources/TensorFlow/Layers/Dropout.swift
texasmichelle: I was actually able to remove these typealiases altogether, which makes this PR much smaller 👍

dan-zheng: Yay, even better!

8bitmp3:
This is super helpful, thanks @texasmichelle

PR #1069, opened 2020-08-31
Revert non-active try-apply differentiation workarounds

efremale:
Since TF-433 was fixed, we can remove all the workarkounds for (non-active) try-apply.
For example, calls to `Array.map(_:)`.


efremale:
@googlebot I signed it!

dan-zheng:
Tests fail because the toolchains used for running CI haven't been updated in a while due to build failures. Once the build failures are fixed and toolchains are updated, CI should pass.

dan-zheng:
Test pass now that the toolchains used for running CI have been updated!

dan-zheng:
Verifying that tests pass one more time.

PR #1076, opened 2020-09-10
Convert more xla ops to be generated from a spec.

pschuh:
Added ops include: log_softmax logical_cast softmax tf_OneHot tf_StatelessRandomUniform tf_UnsortedSegmentSum truncated_normal update_slice where xla_slice

### DIFF
--- a/Sources/CX10/xla_tensor_ops_wrapper_generated.cc.inc
+++ b/Sources/CX10/xla_tensor_ops_wrapper_generated.cc.inc
@@ -132,8 +132,8 @@ class All : public Node {
        return XlaHelpers::ShapeOfXlaOp(result);
      },
              /*num_outputs=*/1, xla::util::MHash(dims, keep_reduced_dimensions)),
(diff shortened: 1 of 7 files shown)


Sources/CX10/xla_tensor_ops_wrapper_generated.cc.inc
```diff
@@ -2038,14 +2128,127 @@ class Tanh : public Node {
  private:
 };
 
+class TfOnehot : public Node {
```
texasmichelle, line 2131: Where does the capitalization for these class name originate? It's a bit off.

pschuh:
```
def snake_to_camel(name):
  return "".join(map(lambda x: x.capitalize(),name.split("_")))
```

pschuh:
@googlebot rescan

texasmichelle (approved):

PR #1077, opened 2020-09-11
Accurate `Tensor.device` for TFEager backends

texasmichelle:
This modifies `Tensor.device` for TFEager backends to retrieve device details from the underlying eager libraries. This is more accurate than returning `Device.defaultTFEager` and reflects the actual location of eager operations.

This was tested by building a CUDA toolchain and verifying that `TFE_TensorHandleDeviceName()` returns:
```
/job:localhost/replica:0/task:0/device:GPU:0
```

and a GPU device is returned by `Tensor.device`:
```
Device(kind: .GPU, ordinal: 0, backend: .TF_EAGER)
```

On a device without GPU, `TFE_TensorHandleDeviceName()` returns:
```
/job:localhost/replica:0/task:0/device:CPU:0
```

and a CPU device is returned by `Tensor.device`:
```
Device(kind: .CPU, ordinal: 0, backend: .TF_EAGER)
```

Fixes [tensorflow/swift#524](https://github.com/tensorflow/swift/issues/524).


Sources/x10/swift_bindings/Device.swift
```diff
@@ -146,8 +150,49 @@ public struct Device {
...
+    let device = Device(kind: kind, ordinal: defaultOrdinal, backend: .TF_EAGER)
+    let tensor = Tensor<Float>(zeros: [1], on: device)
+    let handle = tensor.handle._cTensorHandle
+    let status = TF_NewStatus()
+
+    // Find out what the underlying libraries think the default is.
+    if let cString = TFE_TensorHandleDeviceName(handle, status) {
+      checkOk(status)
+
+      // TODO: What's the best way to deallocate this memory?
```
texasmichelle, line 165: @pschuh As-is will cause memory leaks, right?

Sources/x10/swift_bindings/Device.swift
```diff
@@ -146,8 +150,49 @@ public struct Device {
...
+      // defer { DeleteString(str) }
+      let tfDeviceName = String(cString: cString)
+
+      // Parse type and ordinal from a string with the expected syntax:
+      //   /job:localhost/replica:0/task:0/device:CPU:0
+      let pattern = ".+device:(.+):(\\d+)$"
+      let regex = try! NSRegularExpression(pattern: pattern)
```
pschuh, line 172: This string parsing looks expensive. Might want to double check that this doesn't happen too much.

pschuh (approved):

Sources/x10/swift_bindings/Device.swift
```diff
@@ -146,8 +150,49 @@ public struct Device {
...
+      if let match = regex.firstMatch(in: tfDeviceName, range: nsrange) {
+        if let kindRange = Range(match.range(at: 1), in: tfDeviceName) {
+          switch String(tfDeviceName[kindRange]).uppercased() {
+          case "CPU":
+            kind = .CPU
+          case "GPU":
+            kind = .GPU
+          case "TPU":
+            kind = .TPU
+          case "REMOTE_TPU":
```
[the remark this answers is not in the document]
texasmichelle: Removed.

Sources/x10/swift_bindings/Device.swift
texasmichelle: I can use the status from the execution context instead

Sources/TensorFlow/Core/Tensor.swift
```diff
@@ -768,7 +769,40 @@ extension Tensor: Differentiable & EuclideanDifferentiable where Scalar: TensorF
...
+          let status = _ExecutionContext.global.status
+
+          // Find out what the underlying libraries think the default is.
+          if let cString = TFE_TensorHandleDeviceName(handle._cTensorHandle, status) {
+            checkOk(status)
+            let tfDeviceName = String(cString: cString)
+
+            // Parse type and ordinal from a string with the expected syntax:
+            //   /job:localhost/replica:0/task:0/device:CPU:0
+            let pattern = ".+device:(.+):(\\d+)$"
```
pschuh, line 783: Maybe break this String -> Device out as a separate function?
Also, I'm concerned that the string parsing will be expensive. This function is called a lot. (whenever there is a scalar constant or anything like that). I think it would be best to see if you can add your own TFE_TensorHandleDevice_Type and TFE_TensorHandleDevice_Id. Some benchmarking results might work instead.

texasmichelle:
Different approach coming. Closing this PR

PR #1081, opened 2020-09-15
Tagging zero scalar tensors.

dan-zheng:
A fork of https://github.com/tensorflow/swift-apis/pull/1022 with an attempt to work around SR-13263: a debug info generation crash.


BradLarson (approved):
This looks like it works! Thanks, this will be a huge step forward for performance.

PR #1088, opened 2020-09-21
Add squeezing with no dims.

pschuh:
Should fix: https://github.com/tensorflow/swift/issues/537


saeta (approved):
Thanks @pschuh !

saeta:
Note: it looks like some of the tests are failing...

PR #1095, opened 2020-10-02
Convert some xla ops to have generated xla wrappers.

pschuh:


Sources/x10/swift_bindings/apis/RawOpsManual.swift
```diff
@@ -30,7 +30,7 @@ public enum _RawXLA {
     dims.map { $0 < 0 ? $0 + rank : $0 }
   }
 
-  private static func checkSameDevice(
+  static func checkSameDevice(
```
texasmichelle, line 33: Is this changing or preserving the scope and is it intentional? It would be good to be explicit here.

texasmichelle (approved):

Sources/x10/swift_bindings/apis/RawOpsManual.swift
pschuh: This is private to internal. For some reason this isn't available in extensions

Issue #1098, opened 2020-10-05
Remaining Bidirectional RNN merge modes


kongzii:
After https://github.com/tensorflow/swift-apis/pull/946 there is support for BiRNN with `concatenate` and `sum` mode. It would be awesome to have remaining merge modes from keras: ave, mul and none.

kongzii:
Could someone assign this issue to me, please?

saeta:
Thanks!

PR #1101, opened 2020-10-08
Add average, multiply and stack modes for BiRNNs

kongzii:
This will close https://github.com/tensorflow/swift-apis/issues/1098, more specifically, it adds rest of merge modes available in Keras in bidirectional recurrent layers.

- `ave` as `average`
- `mul` as `multiply`
- `none` as `stack`

If you can think of better names I will happily edit them :) Thanks for reviews!


xihui-wu (approved):
LGTM!

PR #1104, opened 2020-10-09
Refactor Device APIs to be pointer based.

pschuh:
This will allow a different implementation for TPU, CPU and GPU devices.


xihui-wu (approved):
LGTM

PR #1107, opened 2020-10-14
Add per-weight Adam to x10 optimizers

xihui-wu:
The code is mostly copied from https://github.com/tensorflow/swift-models/blob/master/Models/Text/WeightDecayedAdam.swift. The intent of this change is just to move that piece of code to apis here.


texasmichelle (approved):
It's great to see this moving into swift-apis! Thank you @xihui-wu & @shadaj!

PR #1108, opened 2020-10-15
X10 source cleanup

texasmichelle:
Remove hand-coded source that is now autogenerated.


pschuh (approved):

pschuh (commented):
I think this can be staged into a few changes. Maybe:
- Remove view.{h,cpp}
- Remove extra ops from tensor_methods.cpp.
- Delete op implementation.

Issue #1094, opened 2020-09-30
GPU memory growth and logging level flags aren't being set when X10 is used before eager mode


BradLarson:
If you are working with the GPU as an accelerator and you use a manually specified or default XLA device before accessing the eager-mode runtime, the next access to the eager-mode runtime will consume all available GPU memory. TensorFlow's default is to consume all available memory, but that default can be overridden by [setting `gpuMemoryAllowGrowth` to true](https://github.com/tensorflow/swift-apis/blob/15435771466749d96857bc508efa646060b31d94/Sources/TensorFlow/Core/Runtime.swift#L273), which will switch the runtime to only allocate memory as needed.

A trivial access to the eager-mode runtime, such as

```swift
_ = _ExecutionContext.global
```

before any reference to `Device.defaultXLA`, etc. leads to the proper setting of the memory growth and logging parameters. The initialization of the eager-mode global `_ExecutionContext` sets these parameters [via `TFE_ContextOptionsSetConfig()`](https://github.com/tensorflow/swift-apis/blob/15435771466749d96857bc508efa646060b31d94/Sources/TensorFlow/Core/Runtime.swift#L275).

Somehow, if an X10 device is initialized first, the next access to the eager-mode runtime does not obey these settings. I proposed a crude workaround in PR #1017 (closed), but we should find a way to set this correctly within the X10 computation client.

This would allow for the elimination of workaround code across swift-models and elsewhere.

xihui-wu:
As synced with @pschuh offline, we are seeking a solution to share the allocator setting between eager and x10 client. For x10 client, Parker has been planing to migrate to local_computation_client for GPU in OSS, so the allocator fix would go with that.

BradLarson:
As of PRs #1106 and #1109, I can confirm that this has been fixed and the workaround is no longer needed.

PR #1109, opened 2020-10-16
Make local backend default for GPU.

pschuh:


xihui-wu (approved):
Thanks! This should fix https://github.com/tensorflow/swift-apis/issues/1094!

PR #1114, opened 2020-10-29
Tests: add a workaround for type deduction

compnerd:
```
/SourceCache/tensorflow-swift-apis/Tests/TensorFlowTests/TensorAutoDiffTests.swift:616:17: error: cannot convert value of type 'Tensor<Float>.TangentVector' (aka 'Tensor<Float>') to expected argument type 'Tensor<Double>'
	assertEqual(gradient(at: [-1, 0, 1], in: f), [0.1966119, 0.25, 0.1966119], accuracy: 0.0001)
        ^
```

For some reason the gradient is being deduced as `Tensor<Double>` which
does not match the expectation of a `Tensor<Float>`.  Provide an
explicit conversion to workaround the issue.


texasmichelle (approved):
How about adding a TODO to look at the root cause?

PR #1123, opened 2020-11-12
build: make the `CTensorFlow` module more SPM friendly

compnerd:
Add the structure expected by SPM to enable building the CTensorFlow
module as part of a Swift Package.  It is no longer possible to build
this package without X10 support.


pschuh:
Feel free to delete any of the #ifdefs related to X10 support. I don't think we do any builds without x10.

pschuh (approved):

PR #1125, opened 2020-11-16
Adjusting or adding tolerances on tests

BradLarson:
These tests were failing on a test platform due to slight numerical differences. Several of the tests just checked for floating point equality, so tolerances were added to those. Others had results that were just outside of the allowed tolerance, so the tolerances have been slightly relaxed for those.


texasmichelle (approved):
Confirmed that this resolves precision errors on GPU. Thanks, Brad!!

PR #1126, opened 2020-11-18
fix dimension order in TransposedConv2D.init doc comment

marcrasi:
As discovered in https://bugs.swift.org/projects/TF/issues/TF-1349, the documented order is wrong.


BradLarson (approved):

Shashi456:
If there's a need for checking dimension order across the convolutional layers, I'd be happy to take a look ( I wrote a substantial amount of them)

marcrasi:
> If there's a need for checking dimension order across the convolutional layers, I'd be happy to take a look ( I wrote a substantial amount of them)

That would be great! I didn't look at the others but I expect that some of them might also have doc comments that were copy-pasted and not updated to perfectly match their actual code.

PR #1129, opened 2020-11-27
Adding dilations argument inside DepthwiseConv2D API

diwakar-vsingh:
Extended the depthwiseConv2D struct to control dilation rate as well.

### DIFF
--- a/Sources/TensorFlow/Layers/Convolutional.swift
+++ b/Sources/TensorFlow/Layers/Convolutional.swift
@@ -909,12 +909,15 @@ public struct SeparableConv1D<Scalar: TensorFlowFloatingPoint>: Layer {
   public var pointwiseFilter: Tensor<Scalar>
   /// The bias vector.
   public var bias: Tensor<Scalar>
+
   /// The element-wise activation function.
   @noDerivative public let activation: Activation
   /// The strides of the sliding window for spatial dimensions.
   @noDerivative public let stride: Int
   /// The padding algorithm for convolution.
   @noDerivative public let padding: Padding
+  /// The dilation factor for the temporal dimension.
+  @noDerivative public let dilation: Int
   /// Note: `useBias` is a workaround for TF-1153: optional differentiation support.
   @noDerivative private let useBias: Bool
 
@@ -933,13 +936,15 @@ public struct SeparableConv1D<Scalar: TensorFlowFloatingPoint>: Layer {
   ///   - activation: The element-wise activation function.
   ///   - strides: The strides of the sliding window for spatial dimensions.
   ///   - padding: The padding algorithm for convolution.
+  ///   - dilation: The dilation factor for the temporal dimension.
   public init(
     depthwiseFilter: Tensor<Scalar>,
     pointwiseFilter: Tensor<Scalar>,
     bias: Tensor<Scalar>? = nil,
     activation: @escaping Activation = identity,
     stride: Int = 1,
-    padding: Padding = .valid
+    padding: Padding = .valid,
+    dilation: Int = 1
   ) {
     self.depthwiseFilter = depthwiseFilter
     self.pointwiseFilter = pointwiseFilter
@@ -947,6 +952,7 @@ public struct SeparableConv1D<Scalar: TensorFlowFloatingPoint>: Layer {
     self.activation = activation
     self.stride = stride
     self.padding = padding
+    self.dilation = dilation
     useBias = (bias != nil)
   }
 
@@ -960,7 +966,8 @@ public struct SeparableConv1D<Scalar: TensorFlowFloatingPoint>: Layer {
       input.expandingShape(at: 1),
       filter: depthwiseFilter.expandingShape(at: 1),
       strides: (1, stride, stride, 1),
-      padding: padding)
+      padding: padding,
+      dilations: (1, dilation, dilation, 1))
     let x = conv2D(
       depthwise,
       filter: pointwiseFilter.expandingShape(at: 1),
@@ -978,8 +985,9 @@ extension SeparableConv1D {
   /// - Parameters:
   ///   - depthwiseFilterShape: The shape of the 3-D depthwise convolution kernel.
   ///   - pointwiseFilterShape: The shape of the 3-D pointwise convolution kernel.
-  ///   - strides: The strides of the sliding window for temporal dimensions.
+  ///   - stride: The stride of the sliding window for temporal dimensions.
   ///   - padding: The padding algorithm for convolution.
+  ///   - dilation: The dilation factor for the temporal dimension.
   ///   - activation: The element-wise activation function.
   ///   - filterInitializer: Initializer to use for the filter parameters.
   ///   - biasInitializer: Initializer to use for the bias parameters.
@@ -988,6 +996,7 @@ extension SeparableConv1D {
     pointwiseFilterShape: (Int, Int, Int),
     stride: Int = 1,
     padding: Padding = .valid,
+    dilation: Int = 1,
     activation: @escaping Activation = identity,
     useBias: Bool = true,
     depthwiseFilterInitializer: ParameterInitializer<Scalar> = glorotUniform(),
@@ -1006,7 +1015,8 @@ extension SeparableConv1D {
       bias: useBias ? biasInitializer([pointwiseFilterShape.2]) : nil,
       activation: activation,
       stride: stride,
-      padding: padding)
+      padding: padding,
+      dilation: dilation)
   }
 }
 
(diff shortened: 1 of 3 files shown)


Shashi456:
Can you make sure you've run `swift build`?

diwakar-vsingh:
I have tested it locally and everything looks fine.

BradLarson:
Thanks for exposing this parameter at a higher level.

Unfortunately, it looks like there are two locations within Convolutional.swift that still need to be updated with the new `dilations:` parameter (lines 963 and 1076) before this will build cleanly. Also, would it make sense to add a quick test case to [`testDepthwiseConv2D()`](https://github.com/tensorflow/swift-apis/blob/master/Tests/TensorFlowTests/LayerTests.swift#L287) to verify that a non-1 dilation behaves as expected?

brettkoonce:
Does [the requirement that strides and dilations not both be set](https://github.com/tensorflow/tensorflow/blob/abd4c4710933cd52d090e2b8f59b04db1e421ca6/tensorflow/python/keras/layers/convolutional.py#L2258) from Keras apply here as well?

BradLarson:
@brettkoonce - That's a good point. I wonder if a precondition to enforce that might be needed.

diwakar-vsingh:
> Unfortunately, it looks like there are two locations within Convolutional.swift that still need to be updated with the new `dilations:` parameter (lines 963 and 1076) before this will build cleanly.

Thanks for pointing that out. I mainly focused on DepthwiseConv2D and forgot to add dilations argument inside SeparableConv2D. I just modified the Convolutional.swift file as required.

diwakar-vsingh:
> Also, would it make sense to add a quick test case to [`testDepthwiseConv2D()`](https://github.com/tensorflow/swift-apis/blob/master/Tests/TensorFlowTests/LayerTests.swift#L287) to verify that a non-1 dilation behaves as expected?

Will do it as soon as possible.

diwakar-vsingh:
> Does [the requirement that strides and dilations not both be set](https://github.com/tensorflow/tensorflow/blob/abd4c4710933cd52d090e2b8f59b04db1e421ca6/tensorflow/python/keras/layers/convolutional.py#L2258) from Keras apply here as well?

That's quite a good point. We might need to enforce that both strides and dilations cannot be greater than 1 at the same time.

BradLarson:
I just realized that the CI had gotten stuck, so I re-ran it. From the compilation failures, it looks like `/swift-apis/Tests/x10/ops_test.swift` might be missing the new parameters on lines 1079 and 1094.

Were you still intending to also add the preconditions and tests for the new parameter? No rush, just wanted to make sure we were still waiting on those after these compilation issues were taken care of.

diwakar-vsingh:
@BradLarson I just noticed that `_Raw.depthwiseConv2dNative()` implemented in [depthwiseConv2dNative()](https://github.com/tensorflow/swift-apis/blob/main/Sources/x10/swift_bindings/apis/RawOpsManual.swift) is not working as anticipated. It is not giving me the correct result when I change the dilations to something other than [1, 1, 1, 1]. I'm comparing the result with `tf.nn.depthwise_conv2d()`. More specifically, no matter what I input as dilations rate, `_Raw.depthwiseConv2dNative()` gives the same result as using the default dilations argument i.e. [1, 1, 1, 1]. I noticed this when I was adding a quick test case to [testDepthwiseConv2D()](https://github.com/tensorflow/swift-apis/blob/master/Tests/TensorFlowTests/LayerTests.swift#L287) to verify that a non-1 dilation behaves as expected.

diwakar-vsingh:
Test in Swift for Tensorflow
```
let input = Tensor(shape: [1, 2, 8, 2], scalars: (0..<32).map(Float.init)) 
let filter = Tensor(shape: [2, 2, 2, 1], scalars: (0..<8).map(Float.init))
let bias = Tensor<Float>([1, 2])
let strides = (1, 1, 1, 1)
let dilations = (1, 2, 2, 1)

var layer = _Raw.depthwiseConv2dNative(
    input,
    filter: filter,
    strides: [Int32(strides.0), Int32(strides.1), Int32(strides.2), Int32(strides.3)],
    padding: .same,
    dilations: [Int32(dilations.0), Int32(dilations.1), Int32(dilations.2), Int32(dilations.3)])
print("layer: \n", layer + bias)
print(layer.shape)
```

Test in TF2.x
```
input = tf.reshape(tf.range(32, dtype=tf.float32), [1, 2, 8, 2])
filter = tf.reshape(tf.range(8, dtype=tf.float32), [2, 2, 2, 1])
bias = tf.constant([1., 2.])

output = tf.nn.depthwise_conv2d(input=input,
                                filter=filter,
                                strides=[1, 1, 1, 1],
                                data_format="NHWC",
                                padding="SAME",
                                dilations=[2,2]) + bias
print('output: ', output)
print(output.shape)
```

Shashi456:
https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/depthwise-conv2d-native

Take a look at the above for potential reasons @diwakar-vsingh, Maybe dilations more than 1 are not supported in the api yet?

>   ///     - dilations: 1-D tensor of length 4.  The dilation factor for each dimension of
  ///         `input`. If set to k > 1, there will be k-1 skipped cells between each filter
  ///         element on that dimension. The dimension order is determined by the value of
  ///         `data_format`, see above for details. Dilations in the batch and depth
  ///         dimensions must be 1.

diwakar-vsingh:
> /// - dilations: 1-D tensor of length 4. The dilation factor for each dimension of
> /// `input`. If set to k > 1, there will be k-1 skipped cells between each filter
> /// element on that dimension. The dimension order is determined by the value of
> /// `data_format`, see above for details. Dilations in the batch and depth
> /// dimensions must be 1.

This is the exact definition of dilations. The last line indicates that it should be something like this [1, rate.1, rate.2, 1] which is correct.

diwakar-vsingh:
@BradLarson These checks have been running for over a month now. Could you take a look?

BradLarson:
@diwakar-vsingh - Sorry about that, wasn't sure if you were looking to add more code and I forgot to kick the CI system. I restarted the job.

It looks like it's failing in `Tests/x10/ops_test.swift` at lines 1079, 1094, and 1100 due to changes in the argument types. Those tests might need to be adjusted to the new parameters as well.

diwakar-vsingh:
> It looks like it's failing in `Tests/x10/ops_test.swift` at lines 1079, 1094, and 1100 due to changes in the argument types. Those tests might need to be adjusted to the new parameters as well.

There was an error in the dilation argument inside depthwiseConv2D function. I just fixed that.

BradLarson:
@diwakar-vsingh - The status isn't being reported right, but there are two places in swift-models that this change would break, which is causing CI tests to fail. [This line](https://github.com/tensorflow/swift-models/blob/main/Examples/GrowingNeuralCellularAutomata/CellRule.swift#L57) in the GrowingNeuralCellularAutomata example uses `depthwiseConv2D()`, as does [this line](https://github.com/tensorflow/swift-models/blob/main/Examples/Shallow-Water-PDE/TensorConvSolution.swift#L159) in the Shallow Water PDE example. Those both break with this change in the API.

Your best best might be to create a `depthwiseConv2D()` version without the `dilations` parameter that calls your `depthwiseConv2D()` implementation with that parameter, and then mark the old signature as deprecated in favor of the new one. Then we can follow up with a change to swift-models to match this new API once swift-apis passes CI and this change is integrated.

diwakar-vsingh:
> The status isn't being reported right, but there are two places in swift-models that this change would break, which is causing CI tests to fail. [This line](https://github.com/tensorflow/swift-models/blob/main/Examples/GrowingNeuralCellularAutomata/CellRule.swift#L57) in the GrowingNeuralCellularAutomata example uses `depthwiseConv2D()`, as does [this line](https://github.com/tensorflow/swift-models/blob/main/Examples/Shallow-Water-PDE/TensorConvSolution.swift#L159) in the Shallow Water PDE example. Those both break with this change in the API.

To fix this problem, what I did was to add the default value of dilation argument as (1, 1, 1, 1) inside 'depthwiseConv2D()' func in [NN.swift](https://github.com/tensorflow/swift-apis/pull/1129/commits/b2b8f98aaeb1fc4a56827e06c3cbfe84782cbeb3). This way you don't have to explicitly call for dilation argument inside the 'depthwiseConv2D()' func defined in swift-models examples and hence previous implementations won't break.

It looks like all the tests passed this time. Could you take a look at it now?

BradLarson (approved):
Sorry for all the back-and-forth. Now that things are all green, I think we're good to bring this in.

Thanks for the work in improving this interface.

diwakar-vsingh:
It's been my pleasure to contribute. I'll open up another PR in few weeks with DeepLab implementation.

diwakar-vsingh:
@BradLarson I created the new colab notebook to test out the updated 'depthwiseConv2D()' function. However, I am not able to add a dilation argument inside it. When we import TensorFlow, don't we get the most up-to-date APIs?

BradLarson:
@diwakar-vsingh - Right now, Colab uses Swift for TensorFlow toolchains with tensorflow/swift-apis as part of them. That means that you only get the swift-apis capabilities that correspond with those present at the time the toolchain was compiled (in this case, the 0.12 release). It's a difficult process to update the Colab images, so we only do that on stable releases.

If you instead develop locally, using our nightly toolchains, you'll get the latest APIs included in those. Or build swift-apis using CMake locally from a Swift.org toolchain.

PR #1139, opened 2020-12-07
Reflection based PointwiseMultiplicative and VectorProtocol.

pschuh:


Sources/TensorFlow/Core/VectorProtocol.swift
```diff
@@ -0,0 +1,118 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import _Differentiation
+
+#if TENSORFLOW_USE_STANDARD_TOOLCHAIN
+@_spi(Reflection) import Swift
+
+public protocol _VectorProtocol {
+  typealias VectorSpaceScalar = Float
```
marcrasi, line 21: Do you mean `associatedtype`?

Sources/TensorFlow/Core/PointwiseMultiplicative.swift
```diff
@@ -0,0 +1,128 @@
...
+
+import _Differentiation
+
+#if TENSORFLOW_USE_STANDARD_TOOLCHAIN
+@_spi(Reflection) import Swift
+
+infix operator .*: MultiplicationPrecedence
+infix operator .*=: AssignmentPrecedence
+
+public protocol _PointwiseMultiplicative {
```
marcrasi, line 23: A doc comment explaining what this is for would be good.

Sources/TensorFlow/Core/PointwiseMultiplicative.swift
```diff
@@ -0,0 +1,128 @@
...
+  static func pointwiseMult<Root>(_ lhs: inout Root, _ rhs: Root, _ kp: PartialKeyPath<Root>)
+  static func setOne<Root>(_ out: inout Root, _ kp: PartialKeyPath<Root>)
+  static func setReciprocal<Root>(_ out: inout Root, _ kp: PartialKeyPath<Root>)
```
marcrasi, line 26: Also doc comments on these would be good.

Sources/TensorFlow/Core/VectorProtocol.swift
```diff
@@ -0,0 +1,118 @@
...
+  mutating func add(_ x: VectorSpaceScalar)
+  mutating func subtract(_ x: VectorSpaceScalar)
+  mutating func scale(by scalar: VectorSpaceScalar)
+}
```
marcrasi, line 25: And doc comments on all these.

marcrasi (changes requested):

Sources/TensorFlow/Core/VectorProtocol.swift
pschuh: _VectorProtocol cannot have any associated type or Self constrained that would prevent doing type as? _VectorProtocol.Type.

marcrasi: Oh I see. So this only works with Float now? That seems fine to me. `VectorProtocol` isn't very good anyways and we should stop using it eventuallly.

Does this make the `associatedtype VectorSpaceScalar` inside of `VectorProtocol` irrelevant? Can you remove it so that it's not confusing?

pschuh: Cannot. It is needed to be associatedtype so it can be in the where clauses until we can delete them.

marcrasi: Oh, ok

marcrasi (approved):

PR #1140, opened 2020-12-08
Keypath reflection default implementation.

pschuh:


Sources/TensorFlow/Core/KeyPathIterable.swift
```diff
@@ -0,0 +1,174 @@
...
+
+public extension KeyPathIterable {
+  var allKeyPaths: [PartialKeyPath<Self>] {
+    var out = [PartialKeyPath<Self>]()
+    _forEachFieldWithKeyPath(of: Self.self, options: .ignoreUnknown) { name, kp in
+      out.append(kp)
+      return true
+    }
+    return out
+  }
```
dan-zheng, line 51: Note: sneaky default implementation - types still must declare a conformance to `KeyPathIterable` to use `KeyPathIterable.allKeyPaths`, unlike [this alternative implementation](https://github.com/apple/swift-evolution-staging/tree/reflection).

This seems reasonable though for maintaining the same API surface!

dan-zheng (approved):
Would you like to add any tests for this new `KeyPathIterable.allKeyPaths` implementation?

- Upstreamed from [`apple/swift:test/stdlib/KeyPathIterable.swift`](https://github.com/apple/swift/blob/tensorflow/test/stdlib/KeyPathIterable.swift)
- Or adapted from [here](https://github.com/apple/swift-evolution-staging/blob/fbfe1007b45afad8faf9ca8eabcbac78c27863b4/Tests/SE0000_KeyPathReflectionTests/SE0000_KeyPathReflectionTests.swift)

Sources/TensorFlow/Core/KeyPathIterable.swift
pschuh: Yes, I think that a standalone function is fine (ie: `var allKeyPaths: [PartialKeyPath<Self>]`) and people should just use that instead, but we still have code using KeyPathIterable.

dan-zheng (approved):

PR #1141, opened 2020-12-09
work around SR-13945 segfault

marcrasi:
I filed https://bugs.swift.org/browse/SR-13945 for the underlying cause. This seems like a perfectly good workaround to get swift-apis compiling again.

### DIFF
--- a/Sources/TensorFlow/Layers/Recurrent.swift
+++ b/Sources/TensorFlow/Layers/Recurrent.swift
@@ -719,7 +719,7 @@ fileprivate extension Array where Element: Differentiable {
   }
 
   @derivative(of: differentiableMerging)
-  func differentiableMerging(
+  func vjpDifferentiableMerging(
     _ other: Self, mergeFunction: @differentiable (Element, Element) -> Element
   ) -> (value: Self, pullback: (TangentVector) -> (TangentVector, TangentVector)) {
     let valuesWithPullbacks = zip(self, other).map {
@@ -729,9 +729,8 @@ fileprivate extension Array where Element: Differentiable {
     return (
       valuesWithPullbacks.map { $0.value },
       { vs in
-        let resultPairs: [(Element.TangentVector, Element.TangentVector)] = zip(vs.base, pullbacks).map {
-          let (v, pb) = $0
-          return pb(v)
+        let resultPairs = zip(vs.base, pullbacks).map { (v, pb) in
+          pb(v)
         }
         return (.init(resultPairs.map { $0.0 }), .init(resultPairs.map { $0.1 }))
       }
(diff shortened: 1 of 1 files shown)


Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -703,3 +687,54 @@ public typealias SimpleRNNCell = BasicRNNCell
...
+  /// encounter if you tried to implement this at the callsite using a for loop.
+  @differentiable
+  func differentiableMerging(
+    _ other: Self, mergeFunction: @differentiable (Element, Element) -> Element
+  ) -> Self {
+    zip(self, other).map { mergeFunction($0.0, $0.1) }
+  }
+
+  @derivative(of: differentiableMerging)
+  func differentiableMerging(
```
dan-zheng, line 722:
```suggestion
  func vjpDifferentiableMerging(
```

Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -703,3 +687,54 @@ public typealias SimpleRNNCell = BasicRNNCell
...
+  ) -> (value: Self, pullback: (TangentVector) -> (TangentVector, TangentVector)) {
+    let valuesWithPullbacks = zip(self, other).map {
+      valueWithPullback(at: $0.0, $0.1, in: mergeFunction)
+    }
+    let pullbacks = valuesWithPullbacks.map { $0.pullback }
+    return (
+      valuesWithPullbacks.map { $0.value },
+      { vs in
+        let resultPairs: [(Element.TangentVector, Element.TangentVector)] = zip(vs.base, pullbacks).map {
+          let (v, pb) = $0
```
dan-zheng, line 733:
```suggestion
        let resultPairs: [(Element.TangentVector, Element.TangentVector)] = zip(vs.base, pullbacks).map { (v, pb) in
```

Sources/TensorFlow/Layers/Recurrent.swift
```diff
@@ -703,3 +687,54 @@ public typealias SimpleRNNCell = BasicRNNCell
...
+  func differentiableReversed() -> Self {
+    .init(self.reversed())
+  }
+
+  @derivative(of: differentiableReversed)
+  func vjpDifferentiableReversed()
+    -> (value: Self, pullback: (TangentVector) -> TangentVector)
+  {
+    return (self.differentiableReversed(), { .init(.init($0.base.reversed())) })
+  }
```
dan-zheng, line 708: No urgency: I wonder if this derivative could be defined as `@derivative(of: reversed)` and put in the `_Differentiation` module?

dan-zheng (approved):

pschuh (approved):

Sources/TensorFlow/Layers/Recurrent.swift
marcrasi: That would be nice. `reversed()` returns a `ReversedCollection` so we would need to make that differentiable too.

pschuh: reverse() is mutating. Perhaps that could be used instead?

PR #1143, opened 2020-12-10
Implement TensorGroup based on reflection.

pschuh:


Tests/TensorFlowTests/LazyTensorEvaluationTests.swift
```diff
@@ -77,8 +77,8 @@ final class LazyTensorEvaluationTests: LazyTensorTestCase {
   }
 
   struct SimpleOutput: TensorGroup {
-    let a: TensorHandle<Int32>
-    let b: TensorHandle<Int32>
+    var a: TensorHandle<Int32>
+    var b: TensorHandle<Int32>
```
dan-zheng, line 81: Does the new implementation require `TensorGroup`-conforming stored properties to be `var` instead of `let`? I wonder if that breaks anything (e.g. `tensorflow/swift-models`) in practice.

dan-zheng (approved):

Tests/TensorFlowTests/LazyTensorEvaluationTests.swift
pschuh: yes. I don't know, they shouldn't be using any of these functions.

PR #1145, opened 2020-12-11
build: install `_NumericsShims`

compnerd:
The `_NumericsShims` module is required for Numerics as we cannot use
the Numerics module as implementation only as we provide extensions via
`ElementaryFunctions`.  This should allow users to actually build
against the build of swift-apis.


compnerd:
CC: @marcrasi

pschuh (approved):

PR #1160, opened 2020-12-21
workflows: setup a macOS GitHub workflow for CI

compnerd:
Now that we can build with a stock toolchain, we can start investigating
building on macOS with GitHub actions.  This will give us additional
coverage and earlier notifications.


dan-zheng (approved):
Incredible 👩‍💻🐦👨‍💻

PR #1162, opened 2020-12-21
TensorFlow: annotate functions using `_forEachFieldWithKeyPath`

compnerd:
Because the standard toolchain support requires the newest runtime in
order to have access to `_forEachFieldWithKeyPath` and the runtime is
bundled into the OS on macOS, we need to annotate the functions with
availability.

### DIFF
--- a/Sources/TensorFlow/Core/ElementaryFunctions.swift
+++ b/Sources/TensorFlow/Core/ElementaryFunctions.swift
@@ -13,12 +13,18 @@
 // limitations under the License.
 
 #if TENSORFLOW_USE_STANDARD_TOOLCHAIN
+
 import Numerics
 @_spi(Reflection) import Swift
+
 extension ElementaryFunctions {
   internal static func visitChildren(
     _ body: (PartialKeyPath<Self>, ElementaryFunctionsVisit.Type) -> Void
   ) {
+    guard #available(macOS 9999, *) else {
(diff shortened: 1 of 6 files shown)


pschuh (approved):

dan-zheng (commented):
Does this solves usability of the attributed declarations on macOS in practice? I fear end users will still encounter an availability error. We could verify this by adding a test case.

I'm not sure how to fix this in a robust way? May be good to ask the forums.

compnerd:
@dan-zheng it didn't - we needed to restrict this to a runtime check, which I have now done.  This was sufficient to get this to build the package with the latest snapshot locally.

pschuh (approved):

philipturner:
> I fear end users will still encounter an availability error. 

Reverted by https://github.com/s4tf/s4tf/pull/17. This causes runtime crashes and no longer serves a functional purpose. On release toolchains, S4TF uses [swift-reflection-mirror](https://github.com/philipturner/swift-reflection-mirror) for reflection. I currently have a bug in ReflectionMirror exposed by comparing arrays of keypaths, but that should be fixable.

philipturner:
Nevermind. I have to add these guards back in to make S4TF compile on development toolchains. But they will be gated under `#if !TENSORFLOW_USE_RELEASE_TOOLCHAIN`.

PR #1164, opened 2020-12-21
Adjust structure of developer guide

texasmichelle:
Make it clearer when to use each set of instructions.

### DIFF
--- a/Documentation/Development.md
+++ b/Documentation/Development.md
@@ -11,7 +11,7 @@ Swift for TensorFlow APIs can be built with either:
 
 ## Components
 
-Swift for TensorFlow APIs are composed of two distinct components:
+Building Swift for TensorFlow APIs involves two distinct components:
 
 1. X10
 2. Swift APIs
@@ -29,9 +29,20 @@ Swift for TensorFlow APIs are composed of two distinct components:
 
 The two components can be built together (CMake only) or separately.
 
-Follow the instructions below based on your preferred build tool.
+Follow the instructions below based on your preferred build tool (CMake or
+SwiftPM).
 
 ## Building With CMake
+
+With CMake, X10 and Swift APIs can be built either together or separately.
+
+To determine the appropriate build path, first consider whether you are
+modifying X10.
+
+* If you are modifying files in any subdirectory within **Sources/CX10** or
+  **Sources/x10**, you are modifying X10 and must follow
+  [**Option 1**](#option-1-build-x10-and-swift-apis-together).
+* Otherwise, use [**Option 2**](#option-2-use-a-prebuilt-version-of-x10).
 
 > **Note:** In-tree builds are not supported.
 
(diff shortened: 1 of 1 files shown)


dan-zheng (approved):
Splendid! I'll try the macOS toolchain instructions sometime soon.

texasmichelle:
Same here. I haven't verified for correctness of commands yet.

Documentation/Development.md
```diff
@@ -40,20 +106,21 @@ situations where you may need to build X10 yourself.  These include:
...
+##### Requirements
 
 * Bazel. This can be installed [manually][bazel] or with
 [Bazelisk][bazelisk]. You will need a version supported by TensorFlow
 (between `_TF_MIN_BAZEL_VERSION` and `_TF_MAX_BAZEL_VERSION` as specified in
 [tensorflow/configure.py][configure.py]).
 * Python3 with [numpy][numpy].
 
-#### Building
+##### Building
```
compnerd, line 117: Reading over this, I feel like (R) 119-127 should be extracted out.  It is describing what X10 is rather than how to build it.  With the new structure, this seems like it belongs in the new `Components` section.  WDYT?

Documentation/Development.md
```diff
@@ -1,37 +1,103 @@
...
+Building with SwiftPM involves building X10 and Swift APIs separately.
+
+### Building X10
+
+To determine the appropriate build path, first consider whether you are
+modifying X10.
+
+* If you are modifying files in any subdirectory within **Sources/CX10** or
+  **Sources/x10**, you are modifying X10 and must follow **Option 2**.
+* Otherwise, use **Option 1**.
```
compnerd, line 82: I think that we need to be more conservative about Option 1.  We should explicitly call out, if you are not building for Windows x64 or macOS x86_64, then you must use Option 2.  This is the reason that I had the options ordered as such, Option 1 is a special case, and you quickly fall back to option 2.  I don't have a strong opinion on how we name it order it, but I do feel like we need to be overly explicit that Option 1 is a very narrow set of builds.

Documentation/Development.md
```diff
@@ -218,19 +290,15 @@ cp swift-apis/Sources/x10/swift_bindings/xla_tensor_wrapper.h ${DESTDIR}/usr/inc
...
 
-Because X10 is a required component for building the TensorFlow Swift APIs,
-and the project does not build with Swift Package Manager, we must pass along
-the necessary information to the build. We need to pass two items to the
-build:
-
-1. the location of the X10 & TensorFlow headers
-2. the location of the X10 & TensorFlow libraries
+1. The location of the X10 & TensorFlow headers.
+2. The location of the X10 & TensorFlow libraries.
```
compnerd, line 301: One thing that I feel is nice about the CMake section is that we explicitly indicate the command which is for testing, and which is for building.  Do you think it is valuable to replicate that in the SPM section below as well?

Documentation/Development.md
texasmichelle: agreed!

Documentation/Development.md
texasmichelle: I'd like to surface the least complex path first. I added more detail, but it still feels a bit convoluted. I'll keep thinking of a better way to organize this.

compnerd (approved):

PR #1167, opened 2020-12-22
Dev docs fixes/improvements

dabrahams:


Documentation/Development.md
```diff
@@ -317,16 +311,21 @@ export TF_NEED_CUDA=0
...
 # terminate bazel daemon
 bazel --output_user_root ${PWD}/caches/bazel/tensorflow shutdown
+cd ..
 
 # package
 DESTDIR=${PWD}/Library/tensorflow-$(echo $(uname -s) | tr 'A-Z' 'a-z')-$(uname -m)/tensorflow-2.4.0
 
 mkdir -p ${DESTDIR}/usr/lib
-cp tensorflow/bazel-bin/tensorflow/libtensorflow-2.4.0.(dylib|so) ${DESTDIR}/usr/lib/
-cp tensorflow/bazel-bin/tensorflow/compiler/tf2xla/xla_tensor/libx10.(dylib|so) ${DESTDIR}/usr/lib/
```
compnerd, line 329: This should have worked as a copy-paste command in bash.  Did it not work for the shell you use?  (I'm not 100% certain that it would work in the ksh/tsch/csh though).

Documentation/Development.md
```diff
@@ -347,6 +346,39 @@ cp swift-apis/Sources/x10/swift_bindings/xla_tensor_wrapper.h ${DESTDIR}/usr/inc
 ```
 </details>
 
+<details>
+    <summary>What the scripts do.</summary>
```
compnerd, line 350: This isn't really what the scripts do.  It is more an explanation of how X10 is developed as X10 cannot be built standalone - it is built as part of TensorFlow, and is current not distributed as part of the TensorFlow repository.  As such we need to explain how to overlay the source structure.

Documentation/Development.md
```diff
@@ -1,5 +1,23 @@
...
+    * [Building X10](#building-x10)
+      * [Option 1: Build X10](#option-1-build-x10)
+        * [Requirements](#requirements-1)
+        * [Building](#building)
+      * [Option 2: Use a prebuilt version of X10](#option-2-use-a-prebuilt-version-of-x10)
+    * [Building Swift APIs](#building-swift-apis)
+      * [macOS](#macos-1)
+    * [Running tests](#running-tests-1)
+
+
```
texasmichelle, line 20: 👍  to adding a TOC

Documentation/Development.md
```diff
@@ -1,5 +1,23 @@
 # Swift for TensorFlow Developer Guide
 
+
+  * [Requirements](#requirements)
+  * [Components](#components)
+  * [Building With CMake](#building-with-cmake)
+    * [Option 1: Build X10 and Swift APIs together](#option-1-build-x10-and-swift-apis-together)
```
texasmichelle, line 7:
```suggestion
    * [Option 1: Build X10 and Swift APIs together](#option-1-build-x10-and-swift-apis-together)
    * [Option 2: Use a prebuilt version of X10](#option-2-use-a-prebuilt-version-of-x10)
```

Documentation/Development.md
dabrahams: Of course it can't work as copy/paste because of the `.(dylib|so)` hiding in there.

dabrahams: Oh… bash!  I was running the script with `sh`.

Documentation/Development.md
dabrahams: I don't see any instructions in this section.  Who do we need to explain this to, and what are they supposed to do with the information?

Documentation/Development.md
compnerd: Ah, yes, that is definitely not POSIX sh compliant.  It is bash specific (which most Linux distributions default to these days, and macOS used to pre-Catalina) and should be compatible in zsh (which macOS defaults to).

Documentation/Development.md
compnerd: It isn't meant to be instruction, it is covering details of the tree layout to allow a new developer get oriented within the tree to make changes to X10.

Documentation/Development.md
```diff
@@ -80,6 +80,9 @@ whether you are modifying X10.
 > **Note:** If `swiftc` is not in your `PATH`, you must specify the path to it
   using `-D CMAKE_Swift_COMPILER=`.
 
+> **Note:** If the `ninja` build tool is not in your `PATH`, you must [install
```
compnerd, line 83: I think that the phrasing here is odd - if it isn't in the path, you need to add it to your path.  Installation does not guarantee addition to the path (e.g. I may have installed it to `/opt/ninja` which does not get added to your path - at least not by any distribution that I am aware of as `/opt` is not a unified alternate root but rather a dumping ground).

Documentation/Development.md
```diff
@@ -73,6 +94,12 @@ cmake -B out -G Ninja -S swift-apis -D CMAKE_BUILD_TYPE=Release
 cmake --build out
 ```
 
+> **Note:** expect to see this error from the first command. It's fine; we're
```
texasmichelle, line 97: Good call 👍

PR #1175, opened 2021-01-05
build: correct the install rules for Darwin

compnerd:
We were using the generator expression for the installation of the swift
modules on Darwin.  However, with swift-numerics, the Numerics module
has an implicit link against `_NumericsShims`.  This results in the
interface include directories having multiple values when computed late.
Use `get_target_property` to get the value prior to the additions,
allowing us to rename the file into place.

Thanks to @dan-zheng for reporting the issue!


dan-zheng (approved):
Thank you!

PR #1176, opened 2021-01-05
build: enable building X10 support with SPM

compnerd:
Enable building the Swift interfaces for X10 with SPM.  This effectively
hard-enables X10 support in TensorFlow as SPM does not provide
configuration.

In order to accomplish this, we need to create symbolic links to sources
in order to merge the `x10` sources into the TensorFlow module.  It is
not possible to simply move the sources to reflect this structure due to
out-of-tree use cases.


Package.swift
```diff
@@ -50,6 +50,10 @@ let package = Package(
         "PythonKit",
         "CTensorFlow",
         .product(name: "Numerics", package: "swift-numerics"),
+      ],
+      swiftSettings: [
+        .define("USING_X10_BACKEND"),
```
dan-zheng, line 55: From offline discussion: it seems like we can remove the `USING_X10_BACKEND` flag and checks soon, since that is the only path forward for enabling SwiftPM builds for `tensorflow/swift-apis`.

Package.swift
```diff
@@ -50,6 +50,10 @@ let package = Package(
...
+        .define("DEFAULT_BACKEND_EAGER"),
```
dan-zheng, line 56: I wonder if it's possible to eventually enable non-`DEFAULT_BACKEND_EAGER` builds via SwiftPM?

dan-zheng (approved):

Package.swift
compnerd: It is possible, but requires an API break; I'd want such a switch to be planned (and communicated) prior to executing.

dan-zheng:
Intentionally merging commits separately.

PR #1177, opened 2021-01-06
build: add build rules for X10 supplementary modules

compnerd:
Add support to build:
- x10_optimizers_optimizer
- x10_optimizers_tensor_visitor_plan
- x10_trailing_loop

These are used by some dependent projects (such as the BERT CoLA model).

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -31,6 +31,18 @@ let package = Package(
       name: "Tensor",
       type: .dynamic,
       targets: ["Tensor"]),
+    .library(
+      name: "x10_optimizers_optimizer",
+      type: .dynamic,
+      targets: ["x10_optimizers_optimizer"]),
+    .library(
+      name: "x10_optimizers_tensor_visitor_plan",
+      type: .dynamic,
+      targets: ["x10_optimizers_tensor_visitor_plan"]),
+    .library(
+      name: "x10_training_loop",
+      type: .dynamic,
+      targets: ["x10_training_loop"]),
   ],
   dependencies: [
     .package(url: "https://github.com/apple/swift-numerics", .branch("main")),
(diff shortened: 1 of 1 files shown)


dan-zheng (approved):
Thanks!

Confirmed that this unblocks [building `tensorflow/swift-models` with `tensorflow/swift-apis` as a SwiftPM dependency](https://github.com/dan-zheng/swift-models/tree/71adfdd9c97f033b6b069b62c084f6f404d7b53d):
```console
$ swift build -Xcc -I$HOME/Library/tensorflow-2.4.0/usr/include -Xlinker -L$HOME/Library/tensorflow-2.4.0/usr/lib -Xswiftc -DTENSORFLOW_USE_STANDARD_TOOLCHAIN
/Users/danielzheng/swift-models/Examples/BERT-CoLA/main.swift:21:8: error: no such module 'x10_optimizers_optimizer'
import x10_optimizers_optimizer
       ^
# After this patch: no error.
```

philipturner:
The failure is still happening to this day with `x10_training_loop` and the 2022-05-11 development snapshot. It's okay to uncomment the `x10_training_loop` product and target, but not okay to import them into LLDB. In Colab, I get the following runtime error if I run this command (with appropriate SwiftPM flags set):

```swift
%install '.package(url: "https://github.com/s4tf/s4tf", .branch("enable-x10-training-loop-pr1177"))' 
  TensorFlow x10_optimizers_optimizer x10_optimizers_tensor_visitor_plan x10_training_loop
import TensorFlow
print(Tensor<Float>.self)
```

Error:

```
Compile error: expression failed to parse:
Couldn't lookup symbols:
  nominal type descriptor for TensorFlow.Tensor
```

But if I remove `x10_training_loop` from the above install command, I instead get:

```
Tensor<Float>
```

> Edit: the following may not be related to the Swift bug. If you remove `x10_training_loop` from the command, then re-run it without restarting the runtime, the Jupyter kernel crashes. I don't know why.

philipturner:
The error is definitely a Swift bug. My reproducer is difficult to narrow because it involves [apple/swift](https://github.com/apple/swift), SwiftPM, and LLDB, and comes from a massive code base, just like https://github.com/apple/swift/issues/59569. I had no errors when compiling and testing via SwiftPM.

@compnerd is it possible to reproduce the original "failure" you encountered in https://github.com/tensorflow/swift-apis/pull/1177/commits/51bdd2749d225f3a5647ff249eed1cdbffaf5f00? What environment (OS, Swift version, etc.) do I need to reproduce the bug and determine whether it happens on modern toolchains?

compnerd:
Sorry, I don't remember the context at this point.  I would recommend Ubuntu 16.04 and Windows 10.  I would guess that Swift 5.5+ is what you want to test on those platforms.

PR #1179, opened 2021-01-06
Add GitHub workflow for building+installing into toolchain via CMake.

dan-zheng:
Add GitHub Actions workflow for:
- Downloading and installing a swift.org/download macOS development snapshot
- Building tensorflow/swift-apis via CMake
- Installing tensorflow/swift-apis into the toolchain via CMake
- Repackaging the toolchain into a .pkg
- Publishing the toolchain .pkg artifact

This allows users to download standard swift.org/download toolchains with
TensorFlow and X10 installed.


.github/workflows/macOS-CMake-toolchain.yml
```diff
@@ -0,0 +1,88 @@
...
+          TOOLCHAIN_DATE=${{ env.toolchain-date }}
+          # TOOLCHAIN_SHORT_DATE: TOOLCHAIN_DATE with hypens removed. Used for toolchain identifier and version.
+          TOOLCHAIN_SHORT_DATE="${TOOLCHAIN_DATE//-/}"
+          TOOLCHAIN_NAME=swift-DEVELOPMENT-SNAPSHOT-${TOOLCHAIN_DATE}-a
+          TF_TOOLCHAIN_NAME=swift-tensorflow-DEVELOPMENT-SNAPSHOT-${TOOLCHAIN_DATE}-a
+          echo "TOOLCHAIN_DATE=${TOOLCHAIN_DATE}" >> $GITHUB_ENV
+          echo "TOOLCHAIN_SHORT_DATE=${TOOLCHAIN_SHORT_DATE}" >> $GITHUB_ENV
+          echo "TOOLCHAIN_NAME=${TOOLCHAIN_NAME}" >> $GITHUB_ENV
+          echo "TF_TOOLCHAIN_NAME=${TF_TOOLCHAIN_NAME}" >> $GITHUB_ENV
+          echo "SWIFT_APIS_BUILD_DIR=/tmp/BinaryCache/tensorflow-swift-apis" >> $GITHUB_ENV
```
compnerd, line 29: Not sure how much value this has to be honest.

.github/workflows/macOS-CMake-toolchain.yml
```diff
@@ -0,0 +1,88 @@
...
+
+    env:
+      # Hardcoded date selecting the latest toolchain from https://swift.org/download/#snapshots.
+      toolchain-date: 2021-01-04
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set environment variables
+        run: |
```
compnerd, line 19: This seems like a large set of variables to define.  I assume that most of these are only needed for packaging?  It might make sense to make the packaging-only ones be set in the environment of the packaging.

.github/workflows/macOS-CMake-toolchain.yml
```diff
@@ -0,0 +1,88 @@
...
+
+      - name: Install pre-built TensorFlow and X10 libraries
+        run: |
+          curl -sL https://artprodeus21.artifacts.visualstudio.com/A8fd008a0-56bc-482c-ba46-67f9425510be/3133d6ab-80a8-4996-ac4f-03df25cd3224/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL2NvbXBuZXJkL3Byb2plY3RJZC8zMTMzZDZhYi04MGE4LTQ5OTYtYWM0Zi0wM2RmMjVjZDMyMjQvYnVpbGRJZC80NTU3NC9hcnRpZmFjdE5hbWUvdGVuc29yZmxvdy1kYXJ3aW4teDY00/content?format=zip -o tensorflow-darwin-x64.zip
+          unzip tensorflow-darwin-x64.zip
+          mv tensorflow-darwin-x64/Library/tensorflow-2.4.0 ~/Library/
+
+      - name: Build tensorflow/swift-apis via CMake
+        run: |
+          set -euxo pipefail
```
compnerd, line 52: I don't think that we need to worry about `pipefail` here - this isn't really a shell script that we are going to have pipes in :). The tracing is unnecessary (and useless) as GHA will trace everything, and the trace applies to the script itself, not to the contents within CMake.

.github/workflows/macOS-CMake-toolchain.yml
```diff
@@ -0,0 +1,88 @@
...
+          mkdir -p $SWIFT_APIS_BUILD_DIR
```
compnerd, line 53: This is only needed to work around an old CMake bug, I think that with the current GHA builders it should be unnecessary.

.github/workflows/macOS-CMake-toolchain.yml
```diff
@@ -0,0 +1,88 @@
...
+          cmake \
+            -B $SWIFT_APIS_BUILD_DIR \
+            -D BUILD_SHARED_LIBS=YES \
+            -D CMAKE_BUILD_TYPE=Release \
+            -D CMAKE_INSTALL_PREFIX=${TOOLCHAIN_HOME_LOCATION}/usr \
```
compnerd, line 58: If you want to inject this into the toolchain, I think that you should extract a second copy of the pkg and use that.  Something like this should do the trick:

```shell
pkg --expand ${TOOLCHAIN_NAME}-osx.pkg --destination ...
```

.github/workflows/macOS-CMake-toolchain.yml
```diff
@@ -0,0 +1,88 @@
...
+            -D CMAKE_Swift_COMPILER=${TOOLCHAIN_HOME_LOCATION}/usr/bin/swiftc \
```
compnerd, line 59: I think it would be nicer to use `TOOLCHAINS` and `xcrun` to get the path to `swiftc` a la:

```
-D CMAKE_Swift_COMPILER=$(TOOLCHAINS=org.swift.50202101041a xcrun -f swiftc)
```

.github/workflows/macOS-CMake-toolchain.yml
```diff
@@ -0,0 +1,88 @@
...
+            -D CMAKE_OSX_DEPLOYMENT_TARGET=10.9 \
+            -D TENSORFLOW_USE_STANDARD_TOOLCHAIN=YES \
+            -D X10_LIBRARY=${HOME}/Library/tensorflow-2.4.0/usr/lib/libx10.dylib \
+            -D X10_INCLUDE_DIRS=${HOME}/Library/tensorflow-2.4.0/usr/include \
+            -D BUILD_TESTING=NO \
+            -G Ninja \
+            -S .
```
compnerd, line 66: For consistency, since everything else is expanded, might be nice to use `${PWD}` instead of `.`.

.github/workflows/macOS-CMake-toolchain.yml
```diff
@@ -0,0 +1,88 @@
...
+
+      - name: Install tensorflow/swift-apis into toolchain
+        run: |
+          set -euxo pipefail
```
compnerd, line 70: Similar to above - the `set` is meaningless.

.github/workflows/macOS-CMake-toolchain.yml
```diff
@@ -0,0 +1,88 @@
...
+          cmake --build $SWIFT_APIS_BUILD_DIR --target install
+
+      - name: Build toolchain package installer
+        run: |
+          set -euxo pipefail
```
compnerd, line 75: Similar to above, the `set` is meaningless.

.github/workflows/macOS-CMake-toolchain.yml
```diff
@@ -0,0 +1,88 @@
...
+          pkgbuild --identifier org.tensorflow-${TOOLCHAIN_SHORT_DATE} \
+                   --install-location ${TOOLCHAIN_ROOT_LOCATION} \
+                   --version 5.0.${TOOLCHAIN_SHORT_DATE} \
+                   --root ${TOOLCHAIN_HOME_LOCATION} \
+                   ${TF_TOOLCHAIN_NAME}-osx.pkg
+          pwd
+          ls -alh
+
+      - name: Upload toolchain package installer 🐦🔢
```
compnerd, line 84: Not sure I understand the emoji usage here.

compnerd (commented):
I think that if we are doing this, it would be nice to do all the targets (or at least the ones that are on GHA).

.github/workflows/macOS-CMake-toolchain.yml
dan-zheng: Sounds good to me. I'm curious why you prefer calling `xcrun -f swiftc` when the exact path to `swiftc` is already known?

dan-zheng: I guess the advantage of using `TOOLCHAINS=... xcrun -f swiftc` is that it finds where `swiftc` is installed, and is robust against toolchain install location.

However, I actually think it is more portable to not use `xcrun` so that `macOS-CMake-toolchain.yml` can be more easily adapted to other platforms.

.github/workflows/macOS-CMake-toolchain.yml
dan-zheng: `$SWIFT_APIS_BUILD_DIR` is used twice (in the CMake build and install steps), so I think it does add some value as a variable. For example, if `macOS-CMake-toolchain.yml` is cloned for a different platform, it would only be necessary to update `SWIFT_APIS_BUILD_DIR` instead of its two use sites.

I'll proceed as is if that is okay.

.github/workflows/macOS-CMake-toolchain.yml
dan-zheng: I see. I wonder if installing into a separate second toolchain copy has added value? Installing into an existing single toolchain copy seems to work well.

.github/workflows/macOS-CMake-toolchain.yml
dan-zheng: Thanks for the pointing this out! I moved packaging-only variables to the specific packaging step, the list of variables at the top is more reasonable now.

.github/workflows/macOS-CMake-toolchain.yml
compnerd: It does have value - it prevents the toolchain being used to build from being contaminated.  The idea is that the version which is being installed for the time being will eventually be part of the underlying image (which should be immutable).  Although the CI may have the ability to write to `/`, but imagine reproducing this where the pre-installed toolchain is in `/` and you cannot add content to it.

.github/workflows/macOS-CMake-toolchain.yml
compnerd: I dont think that this action is portable at all.  `pkg` is very much a macOS operation.  The installation process, the build, etc are platform dependent.  We could make the rule more sharable by relying on the PATH.  However, all that makes sharable is a single line, which IMO doesn't add much value.

.github/workflows/macOS-CMake-toolchain.yml
compnerd: The single variable isn't particularly useful to hoist IMO.  It is just the location of the build tree.  You could also just replace that with `build` and use a relative path :)

.github/workflows/macOS-CMake-toolchain.yml
dan-zheng: Good points about non-portability, thanks!

.github/workflows/macOS-CMake-toolchain.yml
dan-zheng: Good points, thanks! Done.

(`installer -pkg <toolchain>.pkg -target <destination directory>` was the command for extracting the toolchain)

.github/workflows/macOS-CMake-toolchain.yml
dan-zheng: Done.

.github/workflows/macOS-CMake-toolchain.yml
```diff
@@ -0,0 +1,87 @@
...
+                   --install-location ${TOOLCHAIN_ROOT_LOCATION} \
+                   --version 5.0.${TOOLCHAIN_SHORT_DATE} \
+                   --root ${DEST_TOOLCHAIN_LOCATION} \
+                   ${TF_TOOLCHAIN_NAME}-osx.pkg
+
+      - name: Upload toolchain package installer
+        uses: actions/upload-artifact@v2
+        with:
+          name: swift-tensorflow-DEVELOPMENT-SNAPSHOT-${{ env.toolchain-date }}-a-osx.pkg
+          path: '*.pkg'
```
compnerd, line 87: I think that this needs to be made conditional.  This will do a GH release per PR.

dan-zheng: That's a good point. I wonder what do you think is a good condition, and how to add it?

.github/workflows/macOS-CMake-toolchain.yml
dan-zheng: That's fair, replaced with `build` :)

.github/workflows/macOS-CMake-toolchain.yml
dan-zheng: Using `TOOLCHAINS=... xcrun -f swiftc` doesn't work when the toolchain installer is unpacked twice: once as a source toolchain for compiling `tensorflow/swift-apis`, and second as a destination toolchain for installing `tensorflow/swift-apis` and repackaging.

The two toolchains share the same `com.swift.50202101041a` identifier, so `xcrun` can't resolve to the `swiftc` path that we want (the destination toolchain one).

<details>
<p>

```
  TOOLCHAINS=org.swift.50${TOOLCHAIN_SHORT_DATE}a
  SOURCE_TOOLCHAIN_SWIFTC=$(TOOLCHAINS=${TOOLCHAINS} xcrun -f swiftc)
  if [[ ${SOURCE_TOOLCHAIN_SWIFTC} == *"Xcode"* ]]; then
    echo "Unexpected Xcode toolchain swiftc found; incorrect TOOLCHAINS identifier."
    exit 1
  fi
  cmake \
    -B build \
    -D BUILD_SHARED_LIBS=YES \
    -D CMAKE_BUILD_TYPE=Release \
    -D CMAKE_INSTALL_PREFIX=${DEST_TOOLCHAIN_LOCATION}/usr \
    -D CMAKE_Swift_COMPILER=${SOURCE_TOOLCHAIN_SWIFTC} \
    -D CMAKE_OSX_DEPLOYMENT_TARGET=10.9 \
    -D TENSORFLOW_USE_STANDARD_TOOLCHAIN=YES \
    -D X10_LIBRARY=${HOME}/Library/tensorflow-2.4.0/usr/lib/libx10.dylib \
    -D X10_INCLUDE_DIRS=${HOME}/Library/tensorflow-2.4.0/usr/include \
    -D BUILD_TESTING=NO \
    -G Ninja \
    -S ${PWD}
  shell: /bin/bash -e {0}
  env:
    toolchain-date: 2021-01-04
    TOOLCHAIN_DATE: 2021-01-04
    TOOLCHAIN_SHORT_DATE: 20210104
    TOOLCHAIN_NAME: swift-DEVELOPMENT-SNAPSHOT-2021-01-04-a
    SOURCE_TOOLCHAIN_LOCATION: /Users/runner/Library/Developer/Toolchains/swift-DEVELOPMENT-SNAPSHOT-2021-01-04-a.xctoolchain
    DEST_TOOLCHAIN_LOCATION: /tmp/destination-toolchain/Library/Developer/Toolchains/swift-DEVELOPMENT-SNAPSHOT-2021-01-04-a.xctoolchain
2021-01-07 03:07:17.443 xcodebuild[1284:9049] [MT] DVTToolchain: Failed to register toolchain: <DVTToolchain:0x7f8b0c45eb30:'org.swift.50202101041a':'Swift Development Snapshot 2021-01-04 (a)'>: Error Domain=DVTFoundationErrorDomain Code=-1 "Can't register <DVTFilePath:0x7f8b0c459020:'/Library/Developer/Toolchains/swift-DEVELOPMENT-SNAPSHOT-2021-01-04-a.xctoolchain'>; identifier org.swift.50202101041a conflicts with <DVTFilePath:0x7f8b0c459190:'/Users/runner/Library/Developer/Toolchains/swift-DEVELOPMENT-SNAPSHOT-2021-01-04-a.xctoolchain'>" UserInfo={NSLocalizedDescription=Can't register <DVTFilePath:0x7f8b0c459020:'/Library/Developer/Toolchains/swift-DEVELOPMENT-SNAPSHOT-2021-01-04-a.xctoolchain'>; identifier org.swift.50202101041a conflicts with <DVTFilePath:0x7f8b0c459190:'/Users/runner/Library/Developer/Toolchains/swift-DEVELOPMENT-SNAPSHOT-2021-01-04-a.xctoolchain'>}
2021-01-07 03:07:18.964 xcodebuild[1286:9170] [MT] DVTToolchain: Failed to register toolchain: <DVTToolchain:0x7fe214748ba0:'org.swift.50202101041a':'Swift Development Snapshot 2021-01-04 (a)'>: Error Domain=DVTFoundationErrorDomain Code=-1 "Can't register <DVTFilePath:0x7fe2147433d0:'/Library/Developer/Toolchains/swift-DEVELOPMENT-SNAPSHOT-2021-01-04-a.xctoolchain'>; identifier org.swift.50202101041a conflicts with <DVTFilePath:0x7fe214742c80:'/Users/runner/Library/Developer/Toolchains/swift-DEVELOPMENT-SNAPSHOT-2021-01-04-a.xctoolchain'>" UserInfo={NSLocalizedDescription=Can't register <DVTFilePath:0x7fe2147433d0:'/Library/Developer/Toolchains/swift-DEVELOPMENT-SNAPSHOT-2021-01-04-a.xctoolchain'>; identifier org.swift.50202101041a conflicts with <DVTFilePath:0x7fe214742c80:'/Users/runner/Library/Developer/Toolchains/swift-DEVELOPMENT-SNAPSHOT-2021-01-04-a.xctoolchain'>}
Unexpected Xcode toolchain swiftc found; incorrect TOOLCHAINS identifier.
```

</p>

</details>

Reverted to hardcoding the destination toolchain `swiftc` path. This seems more direct and straightforward.

.github/workflows/macOS-CMake-toolchain.yml
dan-zheng: I tried adding some conditional steps, but they still seem to run unconditionally even when the guard condition is `false`. Any clue why or how to do better? 

https://github.com/tensorflow/swift-apis/blob/0ba6914526a7b65bec1da63e993804e9ff70d4b8/.github/workflows/macOS-CMake-toolchain.yml#L83-L99

https://github.com/tensorflow/swift-apis/runs/1660455199

compnerd: I suppose that we _could_ do this on tags only.  I think that tags are uncommon enough where it may make sense for building on a tag even when its not an actual release but some point that we want to archive.

dan-zheng:
The repackaged toolchains built as part of this GitHub action are currently not right (way too small, much is missing). I'm not sure why, would be good to investigate. This action is still useful for testing CMake-based builds and installs.

PR #1180, opened 2021-01-07
TensorFlow: remove dead code

compnerd:
Both the CMake and SPM builds now unconditionally enable the X10
backend.  As there is no configuration knob to control whether X10
backend support can be disabled, remove the now dead code path.


dan-zheng (approved):
Nice simplification! (`+40,692 −40,839` huh)

BradLarson (approved):

PR #1184, opened 2021-10-23
Patching to enable building swift-apis on current nightly Swift

ProfFan:
Currently this builds in `Debug` on the latest nightly toolchain on Linux. macOS is untested.

All unit tests pass except one, which is not crucial (only a floating point difference of 1e-6).

Building on `Release` is blocked by https://github.com/apple/swift/pull/38745#issuecomment-947890567

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -189,7 +189,7 @@ if(NOT X10_FOUND AND NOT USE_BUNDLED_X10)
       COMMAND
         rm -rf <SOURCE_DIR>/bazel-bin # ${CMAKE_COMMAND} -E rm -Rrf <SOURCE_DIR>/bazel-bin
       COMMAND
-        bazel build ${VISIBILITY_FLAGS} -c opt --define framework_shared_object=false //tensorflow/compiler/tf2xla/xla_tensor:x10 --nocheck_visibility
+        bazel build ${VISIBILITY_FLAGS} -c opt --define framework_shared_object=false //tensorflow:tensorflow //tensorflow/compiler/tf2xla/xla_tensor:x10 --nocheck_visibility
       COMMAND
         bazel shutdown
     INSTALL_COMMAND
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,7 +118,7 @@ if(ENABLE_PYTHON_SUPPORT)
     GIT_REPOSITORY
       git://github.com/pvieito/PythonKit
     GIT_TAG
-      master
+      6a05a15
     CMAKE_ARGS
       -D BUILD_SHARED_LIBS=YES
       -D CMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
--- a/Documentation/X10/SUMMARY.md
+++ b/Documentation/X10/SUMMARY.md
@@ -22,7 +22,8 @@ public struct MyModel: Layer {
   public var dense3 = Dense<Float>(inputSize: 4, outputSize: 4)
   public var flatten = Flatten<Float>()
 
-  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+  @differentiable(reverse)
+  public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
     let layer1 = dense1(input)
     let layer2 = layer1.reshaped(to: [1, 4])
     let layer3 = dense2(layer2)
--- a/Documentation/X10/SUMMARY.md
+++ b/Documentation/X10/SUMMARY.md
@@ -22,8 +22,7 @@ public struct MyModel: Layer {
   public var dense3 = Dense<Float>(inputSize: 4, outputSize: 4)
   public var flatten = Flatten<Float>()
 
-  @differentiable
-  public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
     let layer1 = dense1(input)
     let layer2 = layer1.reshaped(to: [1, 4])
     let layer3 = dense2(layer2)
--- a/README.md
+++ b/README.md
@@ -34,8 +34,7 @@ struct Model: Layer {
     var layer2 = Dense<Float>(inputSize: hiddenSize, outputSize: hiddenSize, activation: relu)
     var layer3 = Dense<Float>(inputSize: hiddenSize, outputSize: 3, activation: identity)
     
-    @differentiable
-    func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+    @differentiable(reverse)    func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
         return input.sequenced(through: layer1, layer2, layer3)
     }
 }
--- a/README.md
+++ b/README.md
@@ -34,7 +34,8 @@ struct Model: Layer {
     var layer2 = Dense<Float>(inputSize: hiddenSize, outputSize: hiddenSize, activation: relu)
     var layer3 = Dense<Float>(inputSize: hiddenSize, outputSize: 3, activation: identity)
     
-    @differentiable(reverse)    func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+    @differentiable(reverse)
+    func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
         return input.sequenced(through: layer1, layer2, layer3)
     }
 }
--- a/Sources/TensorFlow/Core/Tensor.swift
+++ b/Sources/TensorFlow/Core/Tensor.swift
@@ -40,12 +40,12 @@ public struct Tensor<Scalar: TensorFlowScalar> {
   @usableFromInline
   internal var _isScalarZero = false
 
-  /// An internal workaround for SR-13263: debug info generation crash.
-  @usableFromInline
-  class SR13263Workaround {}
+  // /// An internal workaround for SR-13263: debug info generation crash.
+  // @usableFromInline
+  // class SR13263Workaround {}
 
-  /// An internal workaround for SR-13263: debug info generation crash.
-  internal var _sr13263Workaround: SR13263Workaround?
+  // /// An internal workaround for SR-13263: debug info generation crash.
+  // internal var _sr13263Workaround: SR13263Workaround?
   
   @inlinable
   public init(handle: TensorHandle<Scalar>) {
@@ -132,7 +132,7 @@ extension Tensor {
   /// Reshape to scalar.
   /// - Precondition: The tensor has exactly one scalar.
   @inlinable
-  @differentiable(where Scalar: TensorFlowFloatingPoint)
+  @differentiable(reverse where Scalar: TensorFlowFloatingPoint)
   public func scalarized() -> Scalar {
     precondition(
       shape.contiguousSize == 1,
@@ -174,7 +174,7 @@ extension Tensor {
     return handle.makeHostCopy()
   }
 
-  @differentiable(where Scalar: TensorFlowFloatingPoint)
+  @differentiable(reverse where Scalar: TensorFlowFloatingPoint)
   public var scalars: [Scalar] {
     if handle.backend == .XLA {
       let (storage, _) = xlaTensor.fetchTensorValues(Scalar.self)
@@ -203,7 +203,7 @@ extension Tensor where Scalar: TensorFlowFloatingPoint {
 
 extension Tensor {
   /// Creates a 0-D tensor from a scalar value.
-  @differentiable(where Scalar: TensorFlowFloatingPoint)
+  @differentiable(reverse where Scalar: TensorFlowFloatingPoint)
   public init(_ value: Scalar, on device: Device = .default) {
     switch device.backend {
     case .XLA:
@@ -227,7 +227,7 @@ extension Tensor where Scalar: TensorFlowFloatingPoint {
 extension Tensor {
   /// Creates a 1D tensor from scalars.
   @inlinable
-  @differentiable(where Scalar: TensorFlowFloatingPoint)
+  @differentiable(reverse where Scalar: TensorFlowFloatingPoint)
   public init(_ scalars: [Scalar], on device: Device = .default) {
     self.init(shape: [scalars.count], scalars: scalars, on: device)
   }
@@ -247,7 +247,7 @@ extension Tensor {
   ///   - scalars: The scalar contents of the tensor.
   /// - Precondition: The product of the dimensions of the shape must equal the number of scalars.
   @inlinable
-  @differentiable(where Scalar: TensorFlowFloatingPoint)
+  @differentiable(reverse where Scalar: TensorFlowFloatingPoint)
   public init(shape: TensorShape, scalars: [Scalar], on device: Device = .default) {
     precondition(
       shape.contiguousSize == scalars.count,
@@ -628,7 +628,7 @@ extension Tensor: AdditiveArithmetic where Scalar: Numeric {
   /// Adds two tensors and produces their sum.
   /// - Note: `+` supports broadcasting.
   @inlinable
-  @differentiable(where Scalar: TensorFlowFloatingPoint)
+  @differentiable(reverse where Scalar: TensorFlowFloatingPoint)
   public static func + (lhs: Tensor, rhs: Tensor) -> Tensor {
     if lhs._isScalarZero {
       return rhs
@@ -641,7 +641,7 @@ extension Tensor: AdditiveArithmetic where Scalar: Numeric {
   /// Subtracts one tensor from another and produces their difference.
   /// - Note: `-` supports broadcasting.
   @inlinable
-  @differentiable(where Scalar: TensorFlowFloatingPoint)
+  @differentiable(reverse where Scalar: TensorFlowFloatingPoint)
   public static func - (lhs: Tensor, rhs: Tensor) -> Tensor {
     if rhs._isScalarZero {
       return lhs
@@ -745,7 +745,7 @@ public protocol TensorProtocol {
 public protocol DifferentiableTensorProtocol:
   TensorProtocol & Differentiable & EuclideanDifferentiable
 where Scalar: TensorFlowFloatingPoint {
-  @differentiable(wrt: self)
+  @differentiable(reverse, wrt: self)
   func annotate(_ annotation: String) -> Self
 }
 
@@ -773,7 +773,7 @@ where Scalar: TensorFlowFloatingPoint {
   ///
   /// - Parameter annotation: The annotation to be added.
   /// - Returns: The annotated tensor.
-  @differentiable(wrt: self)
+  @differentiable(reverse, wrt: self)
   public func annotate(_ annotation: String) -> Tensor<Scalar> {
     switch handle.backend {
     case .XLA:
--- a/Sources/TensorFlow/Layer.swift
+++ b/Sources/TensorFlow/Layer.swift
@@ -32,14 +32,14 @@ where
   ///
   /// - Parameter input: The input to the layer.
   /// - Returns: The output.
-  @differentiable(wrt: self)
+  @differentiable(reverse, wrt: self)
   func callAsFunction(_ input: Input) -> Output
 
   /// Returns the output obtained from applying the layer to the given input.
   ///
   /// - Parameter input: The input to the layer.
   /// - Returns: The output.
-  @differentiable(wrt: self)
+  @differentiable(reverse, wrt: self)
   func forward(_ input: Input) -> Output
 }
 
@@ -48,7 +48,7 @@ extension Module {
   ///
   /// - Parameter input: The input to the layer.
   /// - Returns: The output.
-  @differentiable(wrt: self)
+  @differentiable(reverse, wrt: self)
   public func forward(_ input: Input) -> Output {
     return callAsFunction(input)
   }
@@ -60,7 +60,7 @@ extension Module where Input: TensorProtocol, Output: DifferentiableTensorProtoc
   ///
   /// - Parameter input: The input to the layer.
   /// - Returns: The annotated output.
-  @differentiable(wrt: self)
+  @differentiable(reverse, wrt: self)
   public func callAsFunction(_ input: Input) -> Output {
     let activation = forward(input)
     return annotated(activation)
@@ -72,8 +72,7 @@ extension Module where Input: TensorProtocol, Output: DifferentiableTensorProtoc
   ///
   /// - Parameter output: The output to the layer.
   /// - Returns: The annotated output.
-  @differentiable
-  public func annotated(_ output: Output) -> Output {
+  @differentiable(reverse)  public func annotated(_ output: Output) -> Output {
     let annotated = output.annotate("type=\(Self.self)")
     return annotated
   }
@@ -153,31 +152,18 @@ public protocol Layer: Module where Input: Differentiable {
   ///
   /// - Parameter input: The input to the layer.
   /// - Returns: The output.
-  @differentiable
-  func callAsFunction(_ input: Input) -> Output
-
-  @differentiable
-  func forward(_ input: Input) -> Output
+  @differentiable(reverse)  func callAsFunction(_ input: Input) -> Output
 }
 
-extension Layer {
-  // Workaround for SR-13455: autodiff undefined symbol linker error.
-  @differentiable(wrt: self)
-  @differentiable
-  public func forward(_ input: Input) -> Output {
-    return callAsFunction(input)
-  }
-}
 
-extension Layer where Input: DifferentiableTensorProtocol, Output: DifferentiableTensorProtocol {
-  // Workaround for SR-13455: autodiff undefined symbol linker error.
-  @differentiable(wrt: self)
-  @differentiable
-  public func callAsFunction(_ input: Input) -> Output {
-    let activation = forward(input)
-    return annotated(activation)
-  }
-}
+// extension Layer where Input: DifferentiableTensorProtocol, Output: DifferentiableTensorProtocol {
+//   // Workaround for SR-13455: autodiff undefined symbol linker error.
+//   @differentiable(reverse, wrt: self)
+//   public func callAsFunction(_ input: Input) -> Output {
+//     let activation = callAsFunction(input)
+//     return annotated(activation)
+//   }
+// }
 
 /// An empty struct representing empty `TangentVector`s for parameterless layers.
 public struct EmptyTangentVector: EuclideanDifferentiable, VectorProtocol, ElementaryFunctions,
@@ -200,12 +186,11 @@ public struct EmptyTangentVector: EuclideanDifferentiable, VectorProtocol, Eleme
 ///
 /// The `TangentVector` of parameterless layers is always `EmptyTangentVector`.
 public protocol ParameterlessLayer: Layer where TangentVector == EmptyTangentVector {
-  @differentiable
-  func callAsFunction(_ input: Input) -> Output
+  @differentiable(reverse) func callAsFunction(_ input: Input) -> Output
 }
 
 extension ParameterlessLayer {
-  public mutating func move(along direction: EmptyTangentVector) {}
+  public mutating func move(by direction: EmptyTangentVector) {}
   public var differentiableVectorView: EmptyTangentVector { EmptyTangentVector() }
 }
 
@@ -269,8 +254,7 @@ extension Differentiable {
   ///   - l1: The first layer.
   ///   - l2: The second layer.
   /// - Returns: The final layer's output after sequential application.
-  @differentiable
-  public func sequenced<L1: Layer, L2: Layer>(through l1: L1, _ l2: L2) -> L2.Output
+  @differentiable(reverse)  public func sequenced<L1: Layer, L2: Layer>(through l1: L1, _ l2: L2) -> L2.Output
   where L1.Input == Self, L1.Output == L2.Input {
     let o1 = l1(self)
     return l2(o1)
@@ -284,8 +268,7 @@ extension Differentiable {
   ///   - l2: The second layer.
   ///   - l3: The third layer.
   /// - Returns: The final layer's output after sequential application.
-  @differentiable
-  public func sequenced<L1: Layer, L2: Layer, L3: Layer>(through l1: L1, _ l2: L2, _ l3: L3)
+  @differentiable(reverse)  public func sequenced<L1: Layer, L2: Layer, L3: Layer>(through l1: L1, _ l2: L2, _ l3: L3)
     -> L3.Output
   where L1.Input == Self, L1.Output == L2.Input, L2.Output == L3.Input {
     let o1 = l1(self)
@@ -302,8 +285,7 @@ extension Differentiable {
   ///   - l3: The third layer.
   ///   - l4: The fourth layer.
   /// - Returns: The final layer's output after sequential application.
-  @differentiable
-  public func sequenced<L1: Layer, L2: Layer, L3: Layer, L4: Layer>(
+  @differentiable(reverse)  public func sequenced<L1: Layer, L2: Layer, L3: Layer, L4: Layer>(
     through l1: L1, _ l2: L2, _ l3: L3, _ l4: L4
   ) -> L4.Output
   where
@@ -326,8 +308,7 @@ extension Differentiable {
   ///   - l4: The third layer.
   ///   - l5: The fifth layer.
   /// - Returns: The final layer's output after sequential application.
-  @differentiable
-  public func sequenced<L1: Layer, L2: Layer, L3: Layer, L4: Layer, L5: Layer>(
+  @differentiable(reverse)  public func sequenced<L1: Layer, L2: Layer, L3: Layer, L4: Layer, L5: Layer>(
     through l1: L1, _ l2: L2, _ l3: L3, _ l4: L4, _ l5: L5
   ) -> L5.Output
   where
@@ -352,8 +333,7 @@ extension Differentiable {
   ///   - l5: The fifth layer.
   ///   - l6: The sixth layer.
   /// - Returns: The final layer's output after sequential application.
-  @differentiable
-  public func sequenced<L1: Layer, L2: Layer, L3: Layer, L4: Layer, L5: Layer, L6: Layer>(
+  @differentiable(reverse)  public func sequenced<L1: Layer, L2: Layer, L3: Layer, L4: Layer, L5: Layer, L6: Layer>(
     through l1: L1, _ l2: L2, _ l3: L3, _ l4: L4, _ l5: L5, _ l6: L6
   ) -> L6.Output
   where
--- a/Sources/TensorFlow/Layer.swift
+++ b/Sources/TensorFlow/Layer.swift
@@ -72,7 +72,8 @@ extension Module where Input: TensorProtocol, Output: DifferentiableTensorProtoc
   ///
   /// - Parameter output: The output to the layer.
   /// - Returns: The annotated output.
-  @differentiable(reverse)  public func annotated(_ output: Output) -> Output {
+  @differentiable(reverse)
+  public func annotated(_ output: Output) -> Output {
     let annotated = output.annotate("type=\(Self.self)")
     return annotated
   }
@@ -152,7 +153,8 @@ public protocol Layer: Module where Input: Differentiable {
   ///
   /// - Parameter input: The input to the layer.
   /// - Returns: The output.
-  @differentiable(reverse)  func callAsFunction(_ input: Input) -> Output
+  @differentiable(reverse)
+  func callAsFunction(_ input: Input) -> Output
 }
 
 
@@ -254,7 +256,8 @@ extension Differentiable {
   ///   - l1: The first layer.
   ///   - l2: The second layer.
   /// - Returns: The final layer's output after sequential application.
-  @differentiable(reverse)  public func sequenced<L1: Layer, L2: Layer>(through l1: L1, _ l2: L2) -> L2.Output
+  @differentiable(reverse)
+  public func sequenced<L1: Layer, L2: Layer>(through l1: L1, _ l2: L2) -> L2.Output
   where L1.Input == Self, L1.Output == L2.Input {
     let o1 = l1(self)
     return l2(o1)
@@ -268,7 +271,8 @@ extension Differentiable {
   ///   - l2: The second layer.
   ///   - l3: The third layer.
   /// - Returns: The final layer's output after sequential application.
-  @differentiable(reverse)  public func sequenced<L1: Layer, L2: Layer, L3: Layer>(through l1: L1, _ l2: L2, _ l3: L3)
+  @differentiable(reverse)
+  public func sequenced<L1: Layer, L2: Layer, L3: Layer>(through l1: L1, _ l2: L2, _ l3: L3)
     -> L3.Output
   where L1.Input == Self, L1.Output == L2.Input, L2.Output == L3.Input {
     let o1 = l1(self)
@@ -285,7 +289,8 @@ extension Differentiable {
   ///   - l3: The third layer.
   ///   - l4: The fourth layer.
   /// - Returns: The final layer's output after sequential application.
-  @differentiable(reverse)  public func sequenced<L1: Layer, L2: Layer, L3: Layer, L4: Layer>(
+  @differentiable(reverse)
+  public func sequenced<L1: Layer, L2: Layer, L3: Layer, L4: Layer>(
     through l1: L1, _ l2: L2, _ l3: L3, _ l4: L4
   ) -> L4.Output
   where
@@ -308,7 +313,8 @@ extension Differentiable {
   ///   - l4: The third layer.
   ///   - l5: The fifth layer.
   /// - Returns: The final layer's output after sequential application.
-  @differentiable(reverse)  public func sequenced<L1: Layer, L2: Layer, L3: Layer, L4: Layer, L5: Layer>(
+  @differentiable(reverse)
+  public func sequenced<L1: Layer, L2: Layer, L3: Layer, L4: Layer, L5: Layer>(
     through l1: L1, _ l2: L2, _ l3: L3, _ l4: L4, _ l5: L5
   ) -> L5.Output
   where
@@ -333,7 +339,8 @@ extension Differentiable {
   ///   - l5: The fifth layer.
   ///   - l6: The sixth layer.
   /// - Returns: The final layer's output after sequential application.
-  @differentiable(reverse)  public func sequenced<L1: Layer, L2: Layer, L3: Layer, L4: Layer, L5: Layer, L6: Layer>(
+  @differentiable(reverse)
+  public func sequenced<L1: Layer, L2: Layer, L3: Layer, L4: Layer, L5: Layer, L6: Layer>(
     through l1: L1, _ l2: L2, _ l3: L3, _ l4: L4, _ l5: L5, _ l6: L6
   ) -> L6.Output
   where
--- a/Sources/TensorFlow/Layers/Convolutional.swift
+++ b/Sources/TensorFlow/Layers/Convolutional.swift
@@ -36,7 +36,7 @@ public struct Conv1D<Scalar: TensorFlowFloatingPoint>: Layer {
   @noDerivative private let useBias: Bool
 
   /// The element-wise activation function type.
-  public typealias Activation = @differentiable (Tensor<Scalar>) -> Tensor<Scalar>
+  public typealias Activation = @differentiable(reverse) (Tensor<Scalar>) -> Tensor<Scalar>
 
   /// Creates a `Conv1D` layer with the specified filter, bias, activation function, stride,
   /// dilation and padding.
@@ -79,8 +79,7 @@ public struct Conv1D<Scalar: TensorFlowFloatingPoint>: Layer {
   /// - Returns: The output of shape [batch size, output width, output channel count].
   ///
   /// - Note: Padding size equals zero when using `.valid`.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
     let conv = conv1D(
       input,
       filter: filter,
@@ -149,7 +148,7 @@ public struct Conv2D<Scalar: TensorFlowFloatingPoint>: Layer {
   @noDerivative private let useBias: Bool
 
   /// The element-wise activation function type.
-  public typealias Activation = @differentiable (Tensor<Scalar>) -> Tensor<Scalar>
+  public typealias Activation = @differentiable(reverse) (Tensor<Scalar>) -> Tensor<Scalar>
 
   /// Creates a `Conv2D` layer with the specified filter, bias, activation function, strides,
   /// dilations and padding.
@@ -201,8 +200,7 @@ public struct Conv2D<Scalar: TensorFlowFloatingPoint>: Layer {
   ///   [batch count, output height, output width, output channel count].
   ///
   /// - Note: Padding size equals zero when using `.valid`.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
     let conv = conv2D(
       input,
       filter: filter,
@@ -273,7 +271,7 @@ public struct Conv3D<Scalar: TensorFlowFloatingPoint>: Layer {
   @noDerivative private let useBias: Bool
 
   /// The element-wise activation function type.
-  public typealias Activation = @differentiable (Tensor<Scalar>) -> Tensor<Scalar>
+  public typealias Activation = @differentiable(reverse) (Tensor<Scalar>) -> Tensor<Scalar>
 
   /// Creates a `Conv3D` layer with the specified filter, bias, activation function, strides, and
   /// padding.
@@ -332,8 +330,8 @@ public struct Conv3D<Scalar: TensorFlowFloatingPoint>: Layer {
   ///   [batch count, output depth, output height, output width, output channel count].
   ///
   /// - Note: Padding size equals zero when using `.valid`.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse) 
+  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
     let conv = conv3D(
       input,
       filter: filter,
@@ -405,7 +403,7 @@ public struct TransposedConv1D<Scalar: TensorFlowFloatingPoint>: Layer {
   @noDerivative private let useBias: Bool
 
   /// The element-wise activation function type.
-  public typealias Activation = @differentiable (Tensor<Scalar>) -> Tensor<Scalar>
+  public typealias Activation = @differentiable(reverse) (Tensor<Scalar>) -> Tensor<Scalar>
 
   /// Creates a `TransposedConv1D` layer with the specified filter, bias,
   /// activation function, strides, and padding.
@@ -436,8 +434,7 @@ public struct TransposedConv1D<Scalar: TensorFlowFloatingPoint>: Layer {
   ///
   /// - Parameter input: The input to the layer.
   /// - Returns: The output.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
     let batchSize = input.shape[0]
     let w = (input.shape[1] - (1 * paddingIndex)) * stride + (filter.shape[0] * paddingIndex)
     let c = filter.shape[2]
@@ -506,7 +503,7 @@ public struct TransposedConv2D<Scalar: TensorFlowFloatingPoint>: Layer {
   @noDerivative private let useBias: Bool
 
   /// The element-wise activation function type.
-  public typealias Activation = @differentiable (Tensor<Scalar>) -> Tensor<Scalar>
+  public typealias Activation = @differentiable(reverse) (Tensor<Scalar>) -> Tensor<Scalar>
 
   /// Creates a `TransposedConv2D` layer with the specified filter, bias,
   /// activation function, strides, and padding.
@@ -538,8 +535,7 @@ public struct TransposedConv2D<Scalar: TensorFlowFloatingPoint>: Layer {
   ///
   /// - Parameter input: The input to the layer.
   /// - Returns: The output.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
     let batchSize = input.shape[0]
     let h = (input.shape[1] - (1 * paddingIndex)) * strides.0 + (filter.shape[0] * paddingIndex)
     let w = (input.shape[2] - (1 * paddingIndex)) * strides.1 + (filter.shape[1] * paddingIndex)
@@ -610,7 +606,7 @@ public struct TransposedConv3D<Scalar: TensorFlowFloatingPoint>: Layer {
   @noDerivative private let useBias: Bool
 
   /// The element-wise activation function type.
-  public typealias Activation = @differentiable (Tensor<Scalar>) -> Tensor<Scalar>
+  public typealias Activation = @differentiable(reverse) (Tensor<Scalar>) -> Tensor<Scalar>
 
   /// Creates a `TransposedConv3D` layer with the specified filter, bias,
   /// activation function, strides, and padding.
@@ -641,8 +637,7 @@ public struct TransposedConv3D<Scalar: TensorFlowFloatingPoint>: Layer {
   ///
   /// - Parameter input: The input to the layer.
   /// - Returns: The output.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
     let batchSize = input.shape[0]
     let w = (input.shape[1] - (1 * paddingIndex)) * strides.0 + (filter.shape[0] * paddingIndex)
     let h = (input.shape[2] - (1 * paddingIndex)) * strides.1 + (filter.shape[1] * paddingIndex)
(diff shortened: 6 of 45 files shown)


CMakeLists.txt
```diff
@@ -118,7 +118,7 @@ if(ENABLE_PYTHON_SUPPORT)
     GIT_REPOSITORY
       git://github.com/pvieito/PythonKit
     GIT_TAG
-      master
+      6a05a15
```
ProfFan, line 121: 6a05a15 is the last commit before CMake is removed.

CMakeLists.txt
```diff
@@ -189,7 +189,7 @@ if(NOT X10_FOUND AND NOT USE_BUNDLED_X10)
       COMMAND
         rm -rf <SOURCE_DIR>/bazel-bin # ${CMAKE_COMMAND} -E rm -Rrf <SOURCE_DIR>/bazel-bin
       COMMAND
-        bazel build ${VISIBILITY_FLAGS} -c opt --define framework_shared_object=false //tensorflow/compiler/tf2xla/xla_tensor:x10 --nocheck_visibility
+        bazel build ${VISIBILITY_FLAGS} -c opt --define framework_shared_object=false //tensorflow:tensorflow //tensorflow/compiler/tf2xla/xla_tensor:x10 --nocheck_visibility
```
ProfFan, line 192: This enables one to extract a standalone X10 after a successful CMake build, so subsequent builds can be made with the Swift Package Manager.

Sources/TensorFlow/StdlibExtensions.swift
```diff
@@ -288,20 +299,56 @@ where Element: Differentiable & PointwiseMultiplicative {
...
+  //   precondition(lhs.count == rhs.count, "Count mismatch: \(lhs.count) and \(rhs.count)")
+  //   return .init(zip(lhs, rhs).map(.*))
+  // }
+
+  // public static func .*= (lhs: inout Self, rhs: Self) {
+  //   precondition(lhs.count == rhs.count, "Count mismatch: \(lhs.count) and \(rhs.count)")
+  //   for (i, x) in zip(lhs.indices, rhs) {
+  //     lhs[i] .*= x
+  //   }
+  // }
```
ProfFan, line 312: Can't get this to work, need investigation

Sources/TensorFlow/StdlibExtensions.swift
```diff
@@ -288,20 +299,56 @@ where Element: Differentiable & PointwiseMultiplicative {
...
 }
+
+/// Applies the given closure `body` to `x`. When used in a context where `x` is
+/// being differentiated with respect to, this function will not produce any
+/// derivative at `x`.
+// FIXME: Support throws-rethrows.
+@inlinable
+@inline(__always)
+@_semantics("autodiff.nonvarying")
+public func withoutDerivative<T, R>(at x: T, in body: (T) -> R) -> R {
```
ProfFan, line 327: `withoutDerivative` is removed in language, so copy-pasted to library

Sources/x10/swift_bindings/training_loop.swift
```diff
@@ -159,8 +159,7 @@ struct Statistics {
   }
 }
 
-@differentiable
-public func _defaultLossFunction(_ ŷ: Tensor<Float>, _ y: Tensor<Int32>) -> Tensor<Float> {
+@differentiable(reverse)public func _defaultLossFunction(_ ŷ: Tensor<Float>, _ y: Tensor<Int32>) -> Tensor<Float> {
```
ProfFan: FIXME

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -98,18 +98,19 @@ public struct BatchNorm<Scalar: TensorFlowFloatingPoint>: Layer {
   ///
   /// - Parameter input: The input to the layer.
   /// - Returns: The output.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
-    let positiveAxis = (input.rank + axis) % input.rank
+  @differentiable(reverse)
+  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```
ProfFan, line 102: This place triggers a compiler crash in release due to debug info being generated twice. Core reason should be that the autodiff is not doing proper variable cloning.

Sources/TensorFlow/StdlibExtensions.swift
```diff
@@ -234,6 +234,17 @@ where Element: Differentiable {
...
 
+  @inlinable
+  public subscript(bounds: Range<Array<Element>.Index>) -> Self.SubSequence {
+    _read { yield base[bounds] }
+    set { base[bounds] = newValue }
+  }
+
+  @inlinable
+  public mutating func replaceSubrange<C>(_ subrange: Range<Self.Index>, with newElements: C) where C : Collection, Self.Element == C.Element {
+    fatalError("withUnsafeBufferPointer unimplemented because TensorBuffer is abstract")
```
ProfFan, line 245: remove this shameless copy-pasta. I just can't get this new protocol requirement to work...

ProfFan:
@googlebot I signed it!

CMakeLists.txt
compnerd: This doesn't seem right.  You should be able to do that irrespective.  This is just building an additional label, which means that there is potentially a dependency that is missing?

README.md
```diff
@@ -34,8 +34,7 @@ struct Model: Layer {
     var layer2 = Dense<Float>(inputSize: hiddenSize, outputSize: hiddenSize, activation: relu)
     var layer3 = Dense<Float>(inputSize: hiddenSize, outputSize: 3, activation: identity)
     
-    @differentiable
-    func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+    @differentiable(reverse)    func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
```
compnerd:
```suggestion
    @differentiable(reverse)
    func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
```

Documentation/X10/SUMMARY.md
```diff
@@ -22,8 +22,7 @@ public struct MyModel: Layer {
   public var dense3 = Dense<Float>(inputSize: 4, outputSize: 4)
   public var flatten = Flatten<Float>()
 
-  @differentiable
-  public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Float>) -> Tensor<Float> {
```

Sources/TensorFlow/Core/Tensor.swift
```diff
@@ -40,12 +40,12 @@ public struct Tensor<Scalar: TensorFlowScalar> {
...
-  @usableFromInline
-  class SR13263Workaround {}
+  // /// An internal workaround for SR-13263: debug info generation crash.
+  // @usableFromInline
+  // class SR13263Workaround {}
 
-  /// An internal workaround for SR-13263: debug info generation crash.
-  internal var _sr13263Workaround: SR13263Workaround?
+  // /// An internal workaround for SR-13263: debug info generation crash.
+  // internal var _sr13263Workaround: SR13263Workaround?
```
compnerd, line 48: Why comment this out?  Is it no longer needed?

Sources/TensorFlow/Layer.swift
```diff
@@ -153,31 +152,18 @@ public protocol Layer: Module where Input: Differentiable {
   ///
   /// - Parameter input: The input to the layer.
   /// - Returns: The output.
-  @differentiable
-  func callAsFunction(_ input: Input) -> Output
-
-  @differentiable
-  func forward(_ input: Input) -> Output
+  @differentiable(reverse)  func callAsFunction(_ input: Input) -> Output
```
compnerd:
```suggestion
  @differentiable(reverse)
  func callAsFunction(_ input: Input) -> Output
```

Sources/TensorFlow/Layer.swift
```diff
@@ -153,31 +152,18 @@ public protocol Layer: Module where Input: Differentiable {
...
-  }
-}
+// extension Layer where Input: DifferentiableTensorProtocol, Output: DifferentiableTensorProtocol {
+//   // Workaround for SR-13455: autodiff undefined symbol linker error.
+//   @differentiable(reverse, wrt: self)
+//   public func callAsFunction(_ input: Input) -> Output {
+//     let activation = callAsFunction(input)
+//     return annotated(activation)
+//   }
+// }
```
compnerd, line 168: Similar to above, why is this commented out?  The SR is closed, so perhaps this is resolved and safe to drop.

Sources/TensorFlow/Layer.swift
```diff
@@ -269,8 +254,7 @@ extension Differentiable {
   ///   - l1: The first layer.
   ///   - l2: The second layer.
   /// - Returns: The final layer's output after sequential application.
-  @differentiable
-  public func sequenced<L1: Layer, L2: Layer>(through l1: L1, _ l2: L2) -> L2.Output
+  @differentiable(reverse)  public func sequenced<L1: Layer, L2: Layer>(through l1: L1, _ l2: L2) -> L2.Output
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func sequenced<L1: Layer, L2: Layer>(through l1: L1, _ l2: L2) -> L2.Output
```

Sources/TensorFlow/Layer.swift
```diff
@@ -284,8 +268,7 @@ extension Differentiable {
   ///   - l2: The second layer.
   ///   - l3: The third layer.
   /// - Returns: The final layer's output after sequential application.
-  @differentiable
-  public func sequenced<L1: Layer, L2: Layer, L3: Layer>(through l1: L1, _ l2: L2, _ l3: L3)
+  @differentiable(reverse)  public func sequenced<L1: Layer, L2: Layer, L3: Layer>(through l1: L1, _ l2: L2, _ l3: L3)
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func sequenced<L1: Layer, L2: Layer, L3: Layer>(through l1: L1, _ l2: L2, _ l3: L3)
```

Sources/TensorFlow/Layer.swift
```diff
@@ -302,8 +285,7 @@ extension Differentiable {
   ///   - l3: The third layer.
   ///   - l4: The fourth layer.
   /// - Returns: The final layer's output after sequential application.
-  @differentiable
-  public func sequenced<L1: Layer, L2: Layer, L3: Layer, L4: Layer>(
+  @differentiable(reverse)  public func sequenced<L1: Layer, L2: Layer, L3: Layer, L4: Layer>(
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func sequenced<L1: Layer, L2: Layer, L3: Layer, L4: Layer>(
```

Sources/TensorFlow/Layer.swift
```diff
@@ -326,8 +308,7 @@ extension Differentiable {
   ///   - l4: The third layer.
   ///   - l5: The fifth layer.
   /// - Returns: The final layer's output after sequential application.
-  @differentiable
-  public func sequenced<L1: Layer, L2: Layer, L3: Layer, L4: Layer, L5: Layer>(
+  @differentiable(reverse)  public func sequenced<L1: Layer, L2: Layer, L3: Layer, L4: Layer, L5: Layer>(
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func sequenced<L1: Layer, L2: Layer, L3: Layer, L4: Layer, L5: Layer>(
```

Sources/TensorFlow/Layer.swift
```diff
@@ -352,8 +333,7 @@ extension Differentiable {
   ///   - l5: The fifth layer.
   ///   - l6: The sixth layer.
   /// - Returns: The final layer's output after sequential application.
-  @differentiable
-  public func sequenced<L1: Layer, L2: Layer, L3: Layer, L4: Layer, L5: Layer, L6: Layer>(
+  @differentiable(reverse)  public func sequenced<L1: Layer, L2: Layer, L3: Layer, L4: Layer, L5: Layer, L6: Layer>(
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func sequenced<L1: Layer, L2: Layer, L3: Layer, L4: Layer, L5: Layer, L6: Layer>(
```

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -79,8 +79,7 @@ public struct Conv1D<Scalar: TensorFlowFloatingPoint>: Layer {
   /// - Returns: The output of shape [batch size, output width, output channel count].
   ///
   /// - Note: Padding size equals zero when using `.valid`.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -201,8 +200,7 @@ public struct Conv2D<Scalar: TensorFlowFloatingPoint>: Layer {
   ///   [batch count, output height, output width, output channel count].
   ///
   /// - Note: Padding size equals zero when using `.valid`.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -332,8 +330,8 @@ public struct Conv3D<Scalar: TensorFlowFloatingPoint>: Layer {
   ///   [batch count, output depth, output height, output width, output channel count].
   ///
   /// - Note: Padding size equals zero when using `.valid`.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse)
```
compnerd:
```suggestion
  @differentiable(reverse)
```

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -436,8 +434,7 @@ public struct TransposedConv1D<Scalar: TensorFlowFloatingPoint>: Layer {
   ///
   /// - Parameter input: The input to the layer.
   /// - Returns: The output.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -538,8 +535,7 @@ public struct TransposedConv2D<Scalar: TensorFlowFloatingPoint>: Layer {
...
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -641,8 +637,7 @@ public struct TransposedConv3D<Scalar: TensorFlowFloatingPoint>: Layer {
...
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -749,8 +744,7 @@ public struct DepthwiseConv2D<Scalar: TensorFlowFloatingPoint>: Layer {
   ///   [batch count, input height, input width, input channel count]
   /// - Returns: The output of shape,
   ///   [batch count, output height, output width, input channel count * channel multiplier]
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -823,8 +817,7 @@ public struct ZeroPadding1D<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer
   ///
   /// - Parameter input: The input to the layer.
   /// - Returns: The output.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -857,8 +850,7 @@ public struct ZeroPadding2D<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer
...
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -891,8 +883,7 @@ public struct ZeroPadding3D<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer
...
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -960,8 +951,7 @@ public struct SeparableConv1D<Scalar: TensorFlowFloatingPoint>: Layer {
...
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Convolutional.swift
```diff
@@ -1082,8 +1072,7 @@ public struct SeparableConv2D<Scalar: TensorFlowFloatingPoint>: Layer {
...
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Core.swift
```diff
@@ -28,8 +28,7 @@ public struct Flatten<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
   ///
   /// - Parameter input: The input to the layer.
   /// - Returns: The output.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Core.swift
```diff
@@ -66,25 +65,23 @@ public struct Reshape<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
...
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Core.swift
```diff
@@ -66,25 +65,23 @@ public struct Reshape<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
...
 
   @noDerivative public let body: Body
 
   public init(_ body: @escaping Body) {
     self.body = body
   }
 
-  @differentiable
-  public func callAsFunction(_ input: Input) -> Output {
+  @differentiable(reverse)  public func callAsFunction(_ input: Input) -> Output {
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Input) -> Output {
```

Sources/TensorFlow/Layers/Dense.swift
```diff
@@ -77,8 +77,7 @@ public struct Dense<Scalar: TensorFlowFloatingPoint>: Layer {
   ///
   /// - Parameter input: The input to the layer.
   /// - Returns: The output.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Dropout.swift
```diff
@@ -54,8 +54,7 @@ public struct Dropout<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
   ///
   /// - Parameter input: The input to the layer.
   /// - Returns: The output.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Dropout.swift
```diff
@@ -81,8 +80,7 @@ public struct GaussianNoise<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer
   }
 
   /// Returns a tensor obtained by adding noise to `input`
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Dropout.swift
```diff
@@ -118,8 +116,7 @@ public struct GaussianDropout<Scalar: TensorFlowFloatingPoint>: ParameterlessLay
   }
 
   /// Applies multiplicative 1-centered Gaussian noise to the input during training only.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Dropout.swift
```diff
@@ -158,8 +155,7 @@ public struct AlphaDropout<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer
   }
 
   /// Adds noise to `input` during training, and is a no-op during inference.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Morphological.swift
```diff
@@ -69,8 +69,7 @@ public struct `Dilation2D`<Scalar: TensorFlowFloatingPoint>: Layer {
   ///   [batch count, output height, output width, output channel count].
   ///
   /// - Note: Padding size equals zero when using `.valid`.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Morphological.swift
```diff
@@ -139,8 +138,7 @@ public struct `Erosion2D`<Scalar: TensorFlowFloatingPoint>: Layer {
...
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -23,8 +23,8 @@ import _Differentiation
 ///   - offset: The tensor to be added to normalized tensor.
 ///   - scale: The tensor to be applied to normalized tensor.
 ///   - varianceEpsilon: The small number to avoid dividing by 0.
-@differentiable(wrt: (input, mean, variance, offset, scale))
-private func normalize<Scalar: TensorFlowFloatingPoint>(
+@differentiable(reverse, wrt: (input, mean, variance, offset, scale))
+private func normalize<Scalar: TensorFlowFloatingPoint> (
```
compnerd:
```suggestion
private func normalize<Scalar: TensorFlowFloatingPoint>(
```

Sources/TensorFlow/Layers/Normalization.swift
compnerd: Could you isolate this into a standalone test case like Dan and I have done in the past (see some of the previously mentioned SRs).

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -98,18 +98,19 @@ public struct BatchNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
       input.shape[positiveAxis] == offset.shape[0],
       "The number of features of the input and the offset doesn't match.")
-    var offset = self.offset
-    var scale = self.scale
+    var (offset, scale) = {x in (x.offset, x.scale) }(self)
     if positiveAxis != input.rank - 1 {
       var broadcastShape = TensorShape([Int](repeating: 1, count: input.rank))
       broadcastShape[positiveAxis] = input.shape[positiveAxis]
       offset = offset.reshaped(to: broadcastShape)
+
```
compnerd:
```suggestion
```

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -98,18 +98,19 @@ public struct BatchNorm<Scalar: TensorFlowFloatingPoint>: Layer {
   ///
   /// - Parameter input: The input to the layer.
   /// - Returns: The output.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
-    let positiveAxis = (input.rank + axis) % input.rank
+  @differentiable(reverse)
+  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+    let inputRank = input.rank
+    let positiveAxis = (inputRank + axis) % inputRank
```
compnerd, line 104:
```suggestion
    let positiveAxis = (input.rank + axis) % input.rank
```

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -98,18 +98,19 @@ public struct BatchNorm<Scalar: TensorFlowFloatingPoint>: Layer {
...
     precondition(
       input.shape[positiveAxis] == offset.shape[0],
       "The number of features of the input and the offset doesn't match.")
-    var offset = self.offset
-    var scale = self.scale
+    var (offset, scale) = {x in (x.offset, x.scale) }(self)
```
compnerd:
```suggestion
    var offset = self.offset
    var scale = self.scale
```

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -240,8 +241,7 @@ public struct LayerNorm<Scalar: TensorFlowFloatingPoint>: Layer {
   ///
   /// - Parameter input: The input to the layer.
   /// - Returns: The output.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -341,8 +341,7 @@ public struct GroupNorm<Scalar: TensorFlowFloatingPoint>: Layer {
   /// - Returns: The output.
   /// - Precondition: The axis cannot be batch axis.
   /// - Precondition: The numbers of features of the input and the offset must be same.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Normalization.swift
```diff
@@ -447,8 +446,7 @@ public struct InstanceNorm<Scalar: TensorFlowFloatingPoint>: Layer {
   ///
   /// - Parameter input: The input to the layer.
   /// - Returns: The output.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Pooling.swift
```diff
@@ -44,8 +44,7 @@ public struct MaxPool1D<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
   ///
   /// - Parameter input: The input to the layer.
   /// - Returns: The output.
-  @differentiable
-  public func forward(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
+  @differentiable(reverse)  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Pooling.swift
```diff
@@ -85,8 +84,7 @@ public struct MaxPool2D<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
...
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Pooling.swift
```diff
@@ -142,8 +140,7 @@ public struct MaxPool3D<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
...
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Pooling.swift
```diff
@@ -204,8 +201,7 @@ public struct AvgPool1D<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
...
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Pooling.swift
```diff
@@ -245,8 +241,7 @@ public struct AvgPool2D<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
...
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Pooling.swift
```diff
@@ -302,8 +297,7 @@ public struct AvgPool3D<Scalar: TensorFlowFloatingPoint>: ParameterlessLayer {
...
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Pooling.swift
```diff
@@ -346,8 +340,7 @@ public struct GlobalAvgPool1D<Scalar: TensorFlowFloatingPoint>: ParameterlessLay
...
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

Sources/TensorFlow/Layers/Pooling.swift
```diff
@@ -365,8 +358,7 @@ public struct GlobalAvgPool2D<Scalar: TensorFlowFloatingPoint>: ParameterlessLay
...
```
compnerd:
```suggestion
  @differentiable(reverse)
  public func callAsFunction(_ input: Tensor<Scalar>) -> Tensor<Scalar> {
```

compnerd (changes requested):
There are still a large number of places where you have unnecessary whitespace changes.  Could you please fix them all and update the diff?

Sources/TensorFlow/Layers/Pooling.swift
brettkoonce: 🦖

Sources/TensorFlow/Core/Tensor.swift
ProfFan: Leaving this as-is triggers a compiler crash, so I removed this workaround.

Sources/TensorFlow/Layer.swift
ProfFan: Yeah I should just remove them in entirety

Sources/TensorFlow/Layers/Normalization.swift
ProfFan: Let me try that after I got more time (have a CVPR deadline)

CMakeLists.txt
ProfFan: Without this I think libtensorflow.so.xxx is not generated and I cannot copy them to a standalone `/Library/usr/lib` for a SPM build, following the *current* `docs/Development.md`. Interestingly this is not needed for the CMake build, which I don't know the reason.

ProfFan:
@compnerd Sorry for all the whitespace diffs, made a mistake in regex replacement.

Sources/TensorFlow/Layers/Normalization.swift
BradLarson: If this is the same duplicate debug info validation failure I've been seeing on internal code (and it seems like it, given that it's triggered in release builds with debug info enabled), we started seeing this with Swift commit [9a8f2ed64221635646541757b7e38b3b4e02e4b3](https://github.com/apple/swift/pull/38736/commits/9a8f2ed64221635646541757b7e38b3b4e02e4b3). Patching out the two lines in InstOptUtils.cpp that call `swift::salvageDebugInfo(inst)` prevents this validation failure, but that's obviously addressing a symptom not the cause. Dan and I had seen some duplicate SIL instructions being generated in the presence of inouts and control flow, but I've had a heck of a time getting a reproducer created. This might provide an excellent clue, so thanks for finding it. I'll take a look at this myself to see if I can minimize something out of it.

Sources/TensorFlow/Operators/Basic.swift
```diff
@@ -1297,7 +1297,7 @@ extension Tensor {
   }
 
   @inlinable
-  @differentiable(wrt: self where Scalar: TensorFlowFloatingPoint)
+  // @differentiable(reverse, wrt: self where Scalar: TensorFlowFloatingPoint)
   internal subscript(_ indexPath: IndexPath) -> Tensor {
```
BradLarson, line 1301: Were there problems with differentiating through subscripts that led to the removal of the attribute and VJP?

ProfFan: Yeah it seems that in nightly the where clause is somewhat broken, and the VJP seems to apply to the non-`TensorFlowFloatingPoint` case as well, resulting in a protocol non-compliance. I haven't got the time to investigate deeper.

BradLarson: Are the where clauses broken in nightlies before or after this PR: https://github.com/apple/swift/pull/39826 ? I'm wondering if that fixed or caused the issue here.

ProfFan: Good point, let me investigate. Thanks Brad :)

ProfFan:
@rxwei @BradLarson @dan-zheng I tried really hard to isolate the failure case, but failed. I believe this is because this bug only happens where there is some kind of inlining optimization, hence that Release only failure mode.

CMakeLists.txt
compnerd: Technically, the tensorflow core itself is not needed for the build, it is only needed at runtime.

Sources/TensorFlow/Core/Tensor.swift
compnerd: This was a workaround for an actual issue.  I think that it would be better to verify if the underlying issue has been resolved or not.

Sources/TensorFlow/StdlibExtensions.swift
compnerd: Why the additions?

compnerd (commented):
I think that this would be easier to review if you split it out into logical commits which add the `reverse` modifier to the `@differentiable`, renames `forward` to `callAsFunction`, removes the `TangentVector`, etc.  Theoretically those would have happened individually when the change was made upstream.

Sources/TensorFlow/StdlibExtensions.swift
BradLarson: `MutableCollection` had a recently fixed loophole where you could inherit default conformances that triggered an infinite recursion. You have to provide these yourself now in recent toolchains, which I believe is what is happening here.

Sources/TensorFlow/Core/Tensor.swift
BradLarson: The underlying issue has migrated in that you can now trigger the same crash by inserting this placeholder, rather than avoiding the crash by having it. I've seen this with other types. Unclear when the behavior inverted.

philipturner:
Seeing as this PR is stale, I'm going to comment about something the people involved with this PR ought to know about. I know this might not be the most appropriate way to reach out, but please bear with me.

I'm working on turning [DL4S](https://github.com/palle-k/dl4s) into a successor to Swift for TensorFlow. I'm rebuilding it from the ground up for GPU acceleration through Metal. I will also heavily modify its current API, adding fragments of S4TF (using `@differentiable` for custom layers and `Tensor<Float>` for tensors), Keras (a SwiftUI-like DSL for declaring models), and new activation functions. @ProfFan would you be willing to redirect your interest in resurrecting S4TF onto our effort at DL4S? I could invite you to our organization.

I just realized that I have an opportunity to bring DL4S to Windows with ArrayFire, which would make it closer to replacing S4TF. If it succeeds, it could be a major milestone in support for Swift on Windows. @compnerd I thought you might want to know about this. Please tell me what you think or if you know of anybody who might want to help.

compnerd:
> Seeing as this PR is stale, I'm going to comment about something the people involved with this PR ought to know about. I know this might not be the most appropriate way to reach out, but please bear with me.

I think that this is rather an inappropriate forum for this.  I don't particularly consider this PR as stale btw, there are still items to be resolved and investigated before it can be merged.

> I'm working on turning [DL4S](https://github.com/palle-k/dl4s) into a successor to Swift for TensorFlow. I'm rebuilding it from the ground up for GPU acceleration through Metal. I will also heavily modify its current API, adding fragments of S4TF (using `@differentiable` for custom layers, `Tensor<Float>` for tensors, including the `Sequential` function builder API) and new activation functions. @ProfFan would you be willing to redirect your interest in resurrecting S4TF onto our effort at DL4S? I could invite you to our organization.
> 
> I just realized that I have an opportunity to bring DL4S to Windows with ArrayFire, which would make it closer to replacing S4TF. If it succeeds, it could be a major milestone in support for Swift on Windows. @compnerd I thought you might want to know about this. Please tell me what you think or if you could suggest anybody who might want to help us.

The project doesn't seem particularly enticing to me - the "rebuilding from the ground up for GPU acceleration through Metal" is a big component of that.  If you were designing it for optimization for DCOM+ (DirectX is effective accessed through COM), that might be different.  GPU environments are quite fragmented and designing around a particular one is a choice that I do not particularly agree with.

This also seems very much as other platforms are an afterthought.  The overall structure of S4TF is far better from a design standpoint - it generically solves the problem in software and has abstractions over a compilation pipeline to separate the concern of optimization and device/runtime independence.

Additionally, I don't think that you realize the complexity of using ArrayFire.  That codebase has a core in C++ - which means that it is a non-starter.  Until the underlying ABI issues are resolved, C++ interop cannot be reliably used.

philipturner:
@compnerd this discussion has been moved into another forum.

Referencing: https://github.com/dl4s-team/dl4s-evolution/issues/8#issuecomment-981155814

philipturner:
> I don't particularly consider this PR as stale btw, there are still items to be resolved and investigated before it can be merged.

The S4TF team archived this project in February and said they were no longer accepting pull requests to it. Merging this PR would seem to be in violation of Google’s policy. Furthermore, this PR is based on a branch called “resurrection", implying that it is an effort to un-archive S4TF. 

When I made that comment, I had only skimmed through the edits and got the wrong impression that @ProfFan was trying to bring S4TF out of archive mode. I thought that because of his intention, you were never going to accept the PR. I apologize for this misunderstanding.

philipturner:
If Swift for TensorFlow is open for changes again, I will gladly help you out with this pull request. I have plenty of time, which is something all of you seem to greatly lack. I assume that’s why this PR has been inactive for 3 weeks. Could you outline what must be done before this pull request is approved?

@rxwei @BradLarson @compnerd I have more to say on this topic and refactored it into an issue to prevent cluttering this PR. Could you please take a look at that?

philipturner:
What bugs specifically are blocking this swift-apis PR? I’m trying to build this repo on Linux by redoing the build system, and I need to know exactly where autodiff will cause trouble.

What workarounds can I do for each of the bugs? I have to use an official Swift release toolchain, and I’m dealing with the bugs present in Swift 5.5.2.

Also, any explanation for this?
https://bugs.swift.org/browse/SR-15675

philipturner:
@ProfFan @rxwei @dan-zheng @BradLarson someone please help me. Please bear with my rushed tone and run-on sentences; I have no way to continue working on S4TF because of this problem. I have tried building S4TF, but all of the solutions are failing. Bazel fails because it tries to dowload an arm64 version of Bazel 3.7.2 (only x86_64 exists). I have Bazelisk installed on my system and can't find out how to get it off. I suspect that if I could uninstall Bazelisk, I could force Bazel to download as something that works.

Then, I tried just using CMake instead of SwiftPM. According to a comment in @ProfFan's PR, I could then use SwiftPM to re-build the non-X10 stuff afterward. I'm trying to rip out X10 and build a new backend in its place (possibly named `PlaceholderDevice`), so I won't modify X10 after building. The Swift development snapshot seemed to have something wrong. I checked and made the terminal use the development snapshot (`swift --version` showed 5.6-dev) by putting `export TOOLCHAINS=...` in the .zshrc file. When it built in the terminal, it emitted errors that there was "no such module _Differentiation" and the froze indefinitely without an explanation. I checked the activity monitor, and no code compilation process was taking up anywhere close to 100% of my CPU. It seemed to stop at completely random points in the build process. Also, the Swift REPL crashed because libArgumentParser.dylib wasn't found somewhere.

---

At this point, I'm thinking of modifying S4TF to alternatively build on a release toolchain, using my [Differentiation](https://github.com/philipturner/differentiation) package. I can't do that yet because I have no way to reproduce the original build process. I also can't download the pre-built X10 binary and run it from Rosetta because I'm building from a personal fork with this PR's commits merged into it. @ProfFan modified something in the X10 directory(ies), and the existing binaries don't have these changes. If you have an Apple silicon Mac, could you send me the arm64 binaries you built in a package structured similarly to the pre-built x64 macOS X10 binaries?

BradLarson:
@philipturner - Regarding blocking compiler crashers, [one was recently fixed](https://github.com/apple/swift/pull/40503), but two still remain (the 12/23/2021 nightly shapshot is the first toolchain to include the linked fix). The two remaining crashers are both due to failing assertions around debug symbols, one that triggers around mutating functions and control flow, and the other due to an optimization pass behaving oddly during release builds. The former can be worked around by moving control flow out of any mutating function that triggers the assertion failure (BatchNorm was one place we hit that here), and the latter can be avoided by disabling the generation of debug symbols when building for release. Both can be prevented if you build a non-assertion-enabled toolchain.

Fixing these has been slow going, because I've had a hard time creating a concrete reproducer for the first (I do have SIL dumps, which points to a possible cause, but I need a failing test to verify a fix), and the latter only occurs under very specific optimization conditions. I do know the exact commit and lines that introduced the latter failure, but not the root cause.

By "release toolchain", do you mean the ones integrated with Xcode, or the swift.org release toolchains? In either case, the 5.5 branch does not have some necessary autodiff fixes in it that have been put in place over the last year (the branching for that was a while ago), and I know that there was another compiler crasher that directly blocked tensorflow/swift-apis from building which has been fixed in main and is present in the nightly toolchain builds. I don't think switching to a release toolchain (Xcode or swift.org) will help you much in trying to build the project.

If you're getting errors of "no such module _Differentiation", that sounds like the build process is still trying to use the default Swift Xcode toolchain and not one from Swift.org, so you could try to manually place the path to the correct toolchain in your PATH ahead of the Xcode toolchain in order to force the use of the right toolchain.

Regarding building TensorFlow, getting Bazel set up can be a bit of a pain, and building TensorFlow from source can take as long as building the entire Swift toolchain. If you're trying to build this on an M1, I'm not sure that all even works with the version of TensorFlow you'd need to build to connect to the old Swift for TensorFlow interfaces (2.4.0 if I'm reading things right). I think you'll have to build that on X86_64. You may also need to match up the exact version of Bazel that had been used to build TensorFlow (somewhere between 3.7.2 and 3.99.0, but it sounds like you've got the correct 3.7.2).

philipturner:
@BradLarson I'm continuing this conversation in the "requesting assistance" issue to prevent cluttering this PR.

brettkoonce:
start with x86 rather than jump to m1!

philipturner:
I have a few hours tomorrow morning where I can access my Intel Mac mini. Hopefully, I can bridge from the Swift dev toolchain to release within that timeframe, then proceed on Colab (also x86) for the rest of the day.

brettkoonce:
@philipturner wrt to https://github.com/tensorflow/swift-apis/issues/1185#issuecomment-1004776357 + going off of what brad was saying, you should make sure you have the same cmake version as well!

philipturner:
@ProfFan (or @compnerd or @BradLarson if you reproduced his steps while reviewing in October) could you let me know which Swift nightly snapshot was used while building this fix? I'm having trouble reproducing your results on Linux. I attached the Colab notebook where I'm facing issues right now. It's untidy, but hopefully you can use it to reproduce my conditions (please run on Google Colab and read the comments). 

[CompilingSwift2.ipynb.zip](https://github.com/tensorflow/swift-apis/files/7825069/CompilingSwift2.ipynb.zip)

If you'd rather not run it, I downloaded CMake 3.22.1, Bazel 3.1.0, and took the CMake route to compiling with a pre-compiled Ubuntu x64 binary. I pulled source code from ProfFan's `fan/resurrection` fork. I got CMake to start compiling the x10 library at runtime, but it was so slow that I expected it to take hours (a bad idea, especially in Colab).

I'm facing problems with it freezing after logging the command below. I faced a similar silent freezing error when attempting to build S4TF on my Mac mini (with a different cause), and I later found that the program had actually stopped and would log the error if run slightly differently. I can't figure out how to reveal that error in the Colab environment though.
```
[18/30] Linking Swift static library lib/libTensorTests.a
```

On the Swift releases website, it doesn't show snapshots earlier than November. Yet, these patches were made in October. I just need the date of the snapshot used, and the snapshot from then might still be downloadable. This is all to narrow down the cause of the error and reproduce the exact same environment that @ProfFan somehow successfully built S4TF in.

philipturner:
@ProfFan there was a compiler failure while compiling swift-apis using SwiftPM on the 2021-11-12 nightly and using a pre-compiled Ubuntu x10 binary:

```
Building for debugging...
/content/swift-apis/Sources/TensorFlow/Core/DataTypes.swift:89:61: error: cannot find type 'ElementaryFunctions' in scope
  TensorFlowScalar & BinaryFloatingPoint & Differentiable & ElementaryFunctions
```

The full stack trace and conditions for how to reproduce are in this Colab notebook:
[CompilingSwift2 (1).ipynb.zip](https://github.com/tensorflow/swift-apis/files/7828625/CompilingSwift2.1.ipynb.zip)

philipturner:
I got S4TF to build successfully once in a Colab environment, although I used a pre-compiled x10.

philipturner:
@ProfFan there was a build failure on the 2021-12-23 snapshot that didn't occur on the 2021-11-12 snapshot. This should be worth investigating to add commits to your PR. The crash log is in one of the later cells of this notebook:
[CompilingSwift2 (3).ipynb.zip](https://github.com/tensorflow/swift-apis/files/7831382/CompilingSwift2.3.ipynb.zip)

The crash log was related to `Differentiable`, `Requirement machine`, and `RecurrentLayerCell`. Ignore the first two cells that instruct you to install CMake and Bazel, as those aren't necessary. @BradLarson is this crash something new?

I also tested it on 2022-01-06, with the same result.

BradLarson:
If I'm reading these notebook results correctly, it looks like the 12/23/2021 nightly snapshot toolchain got a lot farther in building than the 11/12/2021 one. I vaguely remember being able to use the 2021-07-07 nightly (downloadable from [here](https://download.swift.org/development/ubuntu1804/swift-DEVELOPMENT-SNAPSHOT-2021-07-07-a/swift-DEVELOPMENT-SNAPSHOT-2021-07-07-a-ubuntu18.04.tar.gz) for X86 Ubuntu 18.04) to build this branch before, but I could be wrong about that. That was a good, stable snapshot for autodiff for a while.

Looking at the failure on the 12/23/2021 toolchain, it appears that the RequirementMachine is unable to solve out the TangentVector for RecurrentLayer (type defined at swift-apis/Sources/TensorFlow/Layers/Recurrent.swift, lines 370 - 469). If that could be minimized, it'd be great to report upstream for Richard and / or Slava to look at. If this persists with the just-released 1/6/2022 toolchain, let me know and I can minimize and report that.

As a workaround, you might be able to define a custom TangentVector for RecurrentLayer that explicitly states what the compiler is currently generating for that TangentVector and that might help the solver here.

philipturner:
2021-11-12 actually completed the build (in fact, it was the only toolchain that succeeded so far). I had some problems with it previously, but they were my fault and I resolved them. I didn’t verify whether it compiles on 2021-07-07.

The problem persists in the 2022-01-06 toolchain. Below is another Colab notebook that ran on 2021-01-06. Is this sufficient content for you to generate your bug report?
[CompilingSwift2 (4).ipynb.zip](https://github.com/tensorflow/swift-apis/files/7831742/CompilingSwift2.4.ipynb.zip)

philipturner:
> As a workaround, you might be able to define a custom TangentVector for RecurrentLayer that explicitly states what the compiler is currently generating for that TangentVector and that might help the solver here.

@BradLarson could you give super precise details on how to implement the workaround? I anticipate that my lack of experience with differentiation will unnecessarily waste time if I tackle the problem alone. Plus, it takes 2 minutes to reach the build failure, unlike Xcode where you can cache build products (I might change that by moving to my Mac mini if this gets serious enough).

BradLarson:
> @BradLarson could you give super precise details on how to implement the workaround?

I haven't verified that this will compile, but you'll want to add something like the following before [line 376 in Recurrent.swift](https://github.com/tensorflow/swift-apis/blob/main/Sources/TensorFlow/Layers/Recurrent.swift#L376):

```swift
public struct TangentVector: Differentiable, AdditiveArithmetic {
  public typealias TangentVector = RecurrentLayer.TangentVector
  public var cell: Cell.TangentVector

  public init(cell: Cell.TangentVector) {
    self.cell = cell
  }
}

mutating func move(by offset: TangentVector) {
  self.cell.move(by: offset.cell)
}
```

That should provide what you need to satisfy a manual TangentVector for the RecurrentLayer type. Again, not entirely sure that'll compile, but something close to that is what you'd need.

philipturner:
Thanks! That’s what I was asking for! And any progress/insights on the bug with VJP pow derivative? (I assigned it to you on JIRA)

philipturner:
@BradLarson the crash still occurs (2022-01-06), even with your workaround. It says this in the crash log:

```
5.	While generating protocol witness thunk SIL function "@AD__$s10TensorFlow14RecurrentLayerVyxGAA0D0A2aEP14callAsFunctiony6OutputQz5InputQzFTW_jvp_SS".
 for 'callAsFunction(_:)' (at /content/swift-apis/Sources/TensorFlow/Layers/Recurrent.swift:464:10)
```

I noticed that it says `jvp` at the end of the mangled name, which shouldn't exist anymore. @rxwei made requests on JIRA for someone to remove JVP protocol witnesses:
- https://bugs.swift.org/browse/SR-15577

It happened at this place (around line 440), which doesn't have a manually defined derivative. Will manually defining a derivative fix it, and what code should I put as the derivative?

```swift
@differentiable(reverse)
public func callAsFunction(_ inputs: [Cell.TimeStepInput]) -> [Cell.TimeStepOutput] {
  let initialState = withoutDerivative(at: cell.zeroState(for: inputs[0]))
  return self(inputs, initialState: initialState)
}
```

Colab notebook:
[CompilingSwift2 (5).ipynb.zip](https://github.com/tensorflow/swift-apis/files/7833376/CompilingSwift2.5.ipynb.zip)

philipturner:
@ProfFan I can't get my comment on your code above to load when I'm not logged in to my personal account. It says "Pending" in a label just like many code snippets say "Outdated". Can you see the comment?

The comment was to this: https://github.com/tensorflow/swift-apis/pull/1184#discussion_r734912234

ProfFan:
You have to click "Review" after you make review comments.

philipturner:
@ProfFan did you get a notification for the PR I left on your fork? GitHub might do something weird with notifications if it isn't on the main repository.

About the comment, it won't let me click "review" now, probably because I don't have comment access. Here's what I said:

> @ProfFan please take a look at my modifications to the custom autodiff tutorial (https://github.com/tensorflow/swift/pull/625).

philipturner:
@ProfFan do any of your current projects depend on S4TF? If so, wouldn't your work be impacted by the new crash that appeared in 2021-12-23?

ProfFan:
No I am not doing any S4TF in my current work

philipturner:
@ProfFan how to do you prefer to be contacted for discussions not relevant to this PR? I can't find your email anywhere.

ProfFan:
I have no bandwidth for this now. Maybe I will have time after 2 months.

Sources/TensorFlow/Core/Tensor.swift
philipturner: I'm not seeing this happen with or without `SR13263Workaround` commented out, on current Swift toolchains.

Nevermind. I reproduced the crash on recent nightly toolchains, using the reproducer from https://github.com/apple/swift/issues/55703.

philipturner: I'm not seeing the crash _affect S4TF_, with or without `SR13623Workaround` commented out. I'm running `swift build` and `swift test`, and it's not activating the crash on recent toolchains. Am I doing this right?

```bash
export TENSORFLOW_USE_RELEASE_TOOLCHAIN=1
cd s4tf
swift build -Xswiftc -DTENSORFLOW_USE_STANDARD_TOOLCHAIN \
  -c release -Xswiftc -g -Xcc -I${DESTDIR}/usr/include -Xlinker -L${DESTDIR}/usr/lib

rm -f .build/release/libx10.dylib
cp ${DESTDIR}/usr/lib/libx10.dylib .build/release/libx10.dylib

swift test -Xswiftc -DTENSORFLOW_USE_STANDARD_TOOLCHAIN \
    -c release -Xswiftc -g -Xcc -I${DESTDIR}/usr/include -Xlinker -L${DESTDIR}/usr/lib
```

Sources/TensorFlow/Layers/Normalization.swift
philipturner: Wouldn't this suggestion harm performance because you have to call `input.rank` twice? That would increase the total overhead from 3/4 to 6/8 function calls.

Sources/TensorFlow/Operators/Basic.swift
philipturner: This is a manifestation of https://github.com/apple/swift/issues/59135, a bug that appeared in September 2021. All you had to do was comment out the `@differentiable(...)` attribute on `subscript(_: IndexPath)`. The other declarations - `subscript(_: TensorRangeExpression...)` and `_vjpSubscript` - are fine.

PR #1188, opened 2021-12-31
Fix indentation

philipturner:
(not related to this PR) We need to remove the dependency on PythonKit for iOS, which doesn't seem like too much of a hurdle (use the conditional dependency feature added in SwiftPM 5.5). When we start producing pre-compiled binaries again, we'll need separate XCFrameworks for macOS and iOS, with the only difference being that the iOS one leaves out PythonKit. The Colab binary will be special because it requires a different build system than SwiftPM, but it can be imported into Google Colab through a special %install directive.

I have no idea how to make a pre-compiled Swift binary that doesn't need a special Swift toolchain on Windows, even if using WSL. That platform also lacks GPU acceleration for the time being, so running on CPU with a debug build will be extremely slow. If DirectML support happens, it might be feasible to build S4TF from source (in debug mode) on Windows because the GPU does the heavy lifting.

### DIFF
--- a/Sources/TensorFlow/Core/PythonConversion.swift
+++ b/Sources/TensorFlow/Core/PythonConversion.swift
@@ -13,164 +13,162 @@
 // limitations under the License.
 
 #if canImport(PythonKit)
+import PythonKit
 
-  import PythonKit
+/// The `numpy` Python module.
+/// Note: Global variables are lazy, so the following declaration won't produce
+// a Python import error until it is first used.
+private let np = Python.import("numpy")
 
-  /// The `numpy` Python module.
-  /// Note: Global variables are lazy, so the following declaration won't produce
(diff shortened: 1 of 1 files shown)


BradLarson:
Is this the indentation produced by [swift-format](https://github.com/apple/swift-format), per [the project's formatting rules](https://github.com/tensorflow/swift-apis/blob/main/.swift-format)? I recall that all files in this project were run through swift-format at some point, and without the presubmits running any more on this repository we might need to manually verify that this still matches the swift-format style.

philipturner:
There are other conditional import statements that aren’t indented in S4TF, and this entire file is basically a conditional import. I suspect it was an accident and either not caught by swift-format or corrupted after it stopped being used.

PR #1192, opened 2022-06-18
Initial support for Google Colaboratory

philipturner:
Allows the head branch of Swift for TensorFlow to execute Swift package tests in a Google Colab environment. This means using the old TF 2.4 binary, which is slightly API-incompatible with TF 2.9. Some operators in the `_Raw` namespace will cause a runtime crash (outlined) below, although nothing in the public API should crash, unless it is deprecated (like `Dataset`).

- Operators with the `metadata` parameter. One example is `_Raw.tensorSliceDataset`, which was used in Swift package tests. To work around this, add `-Xswiftc -DFALLBACK_X10_BINARY` when running tests on the old X10 binary.
- Operators whose generic parameter constraints changed when [upgrading raw bindings](https://github.com/s4tf/s4tf/pull/10) to TF 2.9. Several ops upgraded from `TensorFlowIndex` (`Int32` and `Int64`) to `TensorFlowInteger` (all Swift integer types).


philipturner:
This was an accident. I meant to make the PR to [s4tf/s4tf:main](https://github.com/s4tf/s4tf).

